PR #15, opened 2016-02-21
deps: Add goop and Goopfile

AbdealiJK:


AbdealiJK:
Ive asked the circleci guys whats up with this - blocked until they reply back.

Issue #32, opened 2016-02-21
XML Bear


sils1297:
_From @sils1297 on February 7, 2016 16:8_

apparently I did https://github.com/coala-analyzer/coala/commit/20fdbd88f3d638365776e75f7af9f01e6d18af29 not sure how good this is reusable

_Copied from original issue: coala-analyzer/coala#1309_

sils1297:
_From @yash14123 on February 9, 2016 2:8_

Can you explain this a little bit more?

sils1297:
_From @Makman2 on February 9, 2016 3:44_

This issue is about implementing a new bear that scans through XML files. For this purpose @sils1297 used the python built-in `xml` module inside his commit.
What actual analyze routines this bear shall implement is up to its writer :)

PR #96, opened 2016-02-24
bears/xml: Add XMLBear

SanketDG:
check for XML syntax errors using xmllint

Fixes https://github.com/coala-analyzer/coala-bears/issues/32

### DIFF
--- a/bears/xml/XMLBear.py
+++ b/bears/xml/XMLBear.py
@@ -1,0 +1,19 @@
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+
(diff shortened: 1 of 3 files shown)


Redridge:
@sils1297 maybe also take a look? I find it ok though

sils1297:
I think it'd be better to use gives_corrected, if you have syntax errors it'll give them via stderr which will make coala yield a warning that the file is unlintable which is fine IMO.

sils1297:
though I'm open to opinions

sils1297:
ignoring scrutinizer as it's not required and http://status.scrutinizer-ci.com/

sils1297:
@rultor merge

Issue #71, opened 2016-02-21
Ignore test-requirements.txt in py.test


SanketDG:
`test-requirements.txt` gets picked up by `py.test` because of the `test-` prefix. This should be avoided.

sils1297:
cool, can be ignored in setup.cfg , a happy newcomer can do this ;)

PR #134, opened 2016-03-01
setup.cfg: Ignore file test-requirements.txt

mr-karan:
`test-requirements.txt` has been added to addopts --ignore flag which ensures
it doesn't get picked up when using `py.test`

Fixes https://github.com/coala-analyzer/coala-bears/issues/71

Sorry for opening a new PR, the old one had a dirty commit history and I no longer had that copy of the repo. Closing the previous one.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,7 @@ python_files = *Test.py
 python_classes = *Test
 python_functions = *_test
 timeout = 35
-addopts = --color=yes
+addopts = --color=yes --ignore=test-requirements.txt
 env =
     PYTHONHASHSEED=0
 # PYTHONHASHSEED=0 is required to use same hashes in pytests-xdist's workers


setup.cfg
```diff
@@ -9,7 +9,7 @@ python_files = *Test.py
 python_classes = *Test
 python_functions = *_test
 timeout = 35
-addopts = --color=yes
+addopts = --color=yes --ignore=test-requirements.txt
```
abhsag24: shouldn't this ignore statement be above for clarity ?

abhsag24: and are you sure this how we ignore files?

AbdealiJK: I was wondering the same. But I can't find any other method of doing it :o
Can you ?

mr-karan: According to [this](http://stackoverflow.com/questions/11117062/how-to-tell-py-test-to-skip-certain-directories) we can use [`norecursedirs`](http://pytest.org/latest/customize.html#confval-norecursedirs)

SanketDG: Yup this is the only way I think.

mr-karan: `norecursedirs` only works for directories and not a single file. So, even I think this is the only possible solution

abhsag24: hmm, so in any case can we move the statement where `python_files` are? or at least in its own line?, it looks confusing right now.

AbdealiJK:
I think it's fine ... I can't think of a better way of doing it, and it's a minor change

AbdealiJK:
ack  25ab965

AbdealiJK:
Rebase needed though

sils1297:
reack 39549ed

sils1297:
@rultor merge

PR #159, opened 2016-03-05
requirements: Update coala version

abhsag24:
Previous coala versions didn't have coalang files
packaged into them, and also had some problems with the keys
version 0.4.2.dev20160304094246 has all that fixed and that is also
required by the upcoming IndentationBear


sils1297:
@rultor merge

Issue #157, opened 2016-03-05
dart sdk - use the cache


AbdealiJK:
We cache the dart sdk folder, but we keep reinstalling it without checking if it's installed already in circleci.

Adrianzatreanu:
i feel guilty

Adrianzatreanu:
ill try working on this, even though i remember i tried doing all these and it failed miserably, but ill try again now :)

Adrianzatreanu:
Hm how exactly are we to check if its already installed?
Check if it exists at that path already?

AbdealiJK:
Just do `dartlint -v` ? or `dart -v` ? and see if the exitcode is 0 or not

Adrianzatreanu:
well for some reason it would have a != 0 exit code even if the command existed, so i tried a different approach. what do you think?

PR #166, opened 2016-03-05
deps.sh: Don't download dart-sdk if it exists

Adrianzatreanu:
Downloads and installs dart-sdk only if it doesn't exist already.

Fixes https://github.com/coala-analyzer/coala-bears/issues/157

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -55,8 +55,10 @@ done
 python setup.py --help
 
 # Dart Lint commands
-wget -nc -O ~/dart-sdk.zip https://storage.googleapis.com/dart-archive/channels/stable/release/1.14.2/sdk/dartsdk-linux-x64-release.zip
-unzip -n ~/dart-sdk.zip -d ~/
+if ! dartanalyzer -v &> /dev/null ; then
+  wget -nc -O ~/dart-sdk.zip https://storage.googleapis.com/dart-archive/channels/stable/release/1.14.2/sdk/dartsdk-linux-x64-release.zip
+  unzip -n ~/dart-sdk.zip -d ~/
+fi
 
 # Julia commands
 julia -e "Pkg.add(\"Lint\")"


.ci/deps.sh
```diff
@@ -55,8 +55,10 @@ done
 python setup.py --help
 
 # Dart Lint commands
-wget -nc -O ~/dart-sdk.zip https://storage.googleapis.com/dart-archive/channels/stable/release/1.14.2/sdk/dartsdk-linux-x64-release.zip
-unzip -n ~/dart-sdk.zip -d ~/
+if [ ! -d "~/dart-sdk/" ] && [ ! -d "~/dart-sdk/bin" ]; then
```
AbdealiJK: Would it not be better to do `if ! dartlint -v ; then` as it will also check if the dartlint is installed correctly ?

Adrianzatreanu: https://github.com/coala-analyzer/coala-bears/issues/157#issuecomment-192712860

Adrianzatreanu: ah well i made it work now nvm

Adrianzatreanu: however, i think it would have a conflict with that? echo 'export PATH=$PATH:~/dart-sdk/bin' >> ~/.circlerc

AbdealiJK: I don't follow. This command runs after the export right ?

Adrianzatreanu: I dont know the order. You sure export runs first and then the download?

AbdealiJK: yep (Notice that the export is in the `pre` section in the config file

Adrianzatreanu:
ok, do you like this now :) ?

AbdealiJK:
ack 50cca07

Adrianzatreanu:
@rultor merge

jayvdb:
This didnt work at all.

PR #168, opened 2016-03-06
bears: Adding ESLint Bear

sudheesh001:


sils1297:
@sudheesh001 ping me if you need help, eslint is important, many people want it and it can autofix stuff

AbdealiJK:
@sudheesh001 can you paste the link of the commit with lots of reviews I had given ? So that it doesnt get lost.

sils1297:
wip as per discussion on gitlab

sils1297:
gitter

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                        newoutput[int(endingValue+replacement_count):]
+                    # print(newoutput)
+                else:
+                    yield Result.from_values(
+                        origin=self,
+                        message=lint_result['message'],
+                        file=filename,
+                        severity=self.severity_map[str(lint_result['severity'])],
+                        line=lint_result['line'])
+                    pass
```
sils1297: unneeded pass

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                if 'fix' in lint_result:
+                    fixStructure = lint_result['fix']
+                    startingValue, endingValue = fixStructure['range']
+                    replacementText = fixStructure['text']
+                    if replacementText != "":
+                        replacement_count+=1
+                    newoutput = newoutput[
+                        :int(startingValue)+replacement_count] + replacementText +\
+                        newoutput[int(endingValue+replacement_count):]
+                    # print(newoutput)
```
sils1297: you never yield a result except for the last correction?

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.misc.Shell import escape_path_argument
+from coalib.results.Result import Result
+from json.decoder import JSONDecodeError
+
+
+class ESLintBear(LocalBear, Lint):
+    executable = 'eslint'
+    severity_map = {
+        "2": RESULT_SEVERITY.MAJOR,
```
AbdealiJK: These should probably be ints.
The json gives ints - not strings

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+            lines = [line for line in open(filename)]
+            lines = "".join(lines)
+            newoutput = lines
+            replacement_count = 0
+            for lint_result in output[0]['messages']:
+                if lint_result['severity'] == 0:
+                    continue
+                if 'fix' in lint_result:
+                    fixStructure = lint_result['fix']
+                    startingValue, endingValue = fixStructure['range']
```
AbdealiJK: these numbers dont seem to be strings in the json - are you sure they are strings here ?
I dont think you need the "int()" later

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                    replacementText = fixStructure['text']
+                    if replacementText != "":
+                        replacement_count+=1
```
AbdealiJK: This logic doesnt seem to make sense. I think it should be `replacement_count += len(replacementText) - len_of_replaced_text` ?
I guess length of replaced text can be found with startValue, endValue

bears/js/ESLintBear.py
AbdealiJK: Our variables are snake cased, not camel case

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+            newoutput = lines
+            replacement_count = 0
+            for lint_result in output[0]['messages']:
+                if lint_result['severity'] == 0:
+                    continue
+                if 'fix' in lint_result:
+                    fixStructure = lint_result['fix']
+                    startingValue, endingValue = fixStructure['range']
+                    replacementText = fixStructure['text']
+                    if replacementText != "":
```
AbdealiJK: snake case

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+        :param eslint_config: The location of the .eslintrc config file.
+        '''
+        self.arguments = '--no-ignore --no-color -f=json --stdin'
+        if eslint_config:
+            self.arguments += (" --config "
+                               + escape_path_argument(eslint_config))
+
+        return self.lint(filename, file)
+
+    def _process_issues(self, output, filename):
```
AbdealiJK: Don't use `_process_output` there's another function which already gives you the `file` also. Then you wont need line 38 below

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+        try:
+            output = json.loads("".join(output))
+            lines = [line for line in open(filename)]
+            lines = "".join(lines)
+            newoutput = lines
+            replacement_count = 0
+            for lint_result in output[0]['messages']:
```
AbdealiJK: Can we make sure that len(output) == 1 ?

AbdealiJK: An assert which throws errors would be good, as if its not 1 - its a bug in the bear

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                    newoutput = newoutput[
+                        :int(startingValue)+replacement_count] + replacementText +\
+                        newoutput[int(endingValue+replacement_count):]
+                    # print(newoutput)
+                else:
+                    yield Result.from_values(
+                        origin=self,
+                        message=lint_result['message'],
+                        file=filename,
+                        severity=self.severity_map[str(lint_result['severity'])],
```
AbdealiJK: str can be removed and severity_map can use ints

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                if lint_result['severity'] == 0:
+                    continue
+                if 'fix' in lint_result:
+                    fixStructure = lint_result['fix']
+                    startingValue, endingValue = fixStructure['range']
+                    replacementText = fixStructure['text']
+                    if replacementText != "":
+                        replacement_count+=1
+                    newoutput = newoutput[
+                        :int(startingValue)+replacement_count] + replacementText +\
```
AbdealiJK: Better formatting would be immensely helpful for readability here

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                lines.splitlines(True), newoutput.splitlines(True))
+
+            yield Result.from_values(
+                origin=self,
+                message=lint_result['message'],
+                file=filename,
+                diffs={filename: diff},
+                severity=self.severity_map[str(lint_result['severity'])],
+                line=lint_result['line'])
+        except JSONDecodeError:
```
AbdealiJK: The JSONDecodeError is only raised in `json.loads` so why add everything into try ?

bears/js/ESLintBear.py
```diff
@@ -0,0 +1,77 @@
...
+                else:
+                    yield Result.from_values(
+                        origin=self,
+                        message=lint_result['message'],
+                        file=filename,
+                        severity=self.severity_map[str(lint_result['severity'])],
+                        line=lint_result['line'])
+                    pass
+
+            diff = Diff.from_string_arrays(
```
AbdealiJK: You could just use `__yield_diffs` in Lint I think ? (Dont remember the name but there definitely exists a helper function for this

tests/js/ESLintBearTest.py
```diff
@@ -0,0 +1,51 @@
...
+        "semi"             : [1, "always"],
+        "space-unary-ops"  : 2
+    }
+}
+""".splitlines(True)
+
+ESLintBearGoodTest = verify_local_bear(ESLintBear,
+                                       valid_files=(test_good,),
+                                       invalid_files=(test_bad,),
+                                       settings={"eslint_config":
```
AbdealiJK: Im still wondering about the config ... Are you sure it makes sense ???
Youre giving a list as a cli arg. I have no clue how it works ...

sudheesh001:
previous commit review at https://github.com/coala-analyzer/coala-bears/pull/168/files#diff-b16213bb59fa1facd911d595fc62fa31R50

Just putting this up so that we don't lose it after the rebase

bears/js/ESLintBear.py
sudheesh001: I didn't understand this, Could you explain further, I've made the other changes, the bear recognizes the warnings correctly but still faces a few issues in the corrections.

AbdealiJK: Use `_process_corrected`.
It gives out a variable named `file` which is the same as `lines` that you create inside this function

sudheesh001:
Thanks @AbdealiJK I'll open a new comer issue, maybe @gopalmeena would be interested to work on it. He sent an initial patch to coala before and was looking at getting started.

sudheesh001:
@rultor merge

PR #186, opened 2016-03-08
Makman2/deps

Makman2:
~~browsed the packages on pypi for the most recent versions, since they seem to work with the latest PRs that got merged.~~
EDIT: Some of the newest packages don't work, let's do this somewhen else...


AbdealiJK:
I am still against it
I think we should file a bug in radon about this. If another package (lets say pytest) wants a higher version - we wont be able to do anything.
I do not think this is a fix.

Makman2:
We can't rely on packages we don't test against. This is the only way to ensure our users a working program.
If they want newer packages they shall try installing them manually or if they have conflicting packages, they can easily setup a venv and install the specific dependencies from coala there.
Especially we have many packages, maintaining this stability gets too difficult.

Makman2:
... I hope we don't spin around with the dependencies now...  @sils1297 could you take a look?

AbdealiJK:
@Makman2 so you're saying if I have a flask application to make some website, I should make a virtualenv for flask and another virtualenv for coala if I want to use both ?
I think thats a major usability issue ...

Also, what do you mean by we do not test against it ?

sils1297:
rejected, fixing all the version will yield lots and lots of more of those conflicts we're experiencing now to our users indeed. I'm personally thinking about "lets fix the radon version as a hotfix", then we have to think about how we make pip on our CI uninstall packages like colorama if they are too new. The problem only occurred because radon added the colorama dependency with == right?

sils1297:
can we close this in favour of https://github.com/coala-analyzer/coala-bears/pull/191 ? A bug with radon was filed.

sils1297:
Related: rubik/radon#82
Related: coala-analyzer/coala#1852

PR #187, opened 2016-03-09
requirements.txt: Fix radon version

Makman2:
radon introduces version conflicts with 1.2.3. Limitting version until
we find a solution to that.


sils1297:
I'm doing a rebased version with a more elaborate explanation in the commit message but the change is the same

sils1297:
superseded by https://github.com/coala-analyzer/coala-bears/pull/191

PR #198, opened 2016-03-09
bears/ruby: Add Ruby-Lint Bear

mr-karan:
Added `ruby-lint` support 

Though there was already a file with the name `RubyLintBear`, I don't know what to name this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,5 @@ source 'http://rubygems.org'
 ruby '2.1.5'
 
 gem "sqlint"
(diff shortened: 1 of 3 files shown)


SanketDG:
I am not sure how reliable is `ruby -wc {file}` compared to this linter. I did rename that to something else,
probably RubySyntaxBear, and keep this as RubyLintBear, since this actually uses a linter.

SanketDG:
48beef8 needs work

PR #207, opened 2016-03-10
bears/typescript: Add TSLintBear

arafsheikh:
Fixes https://github.com/coala-analyzer/coala-bears/issues/201

### DIFF
--- a/bears/typescript/TSLintBear.py
+++ b/bears/typescript/TSLintBear.py
@@ -1,0 +1,45 @@
+import json
+
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+from coalib.misc.Shell import escape_path_argument
+from coalib.results.Result import Result
+from coalib.settings.Setting import path
+
+
+class TSLintBear(LocalBear, Lint):
+    executable = 'tslint'
+
+    def process_output(self, output, filename, file):
+        output = json.loads("".join(output)) if output else []
+        for issue in output:
+            yield Result.from_values(
+                origin="{} ({})".format(self.__class__.__name__,
+                                        issue['ruleName']),
+                message=issue["failure"],
+                file=issue["name"],
+                line=int(issue["startPosition"]["line"]) + 1,
+                end_line=int(issue["endPosition"]["line"]) + 1,
+                column=issue["startPosition"]["character"],
+                end_column=issue["endPosition"]["character"])
+
+    def run(self,
+            filename,
+            file,
+            tslint_config: path="",
+            rules_dir: path=""):
+        '''
+        Checks the code with ``tslint`` on each file separately.
+
+        :param tslint_config: Path to configuration file.
+        :param rules_dir:     Rules directory
+        '''
+        self.arguments = "--format json"
+        if tslint_config:
+            self.arguments += (" --config " +
+                               escape_path_argument(tslint_config))
+        if rules_dir:
+            self.arguments += (" --rules-dir " +
+                               escape_path_argument(rules_dir))
+        self.arguments += " {filename}"
+        return self.lint(filename, file)
--- a/tests/typescript/TSLintBearTest.py
+++ b/tests/typescript/TSLintBearTest.py
@@ -1,0 +1,42 @@
+import os
+
+from bears.typescript.TSLintBear import TSLintBear
+from tests.LocalBearTestHelper import verify_local_bear
+
+good_file = """function findTitle(title) {
+    let titleElement = "hello";
+    return title;
+}
+let t = findTitle("mytitle");
+t.innerHTML = "New title";
+""".splitlines(True)
+
+bad_file = """function findTitle(title) {
+    let titleElement = 'hello';
+    return title;
+}
+let t = findTitle('mytitle');
+t.innerHTML = 'New title';
+""".splitlines(True)
+
+tslintconfig = os.path.join(os.path.dirname(__file__),
+                            "test_files",
+                            "tslint.json")
+
+TSLintBearWithoutConfig = verify_local_bear(TSLintBear,
+                                            valid_files=(good_file,),
+                                            invalid_files=(bad_file,),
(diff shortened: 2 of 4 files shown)


sils1297:
64139e115f8d21efb671a52180561dedad721141 needs work

AbdealiJK:
unack 73f75db

tests/typescript/TSLintBearTest.py
```diff
@@ -0,0 +1,42 @@
...
+                                             valid_files=(bad_file,),
+                                             invalid_files=(good_file,),
+                                             settings={"tslint_config":
+                                                       tslintconfig},
+                                             tempfile_kwargs={"suffix": ".ts"})
+
+TSLintBearOtherOptions = verify_local_bear(TSLintBear,
+                                           valid_files=(good_file,),
+                                           invalid_files=(bad_file,),
+                                           settings={"rules_dir": "/"},
```
arafsheikh: @sils1297 @AbdealiJK For custom rules to work we have to compile the `rule.ts` file and then pass it to the linter. Do we really need this complex test? 
IMO a better option will beto combine the last two tests into:

```python
TSLintBearConfigAndOptions = verify_local_bear(TSLintBear,
                                               valid_files=(good_file,),
                                               invalid_files=(bad_file,),
                                               settings={"tslint_config":tslintconfig,
                                               "rules_dir": "/"},
                                               tempfile_kwargs={"suffix": ".ts"})
```
So this will test if passing the rule argument breaks the code, without adding any complexity.

sils1297: I think I'd just keep it like it is right now

arafsheikh: :+1:

bears/typescript/TSLintBear.py
```diff
@@ -0,0 +1,48 @@
...
+    def run(self,
+            filename,
+            file,
+            tslint_config: str="",
+            rules_dir: str=""):
+        '''
+        Checks the code with ``tslint`` on each file separately.
+
+        :param tslint_config: Path to configuration file.
+        :param rules_dir: Rules directory
```
AbdealiJK: formatting is wrong (The `R` should start at the same point as `P`)

bears/typescript/TSLintBear.py
```diff
@@ -0,0 +1,48 @@
...
+            # This implies that the result is empty and there
+            # are no errors.
+            return
+
+        for issue in output:
+            yield Result.from_values(
+                origin=self,
+                message=issue["failure"],
+                file=issue["name"],
+                line=int(issue["startPosition"]["line"])+1,
```
AbdealiJK: spaces around `+`

bears/typescript/TSLintBear.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class TSLintBear(LocalBear, Lint):
+    executable = 'tslint'
+
+    def process_output(self, output, filename, file):
+        try:
+            output = json.loads("".join(output))
+        except ValueError:
+            # This implies that the result is empty and there
```
AbdealiJK: I'd be much more certain of this if you do `if output == ()` or similar ?
`try` should not be used if youre not catching errors if possible

bears/typescript/TSLintBear.py
```diff
@@ -0,0 +1,48 @@
...
+                message=issue["failure"],
+                file=issue["name"],
+                line=int(issue["startPosition"]["line"])+1,
+                column=issue["startPosition"]["character"],
+                end_column=issue["endPosition"]["character"])
+
+    def run(self,
+            filename,
+            file,
+            tslint_config: str="",
```
AbdealiJK: Can you use the `path` in Settings ? This is something we're trying to get people to use from now on.

bears/typescript/TSLintBear.py
arafsheikh: I'm going to change it to:

```python
       try:
            output = json.loads("".join(output))
        except ValueError as e:
            # This implies that the result is empty and there
            # are no errors.
            if str(e) == "Expecting value: line 1 column 1 (char 0)":
                return
            raise ValueError
```

Is that okay?

bears/typescript/TSLintBear.py
arafsheikh: Can you please elaborate?

bears/typescript/TSLintBear.py
SanketDG: Why would you do that?
Use `if` for checking empty json <- pretty self-explanatory.

There shouldn't even be a `try...except`

AbdealiJK: Hey, so the basic thing is that we have a variable `output` which we are parsing json from right ?
If there is no result, can't we simply check if this is empty ?

SO, we can do something like `output == ""` or `output == ()` or `output == '{}'` or similar ?
This would allow us to do something like:

```
if output == "":
    return []
# the rest of the code
```

arafsheikh: `json.loads("".join(output))` throws a `ValueError`. I can't proceed without catching the exception.

tushar-rishav:
```
output = json.loads("".join(output)) if output else []
```
@arafsheikh Will this help?

arafsheikh: @tushar-rishav Oh didn't think of that. Thanks :smiley:

sils1297:
46f94bf is ready

tushar-rishav:
@arafsheikh  Could you just rebase before we go ahead and merge this?

sils1297:
unack 348be18, general rule: if there's info in the output that you didn't use, it's probable that you missed something :) coala should have as much info as possible in the result

sils1297:
https://github.com/coala-analyzer/coala-bears/commit/1520444923f2c035e4a5e58f0710eeba8b8d570b is ready

sils1297:
@rultor merge

PR #220, opened 2016-03-12
deps: Auto reply yes for prompts in apt-get

AbdealiJK:
Earlier we had removed the `-qq` flag in apt-get and
not added the -y. -qq automatically added the -y and so
If a rebuild without cache is done, this will hang.


abhsag24:
@rultor merge

Issue #223, opened 2016-03-12
Setup doesn't assert requirement of python3 correctly


iamgroot42:
Running the setup with python2 gives the following error message:

`Traceback (most recent call last):`
`  File "setup.py", line 8, in <module>`
`    from urllib.request import urlopen`
`ImportError: No module named request`

One of the imports depends on python3 ; yet, it is placed before the check for 'python_version'.

Submitting a PR to fix the same.

PR #224, opened 2016-03-12
setup: Change position of urllib import

iamgroot42:
When installing via setup.py (using python2), the setup doesn't assert
the requirement of python3. Instead, it throws an error. This is fixed
by shifting one the 'from urllib.request import urlopen' import
statement, as it requires python3.

Fixes https://github.com/coala-analyzer/coala-bears/issues/223

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,6 @@ import sys
 from os.path import exists
 from shutil import copyfileobj
 from subprocess import call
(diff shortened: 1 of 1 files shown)


srisankethu:
@iamgroot42 You are separately commiting your every change! You need to update your previous commit. That's why you are having 3 commits.

Issue #37, opened 2016-02-21
annotation bear


sils1297:
_From @sils1297 on November 13, 2015 13:28_

a bear that takes the language definition and annotates each line with simple properties like "from here to there is a comment", "from here to there is a string" and so on. This is intended to be used by other bears for simpler analysis functionality

_Copied from original issue: coala-analyzer/coala#1040_

sils1297:
_From @AbdealiJK on December 12, 2015 5:45_

I think we may want a abstracted AST bear ?
Something that just creates ASTs (using pre existing packages) and let's each bear do things with them.

sils1297:
The idea with this one was more simple, a real universal AST isn't really viable so the idea was that we have some light annotations here and provide some graphs and other language independent representation at other places so some algorithms might use different abstractions. Or we could try making an extensible AST later.

srisankethu:
@sils1297 Can you assign me to this if you are not working on it

AbdealiJK:
@srisankethu this is a pretty complicated issue (and I have a feeling itll depend on coalang again)
I know that he isnt working on this atm, so Im assigning you

Please share your method of how you plan to do ti before attempting to code it though - as I think this may be a bit complicated

srisankethu:
@AbdealiJK Thank you

srisankethu:
How do I approach this issue(some idea)

AbdealiJK:
I'm not sure myself.
@sils1297 @Makman2 could you help out ?

sils1297:
@srisankethu I'd recommend you rather try to take on a difficulty/low bug like writing a simple bear or so, IIRC you're rather new to coala. One difficulty of this issue is that we can't give you an idea on how to approach it because we don't know. That's part of the task.

sils1297:
@abhsag24 might want to take this on since he's working on something related which actually might go well together with this idea

srisankethu:
@sils1297 Okk

sils1297:
Ok, we probably want:

1. A bear that creates sourceranges to determine which regions are comments
2. A bear that creates sourceranges to determine which regions are strings, depends on the comment marking
3. Other bears dependent on those that search for other useful stuff

So in the end if somebody only needs to know which are comments he just depends on 1 and we don't need to run the whole chain.

PR #227, opened 2016-03-12
AnnotationBear: Add AnnotationBear

abhsag24:
The Bear searches for strings and comments in a file
and yields them as results with the appropriate ranges
Also takes care of:
1. Comments within comments
2. Comments within strings and
3. Strings wihin comments

Fixes https://github.com/coala-analyzer/coala-bears/issues/37

### DIFF
--- a/bears/general/AnnotationBear.py
+++ b/bears/general/AnnotationBear.py
@@ -1,0 +1,129 @@
+from coalib.parsing.StringProcessing.Core import (unescaped_search_in_between,
+                                                  unescaped_search_for)
+from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.HiddenResult import HiddenResult
+from coalib.results.SourceRange import SourceRange
+from coalib.results.AbsolutePosition import AbsolutePosition
+
+
+class AnnotationBear(LocalBear):
+
+    def run(self, filename, file, language: str, language_family: str):
+        """
+        Finds out all the positions of strings and comments in a file.
+        The Bear searches for valid comments and strings and yields their
+        ranges as SourceRange objects in HiddenResults.
+
+        :param language:        Language to be whose annotations are to be
+                                searched.
+        :param language_family: Language family whose annotations are to be
+                                searched.
+        :return:                HiddenResults containing a dictionary with
+                                keys as 'strings' or 'comments' and values as
+                                a tuple of SourceRanges of strings and
+                                a tuple of SourceRanges of comments
+                                respectively.
+        """
+        lang_dict = LanguageDefinition(language, language_family)
+        # Strings
+        # TODO treat single-line and multiline strings differently
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        strings_found = self.find_with_start_end(filename, file, strings)
+        # multiline Comments
+        comments_found = self.find_with_start_end(
+                filename, file, dict(lang_dict["multiline_comment_delimiters"]))
+        # single-line Comments
+        comments_found.update(self.find_singleline_comments(
+                          filename, file, list(lang_dict["comment_delimiter"])))
+
+        matches_found = strings_found | comments_found
+        # Remove Nested
+        unnested_annotations = set(filter(
+                                       lambda arg: not starts_within_ranges(
+                                           arg, matches_found),
+                                       matches_found))
+        # Yield different results for strings and comments
+        strings_found = tuple(filter(lambda arg: arg in unnested_annotations,
+                                     strings_found))
+        comments_found = tuple(filter(lambda arg: arg in unnested_annotations,
+                                      comments_found))
+        yield HiddenResult(self, {'comments': comments_found,
+                                  'strings': strings_found})
+
+    @staticmethod
+    def find_with_start_end(filename, file, annot):
+        """
+        Gives all positions of annotations which have a start and end.
+
+        :param filename: Name of the file on which the bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param annot:    A dict containing start of annotation as key and end of
+                         annotation as value.
+        :return:         A set of SourceRanges giving the range of annotation.
+        """
+        text = ''.join(file)
+        found_pos = set()
+        for annot_type in annot:
+            found_pos.update(unescaped_search_in_between(
+                                           annot_type, annot[annot_type], text))
+        if found_pos:
+            found_pos = set(SourceRange.from_absolute_position(
+                                filename,
+                                AbsolutePosition(file, position.begin.range[0]),
+                                AbsolutePosition(
+                                    file, position.end.range[1] - 1))
+                            for position in found_pos)
+        return found_pos
+
+    @staticmethod
+    def find_singleline_comments(filename, file, comments):
+        """
+        Finds all single-line comments.
+
+        :param filename: Name of the file on which the bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param comments: A list containing different types of
+                         single-line comments.
+        :return:         A set of SourceRange objects with start as the
+                         beginning of the comment and end as
+                         the termination of line.
+        """
+        text = ''.join(file)
+        single_comments = set()
+        for comment_type in comments:
+            for found in unescaped_search_for(comment_type, text):
+                start = found.start()
+                end = text.find('\n', start)
+                end = len(text) - 1 if end == -1 else end
+                single_comments.add(SourceRange.from_absolute_position(
+                                        filename,
+                                        AbsolutePosition(file, start),
+                                        AbsolutePosition(file, end)))
+        return single_comments
+
+
+def starts_within_ranges(inside_range, outside_ranges):
+    """
+    Finds if a particular range starts inside a collection of given ranges.
+
+    :param outside_ranges: SourceRange identifying range to be searched.
+    :param inside_range:   A tuple SourceRange objects identifying ranges
+                           to be searched in.
+    :return:               True if inside_range is found inside any of the
+                           ranges given by outside_ranges, else
+                           False is returned.
+    """
+    for outside_range in outside_ranges:
+        if inside_range == outside_range:
+            continue
+        # Special case of python language.
+        # Where doc strings (""") and strings (") have a similar start.
+        elif inside_range.start == outside_range.start:
+            if inside_range.end < outside_range.end:
+                return True
+        elif (inside_range.start > outside_range.start and
+              inside_range.start <= outside_range.end):
+            return True
+    return False
--- a/tests/general/AnnotationBearTest.py
+++ b/tests/general/AnnotationBearTest.py
@@ -1,0 +1,116 @@
+from queue import Queue
+import unittest
+
+from bears.general.AnnotationBear import AnnotationBear
+from coalib.results.SourceRange import SourceRange
+from coalib.results.AbsolutePosition import AbsolutePosition
+from coalib. results.HiddenResult import HiddenResult
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+from tests.LocalBearTestHelper import execute_bear
+
+
+class AnnotationBearTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section1 = Section("")
+        self.section1.append(Setting('language', 'python3'))
+        self.section1.append(Setting('language_family', 'python3'))
+        self.python_uut = AnnotationBear(self.section1, Queue())
+        self.section2 = Section("")
+        self.section2.append(Setting('language', 'c'))
+        self.section2.append(Setting('language_family', 'c'))
+        self.c_uut = AnnotationBear(self.section2, Queue())
+
+    def test_single_line_string(self):
+        text = ["'from start till the end with #comments'\n", ]
+        compare = (SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(text, 0),
+                                    AbsolutePosition(text, len(text[0]) - 2)),)
+        with execute_bear(self.python_uut, "F", text) as result:
+            self.assertEqual(result[0].contents['strings'], compare)
+            self.assertEqual(result[0].contents['comments'], ())
+
+    def test_multiline_string(self):
+        text = ["'''multiline string, #comment within it'''\n"]
+        compare = (SourceRange.from_absolute_position(
+                        "F",
+                        AbsolutePosition(text, 0),
+                        AbsolutePosition(text, len(text[0])-2)),)
+        with execute_bear(self.python_uut, "F", text) as result:
+            self.assertEqual(result[0].contents['strings'], compare)
+            self.assertEqual(result[0].contents['comments'], ())
+
+    def test_single_line_comment(self):
+        text = ["some #coment with 'string'\n", "and next line"]
+        compare = (SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(text, text[0].find('#')),
+                                    AbsolutePosition(text, len(text[0]) - 1)),)
+        with execute_bear(self.python_uut, "F", text) as result:
+            self.assertEqual(result[0].contents['strings'], ())
+            self.assertEqual(result[0].contents['comments'], compare)
+
+    def test_multiline_comment(self):
+        text = ["some string /*within \n", "'multiline comment'*/"]
+        compare = (SourceRange.from_absolute_position(
+                            "F",
+                            AbsolutePosition(text, text[0].find('/*')),
+                            AbsolutePosition(text, len(''.join(text)) - 1)),)
+        with execute_bear(self.c_uut, "F", text) as result:
+            self.assertEqual(result[0].contents['strings'], ())
+            self.assertEqual(result[0].contents['comments'], compare)
+
+    def test_string_with_comments(self):
+        text = ["some #comment\n", "with 'string' in  next line"]
+        comment_start = text[0].find('#')
+        comment_end = len(text[0]) - 1
+        string_start = ''.join(text).find("'")
+        string_end = ''.join(text).find("'", string_start + 1)
+        compare = [(SourceRange.from_absolute_position(
+                                "F",
+                                AbsolutePosition(text, string_start),
+                                AbsolutePosition(text, string_end)),),
+                   (SourceRange.from_absolute_position(
+                                "F",
+                                AbsolutePosition(text, comment_start),
+                                AbsolutePosition(text, comment_end)),)]
+        with execute_bear(self.python_uut, "F", text) as result:
+            self.assertEqual(result[0].contents['strings'], compare[0])
+            self.assertEqual(result[0].contents['comments'], compare[1])
+
+    def test_combined_strings(self):
+        file_text = ['"some string #with comment"\n',
+                     '"""\n',
+                     "now a multiline string ''' <- this one not\n",
+                     '"""\n',
+                     '"""\n'
+                     'another comment # rather harmless\n',
+                     '"""\n']
+        string1_start = 0
+        string1_end = len(file_text[0]) - 2
+        string1 = SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(file_text, string1_start),
+                                    AbsolutePosition(file_text, string1_end))
+        string2_start = string1_end+2
+        text = ''.join(file_text)
+        string2_end = text.find('"""', string2_start + 1) + 2
+        #+2 for length of """
+        string2 = SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(file_text, string2_start),
+                                    AbsolutePosition(file_text, string2_end))
+        string3_start = text.find('"""', string2_end + 1)
+        string3_end = text.find('"""', string3_start + 1) + 2
+        #+2 for length of """
+        string3 = SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(file_text, string3_start),
+                                    AbsolutePosition(file_text, string3_end))
+        with execute_bear(self.python_uut, "F", file_text) as results:
+            self.assertIn(string1, results[0].contents['strings'])
+            self.assertIn(string2, results[0].contents['strings'])
+            self.assertIn(string3, results[0].contents['strings'])
+            self.assertEqual(results[0].contents['comments'], ())


AbdealiJK:
@abhsag24 can you please split the last commit into multiple commits ?

I think people (like me) are avoiding reviewing this because it's huge...

Maybe 1 function in 1 commit ? That way atleast the function will be reviewed and only the last commit adding `AnnotationBear` will be scrutinized more ?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        yield SourceRange.from_values(filename,
+                                      start_line=i[0][0],
+                                      start_column=i[0][1],
+                                      end_line=i[1][0],
+                                      end_column=i[1][1])
+
+
+def find_string_comment(file, annot, escape):
+    """
+    gives all instances of strings or multiline comments found within
```
AbdealiJK: Capital G

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        yield HiddenResult(self, s_str)
+
+
+def get_range(file, filename, match_pos, string_comment):
+    """
+    Checks if range is valid and then yields it.
+    """
+    search_range = []
+    for i in string_comment:
+        if i in match_pos:
```
AbdealiJK: This is another one of those ineffieicnt code techniques
If you ever have a `for` followed by `if` you should be using `sets`

`for i in set(string_comment) & set(match_pos):` is `O(len(string_comment)+len(match_pos))` while your earlier implementation if `O(len(string_comment)*len(match_pos))`

Note: I'm guessing here that the order doesnt matter and that you dont care of repetitive values.

If you may have repetitive values you can still use `set(match_pos)` as `in` operator of a set is O(1) and for a list it is O(N)

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        annot_dict['comment']["multiline"].update(
+                    lang_dict["multiline_comment_delimiters"])
+
+        annot_dict['string']['singleline'].update(
+                                            lang_dict["string_delimiters"])
+        annot_dict['string']['multiline'].update(
+                                       lang_dict["multiline_string_delimiters"])
+        multi_str = single_str = multi_comm = ()
+        search_dict = annot_dict['string']['multiline']
+        multi_str = find_string_comment(file, search_dict, True)
```
AbdealiJK: could you use `escape=True` for booleans ? It gets confusing to read `True` as a third param (and checking the function later)
`escape` gives it context of what that bool really means

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+
+def get_range(file, filename, match_pos, string_comment):
+    """
+    Checks if range is valid and then yields it.
+    """
+    search_range = []
+    for i in string_comment:
+        if i in match_pos:
+            search_range.append(i)
+    search_range = [(calc_line_col(file, start), calc_line_col(file, end))
```
AbdealiJK: Why another for loop ? Why not just do this in the above for loop ?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+    gives all instances of strings or multiline comments found within
+    the file, even if they are nested in other strings or comments
+    """
+    if escape:
+        search_func = unescaped_search_in_between
+    else:
+        search_func = search_in_between
+    found_pos = ()
+    for annot_type in annot:
+        found_pos += tuple(search_func(annot_type, annot[annot_type], file))
```
AbdealiJK: Wow, I did not know you could "Add" to a tuple. Cool :)

PS: `found_pos = tuple(search_func(annot_type, annot[annot_type], file) for annot_type in annot)`
would probably be faster though. Generators FTW

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    single_comm = []
+    for comment_type in annot:
+        for i in search_for(comment_type, file):
+            if not in_range_list(match_pos, (i.start(), i.end())):
+                end = file.find('\n', i.start())
+                if end != -1:
+                    single_comm.append((i.start(), end))
+                else:
+                    end = len(file) + 1
```
AbdealiJK: `end = len(file) + 1 if end == -1 else end` (readability)

AbdealiJK: And code duplication (of the append line :P)

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+                    single_comm.append((i.start(), end))
+    return tuple(single_comm)
+
+
+def in_range_list(outside_range_list, inside_range):
+    """
+    finds if a given 'range' is inside a any of the 'ranges' in
```
AbdealiJK: capital

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+def calc_line_col(file, pos_to_find):
+    """
+    Calculate line number and column in the file, from position.
+    """
+    line = 1
+    pos = -1
+    pos_new_line = file.find('\n')
+    while True:
+        if pos_new_line == -1:
+            return (line, pos_to_find-pos)
```
AbdealiJK: space around `-`

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        elif inside_range[0] == outside_range[0]:
+            if inside_range[1] < outside_range[1]:
+                return True
+
+        elif inside_range[0] in range(outside_range[0], outside_range[1]):
+            return True
+    return False
+
+
+def calc_line_col(file, pos_to_find):
```
AbdealiJK: This function seems super useful. I remember @sudheesh001 struggling with this in `ESLintBear`.
Can we possible move it into SourceRange or something ? so that bear users can use it ?

I think someone else also needed it for a bear recently.

Currently they seem to be doing "".join(file) but this seems to be better in terms of performance

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        Finds out all the positions of comments and strings in a file.
+        The Bear searches for valid comments and strings
+        and yields them as HiddenResults with the appropriate ranges.
+        Invalid comments or strings are:
+        1. Comments within comments,
+        2. Comments within strings and
+        3. Strings within comments.
+        """
+        file = ''.join(file)
```
AbdealiJK: Never try to change `file`.
I think we may move it into a proxy later. Ideally a bear shouldn't change `file`
Can you rename this ? Maybe `text` or `lines`

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+
+        else:
+            line += 1
+            pos = pos_new_line
+            pos_new_line = file.find('\n', pos_new_line + 1)
+
+
+def remove_nested(match_pos):
+    """
+    removes all the entries from a listthat are nested
```
AbdealiJK: capital

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+    inside some other entry of that list.
+    """
+    unnested_list = []
+    for i in match_pos:
+        if not in_range_list(match_pos, i):
+            unnested_list.append(i)
```
AbdealiJK: `tuple(filter(lambda x: in_range_list(match_pos, x), match_pos))` ?
Does this need a new function ?
It's only used once

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+    Checks if range is valid and then yields it.
+    """
+    search_range = []
+    for i in string_comment:
+        if i in match_pos:
+            search_range.append(i)
+    search_range = [(calc_line_col(file, start), calc_line_col(file, end))
+                    for (start, end) in search_range]
+
+    for i in search_range:
```
AbdealiJK: sils doesnt like variables with 1 letter (although I dont mind them)

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        The Bear searches for valid comments and strings
+        and yields them as HiddenResults with the appropriate ranges.
+        Invalid comments or strings are:
+        1. Comments within comments,
+        2. Comments within strings and
+        3. Strings within comments.
+        """
+        file = ''.join(file)
+        annot_dict = {}
+        annot_dict['comment'] = {}
```
AbdealiJK: Why not
```
annot_dict = {
    'comment': {'multiline': {}, 'singleline': {}},
    'string': {'multiline': {}, 'singleline': {}}}
```
seems SO much easier to read

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+                                       lang_dict["multiline_string_delimiters"])
+        multi_str = single_str = multi_comm = ()
+        search_dict = annot_dict['string']['multiline']
+        multi_str = find_string_comment(file, search_dict, True)
+        search_dict = annot_dict['string']['singleline']
+        single_str = find_string_comment(file, search_dict, True)
+        search_dict = annot_dict['comment']["multiline"]
+        multi_comm = find_string_comment(file, search_dict, False)
+
+        match_pos += single_str + multi_str + multi_comm
```
AbdealiJK: You could remove `match_pos = ()` above and just create it here ?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+                                                  unescaped_search_in_between)
+from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
+from coalib.results.SourceRange import SourceRange
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.HiddenResult import HiddenResult
+
+
+class AnnotationBear(LocalBear):
+
+    def run(self, filename, file: str, language: str, language_family: str):
```
AbdealiJK: `file: str` is invalid because we give a tuple of strings ?

Also `:param:` for language, family needed

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        Finds out all the positions of comments and strings in a file.
+        The Bear searches for valid comments and strings
+        and yields them as HiddenResults with the appropriate ranges.
+        Invalid comments or strings are:
+        1. Comments within comments,
```
AbdealiJK: This doesnt seem to be rst compliant.
Have you checked it ?

AbdealiJK: i think you need a newline above this

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+                    lang_dict["multiline_comment_delimiters"])
+
+        annot_dict['string']['singleline'].update(
+                                            lang_dict["string_delimiters"])
+        annot_dict['string']['multiline'].update(
+                                       lang_dict["multiline_string_delimiters"])
+        multi_str = single_str = multi_comm = ()
+        search_dict = annot_dict['string']['multiline']
+        multi_str = find_string_comment(file, search_dict, True)
+        search_dict = annot_dict['string']['singleline']
```
AbdealiJK: The variable `search_dict` seems completely unnecessary and is confusing because it keeps changing and is only used once before it gets changed.

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+                    lang_dict["comment_delimiter"])
+
+        annot_dict['comment']["multiline"].update(
+                    lang_dict["multiline_comment_delimiters"])
+
+        annot_dict['string']['singleline'].update(
+                                            lang_dict["string_delimiters"])
+        annot_dict['string']['multiline'].update(
+                                       lang_dict["multiline_string_delimiters"])
+        multi_str = single_str = multi_comm = ()
```
AbdealiJK: This is not C ... you dont need to initialize variables like this

bears/general/AnnotationBear.py
AbdealiJK: In fact do you really need this dict ? 
Seems like you just asign the values and use those values later. variables are a better choice ...
You could use `singleline_comment` rather than `single_comm` for easier readability

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,172 @@
...
+        single_comm = find_singleline_comments(file, search_dict, match_pos)
+        match_pos += single_comm
+        match_pos = remove_nested(match_pos)
+        m_comm = s_comm = m_str = s_str = []
+        if match_pos:
+            m_comm = list(get_range(file, filename, match_pos, multi_comm))
+            s_comm = list(get_range(file, filename, match_pos, single_comm))
+            m_str = list(get_range(file, filename, match_pos, multi_str))
+            s_str = list(get_range(file, filename, match_pos, single_str))
+        yield HiddenResult(self, m_comm)
```
AbdealiJK: This can reduce repetition:

```
if match_pos:
    for type in [multi_comm, single_comm, multi_str, single_str]:
        yield HiddenResult(self, list(get_range(file, filename, match_pos, type))
```
PS: I dont really know what these are yet ... so, you should use a better name that `type`.  Just used that as a dummy

bears/general/AnnotationBear.py
AbdealiJK: space between list and that

AbdealiJK:
@abhsag24 Ive commented on code style issues at https://github.com/coala-analyzer/coala-bears/commit/b1fd4359aa7964e2b5653ed26990bf7025669e7d

I have not checked the logic yet ... Could you please fix those ? It seems pretty difficult to read right now

bears/general/AnnotationBear.py
abhsag24: this was super cool! thanks :)

bears/general/AnnotationBear.py
abhsag24: i didn't know this! thanks :)

bears/general/AnnotationBear.py
abhsag24: @AbdealiJK this tuple would just have generator objects, although in my implementation those are "generated" inside the tuple.

bears/general/AnnotationBear.py
abhsag24: actually did this so if nothing is found an empty tuple is yielded, it should never be the case that nothing gets yielded

bears/general/AnnotationBear.py
abhsag24: Maybe we can have an issue later, of moving it out?

bears/general/AnnotationBear.py
Makman2: imo it's really better to provide a meaningful name, one-letter-names are reserved for indices (at least for me :P)

abhsag24: its gone in my next implementation FWIW :P

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,72 @@
+from coalib.parsing.StringProcessing.Core import (search_for,
+                                                  search_in_between,
+                                                  unescaped_search_in_between)
+from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.HiddenResult import HiddenResult
+
+
+class AnnotationBear(LocalBear):
```
sils1297: commit message, rather write `Add AnnotationBear` in the shortlog and some explanation why it's cool and useful in the body

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,72 @@
...
+        strings within comments.
+
+        :param language:        Language to be whose annotations are to be
+                                searched.
+        :param language_family: Language family whose annotations are to be
+                                searche.
+
+        :return:                HiddenResult with content conatining
+                                a tuple of tuples of the form
+                                (annotation_start, annotation_end).
```
sils1297: how should I know if I consume this if it's a string or a comment?

abhsag24: @sils1297  i've added that in another commit

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+
+        text = ''.join(file)
+        lang_dict = LanguageDefinition(language, language_family)
+
+        # Strings
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        matches_found = find_multiline(text, strings, escape=True)
```
sils1297: don't you have to use find single line for single line strings?

abhsag24: @sils1297  the thing is, there's no such thing as a single line string right? i mean people can add \ and extend the string onto another line right?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,72 @@
...
+        found_pos += tuple(search_func(annot_type, annot[annot_type], text))
+    if found_pos:
+        found_pos = tuple((i.begin.range[0], i.end.range[1])
+                          for i in found_pos)
+    return found_pos
+
+
+def find_singleline(text, annot):
+    """
+    Finds all single-line comments.
```
sils1297: param docs missing

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,72 @@
...
+        # single-line Comments
+        matches_found += find_singleline(text,
+                                         dict(lang_dict["comment_delimiter"]))
+        yield HiddenResult(self, matches_found)
+
+
+def find_multiline(text, annot, escape=False):
+    """
+    Gives all instances of strings or multiline comments found within
+    the file, even if they are nested in other strings or comments.
```
sils1297: params doc missing

bears/general/AnnotationBear.py
```diff
@@ -11,8 +11,14 @@ class AnnotationBear(LocalBear):
...
-        This does not check however for comments within strings and
-        strings within comments.
+        The Bear searches for valid comments and strings and yields their
+        positions as HiddenResults.
+        Invalid comments or strings are:
+
+        1. Comments within comments,
+        2. Comments within strings,
+        3. Strings within comments and
+        4. Strings within strings.
```
sils1297: ahm this is internal and shouldn't concern the user right?

abhsag24: ahh yes thanks :)

bears/general/AnnotationBear.py
sils1297: but it belongs in this commit!

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-03-20 17:42 GMT+01:00 Abhay Raizada <notifications@github.com>:

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,100 @@
...
+    :return:            a tuple of the form (line, column).
+    """
+    line = 1
+    pos = -1
+    pos_new_line = text.find('\n')
+    while True:
+        if pos_new_line == -1:
+            return (line, pos_to_find - pos)
+
+        if pos_to_find <= pos_new_line:
```
AbdealiJK: elif maybe ? (Makes it a tiny bit faster ;) )

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,100 @@
...
+            return True
+    return False
+
+
+def calc_line_col(text, pos_to_find):
+    """
+    Calculate line number and column in the file, from position.
+
+    :param text:        A string with all the contents of file.
+    :param pos_to_find: position of character to be found in the
```
AbdealiJK: capital P

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,100 @@
...
+                        line,column form.
```
AbdealiJK: space after comma ? (Maybe add brackets to indicate it like a tuple)
Also - and this is important - mention whether your lines start from 0 or 1 and whether your cols start from 0 or 1

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,100 @@
...
+        elif inside_range[0] in range(outside_range[0], outside_range[1]):
+            return True
+    return False
+
+
+def calc_line_col(text, pos_to_find):
+    """
+    Calculate line number and column in the file, from position.
+
+    :param text:        A string with all the contents of file.
```
AbdealiJK: Mention: With `\n` as the newline character. If `\r\n` is given, your calculations will be wrong.

Also, does this counting approach work for unicode characters too ? We should add a testcase for that.

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def find_singleline_comments(text, comments):
+    """
+    Finds all single-line comments.
+
+    :param text:      A string with all the contents of file.
+    :param comments:  A list contatining different types of
+                      single-line comments.
+
```
AbdealiJK: unnecessary newline

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,63 @@
...
+        text = "we have a #comment here\n #and here, this one spans"
+        "upto end of text"
+        self.assertEqual(find_singleline_comments(text, ["#"]),
+                         ((10, 23), (25, 51)))
+
+    def test_in_ranges(self):
+        # test not in any range
+        self.assertFalse(in_ranges(((5, 10), (20, 30), (40, 50)), (11, 19)))
+
+        #test in range
```
AbdealiJK: space

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertTrue(in_ranges(((5, 10), (20, 30), (40, 50)), (46, 54)))
```
AbdealiJK: After looking at this test case this seems to be a bad name. I think your function name should be `is_overlapping` or something. Because "in" means subset, not overlapping...

intuitively:
"abc" in "xyzab" is not True
"abc" in "xyzabc" is True

AbdealiJK: I'm a little confused now ... (sorry)
L52 says if 11 is not in the ranges, it is False. But this line says even if 53 is not in the ranges it is true ?

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_calc_line_col(self):
+        # no newlines
+        text = "find position of 'z'"
+        z_pos = text.find('z')
+        self.assertEqual(calc_line_col(text, z_pos), (1, z_pos+1))
+
+        # newline
+        text = "find position of\n'z'"
+        z_pos = text.find('z')
```
AbdealiJK: Can you add more test cases with out `Constants.COMPLEX_TEST_STRING` ? Always good to check for weird chars.

Also, can you check with raw strings like `r'a\b'` where searching for 'b' should give line 1 col 3 I think ?
`r'a\n'` where searching for `n` should give L1 C3
`'a\\n'` where searching for `n` should give L1 C3

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_find_single_line_comments(self):
+        text = "we have a comment #here and it spans upto \n end of line"
+        self.assertEqual(find_singleline_comments(text, ["#"]), ((18, 42),))
+
+        text = "we have a comment #here and it spans upto end of text"
+        self.assertEqual(find_singleline_comments(text, ["#"]), ((18, 54),))
+
+        text = "we have a #comment here\n #and here, this one spans"
+        "upto end of text"
```
AbdealiJK: indentation ...
In fact this text seems wrong.
"upto end of text" isnt saved into `text` as python will truncate the last sentence.
Are you sure youre testing correctly here ?

Again, try adding `Constants.COMPLEX_TEST_STRING` to the tests ?

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+        """
+        text = ''.join(file)
+        lang_dict = LanguageDefinition(language, language_family)
+
+        # Strings
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        strings_found = find_start_end(text, strings, escape=True)
+        # multiline Comments
+        comments_found = find_start_end(
```
AbdealiJK: Note that this is also fine:
```
comments_found = find_start_end(
    text, dict(lang_dict["multiline_comment_delimiters"]))
```

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+
+        :return:                 SourceRange Objects of all the valid positions
+                                 of the annotation.
+            """
+        for position in set(matches_found) & set(annotation_found):
+            yield SourceRange.from_values(
+                    filename,
+                    start_line=calc_line_col(text,
+                                             position[0])[0],
+                    start_column=calc_line_col(text,
```
AbdealiJK: something like:

```
kwargs = {}
(kwargs['start_line'],
 kwargs['start_column']) = calc_line_col(text, position[0])
(kwargs['end_line'],
 kwargs['end_column']) = calc_line_col(text, position[1])
# and using **kwargs
```
would be easier to read and faster !
you are actually parsing the file 4 times while you can do it with only 2 times. (Call `calc_line_col` as less as possible)

In fact because of this, I would encourage you to modify the `calc_line_col` to take in a list of positions, then parse the file and return a list of tuples of (line, col)
You can assume they're sorted (and Raise an exception if not) or sort them internally. This would increase efficiency.

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+                                text,
+                                dict(lang_dict["multiline_comment_delimiters"]))
+        # single-line Comments
+        comments_found += find_singleline_comments(
+                                          text,
+                                          list(lang_dict["comment_delimiter"]))
+
+        matches_found = strings_found + comments_found
+        # Remove Nested
+        matches_found = tuple(filter(
```
AbdealiJK: why not just make this a `set` here itself ?

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+                              lambda arg: not in_ranges(matches_found, arg),
+                              matches_found))
+        # Yield results
+        for annotation_found in [strings_found, comments_found]:
+            position_range = tuple(AnnotationBear.get_range(text,
+                                                            filename,
+                                                            matches_found,
+                                                            annotation_found))
+            yield HiddenResult(self, position_range)
```
AbdealiJK: So, we have 1 result with a lot of ranges - is this what we want (I'm not sure what's the design here actually)
Do we want some more metadata ? like mentioning the fact that this is comments and the other is strings ?

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
                                                   unescaped_search_in_between)
+from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.HiddenResult import HiddenResult
+from coalib.results.SourceRange import SourceRange
+
+
+class AnnotationBear(LocalBear):
+
+    def run(self, filename, file, language: str, language_family: str):
```
AbdealiJK: This looks soo much cleaner :)

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+        """
+        Finds out all the positions of strings and comments in a file.
+        The Bear searches for valid comments and strings and yields their
+        ranges as SourceRange objects in HiddenResults.
+
+        :param language:        Language to be whose annotations are to be
+                                searched.
+        :param language_family: Language family whose annotations are to be
+                                searche.
```
AbdealiJK: searched *

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+        :return:                HiddenResult with content conatining
```
AbdealiJK: containing *

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+                                a tuple of tuples of the form
+                                (annotation_start, annotation_end).
+        """
+        text = ''.join(file)
+        lang_dict = LanguageDefinition(language, language_family)
+
+        # Strings
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        strings_found = find_start_end(text, strings, escape=True)
```
AbdealiJK: Why don't we directly return a SourceRange/SourcePosition or TextRange/TextPosition directly ? ?
We already have functions like `overlaps()` in SourceRange and TextRange and that can be used directly ?

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+        yields position of valid annotations as SourceRange objects.
+
+        :param text:             A string with all the contents of file.
+        :param filename:         Name of file.
+        :param matches_found:    All the valid strings and comments found.
+        :param annotation_found: All the positions of the given annotation.
+
+        :return:                 SourceRange Objects of all the valid positions
+                                 of the annotation.
+            """
```
AbdealiJK: formatting

bears/general/AnnotationBear.py
```diff
@@ -1,6 +1,81 @@
...
+    @classmethod
+    def get_range(cls, text, filename, matches_found, annotation_found):
+        """
+        yields position of valid annotations as SourceRange objects.
+
+        :param text:             A string with all the contents of file.
+        :param filename:         Name of file.
+        :param matches_found:    All the valid strings and comments found.
+        :param annotation_found: All the positions of the given annotation.
+
```
AbdealiJK: unnecessary newline

tests/general/AnnotationBearTest.py
abhsag24: @AbdealiJK  it was meant to confuse, if the start of `inside_range` is inside one of the other ranges, it didn't actually start right? so it returns true , `is_overlapping` isn't fair since if the end of `inside_range` overalps then it shouldn't be a problem

bears/general/AnnotationBear.py
abhsag24: i'm not sure at might not work actually

tests/general/AnnotationBearTest.py
AbdealiJK: Ohhhh, that makes sense in this context of comments inside comments.
Hmm, I think you ought to mention this properly in the docstring. It sounded like a very generic function at the beginning. But turns out it is custom logic for this specific scenario.

bears/general/AnnotationBear.py
AbdealiJK: Then we should mention this in the docstring and throw exceptions if strings with unicode is given OR fix it to work :)

bears/general/AnnotationBear.py
abhsag24: maybe we can add it in the docs that first result is for the strings and the other for comments

bears/general/AnnotationBear.py
sils1297: I disagree, there's a return above there shouldn't be an else

bears/general/AnnotationBear.py
abhsag24: thanks :)

tests/general/AnnotationBearTest.py
abhsag24: this wasn't correct for sure, sorry , i will add the COMPLEX_TEST_STRING though

bears/general/AnnotationBear.py
abhsag24: UPDATE: It does work! python is awesome! i'll add tests for it :P

abhsag24: UPDATE: It does work! python is awesome! i'll add tests for it :P

tests/general/AnnotationBearTest.py
abhsag24: works like a charm

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def setUp(self):
+        self.section = Section("")
+        self.section.append(Setting('language', 'c'))
+        self.section.append(Setting('language_family', 'c'))
+
+    def assertRange(self, file, uut, sourcerange):
+        with execute_bear(uut, "filename", file) as results:
+            print("results", results)
+            for result in results:
```
abhsag24: this line checks that no other results are yielded, it was custom for this bear, it could be made generic by including the result in the param

sils1297:
@abhsag24 any progress?

abhsag24:
@sils1297  its blocked , or not actually blocked, but can be improved when #1947 gets finished

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        :param language:        Language to be whose annotations are to be
+                                searched.
+        :param language_family: Language family whose annotations are to be
+                                searched.
+        :return:                Two HiddenResults first containing a tuple of
+                                SourceRanges of strings and the second
+                                containing a tuple of SourceRanges of comments.
+        """
+        lang_dict = LanguageDefinition(language, language_family)
+        print(file)
```
sils1297: forgot to remove?

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+from tests.LocalBearTestHelper import execute_bear
+
+
+class AnnotationBearTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section1 = Section("")
+        self.section1.append(Setting('language', 'python3'))
+        self.section1.append(Setting('language_family', 'python3'))
+        self.uut1 = AnnotationBear(self.section1, Queue())
```
sils1297: rather name it `python_uut` and `c_uut` that's way clearer to understand

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+                                "F",
+                                AbsolutePosition(text, string_start),
+                                AbsolutePosition(text, string_end)),),
+                   (SourceRange.from_absolute_position(
+                                "F",
+                                AbsolutePosition(text, comment_start),
+                                AbsolutePosition(text, comment_end)),)]
+        with execute_bear(self.uut1, "F", text) as result:
+            self.assertEqual(result[0].contents, compare[0])
+            self.assertEqual(result[1].contents, compare[1])
```
sils1297: can we have something like

```
"some string #with comment" '''
now a multiline string """ <- this one not
'''

"""
another comment # rather harmless
"""
```

(python)

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Gives all positions of annotations which have a start and end.
+
+        :param filename: Name of the file on which bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param annot:    A dict containing start of annotation as key and end of
+                         annotation as value.
+        :param escape:   Variable to check wether to ignore escaped annotations.
+
+        :return:         A tuple of SourceRanges giving the range of annotation.
```
sils1297: why not a set?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        :return:                Two HiddenResults first containing a tuple of
+                                SourceRanges of strings and the second
+                                containing a tuple of SourceRanges of comments.
+        """
+        lang_dict = LanguageDefinition(language, language_family)
+        print(file)
+        # Strings
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        strings_found = self.find_with_start_end(
```
Adrianzatreanu: this coding style in the brackets is weird, do we not use each argument on one line when the arguments are over 80 chars?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        text = ''.join(file)
+        if escape:
+            search_func = unescaped_search_in_between
+        else:
+            search_func = search_in_between
+        found_pos = ()
+        for annot_type in annot:
+            found_pos += tuple(search_func(annot_type, annot[annot_type], text))
+        if found_pos:
+            found_pos = tuple(SourceRange.from_absolute_position(
```
Adrianzatreanu: again, the indentation here is quite weird

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+                                SourceRanges of strings and the second
+                                containing a tuple of SourceRanges of comments.
+        """
+        lang_dict = LanguageDefinition(language, language_family)
+        print(file)
+        # Strings
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        strings_found = self.find_with_start_end(
+                                    filename, file, strings, escape=True)
```
sils1297: do I see this right that you treat string delimiters just like multiline string delimiters?

tests/general/AnnotationBearTest.py
sils1297: also a test with

```
a ' test
a ' test
```

should not yield anything
```

File: bears/general/AnnotationBear.py [hunk 54]
```diff
@@ -0,0 +1,134 @@
...
+                         commentand end as end of line.
+        """
+        text = ''.join(file)
+        single_comments = []
+        for comment_type in comments:
+            for found in search_for(comment_type, text):
+                start = found.start()
+                end = text.find('\n', start)
+                end = len(text) - 1 if end == -1 else end
+                single_comments.append(SourceRange.from_absolute_position(
```
Adrianzatreanu: weird coding style here too

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        comments_found = tuple(filter(lambda arg: arg in matches_found,
+                                      comments_found))
+        yield HiddenResult(self, comments_found)
+
+    @staticmethod
+    def find_with_start_end(filename, file, annot, escape=False):
+        """
+        Gives all positions of annotations which have a start and end.
+
+        :param filename: Name of the file on which bear is running.
```
Adrianzatreanu: ``on which **the** bear is running``

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        # multiline Comments
+        comments_found = self.find_with_start_end(
+                filename, file, dict(lang_dict["multiline_comment_delimiters"]))
+        # single-line Comments
+        comments_found += self.find_singleline_comments(
+                           filename, file, list(lang_dict["comment_delimiter"]))
+
+        matches_found = strings_found + comments_found
+        # Remove Nested
+        matches_found = tuple(filter(
```
sils1297: maybe a better name? Not that I can think of one spontaneously :/

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+    @staticmethod
+    def find_with_start_end(filename, file, annot, escape=False):
+        """
+        Gives all positions of annotations which have a start and end.
+
+        :param filename: Name of the file on which bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param annot:    A dict containing start of annotation as key and end of
+                         annotation as value.
+        :param escape:   Variable to check wether to ignore escaped annotations.
```
Adrianzatreanu: whether*

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+        self.section1.append(Setting('language_family', 'python3'))
+        self.uut1 = AnnotationBear(self.section1, Queue())
+        self.section2 = Section("")
+        self.section2.append(Setting('language', 'c'))
+        self.section2.append(Setting('language_family', 'c'))
+        self.uut2 = AnnotationBear(self.section2, Queue())
+
+    def test_single_line_string(self):
+        text = ["'from start till the end with #comments'\n", ]
+        compare = (SourceRange.from_absolute_position(
```
Adrianzatreanu: coding style

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+                                AbsolutePosition(text, text[0].find('/*')),
+                                AbsolutePosition(text, len(''.join(text))-1)),)
+        with execute_bear(self.uut2, "F", text) as result:
+            self.assertEqual(result[0].contents, ())
+            self.assertEqual(result[1].contents, compare)
+
+    def test_string_with_comments(self):
+        text = ["some #comment\n", "with 'string' in  next line"]
+        comment_start = text[0].find('#')
+        comment_end = len(text[0])-1
```
Adrianzatreanu: shouldn't place space between operators? ``len(text[0]) - 1``

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+        with execute_bear(self.uut1, "F", text) as result:
+            self.assertEqual(result[0].contents, compare)
+            self.assertEqual(result[1].contents, ())
+
+    def test_single_line_comment(self):
+        text = ["some #coment with 'string'\n", "and next line"]
+        compare = (SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(text, text[0].find('#')),
+                                    AbsolutePosition(text, len(text[0])-1)),)
```
Adrianzatreanu: space here, ``len(text[0] - 1)``

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+        with execute_bear(self.uut1, "F", text) as result:
+            self.assertEqual(result[0].contents, ())
+            self.assertEqual(result[1].contents, compare)
+
+    def test_multiline_comment(self):
+        text = ["some string /*within \n", "'multiline comment'*/"]
+        compare = (SourceRange.from_absolute_position(
+                                "F",
+                                AbsolutePosition(text, text[0].find('/*')),
+                                AbsolutePosition(text, len(''.join(text))-1)),)
```
Adrianzatreanu: space here too, near operator

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+                                AbsolutePosition(file, position.end.range[1]-1))
+                              for position in found_pos)
+        return found_pos
+
+    @staticmethod
+    def find_singleline_comments(filename, file, comments):
+        """
+        Finds all single-line comments.
+
+        :param filename: Name of the file on which bear is running.
```
Adrianzatreanu: the bear

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param comments: A list contatining different types of
```
Adrianzatreanu: containing*

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+                         single-line comments.
+        :return:         A tuplle of SourceRange objects with start as start of
```
Adrianzatreanu: tuple*

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+                         commentand end as end of line.
```
Adrianzatreanu: space between comment & and

bears/general/AnnotationBear.py
Adrianzatreanu: you sure you can't formulate this a little better? feels insecure with ``start as start`` & ``end as end``

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+    def find_with_start_end(filename, file, annot, escape=False):
+        """
+        Gives all positions of annotations which have a start and end.
+
+        :param filename: Name of the file on which bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param annot:    A dict containing start of annotation as key and end of
+                         annotation as value.
+        :param escape:   Variable to check wether to ignore escaped annotations.
+
```
Adrianzatreanu: why is here a newline? i've checked later and you don't place a newline before return usually :)

tests/general/AnnotationBearTest.py
```diff
@@ -0,0 +1,81 @@
...
+        self.section2.append(Setting('language', 'c'))
+        self.section2.append(Setting('language_family', 'c'))
+        self.uut2 = AnnotationBear(self.section2, Queue())
+
+    def test_single_line_string(self):
+        text = ["'from start till the end with #comments'\n", ]
+        compare = (SourceRange.from_absolute_position(
+                                    "F",
+                                    AbsolutePosition(text, 0),
+                                    AbsolutePosition(text, len(text[0])-2)),)
```
Adrianzatreanu: spaces near operator

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        else:
+            search_func = search_in_between
+        found_pos = ()
+        for annot_type in annot:
+            found_pos += tuple(search_func(annot_type, annot[annot_type], text))
+        if found_pos:
+            found_pos = tuple(SourceRange.from_absolute_position(
+                                filename,
+                                AbsolutePosition(file, position.begin.range[0]),
+                                AbsolutePosition(file, position.end.range[1]-1))
```
Adrianzatreanu: spaces near the operator, ``position.end.range[1] - 1``

bears/general/AnnotationBear.py
abhsag24: aahh yess sorry

bears/general/AnnotationBear.py
abhsag24: i do!, but one can use strings as multiline strings just by escaping using `\` right?

tests/general/AnnotationBearTest.py
abhsag24: sure :)

bears/general/AnnotationBear.py
abhsag24: `annotations_found` ?

sils1297: so this is about nested occurrences, right? Nested should be in there

sils1297:
ping @abhsag24

abhsag24:
@sils1297 was travelling, home now, will push a review-able version by tonight or tomorrow morning

bears/general/AnnotationBear.py
abhsag24: not sure, coala doesn't seem to have a problem, and i think it's a waste of lines if you can have all the other arguments in a line cleanly. I'll happily change this if you insist though :P

bears/general/AnnotationBear.py
abhsag24: any suggestion?

tests/general/AnnotationBearTest.py
abhsag24: @Adrianzatreanu i have to use hanging indents here othere wise the params go way out of max line length , do you have any other suggestions?

bears/general/AnnotationBear.py
Adrianzatreanu: i don't insist. i was trying to give you suggestions. if there's no problem on it from PEP, you can leave it any way you want.

bears/general/AnnotationBear.py
Adrianzatreanu: yes, maybe try to allign them all after ``tuple(`` ? or anything like it, because right now it shows to me as if they are indented randomly :)

bears/general/AnnotationBear.py
Adrianzatreanu: same here, you might want to try to place the indendation on next lines alligned with ``append(`` ? it's just a suggestion, feel free to ignore it if it seems awkward.

tests/general/AnnotationBearTest.py
Adrianzatreanu: yes, same as earlier, maybe make the indendation on the other lines start from ``= (`` ? :D

bears/general/AnnotationBear.py
abhsag24: yes, thanks :)

bears/general/AnnotationBear.py
abhsag24: i did that so there is a distinction between the arguments and the for statement that follows that _needs_ to be indented at that spot

bears/general/AnnotationBear.py
abhsag24: if i do it after append it seems like these parameters are parameters for append while they are actually for `from_absolute_position` same as with other things

tests/general/AnnotationBearTest.py
abhsag24: #2 will fail right now because of the same treatment of single line and multiline strings, i have added a #TODO for now

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+    def find_singleline_comments(filename, file, comments):
+        """
+        Finds all single-line comments.
+
+        :param filename: Name of the file on which the bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param comments: A list containing different types of
+                         single-line comments.
+        :return:         A set of SourceRange objects with start as the
+                         begining of comment and end as the termination of line.
```
Adrianzatreanu: beginning***
and also "of the comment"

Adrianzatreanu: i love this, now it is written way better :)

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+                         annotation as value.
+        :param escape:   Variable to check whether to ignore escaped
+                         annotations.
+        :return:         A set of SourceRanges giving the range of annotation.
+        """
+        text = ''.join(file)
+        if escape:
+            search_func = unescaped_search_in_between
+        else:
+            search_func = search_in_between
```
sils1297: shorten: `search_func = unescaped_search_in_between if escape else search_in_between`

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        The Bear searches for valid comments and strings and yields their
+        ranges as SourceRange objects in HiddenResults.
+
+        :param language:        Language to be whose annotations are to be
+                                searched.
+        :param language_family: Language family whose annotations are to be
+                                searched.
+        :return:                Two HiddenResults first containing a tuple of
+                                SourceRanges of strings and the second
+                                containing a tuple of SourceRanges of comments.
```
sils1297: I'd feel better with a dict with proper keys like "comments" and "strings" as that can be extended easily by the next stage of bears building on this :)

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+        strings = dict(lang_dict["string_delimiters"])
+        strings.update(lang_dict["multiline_string_delimiters"])
+        strings_found = self.find_with_start_end(
+                                    filename, file, strings, escape=True)
+        # multiline Comments
+        comments_found = self.find_with_start_end(
+                filename, file, dict(lang_dict["multiline_comment_delimiters"]))
+        # single-line Comments
+        comments_found.update(self.find_singleline_comments(
+                          filename, file, list(lang_dict["comment_delimiter"])))
```
sils1297: why no escape for comments?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,134 @@
...
+                end = text.find('\n', start)
+                end = len(text) - 1 if end == -1 else end
+                single_comments.append(SourceRange.from_absolute_position(
+                                                  filename,
+                                                  AbsolutePosition(file, start),
+                                                  AbsolutePosition(file, end)))
+        return set(single_comments)
+
+
+def starts_within_range(outside_ranges, inside_range):
```
sils1297: starts_within_range*s*

sils1297: I'd reverse arguments

sils1297:
like it! https://github.com/coala-analyzer/coala-bears/pull/227/commits/8cf3e03deb89d17ef7975e3890c5ae248fed4c8a needs work

sils1297:
I think we can merge this soon :)

bears/general/AnnotationBear.py
abhsag24: that is awesome! i dont why i didnt think about it in the first place!

bears/general/AnnotationBear.py
abhsag24: i didn't see any need for it, i mean people escape `'` when they mean it as a quote, why would they escape comments? and why would the comment need escaping in the first place? i'd change it if it's necessary and just have everything as an `unescaped_search`

sils1297: not that I know a concrete example but better be safe here IMO

abhsag24: i've used `unescaped_search` everywhere :P

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,130 @@
...
+        comments_found = self.find_with_start_end(
+                filename, file, dict(lang_dict["multiline_comment_delimiters"]))
+        # single-line Comments
+        comments_found.update(self.find_singleline_comments(
+                          filename, file, list(lang_dict["comment_delimiter"])))
+
+        matches_found = strings_found | comments_found
+        # Remove Nested
+        unnested_annotations = set(filter(
+                              lambda arg: not starts_within_ranges(
```
sils1297: this should only be indented 4 spaces more than the line above, same below btw.

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,130 @@
...
+
+        :param filename: Name of the file on which the bear is running.
+        :param file:     Contents of the file in the form of tuple of lines.
+        :param annot:    A dict containing start of annotation as key and end of
+                         annotation as value.
+        :return:         A set of SourceRanges giving the range of annotation.
+        """
+        text = ''.join(file)
+        found_pos = set()
+        for annot_type in annot:
```
sils1297: could you use `yield_once(annot)` to avoid duplicates?

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,130 @@
...
+            found_pos.update(unescaped_search_in_between(
+                                           annot_type, annot[annot_type], text))
+        if found_pos:
+            found_pos = set(SourceRange.from_absolute_position(
+                              filename,
+                              AbsolutePosition(file, position.begin.range[0]),
+                              AbsolutePosition(file, position.end.range[1] - 1))
```
sils1297: indentation off, should be 4 spaces higher than the `(`

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,130 @@
...
+    for outside_range in outside_ranges:
+        if inside_range == outside_range:
+            continue
+
+        # Special case of python language.
+        # Where doc strings (""") and strings (") have a similar start.
+        elif inside_range.start == outside_range.start:
+            if inside_range.end < outside_range.end:
+                return True
+
```
sils1297: can we remove this line and 120, IMO it seperates the elif from the if and makes it harder to grasp what belongs to what

bears/general/AnnotationBear.py
abhsag24: i don't understand?

sils1297: see `coalib.misc.Decorators.yield_once`, say I provide the same annot twice?

sils1297: not sure if it pays off though

abhsag24: wouldn't `found_pos` being a set already take care of that?

sils1297: indeed, sorry :/

sils1297: though it'd be good if annot is a set too

sils1297: ah it's a dict so keys must be unique anyway, it's great :) :+1:

bears/general/AnnotationBear.py
```diff
@@ -0,0 +1,129 @@
...
+        for comment_type in comments:
+            for found in unescaped_search_for(comment_type, text):
+                start = found.start()
+                end = text.find('\n', start)
+                end = len(text) - 1 if end == -1 else end
+                single_comments.append(SourceRange.from_absolute_position(
+                                           filename,
+                                           AbsolutePosition(file, start),
+                                           AbsolutePosition(file, end)))
+        return set(single_comments)
```
sils1297: why not use a set in the first place?

abhsag24: point, i do that in the other function i don;t know why it's like this here

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/227/files

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/227/commits/6b47e5a2a0bfa3018a38e46b0c6bca97d1c9ee9a

sils1297:
@rultor merge

PR #230, opened 2016-03-12
Flake8Bear: Work in progress

srisankethu:
Did not default flags like max-line-length, exclude, ignore. @sils1297 insisted on adding a PR for review. Need to add tests if bears is :+1:

### DIFF
--- a/bears/python/Flake8Bear.py
+++ b/bears/python/Flake8Bear.py
@@ -1,0 +1,45 @@
+import sys
+import os
+
+from flake8.engine import get_style_guide
+
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+if sys.platform.startswith('win'):
+    DEFAULT_CONFIG = os.path.expanduser(r'~\.flake8')
+else:
+    DEFAULT_CONFIG = os.path.join(
+        os.getenv('XDG_CONFIG_HOME') or os.path.expanduser('~/.config'),
+        'flake8'
+    )
+
+
+class Flake8Bear(Lint, LocalBear):
+
+    """ Checks the code according to pep8 and pyflake8 standards """
+
+    def lint(self, filename, file):
+        self.gives_corrected = False
+
+        flake8_style = get_style_guide(config_file=DEFAULT_CONFIG)
+        options = flake8_style.options
+
+        if options.install_hook:
+            from flake8.hooks import install_hook
+            install_hook()
+
+        report = flake8_style.check_files()
+        output = report.print_statistics()
+
+        return self.process_output(list(output), filename, file)
+
+    def run(self,
+            filename,
+            file,
+            max_line_length: int=80,
+            ignore=(),
+            flake8_format='',
+            max_complexity=-1):
+
+        return self.lint(filename, file)


AbdealiJK:
review is at https://github.com/coala-analyzer/coala-bears/commit/b4fded46d2387aaee82f0de5283a27bad7005bf5

srisankethu:
I tried importing flake8.main but in check_file() and check_code() are returning int values. So should I try without importing?

AbdealiJK:
It shouldnt be returning an int I think ...
https://github.com/PyCQA/flake8/blob/master/flake8/main.py#L79 says it is a
very simple function.
Maybe we want to use
https://github.com/PyCQA/flake8/blob/master/flake8/engine.py#L230 directly
? it seems more powerful

srisankethu:
param: ouput is NONE. I am not understanding, what's my mistake ?
https://github.com/coala-analyzer/coala-bears/pull/230/files#diff-4a6e8b080ef21dc8145e9386c1d06c3fR33

Makman2:
@srisankethu if this is a WIP, could you close the PR temporarily, since if you change stuff our CI activates everytime, this blasts build times a bit when we have a coding "rush-hour" :D
If you are done you can easily reopen it again :+1: :)

srisankethu:
@Makman2 Ok :+1: . Can you check what i might have missed ?

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,45 @@
...
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+if sys.platform.startswith('win'):
+    DEFAULT_CONFIG = os.path.expanduser(r'~\.flake8')
+else:
+    DEFAULT_CONFIG = os.path.join(
+        os.getenv('XDG_CONFIG_HOME') or os.path.expanduser('~/.config'),
+        'flake8'
+    )
```
Makman2: please align the last bracket onto the line with the last element:
`'flake8')`

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Flake8Bear(Lint, LocalBear):
+
+    """ Checks the code according to pep8 and pyflake8 standards """
+
+    def lint(self, filename, file):
+        self.gives_corrected = False
```
Makman2: this one needs to be at class level

Makman2: this one needs to be at class level

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,45 @@
...
+    DEFAULT_CONFIG = os.path.expanduser(r'~\.flake8')
+else:
+    DEFAULT_CONFIG = os.path.join(
+        os.getenv('XDG_CONFIG_HOME') or os.path.expanduser('~/.config'),
+        'flake8'
+    )
+
+
+class Flake8Bear(Lint, LocalBear):
+
```
Makman2: no newline here :+1:

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,45 @@
+import sys
+import os
+
+from flake8.engine import get_style_guide
+
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+if sys.platform.startswith('win'):
```
Makman2: Please use `platform.system()` to check for the OS

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,45 @@
+import sys
+import os
+
+from flake8.engine import get_style_guide
+
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
```
Makman2: please another newline here :)

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,45 @@
...
+        return self.process_output(list(output), filename, file)
+
+    def run(self,
+            filename,
+            file,
+            max_line_length: int=80,
+            ignore=(),
+            flake8_format='',
+            max_complexity=-1):
+
```
Makman2: superfluos newline

Makman2:
And yeah tests are missing :+1:

srisankethu:
@Makman2 I will check the code-style part later! But the code isn't working :( I wanted to know what i am missing here, why is param: option is None

Makman2:
@srisankethu what do you mean?

srisankethu:
@Makman2 I meant that output is of type None when i am running it, but it shouldn't be the case

Makman2:
I don't know how flake8 works, that's your duty to debug and find out what's the problem to get stuff running :)

srisankethu:
@Makman2 Yes! I'm trying :+1:

srisankethu:
In param: report is a object <flake8.reporter.QueueReport> . This class has a method get_file_results to print results and return an overall count(this method is of no use for us). I tried using it but returns QueueReport has no attribute _deferred_print(This must be a bug in flake8). @AbdealiJK I found no method that stores the error_messages.

AbdealiJK:
@Makman2 _deferred_print is obviously not a bug
If you think thats a bug, please try cloning and testing it locally and check if any error arises.

The Reporters are inheriting from pep8 Reporters.
Which are doing something - https://github.com/PyCQA/pycodestyle/blob/c8e0911f5df9d587e16bccf70cd48cac05d9c49d/pep8.py#L1749

srisankethu:
@AbdealiJK i was checking the code from a .tar.gz that i downloaded. There seems to be difference between these two packages.

Issue #212, opened 2016-03-11
Add args to XMLBear


AbdealiJK:
Related to https://github.com/coala-analyzer/coala-bears/issues/155

We should have args to manually specify the doctype of the XML schema
Also other args should ave configs.

Basically: More configs the better !

PS: Markdown Bear s an example

Issue #110, opened 2016-02-25
prepare_file for configs in tests doesnt work


AbdealiJK:
The issue here is that the `with` statement is wrapping the creation of the class. Not the running of the class.

So, when we do 
```
with prepare_file(config_file,
                  filename=None,
                  force_linebreaks=True,
                  create_tempfile=True) as (conf_lines, conf_file):
    JSHintBearConfigFileTest = verify_local_bear(
        JSHintBear,
        valid_files=(),
        invalid_files=(test_file1, test_file2),
        settings={"jshint_config": conf_file})
```
The file in `config_file` exists only when pytest is collecting for the unittests to run. NOT when it is actually running the unittest.

Not sure how we fix this. Possible solution in using `pytest`'s fixtures.
The other (not so nice) solution is again use `test_files` and create config files manually.

tushar-rishav:
@AbdealiJK Could you elaborate "collecting for the unittests to run...actually running the unittest" ? Are you implying that py.test performs two step process. First collecting all the tests files and then actually run the tests?

AbdealiJK:
@tushar-rishav yes, it is a two step process

tushar-rishav:
@AbdealiJK Cool. Could you elaborate why py.test is ignoring the config_file while running the test? Umm is py.test sort of caching the test files in first step and then using those in second step? I guess not, because that would not be the cause for missing data from config_file. Some more thoughts please?

AbdealiJK:
So, in the first step it just goes through all files and checks which classes are children of `unittest.TestCase`. Once it collects all these classes it uses them one by one and runs them. It must be creating objects out of the classes i guess.
The thing is that when it is collecting the classes, the `with` statement where the `config_file` is created has already completed executing. So, the temp config file which was created has already been deleted when the object is being run - As the control went out of the `with` block.

That's why the config file doesnt exist when the object is being run later.

tushar-rishav:
@AbdealiJK Ah If py.test stores the instances then how about if we bind the conf_file to the class itself and pass that class variable as an argument to settings. Something like
```python
TempClass = object()
with prepare_file(config_file,
                  filename=None,
                  force_linebreaks=True,
                  create_tempfile=True) as (conf_lines, conf_file):
    TempClass = verify_local_bear(
        JSHintBear,
        valid_files=(test_file1, test_file2),
        invalid_files=(),
        settings={"jshint_config": conf_file})
    TempClass.conf_file = conf_file
[...]

PR #232, opened 2016-03-12
XMLBear: Add args

vivek425ster:
Added schema and dtdvalid args to the XMLBear.
Also added tests related to the same.

Fixes https://github.com/coala-analyzer/coala-bears/issues/212

### DIFF
--- a/bears/xml/XMLBear.py
+++ b/bears/xml/XMLBear.py
@@ -1,10 +1,24 @@
 from coalib.bearlib.abstractions.Lint import Lint
 from coalib.bears.LocalBear import LocalBear
+from coalib.misc.Shell import escape_path_argument
+from coalib.settings.Setting import path, url
+
+
+def path_or_url(xml_dtd):
+    '''
+    Coverts the setting value to url or path.
+
+    :param xml_dtd: Setting key.
+    :return:        Returns a converted setting value.
+    '''
+    try:
+        return url(xml_dtd)
+    except ValueError:
+        return path(xml_dtd)
 
 
 class XMLBear(LocalBear, Lint):
     executable = 'xmllint'
-    arguments = '{filename}'
     diff_message = "XML can be formatted better."
     output_regex = r'(.*\.xml):(?P<line>\d+): (?P<message>.*)\n.*\n.*'
     gives_corrected = True
@@ -15,8 +29,19 @@ class XMLBear(LocalBear, Lint):
         if self.stderr_output:  # pragma: no cover
             return self._process_issues(self.stderr_output, filename)
 
-    def run(self, filename, file):
+    def run(self, filename, file,
+            xml_schema: path="",
+            xml_dtd: path_or_url=""):
         '''
-        Checks the code with `xmllint`.
+        Checks the code with ``xmllint``.
+
+        :param xml_schema: ``W3C XML Schema`` file used for validation.
+        :param xml_dtd:    ``Document type Definition (DTD)`` file or
+                           url used for validation.
         '''
+        self.arguments = "{filename} "
+        if xml_schema:
+            self.arguments += " -schema " + escape_path_argument(xml_schema)
+        if xml_dtd:
+            self.arguments += " -dtdvalid " + xml_dtd
         return self.lint(filename, file)
--- a/tests/xml/XMLBearTest.py
+++ b/tests/xml/XMLBearTest.py
@@ -1,3 +1,5 @@
+import os
+
 from bears.xml.XMLBear import XMLBear
 from tests.LocalBearTestHelper import verify_local_bear
 
@@ -18,14 +20,64 @@ valid_xml_chars = """<?xml version="1.0"?>
 <a>hey and hi</a>
 """.splitlines(keepends=True)
 
-InvalidXMLCorrectedTest = verify_local_bear(
-                                        XMLBear,
-                                        valid_files=(valid_xml_file,),
-                                        invalid_files=(invalid_xml_file,),
-                                        tempfile_kwargs={"suffix": ".xml"})
+dtd_file = os.path.join(os.path.dirname(__file__),
+                        "test_files",
+                        "note.dtd")
 
-InvalidXMLSyntaxTest = verify_local_bear(
-                                        XMLBear,
-                                        valid_files=(valid_xml_chars,),
-                                        invalid_files=(invalid_xml_chars,),
-                                        tempfile_kwargs={"suffix": ".xml"})
+schema_file = os.path.join(os.path.dirname(__file__),
+                           "test_files",
+                           "note.xsd")
+
+valid_xml_path = list(open(os.path.join(
+    os.path.dirname(__file__),
+    "test_files",
+    "note.xml"), 'r'))
+
+valid_xml_url = list(open(os.path.join(
+    os.path.dirname(__file__),
+    "test_files",
+    "concept-valid.xml"), 'r'))
+
+invalid_xml_schema = list(open(os.path.join(
+    os.path.dirname(__file__),
+    "test_files",
+    "xsd-error.xml"), 'r'))
+
+invalid_xml_dtd = list(open(os.path.join(
+    os.path.dirname(__file__),
+    "test_files",
+    "dtd-error.xml"), 'r'))
+
+invalid_xml_url = list(open(os.path.join(
+    os.path.dirname(__file__),
+    "test_files",
+    "concept-invalid.xml"), 'r'))
+
+dtd_url = "http://docs.oasis-open.org/dita/v1.0.1/dtd/concept.dtd"
+
+XMLBearCorrectedTest = verify_local_bear(
+    XMLBear,
+    valid_files=(valid_xml_file, valid_xml_chars),
+    invalid_files=(invalid_xml_file, invalid_xml_chars),
+    tempfile_kwargs={"suffix": ".xml"})
+
+XMLBearSchemaTest = verify_local_bear(
+    XMLBear,
+    valid_files=(valid_xml_path,),
+    invalid_files=(invalid_xml_schema,),
+    settings={'xml_schema': schema_file},
+    tempfile_kwargs={"suffix": ".xml"})
+
+XMLBearDTDPathTest = verify_local_bear(
+    XMLBear,
+    valid_files=(valid_xml_path,),
+    invalid_files=(invalid_xml_dtd,),
+    settings={'xml_dtd': dtd_file},
+    tempfile_kwargs={"suffix": ".xml"})
+
+XMLBearDTDUrlTest = verify_local_bear(
+    XMLBear,
+    valid_files=(valid_xml_url,),
+    invalid_files=(invalid_xml_url,),
+    settings={'xml_dtd': dtd_url},
+    tempfile_kwargs={"suffix": ".xml"})
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-coala==0.4.2.dev20160310143931
+coala==0.4.2.dev20160316074840
 setuptools>=19.2
 munkres3>=1.0.5.5
 pylint>=1.5.2
--- a/tests/xml/test_files/concept-invalid.xml
+++ b/tests/xml/test_files/concept-invalid.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE concept PUBLIC "-//OASIS//DTD DITA Concept//EN" "http://docs.oasis-open.org/dita/v1.0.1/dtd/concept.dtd">
+<concept xmlns:ditaarch="http://dita.oasis-open.org/architecture/2005/" id="concept1" ditaarch:DITAArchVersion="1.0" domains="(topic ui-d) (topic hi-d) (topic pr-d) (topic sw-d)                          (topic ut-d)" class="- topic/topic concept/concept ">
+   <conbody class="- topic/body  concept/conbody ">
+      <p class="- topic/p " />
+   </conbody>
+</concept>
--- a/tests/xml/test_files/concept-valid.xml
+++ b/tests/xml/test_files/concept-valid.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE concept PUBLIC "-//OASIS//DTD DITA Concept//EN" "http://docs.oasis-open.org/dita/v1.0.1/dtd/concept.dtd">
+<concept xmlns:ditaarch="http://dita.oasis-open.org/architecture/2005/" xmlns="" id="concept1" ditaarch:DITAArchVersion="1.0" domains="(topic ui-d) (topic hi-d) (topic pr-d) (topic sw-d)                          (topic ut-d)" class="- topic/topic concept/concept ">
+  <title xmlns="" class="- topic/title ">This is a title.</title>
+  <conbody xmlns="" class="- topic/body  concept/conbody ">
+    <p xmlns="" class="- topic/p "/>
+  </conbody>
+</concept>
--- a/tests/xml/test_files/dtd-error.xml
+++ b/tests/xml/test_files/dtd-error.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0"?>
+<note
+  xmlns="http://www.w3schools.com"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:noNamespaceSchemaLocation="note.xsd"
+>
+  <to>Tove</to>
+  <from>Jani</from>
+  <heading>Reminder</heading>
+  <body>Don't forget me this weekend!</body>
+  <footer/>
+</note>
--- a/tests/xml/test_files/note.dtd
+++ b/tests/xml/test_files/note.dtd
@@ -1,0 +1,10 @@
+<!ELEMENT note (to, from, heading, body)>
+<!ATTLIST note xmlns CDATA #IMPLIED>
+<!ATTLIST note xmlns:xsi CDATA #IMPLIED>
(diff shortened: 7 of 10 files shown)


sils1297:
unack https://github.com/vivek425ster/coala-bears/commit/b36c4690251ec32388fd57a9e2aa21ec92d9ec8a

Adrianzatreanu:
unack 85470ea ..seems it needs some work

bears/xml/XMLBear.py
```diff
@@ -1,5 +1,26 @@
...
+    '''
+    Checks if the argument is a ``url`` or a ``path``.
+
+    :param type1:       path
+    :param type2:       str
+    :return:            path or str
+    '''
+    def _path_or_url(dtdvalid):
+        if dtdvalid is not None and os.path.exists(dtdvalid):
+            return type1
```
AbdealiJK: This doesnt look right
the function is supposed to return the converted value.

If I use `path_or_url(path, str)` right now, and the path is valid, you're returning `path`
While actually you should be returning the path-ified dtdvalue - i.e. `path(dtdvalue)`

bears/xml/XMLBear.py
```diff
@@ -1,5 +1,26 @@
...
 from coalib.bears.LocalBear import LocalBear
+from coalib.misc.Shell import escape_path_argument
+from coalib.settings.Setting import path
+
+
+def path_or_url(type1, type2):
+    '''
+    Checks if the argument is a ``url`` or a ``path``.
+
+    :param type1:       path
```
AbdealiJK: Formatting is wrong - unnecessary spaced between `:param type1:` and `path`

bears/xml/XMLBear.py
```diff
@@ -1,5 +1,26 @@
+import os
+import requests
+
 from coalib.bearlib.abstractions.Lint import Lint
 from coalib.bears.LocalBear import LocalBear
+from coalib.misc.Shell import escape_path_argument
+from coalib.settings.Setting import path
+
+
+def path_or_url(type1, type2):
```
AbdealiJK: Why does this take in `type1` and `type2` ?
I mean the function name is path or url ... so, it is obviously a path or str right ? Why do you need to askt he user to give it again ?

bears/xml/XMLBear.py
```diff
@@ -15,8 +36,19 @@ def process_output(self, output, filename, file):
...
 
-    def run(self, filename, file):
+    def run(self, filename, file,
+            schema: path="",
+            dtdvalid: path_or_url(path, str)=""):
         '''
-        Checks the code with `xmllint`.
+        Checks the code with ``xmllint``.
+
+        :param: schema:        Use a W3C XML Schema file named ``schema`` for
```
AbdealiJK: Styling is wrong
Unnecessary spaces
And the `:` after `:param` should be removed

bears/xml/XMLBear.py
vivek425ster: @AbdealiJK I changed it to type1(dtdvalid) still it is giving the same error while running tests
```E           AssertionError: 0 == 0 : The local bear 'XMLBear' yields no result although it should. 
E           Running bear XMLBear...
E           The bear XMLBear cannot be executed. Unable to convert parameter 'dtdvalid' into type <function path_or_url.<locals>._path_or_url at 0x7f5a776421e0>```

File: bears/xml/XMLBear.py [hunk 5]
```diff
@@ -15,8 +36,19 @@ def process_output(self, output, filename, file):
...
+                               validation.
+        :param: dtdvalid:      Use the ``DTD`` specified by URL for validation.
         '''
+        if schema:
```
AbdealiJK: You need to put `arguments = "{filename}"` on top here.
otherwise the things you add using `+=` will be saved to the class (as arguments is a class variable) and will be used next time an object is created from it

bears/xml/XMLBear.py
```diff
@@ -15,8 +36,19 @@ def process_output(self, output, filename, file):
...
+            self.arguments += " -schema " + escape_path_argument(schema)
+        if dtdvalid:
+            self.arguments += (" -dtdvalid " +
+                               dtdvalid)
```
AbdealiJK: Can be moved to 1 line ?

AbdealiJK: Not sure how this should be escaped ... Or whether it should be escaped ...

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,76 @@
...
+                                                       valid_xml_path, 'r')),),
+                                               invalid_files=(
+                                                   list(open(
+                                                       invalid_xml_dtd, 'r')),),
+                                               settings={
+                                                   'dtdvalid': dtd_file},
+                                               tempfile_kwargs={
+                                                   "suffix": ".xml"})
+
+InvalidXMLDtdValid_URLTest =  verify_local_bear(XMLBear,
```
AbdealiJK: Hey, I notice this looks a little weird.

Can you move everything to 4 space indents for easier readability ?

```
InvalidXMLDtdValidTest = verify_local_bear(
    XMLBear,
    valid_files=(list(open(valid_xml_path)),),
    invalid_files=(list(open(invalid_xml_dtd)),),
    settings={'dtdvalid': dtd_file},
    tempfile_kwargs={"suffix": ".xml"})
```

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,76 @@
...
+                                               valid_files=(
+                                                   list(open(
+                                                       valid_xml_url, 'r')),),
```
AbdealiJK: Is `r` necessary ? I thought that's the default

bears/xml/XMLBear.py
AbdealiJK: That's interesting. Can you try printing `dtdvalid` `type1` `type1(dtdvalid)` before the return to get more debugging info

bears/xml/XMLBear.py
```diff
@@ -15,8 +36,19 @@ def process_output(self, output, filename, file):
         if self.stderr_output:  # pragma: no cover
             return self._process_issues(self.stderr_output, filename)
 
-    def run(self, filename, file):
+    def run(self, filename, file,
+            schema: path="",
+            dtdvalid: path_or_url(path, str)=""):
```
AbdealiJK: We probably want this to be names `xml_dtd` or so, as dtdvalid is not a very good name.
Even the above `schema` should be `xml_schema` or so ?

bears/xml/XMLBear.py
AbdealiJK:
```
Use a W3C XML Schema file named ``schema``
```
Why is this named schema ? I didnt quite understand the description.
Isn't `The W3C XML Schema file to use for validation` sufficient ??

bears/xml/XMLBear.py
```diff
@@ -15,8 +36,19 @@ def process_output(self, output, filename, file):
...
         '''
-        Checks the code with `xmllint`.
+        Checks the code with ``xmllint``.
+
+        :param: schema:        Use a W3C XML Schema file named ``schema`` for
+                               validation.
+        :param: dtdvalid:      Use the ``DTD`` specified by URL for validation.
```
AbdealiJK: Again `The document type definition (DTD) to use for validation. This can be a URL or a path to the DTD`

Note that DTD is not code so backticsk shouldnt be used

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,76 @@
...
+                               "concept-invalid.xml")
+
+dtd_url = "http://docs.oasis-open.org/dita/v1.0.1/dtd/concept.dtd"
+
+InvalidXMLCorrectedTest = verify_local_bear(XMLBear,
+                                            valid_files=(valid_xml_file,),
+                                            invalid_files=(invalid_xml_file,),
+                                            tempfile_kwargs={"suffix": ".xml"})
+
+InvalidXMLSyntaxTest = verify_local_bear(XMLBear,
```
AbdealiJK: You can merge this with the above one right ?
As there are no extra configs that need to be provided ?

(You can give 2 files in valid_files)

Not sure why they werent done together before

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,76 @@
...
+                                         valid_files=(valid_xml_chars,),
+                                         invalid_files=(invalid_xml_chars,),
+                                         tempfile_kwargs={"suffix": ".xml"})
+
+InvalidXMLSchemaTest = verify_local_bear(XMLBear,
```
AbdealiJK: Why do all the test names begin with `Invalid` :D 
`XMLBearSchemaTest` makes more sense

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,76 @@
...
+                                         valid_files=(
+                                             list(
+                                                 open(valid_xml_path, 'r')),),
+                                         invalid_files=(
+                                             list(open(
+                                                 invalid_xml_schema, 'r')),),
+                                         settings={'schema': schema_file},
+                                         tempfile_kwargs={"suffix": ".xml"})
+
+InvalidXMLDtdValidTest = verify_local_bear(XMLBear,
```
AbdealiJK: `XMLBearDTDPathTest`

tests/xml/XMLBearTest.py
AbdealiJK: We don't normally use underscorses (snake case) when using camel case.

`XMLBearDTDUrlTest` ?

bears/xml/XMLBear.py
vivek425ster: @AbdealiJK How to print it. A simple print statement isn't displaying anything in the error log of py.test

AbdealiJK: Hm. I thought print would work.
You could try raise AssertionError(stuff to print)
That would force pytest to end as you're reading the error but will give you the debug info you want

-----Original Message-----

vivek425ster: @AbdealiJK Still not able to find it in the log. I don't think it is even reaching that line. 
I have written this python script similar to the above program.https://gist.github.com/vivek425ster/2844337636d7e96391f6
It produces the following error:
```    f(10,s)
  File "annot.py", line 36, in f
    validate(f, locals())
  File "annot.py", line 5, in validate
    msg = 'Var: {0}\tValue: {1}\tTest: {2.__name__}	\tType: {3}'.format(var, value, test, type(test(value)))
  File "annot.py", line 25, in <lambda>
    return lambda dtdvalid: _path_or_url(dtdvalid)
  File "annot.py", line 22, in _path_or_url
    return x(dtdvalid)
  File "annot.py", line 11, in path
    return obj.__path__(*args, **kwargs)
AttributeError: 'str' object has no attribute '__path__'
```

AbdealiJK:
This is blcoked by https://github.com/coala-analyzer/coala/pull/1891

AbdealiJK:
Just merged that one ... Not blocked anymore.
@vivek425ster You will need to change the version of `coala` in the requirements.txt for this to work (as otherwise `url` wont exist)
You can see the development version in rultor's log in https://github.com/coala-analyzer/coala/pull/1891
(Do a ctrl+F of "dev2016" or so)

AbdealiJK:
Note: Remember to change thee version in a new commit which is above your commit
(Order is important)

bears/xml/XMLBear.py
```diff
@@ -15,8 +32,20 @@ def process_output(self, output, filename, file):
...
+
+        :param xml_schema: ``W3C XML Schema`` file used for validation.
+        :param xml_dtd:    ``Document type Definition (DTD)`` file or
+                           url used for validation.
         '''
+        self.arguments = ""
+        if xml_schema:
+            self.arguments += " -schema " + escape_path_argument(xml_schema)
+        if xml_dtd:
+            self.arguments += (" -dtdvalid " + xml_dtd)
```
AbdealiJK: unnecessary brackets

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,59 @@
...
+XMLBearSchemaTest = verify_local_bear(
+    XMLBear,
+    valid_files=(list(open(valid_xml_path, 'r')),),
+    invalid_files=(list(open(invalid_xml_schema, 'r')),),
+    settings={'xml_schema': schema_file},
+    tempfile_kwargs={"suffix": ".xml"})
+
+XMLBearDTDPathTest = verify_local_bear(
+    XMLBear,
+    valid_files=(list(open(valid_xml_path, 'r')),),
```
AbdealiJK: Why dont you create `note.xml` as a variable like `valid_xml_file` ?

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,59 @@
...
+    invalid_files=(list(open(invalid_xml_dtd, 'r')),),
+    settings={'xml_dtd': dtd_file},
```
AbdealiJK: Only the dtd_file cannot be created as a variable because of https://github.com/coala-analyzer/coala-bears/issues/110

AbdealiJK:
ack 2f50dba

AbdealiJK:
unack 36299e5

bears/xml/XMLBear.py
```diff
@@ -1,10 +1,27 @@
...
+from coalib.settings.Setting import path
+from coalib.settings.Setting import url
+
+
+def path_or_url():
+    '''
+    Creates a function which converts the given argument into a ``url`` or
+    ``path``.
+
+    :return: Returns a conversion function.
```
AbdealiJK: Returns a converted setting value ?

bears/xml/XMLBear.py
```diff
@@ -1,10 +1,27 @@
 from coalib.bearlib.abstractions.Lint import Lint
 from coalib.bears.LocalBear import LocalBear
+from coalib.misc.Shell import escape_path_argument
+from coalib.settings.Setting import path
+from coalib.settings.Setting import url
```
AbdealiJK: Just do
`from coalib.settings.Setting import path, url`

bears/xml/XMLBear.py
AbdealiJK: Why do we need a decorated function here ?

just do:
```
    def path_or_url(xml_dtd):
        try:
            return url(xml_dtd)
        except ValueError:
            return path(xml_dtd)
```
and then use:
`xml_dtd: path_or_url=""`

bears/xml/XMLBear.py
```diff
@@ -15,8 +32,20 @@ def process_output(self, output, filename, file):
...
+            xml_dtd: path_or_url()=""):
         '''
-        Checks the code with `xmllint`.
+        Checks the code with ``xmllint``.
+
+        :param xml_schema: ``W3C XML Schema`` file used for validation.
+        :param xml_dtd:    ``Document type Definition (DTD)`` file or
+                           url used for validation.
         '''
+        self.arguments = ""
```
AbdealiJK: You can do `self.arguments = "{filename} "` and remove Line 50

tests/xml/XMLBearTest.py
```diff
@@ -18,14 +20,59 @@
...
-                                        tempfile_kwargs={"suffix": ".xml"})
+dtd_file = os.path.join(os.path.dirname(__file__),
+                        "test_files",
+                        "note.dtd")
+
+schema_file = os.path.join(os.path.dirname(__file__),
+                           "test_files",
+                           "note.xsd")
+
+valid_xml_path = list(open(os.path.join(os.path.dirname(__file__),
```
AbdealiJK: I think you misunderstood me. Why not do this like how `invalid_xml_chars` is done ?

i.e. do something like:

```
valid_xml_path = """
<?xml version="1.0" encoding="UTF-8"?>
<note
 xmlns="http://www.w3schools.com"
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:noNamespaceSchemaLocation="./note.xsd">
   <to>Tove</to>
   <from>Jani</from>
   <heading>Reminder</heading>
   <body>Don't forget me this weekend!</body>
</note>
"""
```

This makes it easier to find the xml file rather than opening a new file.

Note: Same for all the other files given in `valid_files` and `invalid_files`

vivek425ster: @AbdealiJK Putting all the content in the file would be very clustering. 
I think putting it in the test_files is good idea.

AbdealiJK: I felt it is easier to read, because it will maximum make this file 120 lines or so, which is quite ok.
I'll leave it to you though - your call

AbdealiJK: just to be clear, this already takes up 4 lines. Adding the file directly here is just 5 lines. .. quite ok IMO.

Also ... In case you decide to use this `open` thing and not copy the contents here,
do:

```
valid_xml_url = list(open(os.path.join(
    os.path.dirname(__file__), "test_files", "concept-valid.xml"), 'r'))
```
as its easier to read

vivek425ster: @AbdealiJK abother problem in writing it in the file is since the lines are too long. It starts giving PEP8 error for line too long.

AbdealiJK: @vivek425ster you can always split lines I think ? most languages (including html/xml) have method of limiting lines to 80 chars.

AbdealiJK:
ack 66e6b29

AbdealiJK:
Rebase needed

AbdealiJK:
reack  948e58f cdc1b3f

sils1297:
reack 948e58f cdc1b3f

AbdealiJK:
@rultor merge

AbdealiJK:
@rultor merge

AbdealiJK:
@rultor merge

AbdealiJK:
@rultor merge

bears/xml/XMLBear.py
vivek425ster: @SanketDG Here.

Issue #221, opened 2016-03-12
Skip condition for FormatR bear


AbdealiJK:
The skip condition doesnt check if formatR is installed or not. Only checks if R is installed.
Should be added.

mr-karan:
Hold till https://github.com/coala-analyzer/coala/issues/1874 is resolved

PR #237, opened 2016-03-13
bears/r: Add Skip condition for FormatRBear

mr-karan:
`prerequisite_command` and `prerequisite_message` added
to FormatRBear to check prerequisite.

Fixes https://github.com/coala-analyzer/coala-bears/issues/221

### DIFF
--- a/bears/r/FormatRBear.py
+++ b/bears/r/FormatRBear.py
@@ -5,6 +5,8 @@ from coalib.bears.LocalBear import LocalBear
 class FormatRBear(Lint, LocalBear):
     executable = "Rscript"
     arguments = "-e 'library(formatR)' -e 'formatR::tidy_source({filename})'"
+    prerequisite_command = ['Rscript', '-e', "'library(formatR)'"]
+    prerequisite_fail_msg = "Please install formatR for this bear to work."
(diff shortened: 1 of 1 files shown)


SanketDG:
ack b0e7f32

SanketDG:
unack b0e7f32

commit body is wrong, should be `prerequisite_fail_msg`

Adrianzatreanu:
unack b0e7f32

AbdealiJK:
8f425eb ready

AbdealiJK:
I just noticed that actually we also need to change the version in requirements.txt

as currently it is `coala>=0.4.1.dev20160219113548` and this would only work after the 2016-03-05 version or so.

But as it's all dev releases Im letting it slide.

@sils1297 alright ?

sils1297:
for the release we should use `coala~=0.5.0`, what's the problem?

sils1297:
`coala==0.4.2.dev20160310143931` is from requirements.txt, this can't break because of coala not being up to date

AbdealiJK:
Ouch, sorry - then it's fine. Seems like my local coala-bears is an older version.

AbdealiJK:
@rultor merge

Issue #240, opened 2016-03-14
InvalidLinkBear: Implement timeout setting and raise default timeout to 2secs


jarifibrahim:
The `InvalidLinkBear` has a timeout of 1 sec which causes a number of links to fail on executing `coala invalidlink`. The timeout period should be atleast 5 sec.

sils1297:
@SanketDG do you agree? Then this is newcoer.

Makman2:
coala-bears doesn't have an `invalidlink` section, this issue should be migrated to coala^^

sils1297:
@Makman2 the bear needs changing, not the coafile

Makman2:
at least `coala invalidlink` does not apply here. Anyway: Don't we have already an issue for using a setting for the timeout? (searching...)

Makman2:
No can't find it any more

Makman2:
@jarifibrahim the better approach here is imo to supply a setting to the bear for the timeout, so you can manually adjust it. We had a discussion about the timeout and agreed on 2secs (though it's only 1 now...), changing it doesn't solve the backgound issue (there could be people that need 10secs due to an even worse internet connection).
@sils1297 do you agree changing this issue to "InvalidLinkBear - Implement timeout setting"?

sils1297:
yeah implement a setting and probably take 2 for default?

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-03-14 20:05 GMT+01:00 Mischa Krüger <notifications@github.com>:

SanketDG:
^ :+1:

PR #241, opened 2016-03-14
InvalidLinkBear: Update request time out period

jarifibrahim:
The request time out period in InvalidLinkBear file should be 5 sec
at least to verify the link.

Fixes https://github.com/coala-analyzer/coala-bears/issues/240

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -8,24 +8,26 @@ from coalib.results.Result import Result
 
 
 class InvalidLinkBear(LocalBear):
+    DEFAULT_TIMEOUT = 2
 
     @classmethod
     def check_prerequisites(cls):
-        code = InvalidLinkBear.get_status_code_or_error("http://www.google.com")
+        code = cls.get_status_code_or_error(
+            "http://www.google.com", cls.DEFAULT_TIMEOUT)
         return ("You are not connected to the internet."
                 if code is None else True)
 
     @staticmethod
-    def get_status_code_or_error(url):
+    def get_status_code_or_error(url, timeout):
         try:
             code = requests.head(url, allow_redirects=False,
-                                 timeout=1).status_code
+                                 timeout=timeout).status_code
             return code
         except requests.exceptions.RequestException:
             pass
 
     @staticmethod
-    def find_links_in_file(file):
+    def find_links_in_file(file, timeout):
         regex = re.compile(
             r'((ftp|http)s?:\/\/\S+\.(?:[^\s\(\)\'\>\|]+|'
             r'\([^\s\(\)]*\))*)(?<!\.)(?<!\,)')
@@ -33,14 +35,17 @@ class InvalidLinkBear(LocalBear):
             match = regex.search(line)
             if match:
                 link = match.group()
-                code = InvalidLinkBear.get_status_code_or_error(link)
+                code = InvalidLinkBear.get_status_code_or_error(link, timeout)
                 yield line_number + 1, link, code
 
-    def run(self, filename, file):
+    def run(self, filename, file, timeout: int=DEFAULT_TIMEOUT):
         '''
         Yields results for all invalid links in a file.
+
+        :param timeout: Request timeout period.
         '''
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(file):
+        for line_number, link, code in InvalidLinkBear.find_links_in_file(
+                file, timeout):
             if code is None:
                 yield Result.from_values(
                     origin=self,


Makman2:
Please look at issue https://github.com/coala-analyzer/coala-bears/issues/240 , to see whether we want a simple raise of the timeout or implementing the setting for it, so one can change manually the timeout depending on his/her internet connection.

Makman2:
okay respond was quicker than thought :)
@jarifibrahim so please look at the issue.

Makman2:
unack  ccc8b07

jarifibrahim:
@Makman2 I'm working on it.

bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +33,20 @@ def find_links_in_file(file):
...
 
-    def run(self, filename, file):
+    def run(self, filename, file, timeout: str):
         '''
         Yields results for all invalid links in a file.
+        
+        :param timeout: Request time out period(Default: 2 sec).
         '''
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(file):
+        if not timeout:
```
AbdealiJK: `timeout = 2 if not timeout else int(timeout)`
easier to read

bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +33,20 @@ def find_links_in_file(file):
             match = regex.search(line)
             if match:
                 link = match.group()
-                code = InvalidLinkBear.get_status_code_or_error(link)
+                code = InvalidLinkBear.get_status_code_or_error(link, timeout)
                 yield line_number + 1, link, code
 
-    def run(self, filename, file):
+    def run(self, filename, file, timeout: str):
```
AbdealiJK: Why not just set this to `int` ?
And give a default value here ?

bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +33,20 @@ def find_links_in_file(file):
...
         '''
         Yields results for all invalid links in a file.
+        
+        :param timeout: Request time out period(Default: 2 sec).
```
AbdealiJK: space before `(`

bears/general/InvalidLinkBear.py
jarifibrahim: I'll just change it.

bears/general/InvalidLinkBear.py
jarifibrahim: I didn't set it to`int` because I had to check if the value was supplied by user or not. `string` gets a value of `""` if no value is provided by user. 

> And give a default value here ?

If you a set a default there, then there is no prompt for a value. The value of `timeout` is always 2

bears/general/InvalidLinkBear.py
jarifibrahim: I'll fix it.

bears/general/InvalidLinkBear.py
AbdealiJK: I think it's fine if there is no prompt for the value - Did someone mention earlier that they wanted a prompt ?

jarifibrahim: >@jarifibrahim the better approach here is imo to supply a setting to the bear for the timeout, so you can manually adjust it. We had a discussion about the timeout and agreed on 2secs (though it's only 1 now...), changing it doesn't solve the backgound issue (there could be people that need 10secs due to an even worse internet connection).

by @Makman2

AbdealiJK: @jarifibrahim right - he didnt mention that it should be a prompt.
By adding it to the `run()` youve already made it into a setting - i.e. they can set it in the coafile.

jarifibrahim: Alright. I didn't understand that at first. I'll change the data type to `int` and directly set `timeout: int=2` in the param.
Correct?

AbdealiJK: Exactly :)
Sounds good

bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +33,17 @@ def find_links_in_file(file):
...
 
-    def run(self, filename, file):
+    def run(self, filename, file, timeout: int=2):
         '''
         Yields results for all invalid links in a file.
+        
+        :param timeout: Request time out period (Default: 2 sec).
         '''
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(file):
+
```
AbdealiJK: unnecessary newline

AbdealiJK:
The commit looks great, but the commit message is making circleci fail.
Please take a look at coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits.html

bears/general/InvalidLinkBear.py
```diff
@@ -11,21 +11,22 @@ class InvalidLinkBear(LocalBear):
 
     @classmethod
     def check_prerequisites(cls):
-        code = InvalidLinkBear.get_status_code_or_error("http://www.google.com")
+        code = InvalidLinkBear.get_status_code_or_error(
+            "http://www.google.com", 5)
```
abhsag24: why 5 here?, shouldn't this be equal to timeout as well?

Makman2: haven't we said 2 secs btw?

jarifibrahim: @abhsag24 @Makman2 I had set 5 with something else in mind. I'll fix it.

bears/general/InvalidLinkBear.py
```diff
@@ -10,22 +10,23 @@
 class InvalidLinkBear(LocalBear):
 
     @classmethod
-    def check_prerequisites(cls):
-        code = InvalidLinkBear.get_status_code_or_error("http://www.google.com")
+    def check_prerequisites(cls, timeout=2):
```
abhsag24: you can go ahead and remove the timout variable and use 2 diectly, since `check_prerequisites` gets called only on initialization, and doesn't get the timeout setting AFAIK

Makman2: there's no way a user can change the default value, so I would just leave the param here^^

bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +34,17 @@ def find_links_in_file(file):
...
-                code = InvalidLinkBear.get_status_code_or_error(link)
+                code = InvalidLinkBear.get_status_code_or_error(link, timeout)
                 yield line_number + 1, link, code
 
-    def run(self, filename, file):
+    def run(self, filename, file, timeout: int=2):
         '''
         Yields results for all invalid links in a file.
+
+        :param timeout: Request time out period (Default: 2 sec).
```
Makman2: `time out` -> `timeout`

Makman2: And you don't need to document the default, I think we grab the value automatically via the signature and pass that to the help :+1:

bears/general/InvalidLinkBear.py
jarifibrahim: So I'll let the param `timeout=2` be there. Cool?

Makman2: no please remove it, and supply `2` directly below for the timeout. Or even better, make a `DEFAULT_TIMEOUT = 2` field and use that for the default inside `run()` and here :+1:

abhsag24:
please run coala over your code

bears/general/InvalidLinkBear.py
```diff
@@ -8,24 +8,25 @@
 
 
 class InvalidLinkBear(LocalBear):
-
+    DEFAULT_TIMEOUT = 2
     @classmethod
     def check_prerequisites(cls):
-        code = InvalidLinkBear.get_status_code_or_error("http://www.google.com")
+        code = InvalidLinkBear.get_status_code_or_error(
+            "http://www.google.com", InvalidLinkBear.DEFAULT_TIMEOUT)
```
Makman2: `InvalidLinkBear.DEFAULT_TIMEOUT` -> `cls.DEFAULT_TIMEOUT`
that's what the `cls` parameter is for :)
oh the same works also with `get_status_code_or_error` :+1:

jarifibrahim: Done :+1:

sils1297:
I like it and want to have it! https://github.com/coala-analyzer/coala-bears/pull/241/commits/87e4d356141f2e8884c64c37cd2db06f0a302dc6 is ready

sils1297:
@rultor merge

Issue #184, opened 2016-03-08
InvalidLinkBear: Distinguish bad connection from Broken Link 


abhsag24:
so sometimes InvalidLinkBear yields results when we have a bad internet connection, but the links used were alright.

AbdealiJK:
Give examples please

sils1297:
uhm the InvalidLinkBear has a connection check in the prerequisites check
and should thus detect a broken connection, can you elaborate?

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-03-08 10:51 GMT+01:00 Abhay Raizada <notifications@github.com>:

abhsag24:
so the InvalidLinkBear checks the connection only when its instantiated after that if a connection is lost or is unstable, it does have a problem in distinguishing the bad connection from a broken link

AbdealiJK:
So we should have a max tries config in coafile and fail if it doesnt work
after max tries ?

abhsag24:
@AbdealiJK but if the connection is permanently lost it would again yield a result though it shouldn't though it would improve our results on an unstable connection.

we could also  just  improve the message clarifying that it's either a `broken link or a bad connection`

AbdealiJK:
I think if its a bad connection we shuld use `self.err()` and give a
warning rather than a result.
Giving a result message with "or bad connection" isnt a good idea imo

AbdealiJK:
@SanketDG can you solve this quick-ish ?
This is a pretty important bug and I am marking it for milestone v0.1

SanketDG:
Most of the connection problem went away with mocking the tests.

Closing as per as discussion with @abhsag24

PR #242, opened 2016-03-15
InvalidLinkBear: Improve request handling

SanketDG:
For bad connections, this commit will ping google.com everytime
it fails to GET a request. If this fails, then that means the
system has an unstable internet connection.

Fixes https://github.com/coala-analyzer/coala-bears/issues/184

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -10,15 +10,18 @@ from coalib.results.Result import Result
 class InvalidLinkBear(LocalBear):
     DEFAULT_TIMEOUT = 2
 
+    # IP Address of www.google.com
+    check_connection_url = "http://216.58.218.174"
+
     @classmethod
     def check_prerequisites(cls):
-        code = cls.get_status_code_or_error(
-            "http://www.google.com", cls.DEFAULT_TIMEOUT)
+        code = cls.get_status_code(
+            cls.check_connection_url, cls.DEFAULT_TIMEOUT)
         return ("You are not connected to the internet."
                 if code is None else True)
 
     @staticmethod
-    def get_status_code_or_error(url, timeout):
+    def get_status_code(url, timeout):
         try:
             code = requests.head(url, allow_redirects=False,
                                  timeout=timeout).status_code
@@ -26,8 +29,16 @@ class InvalidLinkBear(LocalBear):
         except requests.exceptions.RequestException:
             pass
 
-    @staticmethod
-    def find_links_in_file(file, timeout):
+    def get_status_code_checked(self, url, timeout):
+        code = InvalidLinkBear.get_status_code(url, timeout)
+        # if status code of google.com is None, then its
+        # an unstable connection.
+        if code is None and self.get_status_code(
+                self.check_connection_url, timeout) is None:  # pragma: no cover
+            raise requests.exceptions.NoConnectionError
+        return code
+
+    def find_links_in_file(self, file, timeout):
         regex = re.compile(
             r'((ftp|http)s?:\/\/\S+\.(?:[^\s\(\)\'\>\|]+|'
             r'\([^\s\(\)]*\))*)(?<!\.)(?<!\,)')
@@ -35,7 +46,7 @@ class InvalidLinkBear(LocalBear):
             match = regex.search(line)
             if match:
                 link = match.group()
-                code = InvalidLinkBear.get_status_code_or_error(link, timeout)
+                code = self.get_status_code_checked(link, timeout)
                 yield line_number + 1, link, code
 
     def run(self, filename, file, timeout: int=DEFAULT_TIMEOUT):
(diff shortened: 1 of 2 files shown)


bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +44,15 @@ def find_links_in_file(file):
...
+                code = self.get_status_code_checked(link)
                 yield line_number + 1, link, code
 
     def run(self, filename, file):
         '''
         Yields results for all invalid links in a file.
         '''
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(file):
+        for line_number, link, code in (InvalidLinkBear.find_links_in_file(
+                                        self, file)):
```
abhsag24: i guess using `self.find_links_in_file(file)` would save you a line since its nor longer a staticmethod

SanketDG: ^yeah, I wanted to do it in another commit.

abhsag24: like a newcomer issue?

SanketDG: No in this PR, in a separate commit, since it is not associated with the current commit anyway.

abhsag24: so you had to change the implementation of `find_links_in_file` right? that does correspond to how it is called from within other functions, so i don't see how changing `InvalidLinkBear.find_links_in_file(file)` , to `InvalidLinkBear.find_links_in_file(self, file)`  makes sense in this commit and why changing it to self.find_links_in_file(file) deserves a separate commit?

SanketDG: That makes sense, thanks!

abhsag24:
ack  efeecde

bears/general/InvalidLinkBear.py
```diff
@@ -8,15 +8,16 @@
 
 
 class InvalidLinkBear(LocalBear):
+    check_connection_url = "http://www.google.com"
```
Makman2: Have you seen this: http://stackoverflow.com/questions/3764291/checking-network-connection  ? :)
> 74.125.228.100 is one of the IP-addresses for google.com. Change http://74.125.228.100 to whatever site can be expected to respond quickly. Using a numerical IP-address avoids a DNS lookup, which may block the urllib2.urlopen call for more than a second. Thanks to rzetterberg for pointing this out.

Makman2: Though maybe we need to use a different IP, for me nothing happens :O

bears/general/InvalidLinkBear.py
```diff
@@ -33,14 +44,14 @@ def find_links_in_file(file):
...
-                code = InvalidLinkBear.get_status_code_or_error(link)
+                code = self.get_status_code_checked(link)
                 yield line_number + 1, link, code
 
     def run(self, filename, file):
         '''
         Yields results for all invalid links in a file.
         '''
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(file):
+        for line_number, link, code in (self.find_links_in_file(file)):
```
Makman2: no need for the brackets :)

bears/general/InvalidLinkBear.py
SanketDG: Yeah its actually `216.58.218.174`. I'll add this and add a comment.

AbdealiJK: In wondering here ...
Maybe we want a better check where we test for 8.8.8.8 and then check for google.com?
This way we can test if the dns server is not there or the internet connection itself is not there ?

-----Original Message-----

bears/general/InvalidLinkBear.py
```diff
@@ -24,8 +25,18 @@ def get_status_code_or_error(url):
...
-    @staticmethod
-    def find_links_in_file(file):
+    def get_status_code_checked(self, url):
+        code = InvalidLinkBear.get_status_code(url)
+        if code is None:  # pragma: no cover
+            # if status code of google.com is None, then its
+            # an unstable connection.
+            if self.get_status_code(self.check_connection_url) is None:
+                self.err("Unstable connection detected. "
+                         "InvalidLinkBear will not continue.")
```
AbdealiJK: Please use `self.__class__.__name__` so that we dont have to change this if the bear's name changed

bears/general/InvalidLinkBear.py
```diff
@@ -24,8 +25,18 @@ def get_status_code_or_error(url):
...
             pass
 
-    @staticmethod
-    def find_links_in_file(file):
+    def get_status_code_checked(self, url):
+        code = InvalidLinkBear.get_status_code(url)
+        if code is None:  # pragma: no cover
+            # if status code of google.com is None, then its
+            # an unstable connection.
+            if self.get_status_code(self.check_connection_url) is None:
```
AbdealiJK: `code is None and self.get_status_code(....`
nested ifs are basically an `and` as it's shortcircuited

AbdealiJK:
ack c8f520d5fded99c15e9e44c1a64e7796e5a86217

bears/general/InvalidLinkBear.py
SanketDG: will do

AbdealiJK:
@sils1297 just a note here.
I acked the last commit but not the second last one.
I remember you saying gitmate shouldnt set the PR status to OK in such cases. But it did set the PRs status to OK.
Bug ?

abhsag24:
Same thing happened yesterday definitely a bug

-----Original Message-----

bears/general/InvalidLinkBear.py
SanketDG: I would like to keep it this way. This looks more concise and I don't have to wrap any lines.

bears/general/InvalidLinkBear.py
Makman2: if google works imo you have internet, no need to test for 8.8.8.8 too :)

bears/general/InvalidLinkBear.py
Makman2: then `type(self).__name__` :+1:

Makman2: then `type(self).__name__` :+1:

bears/general/InvalidLinkBear.py
Makman2: I would agree with @AbdealiJK , no need to introduce another nesting level and using short-circuitance (I hope this is actually a word...) makes things slightly faster^^

Makman2:
the github status check only checks the last commit on the PR. If you haven't acked the previous ones github doesn't recognize it^^

AbdealiJK:
Sils had mentioned on gitter that now, with the new gitmate, after that huge down time it should check all the prs

-----Original Message-----

sils1297:
gitmate doesn't track acks PR wide yet

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-03-16 17:29 GMT+01:00 AbdealiJK <notifications@github.com>:

sils1297:
@SanketDG any progress here? Need help?

SanketDG:
@sils1297 Could this wait a few days? I'll get this done soon. I am not stuck, just prioritizing.

sils1297:
sure no pressure, I was just checking up on some PRs :)

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/242/commits/2b6f689ec6baebb27a343a9f03aa6fe763aecc80

bears/general/InvalidLinkBear.py
```diff
@@ -44,8 +57,7 @@ def run(self, filename, file, timeout: int=DEFAULT_TIMEOUT):
 
         :param timeout: Request timeout period.
         '''
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(
-                file, timeout):
+        for line_number, link, code in self.find_links_in_file(file, timeout):
```
sils1297: don't you want to try except here with the NoConnectionError

sils1297:
I think https://github.com/coala-analyzer/coala-bears/pull/242/commits/9f66d38849a59e2af0abb831b5a09f0ebef9bbc6 needs work

bears/general/InvalidLinkBear.py
SanketDG: oh yes, should I `pass` in the except?

sils1297: actually I think you'll rather want to do the `self.err` there and not in the get_status_code_checked function

SanketDG:
reack ded1351

bears/general/InvalidLinkBear.py
```diff
@@ -26,8 +29,16 @@ def get_status_code_or_error(url, timeout):
...
             pass
 
-    @staticmethod
-    def find_links_in_file(file, timeout):
+    def get_status_code_checked(self, url, timeout):
+        code = InvalidLinkBear.get_status_code(url, timeout)
+        # if status code of google.com is None, then its
+        # an unstable connection.
+        if code is None and self.get_status_code(
+                self.check_connection_url, timeout) is None:  # pragma: no cover
```
abhsag24: why no cover? can't we stub and test?

bears/general/InvalidLinkBear.py
```diff
@@ -26,8 +29,16 @@ def get_status_code_or_error(url, timeout):
         except requests.exceptions.RequestException:
             pass
 
-    @staticmethod
-    def find_links_in_file(file, timeout):
+    def get_status_code_checked(self, url, timeout):
+        code = InvalidLinkBear.get_status_code(url, timeout)
+        # if status code of google.com is None, then its
+        # an unstable connection.
+        if code is None and self.get_status_code(
```
abhsag24: isn't `get_status_code` a `staticmethod`?

bears/general/InvalidLinkBear.py
```diff
@@ -26,8 +29,16 @@ def get_status_code_or_error(url, timeout):
         except requests.exceptions.RequestException:
             pass
 
-    @staticmethod
-    def find_links_in_file(file, timeout):
+    def get_status_code_checked(self, url, timeout):
```
abhsag24: i'm sorry if i missed something but i'm not sure why we need this function?, can't we have a `check_connection` param in the `get_status_code`  function and do the check there itself?

bears/general/InvalidLinkBear.py
SanketDG: thanks!

bears/general/InvalidLinkBear.py
SanketDG: the point of this function is to raise a error on the second try.
the point of `get_status_code` is to return the code.

Is that okay?

abhsag24: why not something like
```
@staticmethod
def get_status_code(url, timeout, check_connection = True):
    try:
        code = requests.head(url, allow_redirects=False,
                            timeout=timeout).status_code
        if code is None and check_connection:
            if InvalidLinkBear.get_status_code(url, timeout,
                                              check_connection = False) is None:
                raise raise requests.exceptions.NoConnectionError
        return code
    except requests.exceptions.RequestException:
        pass
```

sils1297:
This has conflicts. Is it still needed?

Makman2:
@SanketDG ping :)

SanketDG:
Closing this as per as #184

Issue #109, opened 2016-02-25
PerlCritic: No arg named config


AbdealiJK:
PerlCritic gives an exception when using a config - It says no arg named `--config` exists.

Our test for config files is bad. hence we didnt catch this.

We need to remove this config

AbdealiJK:
@sudheesh001 (just checked, seems like you wrote that part) - Is the config parameter actually there ? Am I missing something ? My `perlcritic` doesn't have it

ravinimmi:
I think it should be replaced with '--profile' instead of '--config' http://search.cpan.org/~thaljef/Perl-Critic-1.126/bin/perlcritic

PR #246, opened 2016-03-16
PerlCriticBear: Replace inline config with file

hypothesist:
PerlCriticBear currently gives an exception when ran because the
`--config` argument does not exist. This should be `--profile`
instead.

Also, due to a known bug in temporary file creation, the inline
configuration file in PerCriticBearTest is moved to a directory
testfiles/ and the configuration file is placed there.

Fixes https://github.com/coala-analyzer/coala-bears/issues/109

### DIFF
--- a/bears/perl/PerlCriticBear.py
+++ b/bears/perl/PerlCriticBear.py
@@ -21,16 +21,16 @@ class PerlCriticBear(LocalBear, Lint):
     def run(self,
             filename,
             file,
-            perlcritic_config: str=""):
+            perlcritic_profile: str=""):
         '''
         Checks the code with perlcritic. This will run perlcritic over
         each of the files seperately
 
-        :param perlcritic_config: Location of the perlcriticrc config file.
+        :param perlcritic_profile: Location of the perlcriticrc config file.
         '''
         self.arguments = '--no-color --verbose "%l|%c|%s|%p|%m (%e)"'
-        if perlcritic_config:
-            self.arguments += (" --config "
-                               + escape_path_argument(perlcritic_config))
+        if perlcritic_profile:
+            self.arguments += (" --profile "
+                               + escape_path_argument(perlcritic_profile))
         self.arguments += " {filename}"
         return self.lint(filename)
--- a/tests/perl/PerlCriticBearTest.py
+++ b/tests/perl/PerlCriticBearTest.py
@@ -1,3 +1,5 @@
+import os
+
 from bears.perl.PerlCriticBear import PerlCriticBear
 from tests.LocalBearTestHelper import verify_local_bear
 from coalib.misc.ContextManagers import prepare_file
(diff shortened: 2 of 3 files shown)


bears/perl/PerlCriticBear.py
```diff
@@ -30,7 +30,7 @@ def run(self,
         '''
         self.arguments = '--no-color --verbose "%l|%c|%s|%p|%m (%e)"'
         if perlcritic_config:
-            self.arguments += (" --config "
+            self.arguments += (" --profile "
```
AbdealiJK: You need to change the test which uses prepare_file as it doesnt work.

tests/perl/PerlCriticBearTest.py
```diff
@@ -26,28 +27,15 @@
...
-                  force_linebreaks=True,
-                  create_tempfile=True) as (conf_lines, conf_file):
-    PerlCriticBearConfigTest = verify_local_bear(
-        PerlCriticBear,
-        valid_files=(bad_file, good_file),
-        invalid_files=(),
-        settings={"perlcritic_config": conf_file})
+PerlCriticBearConfigTest = verify_local_bear(
+    PerlCriticBear,
+    valid_files=(good_file,),
```
AbdealiJK: This test isnt nice :(
Because with or without the config file, the good file is always valis and bad is invalid.

Can you ask on gitter for people who know perl and change appropriately ?
Maybe check the git blame information ?

hypothesist: You're the only one with blame information... I think that's because you moved everything to the new coala-bears repo a few weeks ago. Anyway I can find the original blame info?

hypothesist: Ah, never mind: https://github.com/coala-analyzer/coala/blame/bb75516f3cea85517b0b786f7a195548181c402c/bears/tests/perl/PerlCriticBearTest.py

tests/perl/testfiles/.perlcriticrc
```diff
@@ -0,0 +1,8 @@
+severity  = 5
+# for signatures
+[-Subroutines::ProhibitSubroutinePrototypes]
+
+[-TestingAndDebugging::RequireUseStrict]
```
hypothesist: @AbdealiJK so doing this (notice the `-` at the start of the line) removes the error shown (which is enabled by default). So this will pass the `bad_file` too, thereby verifying the config file usage.

tests/perl/PerlCriticBearTest.py
```diff
@@ -26,28 +27,16 @@
...
-                  force_linebreaks=True,
-                  create_tempfile=True) as (conf_lines, conf_file):
-    PerlCriticBearConfigTest = verify_local_bear(
-        PerlCriticBear,
-        valid_files=(bad_file, good_file),
-        invalid_files=(),
-        settings={"perlcritic_config": conf_file})
+PerlCriticBearConfigTest = verify_local_bear(
+    PerlCriticBear,
+    valid_files=(good_file,bad_file,),
```
hypothesist: @AbdealiJK now both files will pass, unlike before.

tests/perl/PerlCriticBearTest.py
```diff
@@ -1,6 +1,7 @@
 from bears.perl.PerlCriticBear import PerlCriticBear
 from tests.LocalBearTestHelper import verify_local_bear
 from coalib.misc.ContextManagers import prepare_file
+import os
```
AbdealiJK: python imports are on top.
Then comes an empty newline
Then coala imports
```
import os

from bears.perl.PerlCriticBear import PerlCriticBear
```

tests/perl/PerlCriticBearTest.py
```diff
@@ -26,28 +27,16 @@
...
-                  force_linebreaks=True,
-                  create_tempfile=True) as (conf_lines, conf_file):
-    PerlCriticBearConfigTest = verify_local_bear(
-        PerlCriticBear,
-        valid_files=(bad_file, good_file),
-        invalid_files=(),
-        settings={"perlcritic_config": conf_file})
+PerlCriticBearConfigTest = verify_local_bear(
+    PerlCriticBear,
+    valid_files=(good_file, bad_file,),
```
AbdealiJK: extra comma at end

tests/perl/testfiles/.perlcriticrc
AbdealiJK: :+1:

AbdealiJK:
ack 8a83a63

hypothesist:
@rultor merge

Issue #251, opened 2016-03-17
XMLBear: Results are shown as log messages


vivek425ster:
https://asciinema.org/a/6nt1tuju6hv0b1l5ga8zsdgv7

sils1297:
is this a bugfix? Do we want a backport?

vivek425ster:
@sils1297 It is a bug fix

sils1297:
reopening for backporting

sils1297:
backport happened

PR #255, opened 2016-03-17
XMLBear: Process error correctly

vivek425ster:
Errors were shown as logs instead of errors. Corrected the processing of
output to show the error messages.

Fixes https://github.com/coala-analyzer/coala-bears/issues/251

### DIFF
--- a/bears/xml/XMLBear.py
+++ b/bears/xml/XMLBear.py
@@ -1,3 +1,5 @@
+import itertools
+
 from coalib.bearlib.abstractions.Lint import Lint
 from coalib.bears.LocalBear import LocalBear
 from coalib.misc.Shell import escape_path_argument
@@ -22,11 +24,15 @@ class XMLBear(LocalBear, Lint):
     diff_message = "XML can be formatted better."
     output_regex = r'(.*\.xml):(?P<line>\d+): (?P<message>.*)\n.*\n.*'
     gives_corrected = True
+    use_stderr = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stdout_output:
+            # Return issues from stderr and stdout if stdout is not empty
+            return itertools.chain(
+                self._process_issues(self.stderr_output, filename),
+                self._process_corrected(self.stdout_output, filename, file))
+        else:  # Return issues from stderr if stdout is empty
             return self._process_issues(self.stderr_output, filename)
 
     def run(self, filename, file,
--- a/tests/xml/test_files/xsd-error.xml
+++ b/tests/xml/test_files/xsd-error.xml
@@ -1,4 +1,3 @@
-<?xml version="1.0"?>
 <note
   xmlns="http://www.w3schools.com"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"


sils1297:
cancelling non circle builds here as this needs to be rebased later anyway

sils1297:
repeat: the builds are not failed :) all is well

bears/xml/XMLBear.py
```diff
@@ -24,10 +24,11 @@ class XMLBear(LocalBear, Lint):
     gives_corrected = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stderr_output:
+            self.use_stderr = True
```
sils1297: just define it outside, like gives corrected

sils1297:
unack f7fa889

bears/xml/XMLBear.py
```diff
@@ -24,10 +24,11 @@ class XMLBear(LocalBear, Lint):
     gives_corrected = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        self.use_stderr = True
```
sils1297: this needs to be moved to class level

bears/xml/XMLBear.py
```diff
@@ -22,12 +22,13 @@ class XMLBear(LocalBear, Lint):
...
+    use_stderr = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stderr_output:
             return self._process_issues(self.stderr_output, filename)
+        if self.stdout_output:
+            return self._process_corrected(self.stdout_output, filename, file)
```
sils1297: why do you switch those two and why do you remove the comment in line 27?

vivek425ster: @sils1297 I switched it because if there is an error in stderr then return no need to check stdout else check if something needs to be corrected

AbdealiJK: If this is a bug that you're trying to fix, please make a new test to test this.

Either way it would be good to add a comment here stating this so that someone else doesnt change it later by mistake
Note: You can use `elif` instead of `if` here.

vivek425ster: @AbdealiJK Tests are already there which covers everything.
What more tests you want me to add ?

vivek425ster:
@SanketDG @AbdealiJK Any particular reason why coverage is not 100%. 
From what i know https://github.com/coala-analyzer/coala-bears/blob/master/tests/xml/XMLBearTest.py#L11 this test should cover the elif case .

bears/xml/XMLBear.py
```diff
@@ -24,10 +26,16 @@ class XMLBear(LocalBear, Lint):
...
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stderr_output:
+            self.use_stderr = True
+            if self.stdout_output:  # only yield Result if stdout is not empty
+                return collections.ChainMap(
+                           self._process_issues(self.stderr_output, filename),
```
AbdealiJK: indentation ?

bears/xml/XMLBear.py
```diff
@@ -24,10 +26,16 @@ class XMLBear(LocalBear, Lint):
     gives_corrected = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stderr_output:
+            self.use_stderr = True
+            if self.stdout_output:  # only yield Result if stdout is not empty
+                return collections.ChainMap(
```
AbdealiJK: I don't understand why chainMap is needed. This is a list of Results.
chainMap is for dicts right ?

bears/xml/XMLBear.py
```diff
@@ -24,10 +26,16 @@ class XMLBear(LocalBear, Lint):
     gives_corrected = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stderr_output:
+            self.use_stderr = True
+            if self.stdout_output:  # only yield Result if stdout is not empty
```
AbdealiJK: I think the comment doesnt explain things correctly ?
We yield results even if stdout is empty at L36

bears/xml/XMLBear.py
AbdealiJK: use_stderr isnt really needed. It is used only in `lint()` and wont be used after this line changes it

bears/xml/XMLBear.py
```diff
@@ -24,10 +26,16 @@ class XMLBear(LocalBear, Lint):
...
+                return collections.ChainMap(
+                           self._process_issues(self.stderr_output, filename),
+                           self._process_corrected(self.stdout_output,
+                                                   filename, file))
             return self._process_issues(self.stderr_output, filename)
+        elif self.stdout_output:  # only yield Result if stdout is not empty
```
AbdealiJK: Again, this comment isnt helping - Im not sure what youre trying to say. We're yielding Results everywhere right ?

bears/xml/XMLBear.py
vivek425ster: @AbdealiJK use_stderr is needed because if not set to True it will print logs also

AbdealiJK: Ah. you're right. Sorry - I got confused.
@sils1297 it cannot be moved outside.

bears/xml/XMLBear.py
vivek425ster: @AbdealiJK What's the issue in indentation ?
Isn't it 4 spaces PEP8 doesn't give me any error.

bears/xml/XMLBear.py
vivek425ster: @AbdealiJK yeah ``chainMap`` is for dict. 
Changing it to ``itertools.chain``

bears/xml/XMLBear.py
AbdealiJK: It should be 4 spaces from the beginning of "return"

bears/xml/XMLBear.py
vivek425ster: How does ``only merge corrections if stdout is not empty`` sound ?

bears/xml/XMLBear.py
vivek425ster: @AbdealiJK It is specific to that ``elif`` clause. Won't it yield result only when there is some correction to be made ?

bears/xml/XMLBear.py
AbdealiJK: What about "Return issues from stderr and diffs from stdout if both are not empty"

vivek425ster:
@SanketDG @AbdealiJK Why isn't this elif getting covered ?
https://codecov.io/github/coala-analyzer/coala-bears/bears/xml/XMLBear.py?ref=f0425f051fba52543fd2984750a8277b427e655a#l-37
From what i know https://github.com/coala-analyzer/coala-bears/blob/master/tests/xml/XMLBearTest.py#L11 this test should cover the elif case .

bears/xml/XMLBear.py
vivek425ster: :+1:

bears/xml/XMLBear.py
```diff
@@ -24,10 +26,11 @@ class XMLBear(LocalBear, Lint):
     gives_corrected = True
 
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
-            return self._process_issues(self.stderr_output, filename)
+        if self.stderr_output:
+            self.use_stderr = True
```
SanketDG: not sure if this is needed, why do you keep this here?
The usage of `use_stderr` is determined in the `Lint.lint()` function.

vivek425ster: @SanketDG It needs to be added because if it is not there then logs will also be printed

SanketDG: Then why not just `use_stderr=True`

bears/xml/XMLBear.py
```diff
@@ -24,10 +26,11 @@ class XMLBear(LocalBear, Lint):
...
+        return itertools.chain(
+            self._process_issues(self.stderr_output, filename),
+            self._process_corrected(self.stdout_output, filename, file))
```
SanketDG: This won't work, because you are not checking if `self.stdout_output` is empty or not. Thus if I have:

`test.xml` as:

```xml
<a>hey & hi</a>
```

and I do:

```
❯ coala --files=test.xml --bears=XMLBear -L DEBUG
```

I get:

```
test.xml
|   1| <a>hey•&•hi</a>
|    | [NORMAL] XMLBear:
|    | XML can be formatted better.
|    | The following actions are applicable to this result:
|    |  0: Apply no further actions.
|    |  1: Open the affected file(s) in an editor.
|    |  2: Apply the patch automatically.
|    |  3: Print a diff of the patch that would be applied.
|    | Please enter the number of the action you want to execute. 3
|----|    | /home/sanket/coala-bears/test.xml
|    |++++| /home/sanket/coala-bears/test.xml
|   1|    |-<a>hey & hi</a>
|    | Displayed patch successfully.
|    | The following actions are applicable to this result:
|    |  0: Apply no further actions.
|    |  1: Open the affected file(s) in an editor.
|    |  2: Apply the patch automatically.
|    |  3: Print a diff of the patch that would be applied.
|    | Please enter the number of the action you want to execute. 
```

As you can see, the diff is removing the line, because `stdout` is empty.

SanketDG:
unack 6d6e9e8

bears/xml/XMLBear.py
vivek425ster: @SanketDG  I thought it gives stdout also when it gives error .
Because when you validate it against it a schema if invalid it gives output in stdout as well and also errors in stderr
```
<?xml version="1.0"?>
<note xmlns="http://www.w3schools.com" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="note.xsd">
   <from>Jani</from>
   <heading>Reminder</heading>
   <body>Don't forget me this weekend!</body>
</note>
note-invalid.xml:2: element from: Schemas validity error : Element '{http://www.w3schools.com}from': This element is not expected. Expected is ( {http://www.w3schools.com}to ).
note-invalid.xml fails to validate
```

bears/xml/XMLBear.py
vivek425ster: @SanketDG if i do `use_stderr = True` without checking whether the stderr_output is there or not it will set ``results_output `` to stderr_output even when there are no any errors in stderr. 
https://github.com/coala-analyzer/coala/blob/master/coalib/bearlib/abstractions/Lint.py#L98

vivek425ster:
@SanketDG @AbdealiJK @sils1297 Any idea why elif is not getting covered .From what i know https://github.com/coala-analyzer/coala-bears/blob/master/tests/xml/XMLBearTest.py#L11 this test should cover the elif case .
This test should cover the elif case.

sils1297:
seems to be covered now?

vivek425ster:
@sils1297 Yeah i got the problem it was because if both stdout and stderr were empty nothing was being done

tests/xml/test_files/xsd-error.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0"?>
```
Adrianzatreanu: why is this needed to be removed?

AbdealiJK: If I understand correctly, the stdout (which gives corrected output) adds it to the corrected one.
Hence, for coverage of Line 36.

But I havent followed this very closely. So, would be good to let @vivek425ster confirm

vivek425ster: @Adrianzatreanu yes it is so that stdout gives corrections

bears/xml/XMLBear.py
```diff
@@ -22,11 +24,16 @@ class XMLBear(LocalBear, Lint):
...
     def process_output(self, output, filename, file):
-        if self.stdout_output:  # only yield Result if stdout is not empty
-            return self._process_corrected(self.stdout_output, filename, file)
-        if self.stderr_output:  # pragma: no cover
+        if self.stdout_output:
+            # Return issues from stderr and stdout if stdout is not empty
+            return itertools.chain(
+                self._process_issues(self.stderr_output, filename),
+                self._process_corrected(
+                    self.stdout_output, filename, file))
```
SanketDG: why wrap this to the next line?

SanketDG:
cf51a65 needs work

SanketDG:
ack 06b5761

SanketDG:
ack 06b5761

sils1297:
@rultor merge

PR #274, opened 2016-03-21
Issue#259

vivek425ster:


SanketDG:
ack 69dfba4

Adrianzatreanu:
You have some problems with the commit message on the exitcode one. First of all, I don't think we use past tense in the body, it should be Adds; and also, on the last line, grammatically correct is "built".

sils1297:
in the body past tense is ok if it's needed

Adrianzatreanu:
unack 12eda63

Adrianzatreanu:
correct is still ``built`` :D so unacking it for now

sils1297:
ack 
0a99c06

Adrianzatreanu:
@rultor merge

Issue #81, opened 2016-02-22
ClangComplexityBear


sils1297:
should be easy to use the clang ast to calculate cyclomatic complexity

Redridge:
How do you get the clang ast?

sils1297:
look at ClangASTPrintBear for sample code on iterating over the clang ast.

Redridge:
Should this be able to suggest ways to reduce the cyclomatic complexity? If not I would like to give it a try.

sils1297:
not for now, would be neat to have a general markdown tutorial in the debug
message of the result or so

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-02-23 16:52 GMT+01:00 Alexandros Dimos <notifications@github.com>:

Uran198:
@Redridge Are you still working on this or can I try to do it?

Redridge:
@Uran198 you have my blessing

Uran198:
OK, so the approach here would be to calculate it for each module and display it. But, as for the general markdown tutorial, what should it contain? Something about, that recommended number is 10 and when it is higher there have to be a good reason for it?

sils1297:
yeah calculate the complexity and let the user specify one (or several for minor, normal, major) threshold(s)

PR #278, opened 2016-03-21
Add ClangComplexityBear

Uran198:
Fixes https://github.com/coala-analyzer/coala-bears/issues/81
But, here is still a problem. If we have:

```c
int f(int a) {
    if (a)
        return 0;
    else
        return 1;
    if (!a)
         return 2;
    else
         return 3;
}
```
statement `if (!a)` is unreachable, but I will not catch that and will have 2 decision points. In addition, I'll count 5 exit points (extra at the end of the block) instead of 2.

### DIFF
--- a/bears/c_languages/codeclone_detection/ClangComplexityBear.py
+++ b/bears/c_languages/codeclone_detection/ClangComplexityBear.py
@@ -1,0 +1,94 @@
+from clang.cindex import Index, CursorKind
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result
+from coalib.results.SourceRange import SourceRange
+from bears.c_languages.ClangBear import clang_available
+
+
+class ClangComplexityBear(LocalBear):
+    """
+    Calculates cyclomatic complexity of each function and displays it to the
+    user.
+    """
+    check_prerequisites = classmethod(clang_available)
+    decisive_cursor_kinds = {
+        CursorKind.IF_STMT, CursorKind.WHILE_STMT, CursorKind.FOR_STMT,
+        CursorKind.DEFAULT_STMT, CursorKind.CASE_STMT}
+
+    def function_key_points(self, cursor, top_function_level=False):
+        """
+        Calculates number of function's decision points and exit points.
+
+        :param top_function_level: Whether cursor is in the top level of
+                                   the function.
+        """
+        decisions, exits = 0, 0
+
+        for child in cursor.get_children():
+            if child.kind in self.decisive_cursor_kinds:
+                decisions += 1
+            elif child.kind == CursorKind.RETURN_STMT:
+                exits += 1
+                if top_function_level:
+                    # There is no point to move forward, so just return.
+                    return decisions, exits
+            child_decisions, child_exits = self.function_key_points(child)
+            decisions += child_decisions
+            exits += child_exits
+
+        if top_function_level:
+            # Implicit return statement.
+            exits += 1
+
+        return decisions, exits
+
+    def complexities(self, cursor, filename):
+        """
+        Calculates cyclomatic complexities of functions.
+        """
+
+        file = cursor.location.file
+
+        if file is not None and file.name != filename:
+            # There is nothing to do in another file.
+            return
+
+        if cursor.kind == CursorKind.FUNCTION_DECL:
+            child = next((child for child in cursor.get_children()
+                          if child.kind != CursorKind.PARM_DECL),
+                         None)
+            if child:
+                decisions, exits = self.function_key_points(child, True)
+                complexity = max(1, decisions - exits + 2)
+                yield cursor, complexity
+        else:
+            for child in cursor.get_children():
+                yield from self.complexities(child, filename)
+
+    def run(self, filename, file, max_complexity: int=8):
+        """
+        Calculates cyclomatic complexity of functions in file.
+
+        :param max_complexity:  Maximum cyclomatic complexity that is
+                                considered to be normal. The value of 10 had
+                                received substantial corroborating evidence.
+                                But the general recommendation: "For each
+                                module, either limit cyclomatic complexity to
+                                [the agreed-upon limit] or provide a written
+                                explanation of why the limit was exceeded."
+        """
+
+        root = Index.create().parse(filename).cursor
+        for cursor, complexity in self.complexities(root, filename):
+            if complexity > max_complexity:
+                affected_code = (SourceRange.from_clang_range(cursor.extent),)
+                yield Result(
+                    self,
+                    "The cyclomatic complexity of function {function} is "
+                    "{complexity} which exceeded maximal recommended value "
+                    "of {rec_value}.".format(
+                        function=cursor.displayname,
+                        complexity=complexity,
+                        rec_value=max_complexity),
+                    affected_code=affected_code)
--- a/tests/c_languages/codeclone_detection/ClangComplexityBearTest.py
+++ b/tests/c_languages/codeclone_detection/ClangComplexityBearTest.py
@@ -1,0 +1,91 @@
+import os
+import unittest
+from queue import Queue
+from clang.cindex import Index
+
+from coalib.settings.Section import Section
+from coalib.results.Result import Result
+from coalib.results.SourceRange import SourceRange
+from bears.c_languages.codeclone_detection.ClangComplexityBear import (
+    ClangComplexityBear)
+from tests.BearTestHelper import generate_skip_decorator
+from tests.LocalBearTestHelper import execute_bear
+
+
+@generate_skip_decorator(ClangComplexityBear)
+class ClangComplexityBearTest(unittest.TestCase):
+
+    def setUp(self):
+        self.filename = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                     "conditions_samples.c"))
+        self.file = "fake"
+        self.queue = Queue()
+        self.section = Section("test section")
+        self.bear = ClangComplexityBear(self.section, self.queue)
+
+    def test_calculation(self):
+        """
+        Testing that number of decision points and exit points are calculated
+        correctly.
+        """
+        expected = [
+            ('used(int, int)', 3),
+            ('returned(int, int)', 1),
+            ('loopy(int, int)', 5),
+            ('in_condition(int, int)', 1),
+            ('assignation(int, int)', 2),
+            ('arithmetics(int, int)', 1),
+            ('levels(int, int, int)', 10),
+            ('structing(struct test_struct, struct test_struct *)', 1),
+            ('switching(int, int)', 2)]
+
+        root = Index.create().parse(self.filename).cursor
+        complexities_gen = self.bear.complexities(root, self.filename)
+        results = [(cursor.displayname, complexity)
+                   for cursor, complexity in complexities_gen]
+        self.assertSequenceEqual(results, expected)
+
+    def test_output(self):
+        """
+        Validating that the yielded results are correct.
+        """
+        affected_code = (SourceRange.from_values(
+            self.filename,
+            start_line=111,
+            start_column=1,
+            end_line=143,
+            end_column=2),)
+        expected_result = Result(
+            self.bear,
+            "The cyclomatic complexity of function levels(int, int, int) is "
+            "10 which exceeded maximal recommended value of 8.",
+            affected_code=affected_code)
+        with execute_bear(self.bear, self.filename, self.file, 8) as out:
+            self.assertEqual(len(out), 1)
+        self.assertEqual(out[0], expected_result)
+
+    def test_empty_declared_function(self):
+        """
+        Should not take into account and display empty function declarations.
+        """
+        self.filename = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                     "empty_declarations.c"))
+        expected = [('with_body(int *)', 1)]
+
+        root = Index.create().parse(self.filename).cursor
+        complexities_gen = self.bear.complexities(root, self.filename)
+        results = [(cursor.displayname, complexity)
+                   for cursor, complexity in complexities_gen]
+        self.assertSequenceEqual(results, expected)
+
+    def test_file_does_not_exist(self):
+        """
+        Tests that bear throws TranslationUnitLoadError when file does not
+        exist.
+        """
+        from clang.cindex import TranslationUnitLoadError
+
+        generator = self.bear.execute("not_existing", self.file)
+        self.assertNotEqual(generator, None)
+        with self.assertRaises(TranslationUnitLoadError):
+            yield generator
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Use >= for development versions so that source builds always work
-coala>=0.5.2.dev20160321212109
+coala>=0.6.0.dev20160325124842
 setuptools>=19.2
 munkres3~=1.0.5.5
 pylint~=1.5.2
--- a/tests/c_languages/codeclone_detection/empty_declarations.c
+++ b/tests/c_languages/codeclone_detection/empty_declarations.c
@@ -1,0 +1,17 @@
+#include <stdio.h>
+#include "not_existant.c"
+
+int wihtout_params(void);
+
+void* with_one_param(int x);
+
+char with_two_params(char*, double);
+
+int with_body(int *argc) {
+    if (argc) {
+        return 0;
+    } else {
+        return 1;
+    }
+    return 2;
+}


bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,98 @@
+from clang.cindex import Index, CursorKind
+
+from coalib.bears.GlobalBear import GlobalBear
+from bears.c_languages.ClangBear import clang_available
+from coalib.results.Result import Result
+
+
+class ClangComplexityBear(GlobalBear):
```
sils1297: why not make it local? You read the files one by one anyway, right? Let coala parallelize.

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,98 @@
...
+    check_prerequisites = classmethod(clang_available)
+    decisive_cursor_kinds = {
+        CursorKind.IF_STMT, CursorKind.WHILE_STMT, CursorKind.FOR_STMT
+    }
+
+    def __init__(self, *args, **kwargs):
+        super(ClangComplexityBear, self).__init__(*args, **kwargs)
+        self.exit_points = 0
+        self.decision_points = 0
+        self.complexities = {}
```
sils1297: nnot sure if I'd use class variables for this kind of stuff, also if you need to initialize something in `__init__` in a bear that's usually wrong though I guess you did that because your IDE wnated you to only declare stuff in `__init__` and not later? You don't need this as class state and this IMO those variables should be simple local vars passed around directly.

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class ClangComplexityBear(GlobalBear):
+    """
+    Calculates cyclomatic complexity of each function and displays it to the
+    user.
+    """
+    check_prerequisites = classmethod(clang_available)
+    decisive_cursor_kinds = {
+        CursorKind.IF_STMT, CursorKind.WHILE_STMT, CursorKind.FOR_STMT
```
sils1297: how about switch/case

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: It will be much pain to pass them around, but I agree that their initialization should go into run

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: In this way it is more readable, because it has a name. With `switch` it will be a little obscure.

AbdealiJK: I think he meant you should add the cursor kind for switch-case (because `if else` and `switch case` are similar)
Is that what you're talking about ?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,98 @@
...
+    }
```
AbdealiJK: The formatting here is wrong, the `}` should be on the line above

bears/c_languages/codeclone_detection/ClangComplexityBear.py
sils1297: yes, switch/case is a decision point which is missing here right?

Uran198: Oh, I messed up. I even forgot that there is no `switch` in python.
I'll add it.

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+                                received substantial corroborating evidence.
+                                But the general recommendation: "For each
+                                module, either limit cyclomatic complexity to
+                                [the agreed-upon limit] or provide a written
+                                explanation of why the limit was exceeded."
+        """
+        self.complexities = {}
+        counter = {'decisions': 0, 'exits': 0}
+
+        def _traverse(cursor, filename, top_function_level=False):
```
sils1297: does this function need to be *inside* the other?

tests/c_languages/codeclone_detection/ClangComplexityBearTest.py
```diff
@@ -0,0 +1,80 @@
...
+            'arithmetics(int, int)': (89, 0, 1),
+            'assignation(int, int)': (62, 1, 1),
+            'in_condition(int, int)': (47, 2, 5),
+            'levels(int, int, int)': (111, 9, 1),
+            'loopy(int, int)': (22, 6, 3),
+            'returned(int, int)': (13, 1, 3),
+            'structing(struct test_struct, struct test_struct *)': (149, 0, 1),
+            'switching(int, int)': (154, 3, 3),
+            'used(int, int)': (1, 2, 1)
+        }
```
sils1297: :D you reused my test files

sils1297:
this is awesome :) I'll need to take a deeper look still

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: It really not needed outside :smiley: Also, in this way less variables need to be passed around. To be honest, I wouldn't even make `self.complexities` class attribute, but it's very useful for testing. I'll think on this. maybe, I will find a way to test adequately without it.

sils1297: so how do we want to proceed here?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+            :param filename:           Name of the file being processed.
+            :param top_function_level: Whether cursor is in the top level of
+                                       the function.
+            """
+            file = cursor.location.file
+
+            if file is not None and file.name == filename:
+                cursor_kind = cursor.kind
+            else:
+                cursor_kind = None
```
sils1297: don't you want to `return` if this cursor is in another file?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+                yield Result.from_values(
+                    self,
+                    "The cyclomatic complexity of function {function} "
+                    "exceeded maximal recommended value of {rec_value}."
+                    .format(
+                        function=func_name,
+                        rec_value=max_complexity,
+                    ),
+                    file=filename,
+                    line=line,
```
sils1297: could you maybe make a range that reaches over the whole function so the whole thing is shown to the user? i.e. specify end_line and start/end column even?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+        root = Index.create().parse(filename).cursor
+        _traverse(root, filename)
+
+        for func_name, (line, dec_pts, exit_pts) in self.complexities.items():
+            complexity = max(1, dec_pts - exit_pts + 2)
+            if complexity > max_complexity:
+                yield Result.from_values(
+                    self,
+                    "The cyclomatic complexity of function {function} "
+                    "exceeded maximal recommended value of {rec_value}."
```
sils1297: the actual value should be noted here too

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+                    .format(
+                        function=func_name,
+                        rec_value=max_complexity,
+                    ),
```
sils1297: we usually close brackets right after the last element, we're using PEP8 so you're free to do it this way also

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+
+            for child in cursor.get_children():
+                if cursor_kind == CursorKind.FUNCTION_DECL:
+                    if child.kind == CursorKind.PARM_DECL:
+                        continue
+                    counter['exits'] = counter['decisions'] = 0
+                    _traverse(child, filename, top_function_level=True)
+                    line = cursor.extent.start.line
+                    self.complexities[cursor.displayname] = (
+                        line, counter['decisions'], counter['exits'])
```
sils1297: yeah instead of `line` you could just create a `SourceRange` spanning over the whole function and create the result easily with that

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+
+        :param max_complexity:  Maximum cyclomatic complexity that is
+                                considered to be normal. The value of 10 had
+                                received substantial corroborating evidence.
+                                But the general recommendation: "For each
+                                module, either limit cyclomatic complexity to
+                                [the agreed-upon limit] or provide a written
+                                explanation of why the limit was exceeded."
+        """
+        self.complexities = {}
```
sils1297: I'd rather return the dict with the complexities by that function instead of making it a class attribute, still useful for testing as you can just test traverse directly and check the return value, right?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,88 @@
...
+        counter = {'decisions': 0, 'exits': 0}
```
sils1297: I also dislike this global state, makes things very complicated IMO, you're counting recursively so why not returning the complexity values in the traverse functions recursively?

tests/c_languages/codeclone_detection/ClangComplexityBearTest.py
```diff
@@ -0,0 +1,80 @@
...
+        expected_complexities = {
+            'arithmetics(int, int)': (89, 0, 1),
+            'assignation(int, int)': (62, 1, 1),
+            'in_condition(int, int)': (47, 2, 5),
+            'levels(int, int, int)': (111, 9, 1),
+            'loopy(int, int)': (22, 6, 3),
+            'returned(int, int)': (13, 1, 3),
+            'structing(struct test_struct, struct test_struct *)': (149, 0, 1),
+            'switching(int, int)': (154, 3, 3),
+            'used(int, int)': (1, 2, 1)
```
sils1297: did you check every single value manually?

sils1297:
1fb8e60 needs work

tests/c_languages/codeclone_detection/ClangComplexityBearTest.py
Uran198: Yes, I did.

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: It's very local state. Even the function is local.

sils1297: it's kinda cheated global :D

sils1297: you have the traverse function which is invoked multiple times recursively and this variable is global across all of those "instances" of that function right?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: Not really. The traversal starts in another file so it's necessary to get to out file. Below we're traversing all the children, anyway.

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: It's local variable from an enclosing scope.

sils1297: yes, technically, but from the perspective of the function it's global because it has state over all invocations of the recursive invocation which makes things complicated to think right?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
Uran198: Then, you'll have to carry this variable around through `_traverse` function. Functions with a lot of parameters aren't very readable.
I'd rather make another function `traverse` and enclose `_traverse` into it. From that one I'll return a complexities. What about this variant?

Uran198:
This is strange:
```
E   ImportError: cannot import name 'change_directory'
```
Probably some issue with versioning.

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,91 @@
...
+                if top_function_level:
+                    # There is no point to move forward, so just return.
+                    return decisions, exits
+            child_decisions, child_exits = self.function_key_points(child)
+            decisions += child_decisions
+            exits += child_exits
+
+        if top_function_level:
+            # If return statement was avoided.
+            exits += 1
```
sils1297: I don't fully understand this one

Uran198: In C a return statement at the end of the function can be avoided.

12:12, пт, 25.03.2016 Lasse Schuirmann <notifications@github.com>:

sils1297: but then you need to check if there's a return statement and only add 1 if there is no return statement at top level right?

Uran198: Yeah. If there is the return statement I'll be out of this function in the
loop above.

12:31, пт, 25.03.2016 Lasse Schuirmann <notifications@github.com>:

sils1297: ah right, maybe clarify in the comment?

Uran198: I did it. Probably, not very good, but can't find a better way. :( Any
proposals? :)

12:42, пт, 25.03.2016 Lasse Schuirmann <notifications@github.com>:

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+
+        file = cursor.location.file
+
+        if file is not None and file.name != filename:
+            # There is nothing to do in another file.
+            return
+
+        for child in cursor.get_children():
+            if cursor.kind == CursorKind.FUNCTION_DECL:
```
sils1297: this condition belongs outside the loop for logical and performance reasons likewise

bears/c_languages/codeclone_detection/ClangComplexityBear.py
sils1297: `Implicit return statement`?

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,91 @@
...
+    def complexities(self, cursor, filename):
+        """
+        Calculates cyclomatic complexities of functions.
+        """
+
+        file = cursor.location.file
+
+        if file is not None and file.name != filename:
+            # There is nothing to do in another file.
+            return
```
sils1297: I think you'll get this covered if you add an include to a test file

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/278/commits/f99e263025cafed84a2dc28c23e01f3f8fa3b005

bears/c_languages/codeclone_detection/ClangComplexityBear.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if cursor.kind == CursorKind.FUNCTION_DECL:
+            child = next((child for child in cursor.get_children()
+                          if child.kind != CursorKind.PARM_DECL),
+                         None)
```
sils1297: are you taking only the first child on purpose? Using `next` for that is a it weird IMO...

Uran198: Yes. Why?

sils1297: I've never seen it before, but it's efficient indeed.

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/278/commits/9d43bcc2643fe4cd520c7c639e6b75b221e391b7

sils1297:
@Uran198 wanna merge it?

Uran198:
Let me try. This will be fun:)

Uran198:
@rultor merge

Issue #280, opened 2016-03-22
Add CSSAutoPrefixBear


mr-karan:
> PostCSS plugin to parse CSS and add vendor prefixes to CSS rules using values from Can I Use. It is recommended by Google and used in Twitter, and Taobao.

https://github.com/postcss/autoprefixer

^ Would like to do this

PR #281, opened 2016-03-22
bears/css: Add AutoPrefixBear

mr-karan:
This bear uses PostCSS plugin autoprefixer to add
vendor prefixes to CSS rules.

Fixes https://github.com/coala-analyzer/coala-bears/issues/280

### DIFF
--- a/bears/css/CSSAutoPrefixBear.py
+++ b/bears/css/CSSAutoPrefixBear.py
@@ -1,0 +1,18 @@
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+
+class CSSAutoPrefixBear(Lint, LocalBear):
+    executable = "postcss"
+    arguments = "--use autoprefixer {filename}"
+    prerequisite_command = ['postcss', '--use', 'autoprefixer']
+    prerequisite_fail_msg = "Autoprefixer is not installed."
+    diff_message = "Add vendor prefixes to CSS rules."
+    gives_corrected = True
+
+    def run(self, filename, file):
+        '''
+        This bear adds vendor prefixes to CSS rules using ``autoprefixer``
+        utility.
+        '''
+        return self.lint(filename, file)
--- a/tests/css/CSSAutoPrefixBearTest.py
+++ b/tests/css/CSSAutoPrefixBearTest.py
@@ -1,0 +1,22 @@
+from bears.css.CSSAutoPrefixBear import CSSAutoPrefixBear
+from tests.LocalBearTestHelper import verify_local_bear
+
+
+good_file = """
+.example {
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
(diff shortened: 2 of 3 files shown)


tests/css/CSSAutoPrefixBearTest.py
```diff
@@ -0,0 +1,33 @@
...
+from tests.LocalBearTestHelper import verify_local_bear
+
+
+good_file = """
+.example {
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+    -webkit-transition: all .5s;
```
AbdealiJK: Can't we just have display and remove the background, user-select etc ?

tests/css/CSSAutoPrefixBearTest.py
```diff
@@ -0,0 +1,33 @@
...
+            user-select: none;
+    background: -webkit-linear-gradient(top, white, black);
+    background: linear-gradient(to bottom, white, black);
+}
+""".splitlines(keepends=True)
+
+bad_file = """
+.example {
+    display: flex;
+    transition: all .5s;
```
AbdealiJK: same - why have so many ?

bears/css/CSSAutoPrefixBear.py
```diff
@@ -0,0 +1,16 @@
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+
+class CSSAutoPrefixBear(Lint, LocalBear):
+    executable = "postcss"
```
AbdealiJK: It looks as if `postcss` is the executable and `autoprefixer` is a plugin for it.

Should we have `PostCSSBear` and add an argument for this then ?

bears/css/CSSAutoPrefixBear.py
```diff
@@ -0,0 +1,16 @@
...
+    arguments = "--use autoprefixer {filename}"
```
AbdealiJK: We never test if autoprefixer is installed - which means the test will fail instead of being skipped

bears/css/CSSAutoPrefixBear.py
mr-karan: > It looks as if postcss is the executable and autoprefixer is a plugin for it.

That *is* indeed the case. PostCSS comes with a CLI, hence it's the executable and supports many plugins which can be installed using `npm` and used with `--use` argument. https://github.com/postcss/postcss/blob/master/docs/plugins.md

sils1297:
neat! https://github.com/coala-analyzer/coala-bears/pull/281/commits/7720a739caf287b6ff0336cfff65992b27dc29ef ack

sils1297:
@mr-karan can you give me an asciinema again? :)

sils1297:
@rultor merge

AbdealiJK:
This shouldn't have been merged ...
Have you seen the review on top?

Issue #282, opened 2016-03-22
bears/css: Add PostCSSBear


mr-karan:
https://github.com/postcss/postcss

A Bear for support for PostCSS plugins.

Makman2:
If you don't mind I would take this over, as your PR ist the oldest one :D

Adrianzatreanu:
@Makman2 hey :) ?

PR #283, opened 2016-03-22
bears/css: Add PostCSSBear

mr-karan:
This bear uses PostCSS CLI and plugins are
to be defined in .coafile by the user.
The default plugin is `autoprefixer`

Fixes https://github.com/coala-analyzer/coala-bears/issues/282


bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,28 @@
...
+        if postcss_plugins:
+            for plugin in postcss_plugins:
+                self.prerequisite_command = ['postcss', '--use ', plugin]
+                self.prerequisite_fail_msg = "Plugin {} not installed".format(
+                    plugin)
+                self.arguments += (" --use "+plugin)
+        else:
+            # PostCSS needs atleast one plugin to run.
+            # Autoprefixer is the default plugin in that case.
+            self.arguments = "--use autoprefixer"
```
sils1297: just error out in that case, not?

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,28 @@
...
+from coalib.bears.LocalBear import LocalBear
+from coalib.settings.Setting import typed_list
+
+
+class PostCSSBear(Lint, LocalBear):
+    executable = "postcss"
+    diff_message = "PostCSS plugins in action"
+    gives_corrected = True
+
+    def run(self, filename, file, postcss_plugins: typed_list(str)=[]):
```
sils1297: don't give a default so coala will ask the user for it, you should document the param a bit better also

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,28 @@
...
+        '''
+        This bear adds ``PostCSS`` utilities to analyze CSS
+        code.
```
sils1297: missing newwline

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+from coalib.settings.Setting import typed_list
+
+
+class PostCSSBear(Lint, LocalBear):
+    executable = "postcss"
+    diff_message = "PostCSS plugins in action."
```
AbdealiJK: Really ? No better possible message ?

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+    gives_corrected = True
+
+    def run(self, filename, file, postcss_plugins: typed_list(str)=[]):
+        '''
+        This bear adds ``PostCSS`` utilities to analyze CSS code.
+
+        :param postcss_plugins:     List of PostCSS plugins to use with
```
AbdealiJK: Unnecessary whitespace after param

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+                                    this bear which can be defined in
+                                    .coafile.
+        '''
+        if postcss_plugins:
```
AbdealiJK: Would be nicer to do `if not postcss_plugin` and then you don't need an else at all.
makes it easier to read

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+            for plugin in postcss_plugins:
+                self.prerequisite_command = ['postcss', '--use ', plugin]
+                self.prerequisite_fail_msg = "Plugin {} not installed".format(
+                    plugin)
+                self.arguments += (" --use "+plugin)
```
AbdealiJK: space for `+`
unnecessary ()

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+        '''
+        This bear adds ``PostCSS`` utilities to analyze CSS code.
+
+        :param postcss_plugins:     List of PostCSS plugins to use with
+                                    this bear which can be defined in
+                                    .coafile.
+        '''
+        if postcss_plugins:
+            for plugin in postcss_plugins:
+                self.prerequisite_command = ['postcss', '--use ', plugin]
```
AbdealiJK: This ... isn't used anywhere.
Hmm. I wonder how we can do this.
@sils1297  ?

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+                self.prerequisite_fail_msg = "Plugin {} not installed".format(
+                    plugin)
+                self.arguments += (" --use "+plugin)
+            self.arguments += " {filename}"
+            return self.lint(filename, file)
+        else:
+            self.err("Please specify at least one plugin name.")
```
AbdealiJK: Does postcss not do anything on it's own without plugins ?

AbdealiJK: It basically sounds like coala for css :P

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+                                    this bear which can be defined in
+                                    .coafile.
+        '''
+        if postcss_plugins:
+            for plugin in postcss_plugins:
+                self.prerequisite_command = ['postcss', '--use ', plugin]
+                self.prerequisite_fail_msg = "Plugin {} not installed".format(
+                    plugin)
+                self.arguments += (" --use "+plugin)
+            self.arguments += " {filename}"
```
AbdealiJK: Just wondering ... does it have stdin ?

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,29 @@
...
+    def run(self, filename, file, postcss_plugins: typed_list(str)=[]):
+        '''
+        This bear adds ``PostCSS`` utilities to analyze CSS code.
+
+        :param postcss_plugins:     List of PostCSS plugins to use with
+                                    this bear which can be defined in
+                                    .coafile.
+        '''
+        if postcss_plugins:
+            for plugin in postcss_plugins:
```
AbdealiJK: Um, Can't we just "--use" multiple plugins ? Would be much faster

mr-karan: I tried that, it only uses the first plugin in the argument.

bears/css/PostCSSBear.py
mr-karan: Don't think so, not mentioned in the docs.

bears/css/PostCSSBear.py
mr-karan: > The tool itself is a Node.js module that parses CSS into an abstract syntax tree (AST); passes that AST** through any number of “plugin” functions;** and then converts that AST back into a string, which you can output to a file. Each function the AST passes through may or may not transform it; sourcemaps will be generated to keep track of any changes.
The AST provides a straightforward API that developers can use to write plugins

http://davidtheclark.com/its-time-for-everyone-to-learn-about-postcss/

So yeah, without plugins I don't think it can do anything. Much like coala-bears doesn't do anything without specifying a bear.

postcss

bears/css/PostCSSBear.py
mr-karan: The format is `--use plugin_1 --use plugin_2`

AbdealiJK: Ahh, nice. Let's use that then ?
Also, what happens if I give `--use plugin1 --use non-existent-plugin` ?
if there's an error message can we show that ?

mr-karan: ` self.arguments += " --use " + plugin` Is in a `for` loop, so we are already formatting as required. 

For the second part, yes.
`Error: Cannot find module 'inavlidplugin'`

bears/css/PostCSSBear.py
sils1297: yes the message needs to be better, maybe we can make some custom ones for the different plugins?

bears/css/PostCSSBear.py
sils1297: we can't since we allow the user to use arbitrary plugins, we have to do prerequisites checking at runtime if we wanted to d that

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,27 @@
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+from coalib.settings.Setting import typed_list
+
+
+class PostCSSBear(Lint, LocalBear):
+    executable = "postcss"
+    gives_corrected = True
+
+    def run(self, filename, file, postcss_plugins: typed_list(str)=[]):
```
sils1297: why do you give an empty list as default and then error out? Just give no default so the user will be asked

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,27 @@
...
+
+        :param postcss_plugins: List of PostCSS plugins to use with
+                                this bear which can be defined in
+                                .coafile.
+        '''
+        if not postcss_plugins:
+            self.err("Please specify at least one plugin name.")
+            return []
+
+        for plugin in postcss_plugins:
```
AbdealiJK: You need to do more analysis on the invalidplugins.
If an invalid plugin is given then you mentioned postcss gives an error message.
 - Is that error message in stdout or stderr ?
 - Is the "corrected file" in stdout or stderr ?
We need to catch that and make Lint [_print_errors()](https://github.com/coala-analyzer/coala/blob/master/coalib/bearlib/abstractions/Lint.py#L102) with it

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,27 @@
...
+            self.arguments += " --use " + plugin
+            self.arguments += " {filename}"
```
AbdealiJK: Do we need to `+=` filename for every `--use` ?

bears/css/PostCSSBear.py
```diff
@@ -0,0 +1,27 @@
...
+            self.diff_message = "PostCSS plugin {} is analyzing.".format(
```
AbdealiJK: This really doesn't work like this !!!!
The last diff_message will be used. The intermediate ones are useless.

sils1297:
@mr-karan ping? What's the state?

AbdealiJK:
fc12178 needs work

Makman2:
https://github.com/coala-analyzer/coala-bears/pull/809 takes over

Makman2:
@mr-karan if you still want to work on that bear, ping me and you can take my branch :)

PR #286, opened 2016-03-23
Flake8: WIP

srisankethu:


bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,37 @@
...
+
+    """ Checks the code according to pep8 and pyflake8 standards """
+
+    self.gives_corrected = False
+    def lint(self, filename, file):
+        output_regex = re.compile(r'.*\.py:(?P<line>\d):(?P<column>\d):(?P<message>.*)')
+        flake8_style = get_style_guide(
+        config_file=DEFAULT_CONFIG, ignore=(), max_complexity=-1)
+        report = flake8_style.check_files()
+        output = report.get_file_results()
```
srisankethu: getting an error which says no _dereferred_print attribute in get_file_results
https://github.com/PyCQA/pycodestyle/blob/c8e0911f5df9d587e16bccf70cd48cac05d9c49d/pep8.py#L1749

PR #344, opened 2016-04-24
Sils/pmdcpd

sils1297:


bears/java/JavaPMDBear.py
```diff
@@ -0,0 +1,98 @@
...
+        :param check_strings:
+            Check for String, StringBuffer and StringBuilder instances. See
+            http://pmd.github.io/pmd-5.4.1/pmd-java/rules/index.html#String_and_StringBuffer
+            for more information on this matter.
+        :param check_unnecessary:
+            Checks for unnecessary code. See
+            http://pmd.github.io/pmd-5.4.1/pmd-java/rules/index.html#Unnecessary
+            for more information on this matter.
+        :param check_unused:
+            Check for unused code. See
```
sils1297: what's the difference between `unnecessar` and `unused` code?

bears/general/CPDBear.py
```diff
@@ -0,0 +1,41 @@
+import re
+import os
+
+from coalib.bearlib.abstractions.Lint import Lint
+from coalib.bears.LocalBear import LocalBear
+
+
+class CPDBear(LocalBear, Lint):
```
sils1297: alright this has to become a GlobalBear, can you change it?

bears/general/CPDBear.py
```diff
@@ -0,0 +1,41 @@
...
+    executable = 'run.sh'
+    arguments = "cpd --minimum-tokens 10 --files {filename}"
```
sils1297: minimum tokens should be configurable

bears/general/CPDBear.py
SanketDG: didn't start work on CPDBear yet, that why marked as WIP

bears/java/JavaPMDBear.py
aachiritoaei: Unused code refers effectively to code that is not used - local variables, private methods, formal parameters, private fields. 
Unnecessary code refers to meaningless code - return statements in a constructor, useless parantheses,  operations on immutable objects like `string.replace(something)` instead of `string = string.replace(something)` :)

sils1297: hm can we make this clearer in the comments? Like way clearer!

bears/general/CPDBear.py
```diff
@@ -53,12 +56,21 @@ def run(self, language: str, minimum_tokens: int=10,
...
             return
 
+        options = {
+            "--ignore-annotations": ignore_annotations,
+            "--ignore-identifiers": ignore_identifiers,
+            "--ignore-literals": ignore_literals,
+            "--ignore-usings": ignore_usings,
+            "--skip-duplicate-files": skip_duplicate_files,
+        }
+        rules = ' '.join(key for key in options if options[key])
```
sils1297: that won't work, every argument needs to be an own string in the tuple

bears/general/CPDBear.py
```diff
@@ -63,15 +67,39 @@ def run(self, language: str, minimum_tokens: int=10,
...
+        for option in options:
+            if options[option] is True:
+                arguments += option,
+
+        stdout_output, _ = run_shell_command(" ".join(map(str, arguments)))
+
+        if stdout_output:
+            root = ET.fromstring(stdout_output)
+
+            message = "Duplicate code found."
```
sils1297: rather inline this in line 105, no need for a variavle

bears/general/CPDBear.py
```diff
@@ -63,15 +67,39 @@ def run(self, language: str, minimum_tokens: int=10,
...
+
+            for duplication in root.findall('duplication'):
+
+                length = int(duplication.attrib['lines'])
+
+                affected_code = list()
```
sils1297: I'd remove lines 90 and 92 for readability

bears/general/CPDBear.py
```diff
@@ -75,8 +75,8 @@ def run(self, language: str, minimum_tokens: int=10,
                      "--files", files,
                      "--format", "xml")
 
-        for option in options:
-            if options[option] is True:
+        for option, enable in options.items():
+            if enable is True:
```
Makman2: come on :)

sils1297: write us a filter expression

Makman2: I meant the `if enable is True` -> `if enable`

Makman2: and yeah:
`arguments +=  [option for option, enable in options.items() if enable]`

Makman2: or it's a tuple, so use `tuple()` before generator :)

sils1297: we're using the generator since a while, push happening soon. I still think identity checking for true makes sense

sils1297:
reworked in https://github.com/coala-analyzer/coala-bears/pull/381

PR #379, opened 2016-04-28
Origin/sims1253/jpdcpd

sims1253:


sils1297:
reworked in https://github.com/coala-analyzer/coala-bears/pull/381

PR #380, opened 2016-04-28
GitCommitBear: Divide metadata signature

Makman2:
The metadata is now retrieved from `check_shortlog` and `check_body` to
avoid signature duplication overhead.

### DIFF
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -130,7 +130,7 @@ class GitCommitBear(GlobalBear):
                 yield Result(self,
                              "This commit seems to be marked as work in "
                              "progress and should not be used in production. "
-                             "Tread carefully.", severity=RESULT_SEVERITY.MAJOR)
+                             "Treat carefully.", severity=RESULT_SEVERITY.MAJOR)
 
     def check_imperative(self, paragraph):
         """
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -127,10 +127,10 @@ class GitCommitBear(GlobalBear):
                              "bad words are '{}'".format(bad_word))
         if shortlog_wip_check:
             if "wip" in shortlog.lower()[:4]:
-                yield Result(self,
-                             "This commit seems to be marked as work in "
-                             "progress and should not be used in production. "
-                             "Treat carefully.", severity=RESULT_SEVERITY.MAJOR)
+                yield Result(
+                    self,
+                    "This commit seems to be marked as work in progress and "
+                    "should not be used in production. Treat carefully.")
 
     def check_imperative(self, paragraph):
         """
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -90,8 +90,8 @@ class GitCommitBear(GlobalBear):
         :param shortlog_trailing_period: Whether a dot shall be enforced at end
                                          end or not (or ``None`` for "don't
                                          care").
-        :param shortlog_wip_check:       Whether a wip in the shortlog should
-                                         yield a major result or not.
+        :param shortlog_wip_check:       Whether a "WIP" in the shortlog text
+                                         should yield a result or not.
         """
         diff = len(shortlog) - shortlog_length
         if diff > 0:
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -8,6 +8,7 @@ from coalib.misc.ContextManagers import change_directory
 from coalib.misc.Shell import run_shell_command
 from coalib.results.Result import Result
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.FunctionMetadata import FunctionMetadata
 
 
 class GitCommitBear(GlobalBear):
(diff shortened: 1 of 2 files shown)


sils1297:
scheduling for next release, not user relevant but breaking potential

sils1297:
@Makman2 be sure to cancel your old builds, this stuff is not release relevant, please avoid pushing and thus blocking us with this non release relevant PR!

Adrianzatreanu:
ack 95ed972

bears/vcs/git/GitCommitBear.py
```diff
@@ -172,15 +158,17 @@ def check_imperative(self, paragraph):
                      " maxent_treebank_pos_tagger averaged_perceptron_tagger`")
             return
 
-    def check_body(self, body_line_length, force_body, body):
+    def check_body(self, body,
+                   body_line_length: int=72,
```
Adrianzatreanu: why do we use 72 and not 80?

bears/vcs/git/GitCommitBear.py
```diff
@@ -127,7 +127,7 @@ def check_shortlog(self, shortlog,
                 yield Result(self,
                              "This commit seems to be marked as work in "
                              "progress and should not be used in production. "
-                             "Treat carefully.", severity=RESULT_SEVERITY.MAJOR)
+                             "Treat carefully.", severity=RESULT_SEVERITY.INFO)
```
Adrianzatreanu: should it not be a WARN then?

bears/vcs/git/GitCommitBear.py
Makman2: I copied the earlier value, but I think too that 72 is the standard for the body length^^

SanketDG: yes 72 is the preferred length http://chris.beams.io/posts/git-commit/

bears/vcs/git/GitCommitBear.py
```diff
@@ -102,7 +89,6 @@ def check_shortlog(self,
                                          care").
         :param shortlog_wip_check:       Whether a wip in the shortlog should
                                          yield a major result or not.
```
SanketDG: Could you change the description to
Whether a WIP in the shortlog text should yield a major result or not.

bears/vcs/git/GitCommitBear.py
```diff
@@ -20,46 +21,33 @@ def check_prerequisites(cls):
...
+    def get_shortlog_checks_metadata(cls):
+        return FunctionMetadata.from_function(
+            cls.check_shortlog,
+            omit={"self", "shortlog"})
+
+    @classmethod
+    def get_body_checks_metadata(cls):
+        return FunctionMetadata.from_function(
+            cls.check_body,
+            omit={"self", "body"})
```
sils1297: I'm nnot a fan of this kind of function, merely wrapping another...

bears/vcs/git/GitCommitBear.py
sils1297: yeah the FP rate shouldn't be high, just use severity normal IMO

sils1297:
both other commits need work IMO and a rebase is also needed

bears/vcs/git/GitCommitBear.py
Makman2: it defines which omits to use, I need this function at two places so imo it makes sense^^

Makman2:
reack e2c9964

SanketDG:
ack c67e1ff 5d7ae07

SanketDG:
ack 5d7ae07

bears/vcs/git/GitCommitBear.py
Makman2: @SanketDG you acked the commit, but shall I use `NORMAL` like suggested?

SanketDG: Oh I missed this. yeah `NORMAL` sounds good.

Makman2:
reack 0f0a2d1

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/380/commits/c9d4782beacc1f36871766f2bdc3c32e35826292

bears/vcs/git/GitCommitBear.py
```diff
@@ -20,46 +21,33 @@ def check_prerequisites(cls):
...
+
+    @classmethod
+    def get_metadata(cls):
+        return FunctionMetadata.merge(GlobalBear.get_metadata(),
```
sils1297: why `GlobalBear.get_metadata()`? That's the metadata of the GlobalBear class which isn't relevant is it?

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -81,6 +81,11 @@ def test_check_prerequisites(self):
         finally:
             shutil.which = _shutil_which
 
+    def test_get_metadata(self):
+        self.assertEqual(
+            GitCommitBear.get_metadata().name,
+            "<Merged signature of 'run', 'check_shortlog', 'check_body'>")
```
sils1297: test params too please, I think this implementation doesn't get stuff from run correctly

sils1297:
unack 4bc2596

bears/vcs/git/GitCommitBear.py
Makman2: `run()` contains also settings, that's why I want to use the old "behaviour" too

tests/vcs/git/GitCommitBearTest.py
Makman2: It should because of `GlobalBear.get_metadata()`, and the test would fail if it doesn't merge the signature from `run`. (Otherwise it would be `<Merged signature of 'check_shortlog', 'check_body'>`

sils1297: please test the params, I predict a test failure, `GlobalBear.get_metadata()` doesn't have any self and thus can't hold any info about this classes run definition. This only works because you only test that there *is* a run method merged in.

bears/vcs/git/GitCommitBear.py
sils1297: Wrong. This will get the metadata of `GlobalBear.run` not this class! This is lacking parameters.

Makman2: ah damn right...

Makman2:
reack 458c404, 9439205, 83e3753

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/380/commits/a21010e1223155a2dfdf1df8fbb094b0145fda67

Makman2:
@rultor merge

Issue #433, opened 2016-05-15
FlawFinder integration


sils1297:
Neat stuff: http://www.dwheeler.com/flawfinder/

area/lintbears difficulty/low

PR #428, opened 2016-05-14
Makman2/linterpack3

Makman2:

PR #434, opened 2016-05-15
CSecurityBear

sils1297:
Fixes https://github.com/coala-analyzer/coala-bears/issues/433

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -15,7 +15,7 @@ sudo add-apt-repository -y ppa:staticfloat/juliareleases
 sudo add-apt-repository -y ppa:staticfloat/julia-deps
 sudo add-apt-repository -y ppa:avsm/ppa
 sudo apt-get -y update
-deps="espeak libclang1-3.4 indent mono-mcs chktex hlint r-base julia luarocks verilator cppcheck"
+deps="espeak libclang1-3.4 indent mono-mcs chktex hlint r-base julia luarocks verilator cppcheck flawfinder"
 deps_python_dbus="libdbus-glib-1-dev libdbus-1-dev"
 deps_python_gi="glib2.0-dev gobject-introspection libgirepository1.0-dev python3-cairo-dev"
 deps_perl="perl libperl-critic-perl"
--- a/bears/c_languages/CSecurityBear.py
+++ b/bears/c_languages/CSecurityBear.py
@@ -1,0 +1,24 @@
+from coalib.bearlib.abstractions.Linter import linter
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.Setting import typed_list
+
+
+@linter(executable="flawfinder", output_format="regex",
+        output_regex=r'.+:(?P<line>\d+):(?P<column>\d+):\s*'
+                     r'\[(?P<severity>\d)\]\s*'
+                     r'\((?P<origin>.+)\) (?P<message>.+)',
+        severity_map={"5": RESULT_SEVERITY.MAJOR,
+                      "4": RESULT_SEVERITY.NORMAL, "3": RESULT_SEVERITY.NORMAL,
+                      "2": RESULT_SEVERITY.INFO, "1": RESULT_SEVERITY.INFO})
+class CSecurityBear:
+    """
+    Report possible security weaknesses for C/C++.
+
+    For more information, consult <http://www.dwheeler.com/flawfinder/>.
+    """
+
+    LANGUAGES = "C", "C++"
+
+    @staticmethod
+    def create_arguments(filename, file, config_file):
+        return "--columns", "--dataonly", "--quiet", "--singleline", filename
--- a/bears/c_languages/ClangBear.py
+++ b/bears/c_languages/ClangBear.py
@@ -62,7 +62,7 @@ class ClangBear(LocalBear):
                     affected_code = diffs[filename].affected_code(filename)
 
             # Still no affected code? Position is the best we can get...
-            if len(affected_code) == 0:
+            if len(affected_code) == 0 and diag.location.file is not None:
                 affected_code = (SourceRange.from_values(
                     diag.location.file.name,
                     diag.location.line,
--- a/tests/c_languages/CSecurityBearTest.py
+++ b/tests/c_languages/CSecurityBearTest.py
@@ -1,0 +1,17 @@
+from bears.c_languages.CSecurityBear import CSecurityBear
+from tests.LocalBearTestHelper import verify_local_bear
+
+
+good_file = """
+void demo() {
+}""".splitlines(keepends=True)
+
+bad_file = """
+void demo(char *a) {
+    strcpy(a, '\n');  // CWE 120
+}""".splitlines(keepends=True)
+
+
+CSecurityBearTest = verify_local_bear(CSecurityBear,
+                                      valid_files=(good_file,),
+                                      invalid_files=(bad_file,))


bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,23 @@
...
+@linter(executable="flawfinder", output_format="regex",
+        output_regex=r'.+:(?P<line>\d+):(?P<column>\d+):\s*'
+                     r'\[(?P<severity>\d)\]\s*'
+                     r'\((?P<origin>.+)\) (?P<message>.+)',
+        severity_map={"5": RESULT_SEVERITY.MAJOR,
+                      "4": RESULT_SEVERITY.NORMAL, "3": RESULT_SEVERITY.NORMAL,
+                      "2": RESULT_SEVERITY.INFO, "1": RESULT_SEVERITY.INFO})
+class CSecurityBear:
+    """
+    Reports possible security weaknesses for C link languages.
```
arafsheikh: like?

bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,23 @@
+from coalib.bearlib.abstractions.Linter import linter
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+
+
+@linter(executable="flawfinder", output_format="regex",
```
sils1297: flawfinder is a python binary that is istalled at the system. It doesn't work within our virtualenv. Anyone any idea on how we can circumvent this? Making an own flawfinder binary explicitly using /usr/bin/python2.7 to run the original or something?

CC @AbdealiJK

sils1297: solved with the sed expression above, similar to the archlinux aur package

sils1297:
tests will probably fail because @Makman2 broke coala :)

sils1297:
needs a rebase anyway after https://github.com/coala-analyzer/coala-bears/pull/428 is merged

.ci/deps.sh
```diff
@@ -99,3 +99,6 @@ fi
 # Tailor (Swift) commands
 curl -fsSL https://tailor.sh/install.sh | sed 's/read -r CONTINUE < \/dev\/tty/CONTINUE=y/' > install.sh
 sudo bash install.sh
+
+# FlawFinder will be confused by our python 3 venv
+sudo sed -i 's/env python/python2/' /usr/bin/flawfinder
```
AbdealiJK: Is this ok ?
I honestly don't like things like this ... It's meddling with things it sholdnt.

I'd prefer if we make a `.ci/install.flawfinder.sh` which simply installs it using [this](https://gist.github.com/FlorianWolters/07c866500a1fe6f6098c#flawfinder).

@sils1297 ? Is the shebang hack ok with you ?

bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,23 @@
...
+        output_regex=r'.+:(?P<line>\d+):(?P<column>\d+):\s*'
+                     r'\[(?P<severity>\d)\]\s*'
```
AbdealiJK: Is the square brackets around severity needed ? Isnt that redundant ?
Same with the curved brackets around origin

bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,23 @@
...
+                     r'\((?P<origin>.+)\) (?P<message>.+)',
+        severity_map={"5": RESULT_SEVERITY.MAJOR,
+                      "4": RESULT_SEVERITY.NORMAL, "3": RESULT_SEVERITY.NORMAL,
+                      "2": RESULT_SEVERITY.INFO, "1": RESULT_SEVERITY.INFO})
+class CSecurityBear:
+    """
+    Reports possible security weaknesses for C like languages.
```
AbdealiJK: Report ? (Newcomer ?)

AbdealiJK: I wouldnt call it C like ... C like would include C#, Obj-C, etc right ?
This is for C and C++ only according to the documentation page

bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,23 @@
+from coalib.bearlib.abstractions.Linter import linter
```
AbdealiJK: I think that if we're adding new linter bears we should add all the configs from now on ... The current system is causing double work where someone just writes boilerplate code to get it into the CI and there is another issue to add configs to it ...
Our linter stuff is sufficiently simple that they can be both added to the same issue

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,37 @@
+import json
+
+from coalib.bearlib.abstractions.Linter import linter
+from coalib.results.Result import Result
+from coalib.results.SourceRange import SourceRange
+
+
+@linter(executable='reek',
+        use_stdin=True)
```
AbdealiJK: Unnecessary indentation

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,37 @@
...
+class RubySmellBear:
+    """
+    Detect code smells in Ruby source code.
```
AbdealiJK: Smell to official documentation ?

AbdealiJK: Link * XD

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    LANGUAGES = "Ruby"
+
+    @staticmethod
+    def create_arguments(filename, file, config_file):
+        return '--format', 'json'
+
+    def process_output(self, output, filename, file):
+        output = json.loads(output) if output else []
```
AbdealiJK: Do we need a JSONDecodeError ?
Does it always give a json output even for corner cases like giving a directory instead of a file, giving a binary file, etc ?
Also, can we always assure that the `output` is the complete output from the process ? Can it hang and have a half completed output ?

tests/ruby/RubySmellBearTest.py
```diff
@@ -0,0 +1,11 @@
+from bears.ruby.RubySmellBear import RubySmellBear
+from tests.LocalBearTestHelper import verify_local_bear
+
+
+good_file = ('',)
```
AbdealiJK: Not a great testcase IMO ... Possibly something better ? Or make an issue

PS: The linter may just give no output for an empty file and so on.

bears/c_languages/ClangBear.py
```diff
@@ -62,7 +62,7 @@ def run(self, filename, file, clang_cli_options: typed_list(str)=None):
                     affected_code = diffs[filename].affected_code(filename)
 
             # Still no affected code? Position is the best we can get...
-            if len(affected_code) == 0:
+            if len(affected_code) == 0 and diag.location.file is not None:
```
AbdealiJK: Needs testcase

.ci/deps.sh
SanketDG: This looks pretty specific.

IMHO `make`-ing things could lead to increased CI times.

SanketDG: Also, does your solution work under a python3 venv?
flawfinder is written in python2.

AbdealiJK: We enable both 3 and 2.7 in out pyenv. Check [line](https://github.com/coala-analyzer/coala-bears/pull/434/files/8b518fede94de64930cf2e8c366ed7c66ec29a4b#diff-5c64bb36a9fa177d5dfeb57b93e57a6cR50)

bears/c_languages/CSecurityBear.py
sils1297: they are escaped! They appear like this in the output

bears/ruby/RubySmellBear.py
sils1297: if it hangs there will usually be no json output because you usually first create the results and then output then, solely from a technical perspective

bears/c_languages/ClangBear.py
sils1297: I have no idea how to reproduce that yet :P

.ci/deps.sh
sils1297: `make`ing flawfinder doesn't do anything but the same as the current package installation. The reason this sed is needed is that flawfinder is written with python 2 only systems in mind. It won't work unless you sed it. The ArchLinux AUR package uses a sed expression too for that reason.

bears/c_languages/CSecurityBear.py
sils1297: flawfinder is particularly hard to configure, it features only a regex and then only shows messages that match this regex, I can make a setting out of that I guess

bears/ruby/RubySmellBear.py
AbdealiJK: Still true

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+                        uncommunicative_method_name: bool=True,
+                        uncommunicative_module_name: bool=True,
+                        uncommunicative_param_name: bool=True,
+                        uncommunicative_var_name: bool=True,
+                        unused_params: bool=True,
+                        unused_private_method: bool=False,
+                        utility_function: bool=True):
+        """
+        :param attribute:
+            Warns if a setter in a class is detected.
```
AbdealiJK: Hmmm, isnt this a very generic name which is not applicable to other linters ?
Could we possibly have a better variable name which is better to understand ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+        :param boolean_parameter:
+            Warns if a boolean parameter in a function is detected (control
+            coupling).
+        :param class_variable:
+            Warns if a class variable is detected.
+        :param control_parameter:
+            Warns if a parameter controls the function behaviour (control
+            coupling).
+        :param data_clump:
```
AbdealiJK: Please explain that the items appear in function/class  parameter lists. I initially thought it was checking lines that get repeated frequently

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            Warns when the same two or three items frequently appear together.
+        :param duplicate_method_call:
```
AbdealiJK: Is there no setting for this already in radon ? I think it does a similar thing ? That or CodeCloneBear and others like that ?

AbdealiJK: We **should** have a Generic Bear for this. It's actually a very simple thing to check right ? Basically see if a set of lines appear more than once in a code base.
There should be a good algorithm which can do this (It comes under substring search problems i think). If my understanding is right, could you make an issue please ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            Warns when two fragments of code look nearly identical, or when two
+            fragments of code have nearly identical effects at some conceptual
+            level.
+        :param feature_envy:
```
AbdealiJK: I'm wondering whether these "fun" interesting names should be used or simpler to understand names should be used. Opinion ?
Like feature_envy vs allow_many_feature_references or so ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            Warns when a block contains another block.
+        :param nil_check:
+            Warns about nil checks.
+        :param prima_donna_method:
+            Warns about methods whose names end with an exclamation mark.
+        :param repeated_coditional:
+            Warns about repeated conditionals.
+        :param too_many_instance_variables:
+            Warns for too many instance variables.
+        :param too_many_methods:
```
AbdealiJK: too_many_methods_in_class ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            Warns about ``#initialize`` methods in modules.
+        :param nested_iterators:
+            Warns when a block contains another block.
+        :param nil_check:
+            Warns about nil checks.
+        :param prima_donna_method:
+            Warns about methods whose names end with an exclamation mark.
+        :param repeated_coditional:
+            Warns about repeated conditionals.
+        :param too_many_instance_variables:
```
AbdealiJK: too_many_instance_variables_in_class ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            passed.
+        :param module_initialize:
+            Warns about ``#initialize`` methods in modules.
+        :param nested_iterators:
+            Warns when a block contains another block.
+        :param nil_check:
+            Warns about nil checks.
+        :param prima_donna_method:
+            Warns about methods whose names end with an exclamation mark.
+        :param repeated_coditional:
```
AbdealiJK: Typo: conditional

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+        :param uncommunicative_method_name:
+            Warns about method names which are not communicating the purpose
+            of the method well.
+        :param uncommunicative_module_name:
+            Warns about module names which are not communicating the purpose
+            of the module well.
+        :param uncommunicative_param_name:
+            Warns about parameter names which are not communicating the purpose
+            of the parameter well.
+        :param uncommunicative_var_name:
```
AbdealiJK: bad_variable_name ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            Warns about nil checks.
+        :param prima_donna_method:
+            Warns about methods whose names end with an exclamation mark.
+        :param repeated_coditional:
+            Warns about repeated conditionals.
+        :param too_many_instance_variables:
+            Warns for too many instance variables.
+        :param too_many_methods:
+            Warns if a class has too many methods.
+        :param too_many_statements:
```
AbdealiJK: Wouldnt long_method be more appropriate ?

bears/ruby/RubySmellBear.py
```diff
@@ -0,0 +1,162 @@
...
+            Wanrs about huge methods.
```
AbdealiJK: Typo: Warns

AbdealiJK:
ack f87c680

Also, circleci failed ?

bears/ruby/RubySmellBear.py
sils1297: Not so sure, can you file it and be a bit clearer about the approach?

bears/ruby/RubySmellBear.py
sils1297: naming is hard, I did mostly go for theirs because they have thought about them more than we ever can and I can't think of better names.

bears/ruby/RubySmellBear.py
sils1297: in class is redundant, names should be short

bears/ruby/RubySmellBear.py
sils1297: in class is redundant, names should be short (methods are always in a class)

SanketDG:
@sils1297 From what I understand, the ubuntu version of flawfinder doesn't even contain the `--regex` option. The probable solution is to now just download the file from http://www.dwheeler.com/flawfinder/flawfinder into /usr/bin/ or some other folder (and add it to $PATH),
and give it a `+x` chmod

sils1297:
oh holy shit, that sucks. We could just apply the regex by ourselves.

bears/ruby/RubySmellBear.py
sils1297: we can rename everything with https://github.com/coala-analyzer/coala-bears/issues/460

bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,37 @@
...
+from coalib.settings.Setting import typed_list
+
+
+@linter(executable="flawfinder", output_format="regex",
+        output_regex=r'.+:(?P<line>\d+):(?P<column>\d+):\s*'
+                     r'\[(?P<severity>\d)\]\s*'
+                     r'\((?P<origin>.+)\) (?P<message>.+)',
+        severity_map={"5": RESULT_SEVERITY.MAJOR,
+                      "4": RESULT_SEVERITY.NORMAL, "3": RESULT_SEVERITY.NORMAL,
+                      "2": RESULT_SEVERITY.INFO, "1": RESULT_SEVERITY.INFO})
```
Makman2: that's a weird alignment :O

bears/c_languages/CSecurityBear.py
Makman2: if the message contains actually spaces, use them instead of `\s*`, `\s` feels more programming-language like, be specific :)

bears/c_languages/CSecurityBear.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         exclude_likely_false_positives: bool=True):
+        """
+        :param exclude_likely_false_positives:
+            Exclude results that are likely to be false positives.
+        :param flawfinder_rules:
+            A list of CWE numbers. Only results for those CWEs will be shown. To
+            show e.g. only CWE-78, simply give the ``78`` as a value.
```
Makman2: doesn't exist in the signature!!!

Makman2: and also not used below!

sils1297:
@AbdealiJK I removed the options again as they don't seem to work on ubuntu but still failing. You have an ubuntu right? Any idea?

AbdealiJK:
What @SanketDG said is right. Ubuntu has an older version.
http://manpages.ubuntu.com/manpages/precise/man1/flawfinder.1.html is the mapage for it in precise. It has no regex arg.
Wily has the newer version which has regex arg - http://manpages.ubuntu.com/manpages/wily/man1/flawfinder.1.html

bears/c_languages/CSecurityBear.py
sils1297: major, normal and info are in an own row each

bears/c_languages/CSecurityBear.py
sils1297: why be specific ifI can be generic and not care about what is there actually?

sils1297:
Hey @Asnelchristian @Adrianzatreanu would any of you be interested intaking this over? I currently lack the time.

sils1297:
it's almost ready, some CI issues left to do

Asnelchristian:
sure.

2016-07-05 13:02 GMT+01:00 Lasse Schuirmann <notifications@github.com>:

sils1297:
@Asnelchristian you're my saviour! Ping me if you need any help :)

Adrianzatreanu:
hey @Asnelchristian id be happy to be your reviewer on this one, so ping me once your PR is up :)

PR #423, opened 2016-05-14
upgrade coala

sils1297:
testing it out at least...

PR #436, opened 2016-05-15
Makman2/crosstest

Makman2:
Again a crosstest


sils1297:
not needed, see https://github.com/coala-analyzer/coala-bears/pull/423

Issue #432, opened 2016-05-15
Add link to install instructions in missing prerequisites messages


sils:

sils1297:
not only to the tool

sils1297:
Bluntly assigning @Adrianzatreanu as this is a very simple thing but related to your project which then will automate the installation, right?

Adrianzatreanu:
I guess so.

Adrianzatreanu:
im not so sure where to add which links..

sils1297:
Bears have the prerequisite checking method which has a message, that message should ideally contain instructions how to fix it. Linter has a parameter for it I think? Maybe we need to introduce one here? Something like `install_instructions` CC @Makman2

Adrianzatreanu:
Bear.py has @classmethod
    def check_prerequisites(cls): which returns whether a bear has or not satisfied the prerequisites. i guess i could check using this method for each bear to see if they are satisfied, in which case i throw a link if they arent?

Adrianzatreanu:
ah never mind, linter has @classmethod
        def check_prerequisites(cls): too

Adrianzatreanu:
after discussions, it's been clear what i have to do, ill do this soon :)

Adrianzatreanu:
is this still relevant now that we use cib?

sils1297:
more than ever, we should generate those messages from the Requirement set
instead of manually providig it!

arush0311:
Can I take this?

sils:
@Adrianzatreanu @arush0311 sure, I think the most important thing to address is that the Bear class should generate the installation instructions from the DEPENDENCIES metadata

sils:
though @fneu might have a word to say

sils:
he's about to refactor it. At least he's saying it...

fneu:
No objections from me. Dependency checking will stay in the requirements. I'm not touching those kinds of messages. 


@sils I'm busy driving to Harburg, the GitMate Hangouts call later and our meetup this evening. 
I want the cib stuff done this week. I now there will be GitMate feedback so I'm counting on Sunday.

sils:
but the commands won't be generated by the classes, right? I think it makes sense to wait for your changes and then get as much info as possible about installation and ideally a doc link from whatever you do. E.g. if your function can make sure that an APT package does the job maybe we can generate a URL to the package or so.

fneu:
[...]

PR #445, opened 2016-05-16
WIP: bears: Add prerequirements installation instructions

Adrianzatreanu:
This PR is giving the user a helpful link in case the executable is failing to exist.

Currently WIP as it is only for one bear for test purposes, I'll continue working on others too
in different commits if it is okay. Looking forward to feedback.

Fixes https://github.com/coala-analyzer/coala-bears/issues/432

### DIFF
--- a/bears/dart/DartLintBear.py
+++ b/bears/dart/DartLintBear.py
@@ -2,6 +2,8 @@ from coalib.bearlib.abstractions.Linter import linter
 
 
 @linter(executable='dartanalyzer',
+        executable_check_fail_info='Please see https://github.com/dart-lang/linter#installing '
+                                   'for more information about this.',
         output_format='regex',
         output_regex=r'\[(?P<severity>error|warning)\] (?P<message>.+)\('
                      r'.+, line (?P<line>\d+), col (?P<column>\d+)\)')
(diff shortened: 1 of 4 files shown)


bears/dart/DartLintBear.py
```diff
@@ -2,6 +2,8 @@
 
 
 @linter(executable='dartanalyzer',
+        executable_check_fail_info='Please see the Installation section of page'
+                                   'https://github.com/dart-lang/linter.',
```
sils1297: can't you link directly to that section?

Adrianzatreanu: eh i guess you're right, im going to change this as soon as my circleCI build finishes as i wanna see if it works.

Adrianzatreanu:
@justuswilhelm tests were failing until now, i changed something to see how they're doing.

Adrianzatreanu:
ah this fails because I havent used the coala-decorators module here, will fix this

justuswilhelm:
@Adrianzatreanu You good on this?

Adrianzatreanu:
yes, going to fix this for now:)

Adrianzatreanu:
on second thought, do we really need this? Initially i thought it would help me on decentralizing bears, but it doesn't at all honestly. i'm still doing it if coala needs this, but i dont see when it would? all this does is add a link on executable missing (which we'd check without this attribute too) and copy pasting the first link on google about the executable ;p which probably any developer does anyway.
@sils1297 @Makman2 @justuswilhelm

Adrianzatreanu:
closing for now, may continue later on.

PR #470, opened 2016-05-30
perlcritic stuff

Makman2:


Adrianzatreanu:
why is this needed :( ?

Makman2:
the problem is when you want to execute tests on systems that don't have perlcritic, `shutil.which` yields `None` and linter doesn't allow this (it makes sense you can't pass `None` as an executable). There seem to be only problems with batch files, it seems windows doesn't allow to execute them without using the extension (via `subprocess.call`), but `shutil.which` recognizes it as a command obviously like the windows shell would do^^

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/470/commits/a3cb319c4fa149373fb1a35160337b31b60f6e09

Makman2:
@rultor merge

Makman2:
ack https://github.com/coala-analyzer/coala-bears/pull/470/commits/64795ac0c87171f98acd19f52574178ce613570a (this is lasse from mischas PC)

Makman2:
@rultor merge

PR #493, opened 2016-06-04
Gnuindent

Asnelchristian:
More configs were added to gnuindent


arafsheikh:
@Asnelchristian You checked out from the old branch, so the previous commits came with it. You need to get rid of the first two commits

PR #503, opened 2016-06-06
uploadTool: Add it

Adrianzatreanu:
The uploadTool is supposed to create a ``requirements.txt`` from the REQUIREMENTS tuple of the MockBear (which is CPPCleanBear for testing purposes) and a ``setup.py``. Ideally, this script will upload the bear to PyPi too.

### DIFF
--- a/bears/Constants.py
+++ b/bears/Constants.py
@@ -4,3 +4,67 @@ from bears import VERSION
 
 # Path to the bears directory
 bears_root = os.path.dirname(__file__)
+
+
+# Constants to be used for the uploadTool to improve automation
+
+classifiers = """[
+              'Development Status :: 4 - Beta',
+
+              'Environment :: Console',
+              'Environment :: MacOS X',
+              'Environment :: Win32 (MS Windows)',
+              'Environment :: X11 Applications :: Gnome',
+
+              'Intended Audience :: Science/Research',
+              'Intended Audience :: Developers',
+
+              'License :: OSI Approved :: GNU Affero General Public License '
+              'v3 or later (AGPLv3+)',
+
+              'Operating System :: OS Independent',
+
+              'Programming Language :: Python :: Implementation :: CPython',
+              'Programming Language :: Python :: 3.3',
+              'Programming Language :: Python :: 3.4',
+              'Programming Language :: Python :: 3.5',
+              'Programming Language :: Python :: 3 :: Only',
+
+              'Topic :: Scientific/Engineering :: Information Analysis',
+              'Topic :: Software Development :: Quality Assurance',
+              'Topic :: Text Processing :: Linguistic']"""
+imports = """
+# Start ignoring PyImportSortBear as imports below may yield syntax errors
+# Stop ignoring
+
+import locale
+import sys
+from subprocess import call
+
+import setuptools.command.build_py
+from setuptools import find_packages, setup
+from setuptools.command.test import test as TestCommand"""
+
+encoding = """try:
+    locale.getlocale()
+except (ValueError, UnicodeError):
+    locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')"""
+
+requirements = """with open('requirements.txt') as requirements:
+    required = requirements.read().splitlines()"""
+
+setup_constants = """author="The coala developers",
+          maintainer="Lasse Schuirmann, Fabian Neuschmidt, Mischa Kruger",
+          maintainer_email=('lasse.schuirmann@gmail.com, '
+                            'fabian@neuschmidt.de, '
+                            'makman@alice.de'),
+          url='http://coala.rtfd.org/',
+          platforms='any',
+          packages=find_packages(exclude=("build.*", "tests", "tests.*")),
+          install_requires=required,
+          package_data={'bears': ["VERSION"],
+                        'bears.java': ['checkstyle.jar', 'google_checks.xml'],
+                        'bears.scala': ['scalastyle.jar',
+                                        'scalastyle_config.xml']},
+          license="AGPL-3.0",
+          entry_points={"coalabears": ["coala_official_bears = bears"]}"""
--- a/bears/uploadTool.py
+++ b/bears/uploadTool.py
@@ -1,0 +1,68 @@
+from Constants import classifiers, imports, encoding, requirements
+from Constants import setup_constants
+from coalib.bears.requirements.PythonRequirement import PythonRequirement
+import shutil
+import os
+import glob
+import re
+
+if not os.path.exists('upload/'):
+    os.mkdir('upload/')
+
+#Reqrs = MockBear.REQUIREMENTS
+
+# for req in Reqrs:
+#    if type(req) == PythonRequirement:
+#        if req.version:
+#            reqs.write(req.package + '==' + req.version + '\n')
+#        else:
+#            reqs.write(req.package + '\n')
+
+bears = glob.glob('**/*Bear.py')
+
+for bear in bears:
+    file = bear
+    bear = re.match(r'.+/(.*)\.py', bear).group(1)
+    if not os.path.exists('upload/' + bear):
+        os.mkdir('upload/' + bear)
+        if not os.path.exists('upload' + bear + '/' + bear):
+            os.mkdir('upload/' + bear + '/' + bear)
+    if not os.path.exists('upload/' + bear + '/' + bear + '/__init__.py'):
+        init = open('upload/' + bear + '/' + bear + '/__init__.py', 'w')
+        init.write(' ')
+        init.close()
+    shutil.copyfile(file, 'upload/' + bear + '/' + bear + '/' + bear + '.py')
+
+    setup = open('upload/' + bear + '/setup.py', "w")
+    reqs = open('upload/' + bear + '/requirements.txt', "w")
+
+    # set env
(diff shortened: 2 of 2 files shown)


bearuploader/setup.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
```
justuswilhelm: What is the purpose of this file?

justuswilhelm: Nevermind, it seems to be the one generated by uploadTool.py

bearuploader/Constants.py
```diff
@@ -0,0 +1,72 @@
...
+          maintainer_email=('lasse.schuirmann@gmail.com, '
+                            'fabian@neuschmidt.de, '
+                            'makman@alice.de'),
+          url='http://coala.rtfd.org/',
+          platforms='any',
+          packages=find_packages(exclude=("build.*", "tests", "tests.*")),
+          install_requires=required,
+          tests_require=test_required,
+          package_data={'bears': ["VERSION"],
+                        'bears.java': ['checkstyle.jar', 'google_checks.xml'],
```
justuswilhelm: Having all this bear specific data disappear is going to be a TODO. In principle the whole setup.py should be tailored exactly to what an individual bear needs, not more, not less.

uploadTool.py
```diff
@@ -0,0 +1,47 @@
...
+
+for req in Reqrs:
+    if type(req) == PythonRequirement:
+        if req.version:
+            reqs.write(req.package + '==' + req.version + '\n')
+        else:
+            reqs.write(req.package + '\n')
+
+# set env
+setup.write('#!/usr/bin/env python3\n\n')
```
justuswilhelm: I have a feeling this can be abbreviated with either a plain Python `.format` or perhaps even an intricate jinja template.

Adrianzatreanu: does it harm if its here :D ? i think it helps when it is run as a script, to force python3 instead of python, as you do in bash scripting, not sure

bearuploader/Constants.py
Adrianzatreanu: and it will, but currently its purpose is to work for at least one bear, isnt it :) ?

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+    # encoding UTF-8
+    setup.write(encoding + '\n\n')
+
+    # requirements
+    setup.write(requirements + '\n\n')
+
+    # setuptools
+    setup.write('if __name__== "__main__":\n')
+    setup.write("\tsetup(name=" + "'" + bear + "',\n")
+    setup.write('\t\t  version="0.1.30",\n')
```
Adrianzatreanu: currently this is the version i used for deployment to test it, it should be held in another file within a dictionary or so.

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+    setup.write(" Analysis Application)',\n")
+    setup.write('\t\t  ' + setup_constants + ',\n')
+    setup.write('\t\t  classifiers=' + classifiers + ')')
+
+    # close the files
+    setup.close()
+    reqs.close()
+    # upload everything
+    
+    os.chdir('upload/' + bear)
```
Adrianzatreanu: i dont know if this is the optimal way, but i found no other way to run the command for each package and for it to upload only that bear.

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+#            reqs.write(req.package + '==' + req.version + '\n')
+#        else:
+#            reqs.write(req.package + '\n')
+
+bears = glob.glob('**/*Bear.py')
+
+for bear in bears:
+    file = bear
+    bear = re.search('/.*.py', bear)
+    bear = bear.group(0)[:-3]
```
Adrianzatreanu: removes the ``.py``

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+    bear = bear[1:]
```
Adrianzatreanu: removes the ``/``

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+from coalib.bears.requirements.PythonRequirement import PythonRequirement
+import shutil
+import os
+import glob
+import re
+
+if not os.path.exists('upload/'):
+    os.mkdir('upload/')
+
+#Reqrs = MockBear.REQUIREMENTS
```
Adrianzatreanu: need to write the requirements for each bear first

Adrianzatreanu:
leaving it as pending review as I would like it to be reviewed in terms of approaches and mistakes. Still have TODOs but those are for external files, so the whole PR is a wip, but the first commit is pending review.

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+
+# for req in Reqrs:
+#    if type(req) == PythonRequirement:
+#        if req.version:
+#            reqs.write(req.package + '==' + req.version + '\n')
+#        else:
+#            reqs.write(req.package + '\n')
+
+bears = glob.glob('**/*Bear.py')
```
justuswilhelm: `bears/**/*Bear.py` would pin it down a bit more

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+
+for bear in bears:
+    file = bear
+    bear = re.search('/.*.py', bear)
```
justuswilhelm: Try using capturing to directly retrieve the name you want, e.g. `r'/(.*)\.py'`

bears/uploadTool.py
```diff
@@ -0,0 +1,70 @@
...
+    setup.write('\t\t  ' + setup_constants + ',\n')
+    setup.write('\t\t  classifiers=' + classifiers + ')')
+
+    # close the files
+    setup.close()
+    reqs.close()
+    # upload everything
+
+    os.chdir('upload/' + bear)
+    os.system('python setup.py register -r pypitest')
```
justuswilhelm: You might want to use something like `subprocess.check_call` to verify that the results are OK

bears/uploadTool.py
Adrianzatreanu: are you sure? because you run the script in the ``bears/`` directory

bears/uploadTool.py
Adrianzatreanu: i want to change os.system to subprocess.call anyway.

bears/Constants.py
```diff
@@ -4,3 +4,67 @@
...
+
+encoding = """try:
+    locale.getlocale()
+except (ValueError, UnicodeError):
+    locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')"""
+
+requirements = """with open('requirements.txt') as requirements:
+    required = requirements.read().splitlines()"""
+
+setup_constants = """author="The coala developers",
```
sils1297: this kind of data belongs in the bear

bears/Constants.py
```diff
@@ -4,3 +4,67 @@
...
+                            'fabian@neuschmidt.de, '
+                            'makman@alice.de'),
+          url='http://coala.rtfd.org/',
+          platforms='any',
+          packages=find_packages(exclude=("build.*", "tests", "tests.*")),
+          install_requires=required,
+          package_data={'bears': ["VERSION"],
+                        'bears.java': ['checkstyle.jar', 'google_checks.xml'],
+                        'bears.scala': ['scalastyle.jar',
+                                        'scalastyle_config.xml']},
```
sils1297: those are specific to bears, should be generated dynamcally

bears/uploadTool.py
justuswilhelm: Yeah that's right. I was confused.

bears/uploadTool.py
```diff
@@ -0,0 +1,68 @@
...
+#            reqs.write(req.package + '\n')
+
+bears = glob.glob('**/*Bear.py')
+
+for bear in bears:
+    file = bear
+    bear = re.match(r'.+/(.*)\.py', bear).group(1)
+    if not os.path.exists('upload/' + bear):
+        os.mkdir('upload/' + bear)
+        if not os.path.exists('upload' + bear + '/' + bear):
```
justuswilhelm: Try `os.path.join` for some more idiomatic Python bliss 👍

bears/uploadTool.py
```diff
@@ -0,0 +1,68 @@
...
+            os.mkdir('upload/' + bear + '/' + bear)
+    if not os.path.exists('upload/' + bear + '/' + bear + '/__init__.py'):
+        init = open('upload/' + bear + '/' + bear + '/__init__.py', 'w')
+        init.write(' ')
+        init.close()
+    shutil.copyfile(file, 'upload/' + bear + '/' + bear + '/' + bear + '.py')
+
+    setup = open('upload/' + bear + '/setup.py', "w")
```
justuswilhelm: This looks like a great place to make use of `with open(...) as fd:` to save yourself the `.close()` in line 61

Adrianzatreanu: good idea, will do.

Adrianzatreanu:
Marking WIP until placing attributes on all bears, therefore to make this script actually useful.

Adrianzatreanu:
moved to https://github.com/coala-analyzer/coala-bears/pull/561

PR #524, opened 2016-06-14
bears: Add metadata

Adrianzatreanu:
Adds values to the sets and bears contain
metadata about them.

### DIFF
--- a/bears/c_languages/CPPCheckBear.py
+++ b/bears/c_languages/CPPCheckBear.py
@@ -19,6 +19,9 @@ class CPPCheckBear:
     """
 
     LANGUAGES = {"C", "C++"}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
 
     @staticmethod
     def create_arguments(filename, file, config_file,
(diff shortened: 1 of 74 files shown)


bears/c_languages/CPPCheckBear.py
```diff
@@ -19,6 +19,13 @@ class CPPCheckBear:
     """
 
     LANGUAGES = {"C", "C++"}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'https://groups.google.com/forum/#!forum/coala-devel'}
```
sils1297: that's no e-mail address. Rather use `coala-devel@googlegroups.com`

bears/c_languages/CPPCheckBear.py
```diff
@@ -19,6 +19,13 @@ class CPPCheckBear:
...
+    MAINTAINERS = {'Lasse Schuirmann', 'Fabian Neuschmidt', 'Mischa Kr\xfcger'}
```
sils1297: isn't this the coala devs as well? Could we maybe default it to  the AUTHORS if not given?

sils1297:
https://github.com/coala-analyzer/coala-bears/pull/524/commits/db5ef317e38cd455a19ceecaa0a8b17a4ee60ae4 needs work

bears/c_languages/CPPCheckBear.py
Adrianzatreanu: this is what the current setup.py says .. you want me to make it equal to authors if not given? in Bear.py?

sils1297:
ack 9e7547d

sils1297:
@rultor merge

PR #552, opened 2016-06-23
deps.sh: Rewrite deps in Python

Adrianzatreanu:


Adrianzatreanu:
hey @Makman2 this is what i've came up with, but it is a little unclear to me where i'd have to write this python script, since .ci/deps.sh is a shell script and it wont work if i do it there as it seems. can you explain a little more please :D ?

PR #574, opened 2016-07-10
tests: Add BearSerializationTest

arafsheikh:
This test ensures that all the bears in this repository are always serializable.


tests/BearSerializationTest.py
```diff
@@ -0,0 +1,34 @@
+import itertools
+import json
+import os
+import json
```
tushar-rishav: Redundant imports :stuck_out_tongue:

tests/BearSerializationTest.py
```diff
@@ -0,0 +1,34 @@
...
+            ["**"],
+            [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
+            printer,
+            warn_if_unused_glob=False)
+        self.bears = list(itertools.chain(local_bears, global_bears))
+        self.JSONEncoder = create_json_encoder(use_relpath=False)
+
+    def test(self):
+        json.dumps(self.bears, cls=self.JSONEncoder)
+        # If no exception is thrown then the test passes
```
tushar-rishav: @arafsheikh So IIUC for non-serialisation case you are depending on an exception to be raised? Can we rather emulate a bear that is non serialisable and test for that too? Otherwise, at least catch the particular exception which is raised for non serialisable bears and assert for errors neatly. 
I guess it would be little near as it won't just populate the console with tracebacks. Thoughts?

tests/BearSerializationTest.py
arafsheikh: json? It's used at L32

tushar-rishav: You are importing json twice :)

tests/BearSerializationTest.py
arafsheikh: The mechanism of serialization is written in `coala`. If we want to test unserializable bears here then we'll have to write mock bears. I wrote this test with the intention of not testing the mechanism, but to keep all bears in this repo serializable. What do you think?

Secondly I guess we can catch the exceptions and neatly display the error, but then we'll need to do it exhaustively. e.g. `TypeError` is thrown when the object can't be serialized but there's also `ValueError` for circular reference.

tushar-rishav: Hmm alright. I think we go with latter one then. Catch the two exceptions.
Are there any more?

arafsheikh: This is exactly why I didn't go with catching exceptions. I'm not sure how many exceptions can be thrown.

tushar-rishav: Um I am not on PC now. Um alright. Let's at least catch the exception, log
the error neatly and then raise assertion error. :D

tushar-rishav: I mean just catch the broad exceptions.

tests/BearSerializationTest.py
```diff
@@ -0,0 +1,40 @@
...
+    def test(self):
+        ERROR_STR = ("At least one bear in the \"bears\" directory is not "
+                     "JSON serializable. If you have modified any bear then "
+                     "make sure all non-serializable attributes are private. "
+                     "More information about serializable attributes can be "
+                     "found at https://docs.python.org/3.4/library/"
+                     "json.html#encoders-and-decoders")
+        try:
+            json.dumps(self.bears, cls=self.JSONEncoder)
+        except BaseException:
```
tushar-rishav: @arafsheikh Neat. Sorry but just add one more thing. Generally it's bad practice to catch broad exceptions. But since here we need to I am thinking why not also add a small comment explaining why we are catching broad exception? What do you think?
Once done, the PR is ready :)

tushar-rishav:
ack 123c40e def0d53

tushar-rishav:
@rultor merge

tests/BearSerializationTest.py
```diff
@@ -0,0 +1,41 @@
...
+    def test(self):
+        ERROR_STR = ("{} is not JSON serializable. If you have added or "
+                     "modified this bear then make sure all non-serializable "
+                     "attributes are private. More information about "
+                     "serializable attributes can be found at https://docs."
+                     "python.org/3.4/library/json.html#encoders-and-decoders")
+        for bear in self.bears:
+            try:
+                json.dumps(bear, cls=self.JSONEncoder)
+            except BaseException:
```
Uran198: Any reason to catch `BaseException`?

Uran198: With it you will basically won't be able to interrupt test, which is bad.

arafsheikh: See this thread https://github.com/coala-analyzer/coala-bears/pull/574#discussion_r70425826

Uran198: Catching broad exceptions is bad. Check documentation to see what can be throws. IIRC only `TypeError` is raised when something is not serializable. @tushar-rishav

tushar-rishav: @Uran198  Yeah I had [similar](https://github.com/coala-analyzer/coala-bears/pull/574#discussion_r70562782) opinion :sweat_smile: 
I guess catching `TypeError` and `OverflowError` (for circular ref) might suffice for now. 
Also, we may rather use [skipkeys](https://docs.python.org/3.4/library/json.html#json.JSONEncoder) to ignore the errors while encoding non acceptable chars. This would require to tweak the `create_json_encoder` a bit. Thoughts?
CC @arafsheikh

tushar-rishav: Obviously that should go in separate commit ^

Uran198: @tushar-rishav you don't need to catch OverflowError, cause `check_circular` defaults to true. As it is there is basically only TypeError, just checked the [docs](https://docs.python.org/3/library/json.html#json.dump). There is no need for `skipkeys` IMO, but feel free to open an issue for that with your arguing.

tushar-rishav:
@arafsheikh Ping.

arafsheikh:
@tushar-rishav Got a little busy with coala-eclispe. I don't know why the tests fail. Here's what I found:
- `py.test -k generate_packageTest` would pass.
- `py.test -k BearSerializationTest` would pass.
- `py.test -k "generate_packageTest or BearSerializationTest"` would fail.
- If I rename this test to, say `xBearSerializationTest` then the above would pass too.
So running `BearSerializationTest` before `generate_packageTest` would result in an error. Any ideas why?

arafsheikh:
Also `CSSAutoPrefixBearTest` fails only on this PR and locally for me but passes for everyone else.

Adrianzatreanu:
ping as its been a while.

jayvdb:
@gitmate-bot rebase

PR #584, opened 2016-07-18
bears: Use coala-utils for string processing

arafsheikh:


abhsag24:
@arafsheikh shouldn't the commits be reordered? i find it more logical to update coala first and then change the things broken by the update? or is it something that i don't get?

arafsheikh:
@abhsag24 It won't have any effect, but yeah it's logically correct. Updated :)

abhsag24:
ack d024ce5, 359afc5

arafsheikh:
@rultor merge

PR #649, opened 2016-08-03
Add Installation Tool

Adrianzatreanu:


bears/installation_tool.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def get_output(command):
+    """
+    Runs the command and decodes the output and returns it.
+
+    :param command: The command to be run.
+    :return:        The output of the command, decoded.
+    """
+    function = subprocess.Popen(command, stdout=subprocess.PIPE)
```
justuswilhelm: Popen can also be used as a context manager, in case you're looking for some mad 1337 Python code: https://docs.python.org/3/library/subprocess.html#subprocess.Popen

bears/installation_tool.py
```diff
@@ -0,0 +1,131 @@
...
+        output = get_output(['pip', 'search', "coala.rtfd.org"])
+        bear_name_list = re.findall(r'(.+) \(\d.*\n', output)
+
+        bear_name_completer = WordCompleter(bear_name_list, ignore_case=True)
+
+        for bear_name in sorted(bear_name_list):
+            print(bear_name)
+
+        answer = prompt('Which ones? (You can add more, separated by commas)\n',
+                        lexer=SqlLexer,
```
justuswilhelm: Still not sure whether you need a lexer at all here. Can you just leave this line out?

bears/installation_tool.py
```diff
@@ -0,0 +1,131 @@
...
+                    try:
+                        subprocess.call([requirement.install_command()])
+                    except:
+                        bears_failed_list.append(bear_name)
+
+            # check if any bear failed installing
+            if bears_failed_list:
+                print('Bears that failed installing:')
+                for bear_failed in bears_failed_list:
+                    print(bear_failed)
```
justuswilhelm: `return 1` here

bears/installation_tool.py
```diff
@@ -0,0 +1,131 @@
...
```
justuswilhelm: `return 1` here. Is this duplicate code?

bears/installation_tool.py
```diff
@@ -0,0 +1,136 @@
...
+
+        for bear_name in sorted(bear_name_list):
+            print(bear_name)
+
+        # HACK using sqllexer to color words which are correct
+        answer = prompt('Which ones? (You can add more, separated by commas)\n',
+                        lexer=SqlLexer,
+                        completer=bear_name_completer).split(',')
+
+        for bear_name in answer:
```
justuswilhelm: This is duplicate code with the block in line 78

Adrianzatreanu: not really, is it? this one parses a different list. this one is for bears given, that one is for all bears. do you have an idea how to improve this overall :D ?

Asnelchristian: maybe you can define function that take the value of ``answer`` as parameter

bears/installation_tool.py
```diff
@@ -0,0 +1,136 @@
...
+    :return:        The output of the command, decoded.
+    """
+    function = subprocess.Popen(command, stdout=subprocess.PIPE)
+    output, err = function.communicate()
+    return output.decode("utf-8")
+
+
+def check_failed_bears(bear_list):
+    """
+    Check if any bear failed installation. Exits with 1 if so.
```
Asnelchristian: ``Checks``

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+def get_output(command):
+    """
+    Runs the command and decodes the output and returns it.
+
+    :param command: The command to be run.
+    :return:        The output of the command, decoded.
+    """
+    function = subprocess.Popen(command, stdout=subprocess.PIPE)
+    output, err = function.communicate()
+    return output.decode("utf-8")
```
hypothesist: return `function.communicate()[0].decode("utf-8")` is slightly faster, but slight more unreadable too. Your call

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def check_failed_bears(bear_list):
+    """
+    Checks if any bear failed installation. Exits with 1 if so.
+    :param bear_list: List of the failed bears.
```
hypothesist: Newline between param and description

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    if bear_list:
+        print('Bears that failed installing:')
+        for bear_failed in bear_list:
+            print(bear_failed)
```
hypothesist: `for ...` -> `print("\n".join(bear_list))`?

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+
+def main():
+    args = create_installation_parser().parse_args()
+
+    option_completer = WordCompleter(['All', 'Some', 'None'], ignore_case=True)
+
+    # HACK using SqlLexer to color the words which are correct
+    answer = prompt('What bears would you like to install?\n'
+                    'All / Some / None\n',
+                    lexer=SqlLexer, completer=option_completer)
```
hypothesist: Do `answer = answer.lower()` or append `.lower()` to the `prompt(...)`. Then you don't need to check `answer in ("All", "all")`, `answer in ("Some", "some")` and so on; just do `answer == "all"`, ...

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def install_bears(bear_names_list, arg):
+    """
+    Tries to install each bear from the bear_names_list. Will also check for
+    bears which failed to be installed, or their requirements failed to be
+    installed.
+
+    :param bear_names_list: The list which contains the names of the bears.
+    :param arg:             An arg which is given to also install the bears'
```
hypothesist: IMO this could be better named; `arg` -> `install_reqs`?

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+    for bear_name in bear_names_list:
+        bear_name = bear_name.replace(' ', '')
+        bears_failed_list = []
+        try:
+            install_pip_package(bear_name)
+        except:
+            bears_failed_list.append(bear_name)
+
+        if arg:
+            bears_failed_list = install_requirements(
```
hypothesist: Would rather not give`bears_failed_list` as a param to `install_requirements` . Instead let it build its own failure list and then you do: `bears_failed_list += install_requirements(bear_name)`

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+                        action='store_true')
+    return parser
+
+
+def main():
+    args = create_installation_parser().parse_args()
+
+    option_completer = WordCompleter(['All', 'Some', 'None'], ignore_case=True)
+
+    # HACK using SqlLexer to color the words which are correct
```
hypothesist: Will this be changed in the future? If yes, add a `TODO`. Else ignore and delete this comment :P

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+
+        install_bears(sorted(get_all_bears_names_from_PyPI()), args.install)
+
+    elif answer in ("Some", "some"):
+        print('This is a list of all the bears you can install:')
+
+        bear_name_completer = WordCompleter(
+            get_all_bears_names_from_PyPI(), ignore_case=True)
+
+        for bear_name in sorted(get_all_bears_names_from_PyPI()):
```
hypothesist: You're doing `get_all_bears_names_from_PyPI` twice, which will take a lot of time since it's a network request. Maybe store it once and reuse it everywhere?

bears/installation_tool.py
```diff
@@ -0,0 +1,158 @@
...
+            print(bear_name)
+
+        # HACK using sqllexer to color words which are correct
+        answer = prompt('Which ones? (You can add more, separated by commas)\n',
+                        lexer=SqlLexer,
+                        completer=bear_name_completer).split(',')
```
hypothesist: Just a `,` split won't do. There may be spaces after or before commas. `StringConverter`: https://gitlab.com/coala/coala-utils/blob/master/coala_utils/Question.py#L40

bears/installation_tool.py
justuswilhelm: @hypothesist What do you think about the unused `err` variable?

hypothesist: He's not using it anyway. If he's still going for the unpack, may be do `output, _ = ...`? Or at least `_err`?

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def check_failed_bears(bear_list):
+    """
+    Checks if any bear failed installation. Exits with 1 if so.
+
+    :param bear_list: List of the failed bears.
+    """
+    if bear_list:
+        print('Bears that failed installing:' + "\n".join(bear_list))
```
hypothesist: I know I should have mentioned this before, but log this to `stderr` instead? Since it's actually an error. Just nitpicking from now on :P

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+    return package_failed_list
+
+
+def get_all_bears_names_from_PyPI():
+    """
+    Gets all the bears names from PyPI, using the link in the description.
+
+    :return: A list with all the bear names.
+    """
+    output = get_output(['pip', 'search', "coala.rtfd.org"])
```
hypothesist: Use the `sys.executable` approach

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+    Tries to install each bear from the bear_names_list. Will also check for
+    bears which failed to be installed, or their requirements failed to be
+    installed.
+
+    :param bear_names_list: The list which contains the names of the bears.
+    :param install_deps:    An arg which is given to also install the bears'
+                            dependencies.
+    """
+    for bear_name in bear_names_list:
+        bear_name = bear_name.replace(' ', '')
```
hypothesist: Where would this be needed? Wouldn't the bears already be one word with no space?

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+                                requirements failed installing.
+    :return:                    A list with the packages which had their
+                                requirements failing to be installed.
+    """
+    package_failed_list = []
+    package_object = importlib.import_module(
+        package_name + "." + package_name)
+    for requirement in getattr(package_object, package_name).REQUIREMENTS:
+        try:
+            subprocess.call([requirement.install_command()])
```
hypothesist: Um, this already returns a list? https://github.com/coala-analyzer/coala/blob/fee67b3089626b6b4f2e57e2ef1c6b6f335b7802/coalib/bears/requirements/PipRequirement.py#L36

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def install_pip_package(package_name):
+    """
+    Uses call to install a PyPi package.
+
+    :param package_name: The package to be installed.
+    """
+    print('To install the dependencies correctly, we will need administrator '
+          'rights from you.')
```
hypothesist: I *think* `root permission` is the correct word, not `admin rights`. Not sure, confirm with others.

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+            print(bear_name)
+
+        # HACK using sqllexer to color words which are correct
+        answer = prompt('Which ones? (You can add more, separated by commas)\n',
+                        lexer=SqlLexer,
+                        completer=bear_name_completer)
+
+        answer = list(StringConverter(answer))
+
+        install_bears(answer, args.install)
```
hypothesist: return code?

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # HACK using SqlLexer to color the words which are correct
+    answer = prompt('What bears would you like to install?\n'
+                    'All / Some / None\n',
+                    lexer=SqlLexer, completer=option_completer).lower()
+
+    if answer == "all":
+        print('Great idea, we are installing all the bears right now.')
+
+        install_bears(sorted(get_all_bears_names_from_PyPI()), args.install)
```
hypothesist: return code?

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+                        lexer=SqlLexer,
+                        completer=bear_name_completer)
+
+        answer = list(StringConverter(answer))
+
+        install_bears(answer, args.install)
+
+    else:
+        print('See ya!')
+        return 0
```
hypothesist: Or don't give the return code here :P

(Delete the other two comments if you're opting for no return codes, else delete this comment if you're adding return codes for the other two branches)

bears/installation_tool.py
```diff
@@ -0,0 +1,161 @@
...
+
+    :return: A list with all the bear names.
+    """
+    output = get_output(['pip', 'search', "coala.rtfd.org"])
+    return re.findall(r'(.+) \(\d.*\n', output)
+
+
+def install_bears(bear_names_list, install_deps):
+    """
+    Tries to install each bear from the bear_names_list. Will also check for
```
hypothesist: Double backticks around `bear_names_list`

bears/installation_tool.py
Adrianzatreanu: what do you mean?

Adrianzatreanu: nvm i read the other comments too now. i got it

hypothesist: `return 0` since you've done a `return 0` in the else branch

bears/installation_tool.py
```diff
@@ -0,0 +1,180 @@
...
+from coalib.misc.Shell import call_without_output
+from prompt_toolkit import prompt
+from prompt_toolkit.contrib.completers import WordCompleter
+from pygments.lexers import SqlLexer
+from pyprint.ConsolePrinter import ConsolePrinter
+
+
+def install_pip_package(package_name):
+    """
+    Uses call to install a PyPi package.
```
hypothesist: Double backticks around `call`

tests/installation_toolTest.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_parser(self):
+        self.assertTrue(
+            create_installation_parser().parse_args(['--install']).install)
+
+
+class install_requirementsTest(unittest.TestCase):
+
+    def test_requirements_succeeded(self):
+        install_pip_package('PEP8Bear')
+        self.assertTrue(install_requirements('PEP8Bear') == [])
```
hypothesist: `self.assertEqual(..., [])`

Adrianzatreanu:
moved here https://gitlab.com/coala/bear_installation_tool/merge_requests/2

Issue #605, opened 2016-07-23
PEP8Bear for Jupyter Notebooks


caichinger:
Jupyter Notebooks are widely used for tutorials.

It would therefore be nice to have PEP8Bear for notebooks to make sure basic code style standards are met.

I will give this a try and join your sprint later. ;)

sils1297:
Neat, see you!

sils1297:
We could also add grammar checking for the other parts with languagetool :)

sils1297:
CC @Udayan12167 and @mr-karan this is the first bear we'd have that should reuse other bears for parts of the file with different languages. We should think of this as a potential new concept, just write a bear for that particular file and that bear then could deal with invoking the others for the parts?

caichinger:
I started working on this issue and the current draft can be found here:
https://github.com/caichinger/coala-bears/tree/605_ipynb_pep8bear

Currently I am on holidays but I will continue when I come back at beginning of August.

Some thoughts:
- Reading/writing Jupyter Notebooks most easily done with https://github.com/jupyter/nbformat (no problem with different versions, etc.). Everybody working with Jupyter automatically has this lib installed. Is nbformat a feasible requirement?
- The Notebook related part is independent of the bear (it could e.g. be a Julia-Notebook). Where is a good place for the Notebook-part?
- At the moment I encounter problems when supplying the test files as plain strings (due to \n). I will take a look at this.
- When I come back I will prepare a PR for easier reviewing.

Makman2:
Nice thx @caichinger ! cool you are working on this :)
We will send out an invitation to our org, then you can be assigned to this issue so no one starts doing redundant work :+1:

Makman2:
for help, conversion whatsoever feel free to drop by at our gitter channel: https://gitter.im/coala-analyzer/coala

there we can help you too with proceeding on this issue :+1: :smile:

sils1297:
Hi! Welcome to our community and thank you for your interest!

I have sent you an invitation to join our community. This is needed so we can coordinate tasks with you and make sure no two persons are working on the same thing. If you can't find the mail, you should be able to join us at http://github.com/coala-analyzer/!

We'd love to be able to help you when problems occur. Just join us on [our gitter channel](https://gitter.im/coala-analyzer/coala) and tell us about them :)

sils1297:
@Makman2 feel free to send out invitations on your own.

sils1297:
[...]

PR #655, opened 2016-08-06
605 ipynb pep8bear

caichinger:
Hi,

this PR adresses #605 (PEP8Bear for Jupyter Notebooks).

Some thoughts/questions regarding the current implementation:
- I use https://github.com/jupyter/nbformat for to read/write the notebook, thereby we avoid problems with different notebook versions. However, nbformat is then a necessary requirement, is this okay?
- Since the notebook related part is independent of the bear, I think it should be located somewhere else. However, I do not know where. :)
- There is currently some code duplication related to the original PEP8Bear. At the sprint at the EuroPython @sils1297 said that this approach is easier for now. I do not know whether the bears-using-bears idea is currently being worked on.

Thank you for your help!

### DIFF
--- a/bears/python/PEP8NotebookBear.py
+++ b/bears/python/PEP8NotebookBear.py
@@ -1,0 +1,121 @@
+import autopep8
+import nbformat
+
+from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
+from coalib.bears.LocalBear import LocalBear
+from coalib.bears.requirements.PipRequirement import PipRequirement
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.settings.Setting import typed_list
+
+# Comments regarind Jupyter Notebooks:
+# The `nbformat` module contains the reference implementation of the Jupyter
+# Notebook format, and Python APIs for working with notebooks.
+# On the file level, a notebook is a JSON file, i.e. dictionary with a few
+# keys.
+# The functions in `nbformat` work with `NotebookNode` objects, which are like
+# dictionaries, but allow attribute access. The structure of these objects
+# matches the notebook format specification.
+
+
+def notebook_node_from_string_list(string_list):
+    """
+    Reads a notebook from a string list and returns the NotebookNode
+    object.
+
+    :param string_list: The notebook file contents as list of strings
+                        (linewise).
+    :return:            The notebook as NotebookNode.
+    """
+    return nbformat.reads(''.join(string_list), nbformat.NO_CONVERT)
+
+
+def notebook_node_to_string_list(notebook_node):
+    """
+    Writes a NotebookNode to a list of strings.
+
+    :param notebook_node: The notebook as NotebookNode to write.
+    :return:              The notebook as list of strings (linewise).
+    """
+    return nbformat.writes(notebook_node, nbformat.NO_CONVERT).splitlines(True)
+
+
+def autopep8_fix_code_cell(source, options=None, apply_config=None):
+    """
+    Applies autopep8.fix_code and takes care of newline characters.
+
+    autopep8.fix_code automatically adds a final newline at the end,
+    e.g. ``autopep8.fix_code('a=1')`` yields 'a = 1\\n'.
+    Note that this is not related to the 'W292' flag, i.e.
+    ``autopep8.fix_code('a=1', options=dict(ignore=('W292',)))`` gives
+    the same result.
+    For notebook code cells, this behaviour does not make sense, hence
+    newline is removed if ``source`` does not end with one.
+    """
+    source_corrected = autopep8.fix_code(source,
+                                         apply_config=apply_config,
+                                         options=options)
+
+    if not source.endswith('\n'):
+        return source_corrected[:-1]
+
+    return source_corrected
+
+
+class PEP8NotebookBear(LocalBear):
+    LANGUAGES = {"Python", "Python 2", "Python 3"}
+    REQUIREMENTS = {PipRequirement('autopep8', '1.*'),
+                    PipRequirement('nbformat', '4.*')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/83333'
+    CAN_FIX = {'Formatting'}
+
+    def run(self, filename, file,
+            max_line_length: int=79,
+            indent_size: int=SpacingHelper.DEFAULT_TAB_WIDTH,
+            pep_ignore: typed_list(str)=(),
+            pep_select: typed_list(str)=(),
+            local_pep8_config: bool=False):
+        """
+        Detects and fixes PEP8 incompliant code in Jupyter Notebooks. This bear
+        will not change functionality of the code in any way.
+
+        :param max_line_length:   Maximum number of characters for a line.
+        :param indent_size:       Number of spaces per indent level.
+        :param pep_ignore:        A list of errors/warnings to ignore.
+        :param pep_select:        A list of errors/warnings to exclusively
+                                  apply.
+        :param local_pep8_config: Set to true if autopep8 should use a config
+                                  file as if run normally from this directory.
+        """
+        options = {"ignore": pep_ignore,
+                   "select": pep_select,
+                   "max_line_length": max_line_length,
+                   "indent_size": indent_size}
+        notebook_node = notebook_node_from_string_list(file)
+        cells = notebook_node['cells']
+
+        for cell in cells:
+            if cell['cell_type'] != 'code':
+                continue
+            cell['source'] = autopep8_fix_code_cell(cell['source'],
+                                                    local_pep8_config,
+                                                    options)
+
+        corrected = notebook_node_to_string_list(notebook_node)
+
+        # If newline at eof in `file` but not in `corrected`, add
+        # final newline character to `corrected` to make sure this difference
+        # does not pop up in `diffs`.
+        if file[-1].endswith('\n') and not corrected[-1].endswith('\n'):
+            corrected[-1] += '\n'
+
+        diffs = Diff.from_string_arrays(file, corrected).split_diff()
+
+        for diff in diffs:
+            yield Result(self,
+                         "The code does not comply to PEP8.",
+                         affected_code=(diff.range(filename),),
+                         diffs={filename: diff})
--- a/tests/python/PEP8NotebookBearTest.py
+++ b/tests/python/PEP8NotebookBearTest.py
@@ -1,0 +1,91 @@
+from tests.LocalBearTestHelper import verify_local_bear
+from bears.python.PEP8NotebookBear import PEP8NotebookBear
+
+
+# metadata field deleted manually
+good_file = r"""{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "x = 'good'"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "a markdown cell is not a code cell"
+   ]
+  }
+ ],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
+"""
+
+# metadata field deleted manually
+bad_file = r"""{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "x =    1  # <-- PEP8 Error"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "x =    1  # <-- PEP8 Error but we are in a Markdown cell"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "1\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(x)  # <-- here everything is fine\n"
+   ]
+  }
+ ],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
+"""
+
+PEP8NotebookBearTest = \
+    verify_local_bear(PEP8NotebookBear,
+                      valid_files=(good_file,),
+                      invalid_files=(bad_file,)
+                      )
+
+PEP8NotebookBearWithoutTrailingNewlineTest = \
+    verify_local_bear(PEP8NotebookBear,
+                      valid_files=(good_file[:-1],),
+                      invalid_files=(bad_file[:-1],),
+                      force_linebreaks=False,
+                      )
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,3 +29,4 @@ nltk==3.2.*
 appdirs==1.*
 pyyaml==3.*
 vulture==0.10.*
+nbformat>=4.*


tests/python/PEP8JupyterNotebookBearTest.py
```diff
@@ -0,0 +1,142 @@
...
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.5.2"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
+
+"""
```
sils1297: two newlines here?

bears/python/PEP8JupyterNotebookBear.py
```diff
@@ -0,0 +1,128 @@
...
+        Writes notebook in dict form to string array.
+
+        :param file_dct: The notebook in dictionary form.
+        :return:         The notebook in string array form.
+        """
+        file_lst = (nbformat.writes(file_dct, nbformat.NO_CONVERT)
+                            .splitlines(True))
+        # nbformat.writes does not add a final \n, we do this manually
+        # necessary for the diff because original file has it
+        if self.file[-1].endswith('\n') and not file_lst[-1].endswith('\n'):
```
sils1297: youu could add a testcase without a trailing newline at all to cover this branch

caichinger: I will do that

tests/python/PEP8JupyterNotebookBearTest.py
caichinger: I thought it would simplify things but it actually doesn't, will be changed

sils1297:
ouh this isn't WIP, you changed the title after pushing which is why this was unnoticed for so long - sorry

sils1297:
@caichinger are you still around? Otherwise I'd take this over as we would love it for the release :)

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class PEP8NotebookBear(LocalBear):
+    LANGUAGES = {"Python", "Python 2", "Python 3"}
+    REQUIREMENTS = {PipRequirement('autopep8', '1.*'),
+                    PipRequirement('nbformat', '4.*')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Formatting'}
```
sils1297: I think we'll want an ASCIINEMA_URL here :)

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def run(self, filename, file,
+            max_line_length: int=79,
+            tab_width: int=SpacingHelper.DEFAULT_TAB_WIDTH,
```
sils1297: this has been renamed `indent_size`

sils1297: CC @Asnelchristian

caichinger:
@sils1297 yes I am still around and will incorporate your comments during the weekend (most likely tomorrow) -  next release sounds great! :)
ad "changing title after pushing": I thought changing title alone is sufficient, next time I know.

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Reads a notebook from a string array and returns the NotebookNode
+        object.
+
+        :param file: The notebook file contents as JSON string array.
+        :return:     The notebook as NotebookNode.
+        """
+        return nbformat.reads(''.join(file), nbformat.NO_CONVERT)
+
+    def write_to_list(self, file_nb):
```
sils1297: shouldn't this function as a member function simply use self.file_nb instead of receiving it as a param?

sils1297: why does this receive a parameter that is retrieved from the class? Seems super weird to me

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,130 @@
...
+        file_lst = (nbformat.writes(file_nb, nbformat.NO_CONVERT)
+                            .splitlines(True))
+        # nbformat.writes does not add a final \n, we do this manually
+        # necessary for the diff because original file has it
+        if self.file[-1].endswith('\n') and not file_lst[-1].endswith('\n'):
+            file_lst[-1] += '\n'
+        return file_lst
+
+    @property
+    def file_nb(self):
```
sils1297: it isn't particularly obvious what this is

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,130 @@
...
+
+        :param filename: The filename of the file.
+        :param file:     The file contents as string array.
+        """
+        self.filename = filename
+        self.file = file
+        self._file_nb = self.read_from_list(file)
+
+    @staticmethod
+    def read_from_list(file):
```
sils1297: is this meant to be public?

AbdealiJK:
I think that the Jupyter notebook idea is awesome !
Would it not e cooler to make a geeric NoteBookCellBear which gives HiddenResults of cells and then we can do a JupyterMarkdownBear, JupyterPythonBear, etc. ? Which takes these values and feeds it into MarkdownBear, PEP8Bear, etc. for analysis.

@sils1297 Any idea on how to do this cleanly ?

Te reason for this is because jupyter notebooks can benefit from all the bears related to R, Julia, Python, Markdown, and Natural Language !

sils1297:
@AbdealiJK we've thought a long time about this so far. Long story short: coala doesn't support this yet. I think @mr-karan wanted to look into this long term. It's complicated - but indeed this bear is a very good thing to drive this concept forward: a nice example. We now know way better about the problems that come with that, e.g. you'd probably need to translate the locations of the results somehow or the bear would have to know about "regions" that are to be analyzed. Great things ahead.

AbdealiJK:
Nod, agree. Is there an issue somewhere for the long term plan ?

sils1297:
https://github.com/coala-analyzer/coala/issues/2538

bears/python/PEP8NotebookBear.py
caichinger: I am not entirely sure what you mean by *a parameter that is retrieved from the class* but I had two things in mind:

1. the Bear works with a copy of `_file_nb` and when he has done his job, `file_nb` has changed, i.e. 
`file_nb != NotebookHandler(filename, file).file_nb`.
1. due to the issue with the final newline charater '\n' (cf. line 50) the `write_to_list` method needs to know how the original `file` looked like.

since I need both the new `file_nb` as well as the original `file` to properly construct the needed list (string array) I thought this is simplest solution. does this answer your question?

in any case, I have no objections against changing the approach :)
maybe rename the parameter to something like `file_nb_changed` to avoid confusion?

bears/python/PEP8NotebookBear.py
caichinger: it returns the file in notebook-node-form and I used `@property` to make sure that one only receives a copy to work with and does not change the original version of the 
notebook, i.e. `_file_nb` which is created at initialization.

what is you opinion?

apart from that: do you think this is redundant or maybe `file` be treated the same (to be consistent)?

bears/python/PEP8NotebookBear.py
caichinger: since nobody else uses it probably not ;)
I will change it to `_read_from_list`.

bears/python/PEP8NotebookBear.py
caichinger: of course we do :)
I will look into this

caichinger:
@sils1297 : just in case you overlooked the changes, I am still around (and 
willing) to work on this. However, if you prefer to take over to speed things
up feel free to do so. Otherwise I am happy to receive your feedback. :)


When this Bear is merged I would open a new issue as shown below. I think it is a 
necessary (and feasible) intermediate step to arrive at more general applications like the 
ones already mentioned in https://github.com/coala-analyzer/coala/issues/2538 and @AbdealiJK comment above. 
Are you both fine with that?


Prospective issue:
> Refactor notebook handling from PEP8NotebookBear into separate module

> To pave the way for future coala features dealing with Jupyter Notebooks,
some notebook-specific io functionality has to be provided.

> As the first bear dedicated to notebooks, the `PEP8NotebookBear` module currently 
holds the `NotebookHandler` class. However, since the Jupyter Notebook is language 
agnostic (note that Jupyter stands for Julia, Python and R) it makes sense to 
refactor the notebook handling into a separate module.

sils1297:
I'm kind of around but I'm today busy with a client and tomorrow as well
maybe somebody else can help you? CC @AbdealiJK

caichinger:
No problem! I am not in a hurry. :) If one of you posts some comments I will try to deal with them during the weekend.

sils1297:
For now I'd keep it in the same module

bears/python/PEP8NotebookBear.py
sils1297: so at least if I look at this initially I do get the feeling of "complication" but I can't objectively reason about this yet because I don't fully understand what exactly happens where. Usually if my intuition goes like that there is at least a minor simplification that helps understanding a lot.

I'll have to look at it later with more time :/

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,131 @@
...
+        self.file = file
+        self._file_nb = self._read_from_list(file)
+
+    @staticmethod
+    def _read_from_list(file):
+        """
+        Reads a notebook from a string array and returns the NotebookNode
+        object.
+
+        :param file: The notebook file contents as JSON string array.
```
AbdealiJK: A json string array is confusing IMO. "Lines of the notebook file contents which uses JSON format" maybe ?
(I thought `file` was a JSON string array ... hence the confusion)

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,131 @@
...
+
+class NotebookHandler:
+    """Utility class for Jupyter Notebook handling."""
+
+    def __init__(self, filename, file):
+        """
+        Constructs a notebook handler to deal with NotebookNode objects.
+        A NotebookNode is a dict-like node with attribute-access.
+
+        :param filename: The filename of the file.
```
AbdealiJK: Is the filename and file needed ? It seems like filename is not being used in the NotebookHandler.

bears/python/PEP8NotebookBear.py
AbdealiJK: A docstring here is useful in such cases.
Possibly just `notebook_node` would be a fine name too, but I have nothing against `file_nb`.

The only reason I have to prefer notebook_node is that it  doesnt confuse with the `file` which is a list of lines

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,131 @@
...
+        :param file:     The file contents as string array.
+        """
+        self.filename = filename
+        self.file = file
+        self._file_nb = self._read_from_list(file)
+
+    @staticmethod
+    def _read_from_list(file):
```
AbdealiJK: Better name would be appreciated here !
What about just `_create_notebook_node` as it literally creates it ? Or maybe `_parse_notebook` or `_parse_file` ? Because nbformat does parse the JSON file and gives a nice API for it

bears/python/PEP8NotebookBear.py
AbdealiJK: I can see how this makes sense.
I'd recommend changing this function to `_format_notebook` so that it complements the earlier `_parse_notebook`.
Other alternatives: `_format_file()`, `_create_notebook_file` ?

AbdealiJK: I think to avoid confusion, it would be a good idea to add `newline__at_eof` or so. If it is True/False, use that.
If it is `None`, use `self.file[-1].endswith('\n')`

Also, a more descriptive docstring mentioning it auto-understands the EOF \n

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,131 @@
...
+            indent_size: int=SpacingHelper.DEFAULT_TAB_WIDTH,
+            pep_ignore: typed_list(str)=(),
+            pep_select: typed_list(str)=(),
+            local_pep8_config: bool=False):
+        """
+        Detects and fixes PEP8 incompliant code in Jupyter Notebooks. This bear
+        will not change functionality of the code in any way.
+
+        :param max_line_length:   Maximum number of characters for a line.
+        :param indent_size:         Number of spaces per indent level.
```
AbdealiJK: 2 extra spaces before "Number"

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,131 @@
...
+        diffs = Diff.from_string_arrays(file, corrected).split_diff()
+
+        for diff in diffs:
+            yield Result(self,
+                         "The code does not comply to PEP8.",
+                         affected_code=(diff.range(filename),),
+                         diffs={filename: diff})
+
+    @staticmethod
+    def _bypass_no_newline_at_end_of_file(f):
```
AbdealiJK: This is unusual. Why a decorator ? It may make sense to have a contextmanager here ?
Or just put those 2-3 lines in the code itself ... as it is being used only once.

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        Utility function to first append and finally remove a '\n' to
+        a given source string in order to avoid 'no newline at end of file'
+        complaint which does not make sense for notebook cells.
+        """
+        def handle_line_breaks(source):
+            source += '\n'
```
AbdealiJK: Don't you want to check if there is already a \n ?

tests/python/PEP8NotebookBearTest.py
```diff
@@ -0,0 +1,133 @@
...
+
+def prepare_string(file):
+    """
+    Utility function to prepare Notebook strings for further processing.
+
+    replace() necessary to avoid
+    nbformat.reader.NotJSONError: Notebook does not appear to be JSON
+    which is caused by
+    json.decoder.JSONDecodeError: Invalid control character
+    which is caused by the \n characters inside the input cells
```
AbdealiJK: This is interesting. Does this not happen when you read the file directly ? Shouldnt this be in the bear ?

caichinger:
@sils1297 @AbdealiJK thank you for your comprehensive comments!
I already have some ideas and should be able to push tomorrow (please wait until have updated :)).

tests/python/PEP8NotebookBearTest.py
caichinger: This is not only interesting, it is also annoying. ;)
It does not happen when you read the file directly (i.e. when you pass a file directly to the bear).
If you know a better solution I am happy to change this.

caichinger:
Hi there,

after considering your valuable feedback (thanks again!) I decided to rework the module a bit.

What changed and why:

- I dropped the NotebookHandler class and went for two simple functions for 
reading (`read_notebook_node_from_file`) and writing 
(`write_notebook_node_to_file`) the notebook.
Both are named read*/write* because the underlying nbformat functions are named
almost the same (for me this is more natural than calling them parse, or format
but if you disagree I will of course rename them):
	- http://nbformat.readthedocs.io/en/latest/api.html#nbformat.reads
	- http://nbformat.readthedocs.io/en/latest/api.html#nbformat.writes

Both are named *_from_file/*_to_file because within a Bear a file is understood
as a list containing the content of a file linewise.

- The issue with final newline characters within a notebook cell is now dealt
with directly in the `autopep8_fix_code`(the decorator is gone) (thanks for pointing out the case that there also could be too many). See also the 
corresponding comment within the function. There might be a more elegant 
solution but I do not know one.

- The issue related to the final newline character in the corrected file 
(not cell) is handled just before creating the diff (which led me to removing
the NotebookHandler).

- The `run()` method is now a bit more verbose but I am not sure if moving there 
is really a benefit in moving parts aways.

To me current solution is cleaner (and simpler). Please let me know what you 
think about it. :)

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,116 @@
...
+    Reads a notebook from a string array and returns the NotebookNode
+    object.
+
+    :param file: The notebook file contents as string array.
+    :return:     The notebook as NotebookNode.
+    """
+    return nbformat.reads(''.join(file), nbformat.NO_CONVERT)
+
+
+def write_notebook_node_to_file(notebook_node):
```
sils1297: it doesn't write to file

sils1297:
@caichinger yes indeed - way better!

bears/python/PEP8NotebookBear.py
sils1297: i.e. I think you'll want to rename it (sorry for being so brief, hope it's ok)

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,116 @@
...
+            # autopep8.fix_code automatically adds a final newline at the end,
+            # e.g. `autopep8.fix_code('a=1')` yields 'a = 1\n'.
+            # This is not related to the 'W292' flag, i.e.
+            # `autopep8.fix_code('a=1', options=dict(ignore=('W292',)))` gives
+            # the same result.
+            # For notebook cells, this behaviour does not make sense, hence
+            # newline is removed if `source` does not end with one.
+            if not source.endswith('\n'):
+                return source_corrected[:-1]
+            return source_corrected
```
sils1297: hm might it be cleaner to have this outside the class and pass the variables explicitly rather than in the closure? https://www.python.org/dev/peps/pep-0020/

bears/python/PEP8NotebookBear.py
caichinger: Well it writes what is called a file within a Bear, namely a list containing the content of the code-text-file linewise. The *write* is motivated by the use of http://nbformat.readthedocs.io/en/latest/api.html#nbformat.writes.
I wanted to stay with the names used in the Bears and `nbformat`.

However, if this remains a source of confusion we rename to e.g.:
- read -> parse?
- write -> format?
- file -> string_array?

@AbdealiJK : you would prefer it like this?

caichinger: @sils1297 : no need to worry about brevity! :)
okay, I will await your opinions on the new names and then rename both functions.

sils1297: notebook_node_to_string or to_list or so?

caichinger: I will give `notebook_node_to_list` and `notebook_node_from_list` a try. Should not sound fancier than it is...

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,125 @@
...
+# dictionaries, but allow attribute access. The structure of these objects
+# matches the notebook format specification.
+
+
+def notebook_node_from_list(file):
+    """
+    Reads a notebook from a string array and returns the NotebookNode
+    object.
+
+    :param file: The notebook file contents as string array.
```
sils1297: rename the parameter to string_list maybe?

sils1297:
oh btw. circleci fails because there's some issue with one of the doc comments, see https://circleci.com/gh/coala-analyzer/coala-bears/2824 for more info, it's apparently not valid restructured text

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,125 @@
...
+                                     .splitlines(True))
+    return notebook_node_as_string_array
+
+
+def autopep8_fix_code_cell(source, options=None, apply_config=None):
+    """
+    Applies autopep8.fix_code and takes care of newline characters.
+
+    autopep8.fix_code automatically adds a final newline at the end,
+    e.g. `autopep8.fix_code('a=1')` yields 'a = 1\n'.
```
sils1297: use double backticks as we're in rst here

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class PEP8NotebookBear(LocalBear):
+    LANGUAGES = {"Python", "Python 2", "Python 3"}
+    REQUIREMENTS = {PipRequirement('autopep8', '1.*'),
+                    PipRequirement('nbformat', '4.*')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/83333'
```
sils1297: I just watched this - it's super awesome!

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,125 @@
...
+                   "select": pep_select,
+                   "max_line_length": max_line_length,
+                   "indent_size": indent_size}
+        notebook_node = notebook_node_from_list(file)
+        cells = notebook_node['cells']
+
+        for i, cell in enumerate(cells):
+            if cell['cell_type'] != 'code':
+                continue
+            source = cell['source']
```
sils1297: ths doesn't get reused, can't you pass it directly?

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,125 @@
...
+            cells[i]['source'] = autopep8_fix_code_cell(source,
```
sils1297: can't you do `cell['source']` instead and then you don't need enumerate

tests/python/PEP8NotebookBearTest.py
```diff
@@ -0,0 +1,133 @@
...
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.5.2"
```
sils1297: can you remove unrelated stuff so it get's shorter and only holds the relevant parts? Maybe one or two unrlated cells to show that they're not affected

bears/python/PEP8NotebookBear.py
caichinger: ok, thanks for pointing that out

bears/python/PEP8NotebookBear.py
caichinger: > How to name things: the hardest problem in programming

;) 
okay, let's call it a *string_list*

bears/python/PEP8NotebookBear.py
caichinger: of course!
pls do not ask why it was like this ^^

sils1297: heh, I've seen worse. Look what I found during another code review: https://www.reddit.com/r/badcode/comments/4ufhnz/thats_how_we_roll_at_coala/

tests/python/PEP8NotebookBearTest.py
caichinger: I removed the content from the metadata field, it is now much shorter.
When you apply the bear to the example file you see that it will skip the markdown cell and the one where there is no error. However, you do not see in in the test if this is what you meant?

bears/python/PEP8NotebookBear.py
caichinger: very comforting ;)

AbdealiJK:
@caichinger This is some awesome work. Some usability issues I can think of:

 - Do we check anywhere that the notebook/cell is actually a python notebook ? It can also be a R, Julia, etc notebook correct ?
 - Can't users use the magic keyworks like %pylab, %sql, etc. to actually inject non-python code into python notebooks ? What do we do about such things ?

I think these are something to look at later, so possibly create issues for them (doesn't need to be fixed in this PR)

sils1297:
> Do we check anywhere that the notebook/cell is actually a python notebook ? It can also be a R, Julia, etc notebook correct ?

If the user feeds it to this bear it's his responsibility, in general our bears don't check the language, the user gives the language and the bear assumes that that's true. I think that's the right thing to do.

> Can't users use the magic keyworks like %pylab, %sql, etc. to actually inject non-python code into python notebooks ? What do we do about such things ?

Good question, IMO let's test it out once, see if it breaks everything - and it probably doesnt, worst case I think will be that that line might be treated wrongly - and just file an issue for everything that comes up.

AbdealiJK:
While it's true that we let the use decide which bear to use when,
generally is easy because of the file extension. Here, all of them would
have .ipynb and the user would need to specify each file individually.
Hence a usability issue

On Sunday, August 28, 2016, Lasse Schuirmann <notifications@github.com>
wrote:
> Do we check anywhere that the notebook/cell is actually a python notebook
? It can also be a R, Julia, etc notebook correct ?
>
> If the user feeds it to this bear it's his responsibility, in general our
bears don't check the language, the user gives the language and the bear
assumes that that's true. I think that's the right thing to do.
>
> Can't users use the magic keyworks like %pylab, %sql, etc. to actually
inject non-python code into python notebooks ? What do we do about such
things ?
>
> Good question, IMO let's test it out once, see if it breaks everything -
and it probably doesnt, worst case I think will be that that line might be
treated wrongly - and just file an issue for everything that comes up.

sils1297:
are the other languaged notebooks really called I*python*notebooks?

sils1297:
does the notebook accidentally have the language embedded for each cell or something?

mr-karan:
> https://github.com/coala-analyzer/coala-bears/pull/655#issuecomment-242964926

No iJulia, iRKernel etc

AbdealiJK:
No idea what they're called but the extension is .ipynb
I think there all just called 'notebooks' with different kernels. I can dynamically change the kernel for a notebook

caichinger:
> @caichinger This is some awesome work.

Thank you for your support :)

There are indeed a few things that should be kept in mind, however, they are imo out of the scope of this bear.

Sooner or later I would like to move this (necessary) discussion to some other place because it is more general in nature. There we could set up/maintain a document (place for discussion) to specify requirements and collect possible issues we want to keep in mind. @sils1297 @AbdealiJK : what do you think? I can prepare a draft it you tell me where to put it. Is Jupyter Notebook stuff a case for a coala enhancement proposal or is this topic too specific?

---

Regarding magics (line magic % (only affects the line), cell magic %% (affects the entire cell)):
- Since the user is asked whether a change should be applied or not, I think for now it is okay to leave it like it is.
- However, to provide a better user experience we should take care of them in the long run. For example you can insert %%javascript or %%cython cells which should then be skipped automatically. Note that these magics are part of the IPython kernel and not of Juypter (maybe there are similar features for other kernels but I do not know about them).

---

Regarding programming language:
- I also think it is okay to assume that if the user applies a PEP8 bear then it is about Python. However, for cross-language projects (e.g. I mix Python and R) it would be handy to say something like: apply all these checks to these notebooks and select the appropriate bear for each case/file (yet, this is not something the PEP8NotebookBear has do deal with).
- As already pointed out, the extension is *.ipynb*.
- The information about the used kernel is provided in the notebook's metadata field (which is empty an notebook creation but filled as soon as you put some content in). The matadata field looks like this:

**R**:
```
 "metadata": {
  "kernelspec": {
   "display_name": "R",
   "language": "R",
   "name": "ir"
  },
  "language_info": {
   "codemirror_mode": "r",
   "file_extension": ".r",
   "mimetype": "text/x-r-source",
   "name": "R",
   "pygments_lexer": "r",
   "version": "3.2.2"
  }
 },
```

**Python**:
```
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.5.2"
  }
 },
```
- As already mentioned, each cell has a cell type which can be e.g. 'code', 'markdown', ... There is no language information attached to each cell.

bears/python/PEP8NotebookBear.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Writes a NotebookNode to a list of strings.
+
+    :param notebook_node: The notebook as NotebookNode to write.
+    :return:              The notebook as list of strings (linewise).
+    """
+    notebook_node_as_string_list = (nbformat.writes(notebook_node,
+                                                    nbformat.NO_CONVERT)
+                                    .splitlines(True))
+    return notebook_node_as_string_list
```
sils1297: just return it directly

chauffer:
Hi, I'm sils trying out if our access policies are correct

chauffer:
@rultor merge

sils1297:
commit message Fixes -> Closes please

bears/python/PEP8NotebookBear.py
caichinger: sure! :)

sils1297:
Hey,

we use some bug prediction in coala which relies on the `Fixes` keyword in commit messages. To get good results from that we need to use `Closes` for normal issues instead of `Fixes` which should only be used for real bugs. (See also [the commit message docs](http://coala.readthedocs.io/en/latest/Getting_Involved/Writing_Good_Commits.html).)

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix up the message for you - just tell us :).

Cheers!

tests/python/PEP8NotebookBearTest.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Utility function to prepare Notebook strings for further processing.
+
+    replace() necessary to avoid
+    nbformat.reader.NotJSONError: Notebook does not appear to be JSON
+    which is caused by
+    json.decoder.JSONDecodeError: Invalid control character
+    which is caused by the \n characters inside the input cells
+    """
+    return file.replace("\n\"", "\\n\"")
```
sils1297: instead of preparing it with that function you could just do a raw string, right? `r"\n" == "\\n"`

sils1297:
Hi,

it looks like your PR is out of date and needs a rebase. [This page](http://coala.readthedocs.org/en/latest/Users/Tutorials/Git_Help.html) may help you to get started on this. (*Do not press the update branch button. It's evil.*)

Add the original coala repository as a remote:

```
git remote add upstream https://github.com/coala-analyzer/coala.git
```

Then go to your branch and rebase it onto our master:

```
git checkout <yourbranch>
git fetch upstream
git rebase --interactive upstream/master
```

Now push it up. You will need to do a force push because you change history here (do not do this on master branches :)):

```
git push --force
```

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix it up for you - just tell us :).

Ping us on [our gitter channel](https://gitter.im/coala-analyzer/coala) if you need help!

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/655/commits/a94f6dce53df982116457c47db480a52996b0efa

sils1297:
but needs a rebase

tests/python/PEP8NotebookBearTest.py
caichinger: Omg, I wrote a function for that, yes. At the time of writing I did not see through it...
Thanks! :+1:

AbdealiJK:
Sorry, but you need a rebase again.
It can get merged as soon as the rebase is done though

caichinger:
@AbdealiJK no problem, I have just rebased :)

sils1297:
@caichinger let's see if everything passes now, then we shouldbe able to merge it right ahead

sils1297:
I'll cherry pick this for your convenience :)

sils1297:
cherry picked to https://github.com/coala-analyzer/coala-bears/pull/766

PR #660, opened 2016-08-07
WIP:  Add deprecate_settings

Asnelchristian:
I need plenty of reviews ( I am not quite sure about doing this here i am thinking on moving it to coala repo But first please review it  :) )

### DIFF
--- a/bears/deprecate_settings.py
+++ b/bears/deprecate_settings.py
@@ -1,0 +1,20 @@
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def deprecate_settings(**depr_args):
+    def _deprecate_decorator(func):
+        def wrapping_function(*args, **kwargs):
+            for arg, deprecated_arg in depr_args.items():
+                if deprecated_arg in kwargs and arg not in kwargs:
+                    print("The setting `{}` is deprecated. Please use `{}` "
+                          "instead.".format(deprecated_arg, arg))
+                    kwargs[arg] = kwargs[deprecated_arg]
+                    del kwargs[deprecated_arg]
+            return func(*args, **kwargs)
+        wrapping_function.__metadata__ = FunctionMetadata.from_function(func)
+        for arg, deprecated_arg in depr_args.items():
+            wrapping_function.__metadata__.optional_params[
+                deprecated_arg] =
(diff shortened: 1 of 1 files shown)


sils1297:
This should definitely be under coala, not in bears!

bears/deprecate_settings.py
```diff
@@ -0,0 +1,20 @@
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def deprecate_settings(**depr_args):
```
Makman2: doccomment missing

bears/deprecate_settings.py
```diff
@@ -0,0 +1,20 @@
...
+    def _deprecate_decorator(func):
+        def wrapping_function(*args, **kwargs):
```
Makman2: I would add `@wraps(func)`

bears/deprecate_settings.py
```diff
@@ -0,0 +1,20 @@
...
+            for arg, deprecated_arg in depr_args.items():
+                if deprecated_arg in kwargs and arg not in kwargs:
```
Makman2: so if the non-deprecated-arg was given, we just ignore the deprecated one?

bears/deprecate_settings.py
```diff
@@ -0,0 +1,20 @@
...
+                    print("The setting `{}` is deprecated. Please use `{}` "
+                          "instead.".format(deprecated_arg, arg))
+                    kwargs[arg] = kwargs[deprecated_arg]
+                    del kwargs[deprecated_arg]
+            return func(*args, **kwargs)
+        wrapping_function.__metadata__ = FunctionMetadata.from_function(func)
+        for arg, deprecated_arg in depr_args.items():
+            wrapping_function.__metadata__.optional_params[
+                deprecated_arg] =
+                wrapping_function.__metadata__.optional_params[arg]
```
Makman2: what about required params?

sils1297: also he said this doesn't work yet but I don't have time to fix it up right now, I would guess at least the assignment won't work, you'll have to do a setattr manually because you can't just add a metadata attribute

Asnelchristian:
okay i took note 
i am going to close it and move the pr to coala repo

Issue #597, opened 2016-07-21
generate_package: Test the version variable


Adrianzatreanu:
Currently it has a pragma no cover as the testing should be done with mocking probably, so those should be removed.

``            if 'dev' in bear_version:  # pragma: no cover ``

``                  bear_version = bear_version[:bear_version.find("dev")] + ( 
                     nano_version)``
``
              else:  # pragma: no cover
``

tushar-rishav:
I think you may be able to mock the version by containing it within an object (dict) and then mocking that object. Something like:
```python
from unittest.mock import patch
from coalib import VERSION

version_dict = {'v':VERSION}
print(version_dict) # Original version
with patch.dict(version_dict, v=458):
    print(version_dict['v'])  # Mocked version now do whatever

print(version_dict) # Original version again

```
CC @Adrianzatreanu

chriscoyfish:
I can work on this issue

Adrianzatreanu:
All yours!

PR #661, opened 2016-08-08
generate_packageTest: Refactor sys.argv mock

chriscoyfish:
Fixes: https://github.com/coala-analyzer/coala-bears/issues/597

### DIFF
--- a/tests/generate_packageTest.py
+++ b/tests/generate_packageTest.py
@@ -3,7 +3,8 @@ import shutil
 import sys
 import unittest
 from unittest.mock import patch
-from bears.generate_package import (touch, create_file_from_template,
+
+from bears.generate_package import (VERSION, touch, create_file_from_template,
                                     create_file_structure_for_packages,
                                     perform_register, perform_upload, main,
                                     create_upload_parser)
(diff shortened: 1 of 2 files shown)


tests/generate_packageTest.py
```diff
@@ -90,25 +91,41 @@ class mainTest(unittest.TestCase):
...
+        main()
+        with open(self.CSS_BEAR_SETUP_PATH) as fl:
+            setup_py = fl.read()
+        self.assertNotIn(VERSION, setup_py)
+        self.assertIn(VERSION[:VERSION.find("dev")], setup_py)
+
     @patch('bears.generate_package.perform_upload')
     def test_upload(self, call_mock):
-        sys.argv = ["generate_package.py", "--upload"]
+        sys.argv.append("--upload")
```
Adrianzatreanu: hey this commit looks amazing, but i think the changes in lines 121, 128 and 94 are unrelated to this commit and should be changed to another commit accordingly :)

chriscoyfish: I'm happy to omit moving the argv to setUp, but my new tests also need that set, so it seemed cleaner to move it to setUp and save some LOCs. Please let me know if this seems reasonable?

Adrianzatreanu: it sounds really good, but to fix this, you could first move the set in a commit, and THEN do your changes. this way your changes will use the set and you will also have correct commits. is this okay :) ?

PR #681, opened 2016-08-12
Look for `pmd` binary for CPDBear

julianwachholz:
Using brew to install PMD on MacOS will provide a `pmd` binary
instead of the `run.sh` mentioned in the official docs.
This change will allow CPDBear to use whichever variant is available.

### DIFF
--- a/bears/general/CPDBear.py
+++ b/bears/general/CPDBear.py
@@ -41,7 +41,7 @@ class CPDBear(GlobalBear):
     def check_prerequisites(cls):  # pragma: no cover
         if which("bash") is None:
             return "bash is not installed."
-        elif which("run.sh") is None:
+        if which("pmd") is None and which("run.sh") is None:
             return ("CPD is missing. Make sure to install it from "
                     "<https://pmd.github.io/>")
         else:
@@ -89,7 +89,8 @@ class CPDBear(GlobalBear):
             "--skip-duplicate-files": skip_duplicate_files}
 
         files = ",".join(self.file_dict.keys())
-        arguments = ("bash", which("run.sh"), "cpd", "--skip-lexical-errors",
+        executable = which("pmd") or which("run.sh")
+        arguments = ("bash", executable, "cpd", "--skip-lexical-errors",
                      "--minimum-tokens", str(minimum_tokens),
                      "--language", self.lowered_lang_dict[language],
                      "--files", files,


mr-karan:
Can you edit the shortlog of your commit message to include the filename? https://coala.readthedocs.io/en/latest/Getting_Involved/Writing_Good_Commits.html

rest looks good :+1:

bears/general/CPDBear.py
```diff
@@ -41,7 +41,7 @@ class CPDBear(GlobalBear):
     def check_prerequisites(cls):  # pragma: no cover
         if which("bash") is None:
             return "bash is not installed."
-        elif which("run.sh") is None:
+        if which("pmd") is None and which("run.sh") is None:
```
mr-karan: so what happens when the user doesn't have `pmd`? `which("run.sh")` won't be even evaluated

julianwachholz: On the contrary, both calls to `which` will always be called because of the `and`. You want this condition to evaluate to `True` in the case where a user has neither installed, making the Bear unusable.

mr-karan: Sorry, I don't really get this. Is `PMD` essential for CPDBear to run? Or only on MacOS? Because if it's not and the user has `run.sh` but not `PMD`, your method won't work because of the `and`

sils1297: PMD contains CPD, run.sh is a script that runs PMD/CPD on linux.

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/681/commits/fd2d0c029a0ee9782b652cfbd74856a74c638b66 - thanks for your help and thanks for bearing with us!

sils1297:
It looks like your PR is out of date and needs a rebase. [This page](http://coala.readthedocs.org/en/latest/Users/Tutorials/Git_Help.html) may help you to get started on this. (*Do not press the update branch button. It's evil.*)

Add the original coala repository as a remote:

```
git remote add upstream https://github.com/coala-analyzer/coala.git
```

Then go to your branch and rebase it onto our master:

```
git checkout <yourbranch>
git fetch upstream
git rebase --interactive upstream/master
```

Now push it up. You will need to do a force push because you change history here (do not do this on master branches :)):

```
git push --force
```

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix it up for you - just tell us :).

Ping us on [our gitter channel](https://gitter.im/coala-analyzer/coala) if you need help!

sils1297:
https://github.com/coala-analyzer/coala-bears/pull/681/commits/fa2df1fc11df3e375736b51f67d7bf1b184da61c reack

sils1297:
@rultor merge

PR #692, opened 2016-08-15
Sils/links

sils1297:


SanketDG:
ack c249561

SanketDG:
ack a9b1947

SanketDG:
unack 7f6a706, can you explain the git commit message body what you do.

SanketDG:
ack 4be2d76

tests/general/InvalidLinkBearTest.py
```diff
@@ -74,72 +74,39 @@ def test_run(self):
...
-        valid_file = """
-            https://en.wikipedia.org/wiki/Hello_(Adele_song)/200
-            "https://github.com/coala-analyzer/coala-bears/issues/200"
-            http://httpbin.org/status/200\n
-            http://httpbin.org/status/201
-            <http://httpbin.org/status/202>
-            'http://httpbin.org/status/203'
-            http://httpbin.org/status/204.....
-            http://not a link dot com
+        # Not existing
```
SanketDG: Not existing? It wont find the URL.

tests/general/InvalidLinkBearTest.py
```diff
@@ -74,72 +74,39 @@ def test_run(self):
...
+        http://not a link dot com
         """.splitlines()
 
         self.assertResult(valid_file=valid_file)
 
-        # Markdown Links
-        valid_file = """
-            [httpbin](http://httpbin.org/status/200)
-        """.splitlines()
+        invalid_file = """http://coalaisthebest.com/
```
SanketDG: Here you can put `Not existing`

SanketDG:
unack 7db72a6

tests/general/InvalidLinkBearTest.py
sils1297: nope, a comment would be a line that would have to fail the check.

sils1297:
reack
45c4ca5 InvalidLinkBear: Disallow backslashes in URLs
85d2aad InvalidLinkBear: Stop at <
faa6c3d InvalidLinkBear: Remove redundant escapes

bears/general/InvalidLinkBear.py
```diff
@@ -42,7 +42,7 @@ def get_status_code(url, timeout):
     def find_links_in_file(file, timeout, ignore_regex):
         ignore_regex = re.compile(ignore_regex)
         regex = re.compile(
-            r'((ftp|http)s?://\S+\.(?:[^\s()\'\"<>|]+|'
+            r'((ftp|http)s?://[^\.:\s_/?#\[\]@]+\.(?:[^\s()\'\"<>|]+|'
```
Makman2: every metacharacter excecpt `[` and `]` loses its meaning, so you don't need to escape the stuff (except the brackets)

Makman2: oh and `[` too (does not need backslash)

tests/general/InvalidLinkBearTest.py
```diff
@@ -82,6 +82,7 @@ def test_run(self):
         # Quotes
         "https://github.com/coala-analyzer/coala-bears/issues/200"
         'http://httpbin.org/status/203'
+        ('http://httpbin.org/status/200').install_command()
```
Makman2: does this one belong to this commit?

tests/general/InvalidLinkBearTest.py
```diff
@@ -74,72 +74,39 @@ def test_run(self):
...
-        valid_file = """
-            [httpbin](http://httpbin.org/status/200)
-        """.splitlines()
+        invalid_file = """http://coalaisthebest.com/
+        http://httpbin.org/status/301  # Redirect
+        http://httpbin.org/status/302  # Redirect
+        http://httpbin.org/status/404
+        http://httpbin.org/status/410
+        http://httpbin.org/status/500
+        http://httpbin.org/status/503""".splitlines()
```
Makman2: I would move the `spltlines()` below, makes it more obvious imo

tests/general/InvalidLinkBearTest.py
```diff
@@ -74,72 +74,39 @@ def test_run(self):
...
-            http://httpbin.org/status/201
-            <http://httpbin.org/status/202>
-            'http://httpbin.org/status/203'
-            http://httpbin.org/status/204.....
-            http://not a link dot com
+        # Not a link
+        http://not a link dot com
         """.splitlines()
 
         self.assertResult(valid_file=valid_file)
```
Makman2: maybe move the `splitlines()` to the expression inside braces?

tests/general/InvalidLinkBearTest.py
sils1297: yes, previously this failed because it separated at the wrong dot - actually gmail web UI highlights this one wrong as well :P

tests/general/InvalidLinkBearTest.py
Makman2: btw shouldn't the test lines also contain special characters (like enclosing in `<>`) so we can actually see if links are correctly pick up?

sils1297: yeah, can you file an issue? This is a refactoring commit only

sils1297:
reack
a8e4bab InvalidLinkBear: Stop at <
28978f6 InvalidLinkBear: Remove redundant escapes

tests/general/InvalidLinkBearTest.py
sils1297: nah don't think so, actually we should change assertResult to do it on it's own

sils1297: IMO

tests/general/InvalidLinkBearTest.py
```diff
@@ -74,72 +74,37 @@ def test_run(self):
...
+        http://httpbin.org/status/301  # Redirect
+        http://httpbin.org/status/302  # Redirect
+        http://httpbin.org/status/404
+        http://httpbin.org/status/410
+        http://httpbin.org/status/500
+        http://httpbin.org/status/503"""
 
-        self.assertResult(valid_file=valid_file)
+        for line in invalid_file.splitlines():
+            self.assertResult(invalid_file=[line])
```
hypothesist: We should have a mixed case test too, with most links working but some invalid ones thrown in too

sils1297: can you file an issue? Outside scope.

bears/general/InvalidLinkBear.py
```diff
@@ -42,7 +42,7 @@ def get_status_code(url, timeout):
     def find_links_in_file(file, timeout, ignore_regex):
         ignore_regex = re.compile(ignore_regex)
         regex = re.compile(
-            r'((ftp|http)s?://\S+\.(?:[^\s()\'\"<>|]+|'
+            r'((ftp|http)s?://[^.:\s_/?#[\]@]+\.(?:[^\s()\'\"<>|]+|'
```
hypothesist: Is there an RFC or something that verifies this? That everything except these chars are allowed

tests/general/InvalidLinkBearTest.py
```diff
@@ -74,72 +74,37 @@ def test_run(self):
...
+        <http://httpbin.org/status/202>
+        http://httpbin.org/status/204.....
+        [httpbin](http://httpbin.org/status/200)
+        |http://httpbin.org/status/200|
 
-        # Markdown Links
-        valid_file = """
-            [httpbin](http://httpbin.org/status/200)
+        # Not a link
+        http://not a link dot com
```
hypothesist: these are sometimes links. For instance, our university used to have this internal site like `http://something` without any GTLD or trailing slash or anything. When entered in the browser, they actually work because there is a internal DNS record that matches.

sils1297: that doesn't sound like something we'll want to verify though

bears/general/InvalidLinkBear.py
sils1297: yeah we looked it up, not sure where it is. In the later part we have added a bit extra (in another commit) to make a heuristic which actually seems to work better then GMails one in terms of finding the URL that the user "means" ;)

tests/general/InvalidLinkBearTest.py
hypothesist: https://github.com/coala-analyzer/coala-bears/issues/694

tests/general/InvalidLinkBearTest.py
```diff
@@ -90,6 +90,7 @@ def test_run(self):
         [httpbin](http://httpbin.org/status/200)
         |http://httpbin.org/status/200|
         <h3>Something http://httpbin.org/status/200</h3>
+        repo=\\"http://httpbin.org/status/200\\"
```
hypothesist: Wait, why's this in the valid links list?

hypothesist: If we're disallowing backslashes shouldn't InvalidLinkBear flag this?

hypothesist:
ack https://github.com/coala-analyzer/coala-bears/pull/692/commits/4dec4b162832fd85d1683c02b76018393a903381 https://github.com/coala-analyzer/coala-bears/pull/692/commits/99e230b63bad9455f12f6177fb2a8cbfe11591b2

tests/general/InvalidLinkBearTest.py
hypothesist: Ah jeez, right. But then shouldn't there be a test verifying the backslash part?

sils1297: this is the test, if that doesn't work this URL will not pass the check because `...200\` is invalid

hypothesist:
ack https://github.com/coala-analyzer/coala-bears/pull/692/commits/6ea1bf2a020f9d92e31abd5c68144115c0030cea

hypothesist:
@rultor merge

PR #699, opened 2016-08-17
Add DocumentationStyleBear for doc styles

SanketDG:
This bear checks the documentation comments in the given file
against a particular style.

https://asciinema.org/a/7sfk3i9oxs1ixg2ncsu3pym0u
https://asciinema.org/a/9skd6d9c826ioibeobpz4og1x

### DIFF
--- a/bears/documentation/DocumentationStyleBear.py
+++ b/bears/documentation/DocumentationStyleBear.py
@@ -1,0 +1,105 @@
+from coalib.bearlib.languages.documentation.DocumentationComment import (
+    DocumentationComment)
+from coalib.bearlib.languages.documentation.DocstyleDefinition import (
+    DocstyleDefinition)
+from coalib.bearlib.languages.documentation.DocumentationExtraction import (
+    extract_documentation)
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.results.TextRange import TextRange
+
+
+class DocumentationStyleBear(LocalBear):
+    LANGUAGES = {language for docstyle, language in
+                 DocstyleDefinition.get_available_definitions()}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/7sfk3i9oxs1ixg2ncsu3pym0u'
+    CAN_DETECT = {'Documentation'}
+
+    def run(self, filename, file, language: str, docstyle: str='default'):
+        """
+        Checks for certain in-code documentation styles.
+
+        It currently checks for the following style: ::
+
+            The first line needs to have no indentation.
+                - Following lines can have indentation
+
+            :param x:
+                4 space indent
+            :return:
+                also 4 space indent
+                following lines are also 4 space indented
+
+        :param language: The programming language of the file(s).
+        :param docstyle: The docstyle to use. For example ``default`` or
+                         ``doxygen``. Docstyles are language dependent, meaning
+                         not every language is supported by a certain docstyle.
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+            parsed = doc_comment.parse()
+            metadata = iter(parsed)
+
+            # Assuming that the first element is always the only main
+            # description.
+            main_description = next(metadata)
+
+            # 1 empty line shall follow main description (except it's empty or
+            # no annotations follow).
+            if main_description.desc.strip() != '':
+                main_description = main_description._replace(
+                    desc='\n' + main_description.desc.strip() + '\n' *
+                         (1 if len(parsed) == 1 else 2))
+
+            new_metadata = [main_description]
+            for m in metadata:
+                # Split newlines and remove leading and trailing whitespaces.
+                stripped_desc = list(map(str.strip, m.desc.splitlines()))
+
+                if len(stripped_desc) == 0:
+                    # Annotations should be on their own line, though no
+                    # further description follows.
+                    stripped_desc.append('')
+                else:
+                    # Wrap parameter description onto next line if it follows
+                    # annotation directly.
+                    if stripped_desc[0] != '':
+                        stripped_desc.insert(0, '')
+
+                # Indent with 4 spaces.
+                stripped_desc = ('' if line == '' else ' ' * 4 + line for line
+                                 in stripped_desc)
+
+                new_desc = '\n'.join(stripped_desc)
+
+                # Strip away trailing whitespaces and obsolete newlines (except
+                # 1 newline which is mandatory).
+                new_desc = new_desc.rstrip() + '\n'
+
+                new_metadata.append(m._replace(desc=new_desc.lstrip(' ')))
+
+            new_comment = DocumentationComment.from_metadata(
+                new_metadata, doc_comment.docstyle_definition,
+                doc_comment.marker, doc_comment.indent, doc_comment.range)
+
+            if new_comment != doc_comment:
+                # Something changed, let's apply a result.
+                diff = Diff(file)
+
+                # We need to update old comment positions, as `assemble()`
+                # prepends indentation for first line.
+                old_range = TextRange.from_values(
+                    doc_comment.range.start.line,
+                    1,
+                    doc_comment.range.end.line,
+                    doc_comment.range.end.column)
+                diff.replace(old_range, new_comment.assemble())
+
+                yield Result(
+                    origin=self,
+                    message='Documentation does not have correct style.',
+                    affected_code=(diff.range(filename),),
+                    diffs={filename: diff})
--- a/tests/documentation/DocumentationStyleBearTest.py
+++ b/tests/documentation/DocumentationStyleBearTest.py
@@ -1,0 +1,55 @@
+from queue import Queue
+import os.path
+import unittest
+
+from coalib.results.Diff import Diff
+from coalib.settings.Section import Section
+from coalib.testing.LocalBearTestHelper import execute_bear
+
+from bears.documentation.DocumentationStyleBear import DocumentationStyleBear
+
+
+def load_testfile(filename):
+    filepath = os.path.join(os.path.dirname(__file__), 'test_files',
+                            'DocumentationStyleBear', filename)
+    with open(filepath) as fl:
+        return fl.read()
+
+
+def test(test_file, expected_file):
+    def test_function(self):
+        test_file_content = load_testfile(test_file).splitlines(True)
+
+        arguments = {'language': 'python', 'docstyle': 'default'}
+        section = Section('test-section')
+        for key, value in arguments.items():
+            section[key] = value
+
+        with execute_bear(
+                DocumentationStyleBear(section, Queue()),
+                test_file,
+                test_file_content,
+                **arguments) as results:
+
+            diff = Diff(test_file_content)
+            for result in results:
+                # Only the given test file should contain a patch.
+                self.assertEqual(len(result.diffs), 1)
+
+                diff += result.diffs[test_file]
+
+        correct_file_content = load_testfile(expected_file).splitlines(True)
+
+        self.assertEqual(correct_file_content, diff.modified)
+
+    return test_function
+
+
+class DocumentationStyleBearTest(unittest.TestCase):
+    test_bad1 = test('bad_file.py.test', 'bad_file.py.test.correct')
+    test_bad2 = test('bad_file2.py.test', 'bad_file2.py.test.correct')
+    test_bad3 = test('bad_file3.py.test', 'bad_file3.py.test.correct')
+    test_bad4 = test('bad_file4.py.test', 'bad_file4.py.test.correct')
+    test_bad5 = test('bad_file5.py.test', 'bad_file5.py.test.correct')
+    test_good1 = test('good_file.py.test', 'good_file.py.test')
+    test_good2 = test('good_file2.py.test', 'good_file2.py.test')
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test.correct
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test.correct
@@ -1,0 +1,12 @@
+def hello_there(x):
+    """
+    hello and no following empty lines
+
+    :param x:
+        5 space indent
+        just a second test
+    :return:
+        3 space indent
+        following lines are 4 space indented
+    """
+    return None
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
@@ -1,0 +1,12 @@
+def myfunc(a, b, x):
+    """
+
+    :param a:
+        slope
+    :param b:
+        offset
+    :param x:
+    :return:
+    """
+
+    return a * x + b
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test
@@ -1,0 +1,12 @@
+def hello_there(x):
+    """
+    hello and no following empty lines
+    :param x:
+         5 space indent
+        just a second test
+
+    :return:
+       3 space indent
+        following lines are 4 space indented
+    """
+    return None
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file2.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file2.py.test
@@ -1,0 +1,15 @@
+def x():
+    return 55
+
+
+def qr():
+    """
+       this is heavy
+
+    :param x:
+        ss
+    :return:abc"""; return 88
+
+
+def w():
+    """ Hello world """
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file2.py.test.correct
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file2.py.test.correct
@@ -1,0 +1,19 @@
+def x():
+    return 55
+
+
+def qr():
+    """
+    this is heavy
+
+    :param x:
+        ss
+    :return:
+        abc
+    """; return 88
+
+
+def w():
+    """
+    Hello world
+    """
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file3.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file3.py.test
@@ -1,0 +1,3 @@
+"""
+:param t:xyz
+:return:123"""
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file3.py.test.correct
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file3.py.test.correct
@@ -1,0 +1,6 @@
+"""
+:param t:
+    xyz
+:return:
+    123
+"""
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test
@@ -1,0 +1,9 @@
+def myfunc(a, b, x):
+    """
+
+    :param a:slope
+    :param b:offset
+    :param x:
+    :return:"""
+
+    return a * x + b
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file5.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file5.py.test
@@ -1,0 +1,8 @@
+def empty():
+    """
+        first line not correctly aligned.
+        But second can be aligned differently.
+
+    :param x:
+    :return:
+    """; return 88
--- a/tests/documentation/test_files/DocumentationStyleBear/bad_file5.py.test.correct
+++ b/tests/documentation/test_files/DocumentationStyleBear/bad_file5.py.test.correct
@@ -1,0 +1,8 @@
+def empty():
+    """
+    first line not correctly aligned.
+        But second can be aligned differently.
+
+    :param x:
+    :return:
+    """; return 88
--- a/tests/documentation/test_files/DocumentationStyleBear/good_file.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/good_file.py.test
@@ -1,0 +1,12 @@
+def hello_there(x):
+    """
+    hello
+
+    :param x:
+        4 space indent
+        just a second test
+    :return:
+        4 space indent
+        following lines are 4 space indented
+    """
+    return None
--- a/tests/documentation/test_files/DocumentationStyleBear/good_file2.py.test
+++ b/tests/documentation/test_files/DocumentationStyleBear/good_file2.py.test
@@ -1,0 +1,7 @@
+def empty():
+    """
+    this is heavy
+
+    :param x:
+    :return:
+    """; return 88


bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,94 @@
...
+                doc_comment.marker, doc_comment.indent, doc_comment.range)
+
+            def occurrences(string, sub):
+                count = start = 0
+                while True:
+                    start = string.find(sub, start) + 1
+                    if start > 0:
+                        count += 1
+                    else:
+                        return count
```
Makman2: there's actually `"my_string".count(substring)` :D

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,94 @@
...
+
+            if new_comment != doc_comment:
+                # Something changed, let's apply a result.
+                diff = Diff(file)
+
+                # Delete the old comment
+                for i in range(doc_comment.range.start.line,
+                               doc_comment.range.end.line + 1):
+                    diff.delete_line(i)
```
Makman2: https://github.com/coala-analyzer/coala/pull/2620 is going to come, if it gets merged earlier than yours we could use that :)+

bears/documentation/DocumentationStyleBear.py
SanketDG: oh sorry I need to add comments in here

this is for overlapping counting, `str.count` is for non-overlapping counts.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,94 @@
...
+
+                # Apply the new comment
+                assembled = new_comment.assemble().splitlines(True)
+
+                newline_count = occurrences(new_comment.documentation, "\n\n")
+                end_marker = new_comment.marker[2]
+                end_line = new_comment.range.end.column + newline_count + 1
+                end_marker_pos = file[end_line:][0].find(end_marker)
```
Makman2: isn't this the same?
```
end_marker_pos = file[end_line].find(end_marker)
```

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,94 @@
...
+                end_marker_part = (file[end_line:][0]
```
Makman2: here too

tests/documentation/test_files/DocumentationStyleBear/bad_file.py
```diff
@@ -0,0 +1,11 @@
...
+    """
+    hello
+    :param x:
+         5 space indent
+        just a second test
+    :return:
+       3 space indent
+        following lines are 4 space indented
+    """
+    return None
```
Makman2: though this is at least a good test, we should actually also assert for the diff/corrected version, otherwise we don't know what actually happens :D

but this can be also done after merge :+1:

SanketDG: okay opening an issue :smile:

bears/documentation/DocumentationStyleBear.py
SanketDG: uh perks of search and replacing :3
will fix

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,100 @@
...
+
+            hello
+            :param x:
+                4 space indent
+            :return:
+                also 4 space indent
+                following lines are also 4 space indented
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+
```
Makman2: unneeded newline

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,100 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/7sfk3i9oxs1ixg2ncsu3pym0u'
+    CAN_DETECT = {'Documentation'}
+
+    def run(self, filename, file, docstyle: str, language: str):
+        """
+        Checks for certain in-code documentation styles.
+
+        It currently checks for the following styles:
```
Makman2: styles -> style

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,100 @@
...
+            new_metadata = []
+            for m in metadata:
+                stripped_desc = map(str.lstrip, m.desc.splitlines(True))
+                new_desc = ""
+                for line in stripped_desc:
+                    new_desc += (("\n" if not line.endswith("\n") else "") +
+                                 (" " * 4 if line != "" else "") + line)
+                new_metadata.append(m._replace(desc=new_desc))
+
+            new_metadata = [description] + new_metadata
```
Makman2: I would refactor a little bit in this case:
```python
# delete line:
# description = next(metadata)

new_metadata = [next(metadata)]
for m in metadata:
    ...

# now you can also delete
# new_metadata = [description] + new_metadata
```

bears/documentation/DocumentationStyleBear.py
Makman2: also please define this function at class level or module level. Later we could also move it to the StringProcessing module in `coala-utils`

SanketDG: I dont see why I need to do that.

I defined it there because it was used there.

If it was used at the class level or module level, I would define it in there.

If its used in more than one class, we will move it to StringProcessing.

Makman2: This function gets assigned inside the body on every run to the name `occurrences`. Also it increases nesting which adds additional complexity.
Conceptually this function is not bound to this class, and especially not to a specific function run --> class level (staticmethod) or module level, where it gets initialized once :)

SanketDG: Right that makes sense, so I am making this module level.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,97 @@
...
+
+            # Assuming that the first element is always the only description.
+            new_metadata = [next(metadata)]
+
+            for m in metadata:
+                stripped_desc = map(str.lstrip, m.desc.splitlines(True))
+                new_desc = ""
+                for line in stripped_desc:
+                    new_desc += (("\n" if not line.endswith("\n") else "") +
+                                 (" " * 4 if line != "" else "") + line)
```
Makman2: this can be made to a join :)
```
new_desc = ''.join(... for lin in stripped_desc)
```

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,97 @@
...
+                diff = Diff(file)
+
+                # Delete the old comment
+                diff.delete_lines(doc_comment.range.start.line,
+                                  doc_comment.range.end.line + 1)
+
+                # Apply the new comment
+                assembled = new_comment.assemble().splitlines(True)
+
+                newline_count = count_overlapping_occurrences(
```
Makman2: maybe not `newline_count`, as you count for double-newlines? maybe `double_newline_count`?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,97 @@
...
+                    new_comment.documentation, "\n\n")
+                end_marker = new_comment.marker[2]
+                end_line = new_comment.range.end.column + newline_count + 1
```
Makman2: `end.column`?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,97 @@
...
+                end_marker_pos = file[end_line].find(end_marker)
+                end_marker_part = (file[end_line]
+                                   [:end_marker_pos + len(end_marker) + 1])
+
+                diff.add_lines(new_comment.range.start.line, assembled[:-1])
+                diff.add_lines(new_comment.range.end.line,
+                               [assembled[-1] + "\n"]
+                               if end_marker_part.endswith("\n")
+                               else file[end_line:])
```
Makman2: wait this adds all remaining lines of the file! not the rest of the line!

tests/documentation/DocumentationStyleBearTest.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from bears.documentation.DocumentationStyleBear import DocumentationStyleBear
+from tests.LocalBearTestHelper import verify_local_bear
+
```
Makman2: missing newline

tests/documentation/DocumentationStyleBearTest.py
```diff
@@ -0,0 +1,22 @@
...
+                       "test_files",
+                       "DocumentationStyleBear",
+                       "bad_file.py")) as bad_f:
+    bad_file = bad_f.read()
+
+with open(os.path.join(os.path.dirname(__file__),
+                       "test_files",
+                       "DocumentationStyleBear",
+                       "good_file.py")) as good_f:
+    good_file = good_f.read()
```
Makman2: wanna do a `load_testfile` function? :)

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,80 @@
...
+                following lines are also 4 space indented
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+            metadata = iter(doc_comment.parse())
+
+            # Assuming that the first element is always the only description.
+            new_metadata = [next(metadata)]
+
+            for m in metadata:
+                stripped_desc = map(str.lstrip, m.desc.splitlines(True))
```
Makman2: I think you can do
```python
stripped_desc = (line.lstrip(' \t') for line in m.desc.splitlines(True))  # <-- \n's don't get removed now
new_desc = ''.join((" " * 4 if line != "" else "") + line for line in stripped_desc)
if new_desc[-1] != '\n':
    new_desc += '\n'
new_metadata.append(...)
```

SanketDG: for strings like `\n        hello` it wont work. I have to remove the newlines anyway. Or more cleanly,
I have to remove the spaces between the `\n` and the text.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,80 @@
...
+                end_line = new_comment.range.end.line - 1
+                end_marker_pos = file[end_line].find(end_marker)
+                end_marker_part = (file[end_line]
+                                   [:end_marker_pos + len(end_marker) + 1])
+
+                diff.add_lines(new_comment.range.start.line, assembled[:-1])
+                diff.add_lines(new_comment.range.end.line,
+                               [assembled[-1] + "\n"]
+                               if end_marker_part.endswith("\n")
+                               else file[end_line])
```
Makman2: -->
```python
assembled_string = new_comment.assemble()
assembled = assembled_string.splitlines(True)

end_line = new_comment.range.start.line + assembled_string.count('\n')
diff.add_lines(new_comment.range.start.line, assembled[:-1]
diff.add_line(end_line, assembled[-1] + file[end_line - 1][len(assembled[-1]):])
```
but test this before^^

tests/documentation/DocumentationStyleBearTest.py
```diff
@@ -0,0 +1,22 @@
...
+from tests.LocalBearTestHelper import verify_local_bear
+
+
+def load_testfiles(test, good_file, bad_file):
+    good_file_path = os.path.join(os.path.dirname(__file__), "test_files",
+                                  test, good_file)
+    bad_file_path = os.path.join(os.path.dirname(__file__), "test_files",
+                                 test, bad_file)
+    with open(bad_file_path) as bad_f, open(good_file_path) as good_f:
+        return good_f.read(), bad_f.read()
```
Makman2: hehe I've more though about this one: :D 
```python
def load_testfile(name):
    filename = os.path.join(os.path.dirname(__file__), "test_files", "DocumentationStyleBear", name)
    with open(filename) as fl:
        return fl.read()
```

good_file = load_testfile("good_file.py")
bad_file = load_testfile("bad_file.py")
```

Comment (SanketDG) [inline tests/documentation/DocumentationStyleBearTest.py]: haha yeah this is way better.

but I'd also like to have a "test" parameter, because each of the docmentation bears will have tests like this, and I will also be moving this function to a testutils module in a future PR.

File: bears/documentation/DocumentationStyleBear.py [hunk 18]
```diff
@@ -0,0 +1,85 @@
...
+            :param x:
+                4 space indent
+            :return:
+                also 4 space indent
+                following lines are also 4 space indented
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+            metadata = iter(doc_comment.parse())
+            from pprint import pprint
+            pprint(doc_comment.parse())
```
SanketDG: noooooooooooooooooo

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,77 @@
...
+                                  doc_comment.range.end.line)
+
+                # Apply the new comment
+                assembled_string = new_comment.assemble()
+                assembled = assembled_string.splitlines(True)
+
+                end_line = new_comment.range.end.line
+                diff.add_lines(new_comment.range.start.line, assembled[:-1])
+                diff.add_lines(end_line, assembled[-1] +
+                               file[end_line - 1][len(assembled[-1]):])
```
Makman2: hm I'm wondering why it does work... `add_lines` should only allow lists

Makman2: especially I'm wondering about `end_line`, as it's set to `new_comment.range.end.line`, and this is the same range like in the old doccomment^^

SanketDG: > hm I'm wondering why it does work... add_lines should only allow lists

you are right it thus breaks up the string because:
```
>>> x = "hello"
>>> list(x)
['h', 'e', 'l', 'l', 'o']
>>> 
```



> especially I'm wondering about end_line, as it's set to new_comment.range.end.line, and this is the same range like in the old doccomment^^

Even I am not sure what happens here :confused:, the old comment's range is always equal to the new_comment's range.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+            metadata = iter(doc_comment.parse())
+
+            # Assuming that the first element is always the only description.
+            new_metadata = [next(metadata)]
+
+            for m in metadata:
+                stripped_desc = list(map(str.lstrip, m.desc.splitlines(True)))
+                stripped_desc[0] = (('\n' + ' ' * 4 + stripped_desc[0])
```
Makman2: this crashes on following docstring:
```
"""
stuff

:param x:
:return:
```
as `splitlines()` returns an empty list on empty strings^^

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,77 @@
...
+                                    if not stripped_desc[0] == ''
+                                    else stripped_desc[0])
+                new_desc = ''.join(('\n' if not line.endswith('\n') else '') +
+                                   (' ' * 4 if line != '' else '') + line
+                                   for line in stripped_desc)
```
Makman2: oh I found a little bug:
If I have this doc:
```
:param x: On the same line
```
it gets wrapped accordingly, but leaves some whitespaces after `:param x:`, I think it's the indentation.
after running coala again, they get removed
this should happen on the first run :grinning:

Makman2:
unack 5ddd06d

bears/documentation/DocumentationStyleBear.py
SanketDG: I dont see it crashing. It gives no output instead?

bears/documentation/DocumentationStyleBear.py
SanketDG: Yeah, could I do this later, this is something entirely different to the documentation style.

bears/documentation/DocumentationStyleBear.py
Makman2: oh forgot to apppend the closing `"""`

Makman2:
```
mak@hp-elitebook:~/dev/test$ cat test.py
def x:
 return 55
 
def qr:
    """
       this is heavy

    :param x:
    :return:"""return 88
    
def w:
    """ Hello world """
mak@hp-elitebook:~/dev/test$ cat .coafile
bears = DocumentationStyleBear
files = test.py
docstyle = default
language = python
default_actions = DocumentationStyleBear: ApplyPatchAction
bear_dirs = ../coala-bears/bears
mak@hp-elitebook:~/dev/test$ coala -L DEBUG
[DEBUG][18:36:19] Platform Linux -- Python 3.4.3, pip 8.1.2, coalib 0.8.0.dev99999999999999
[DEBUG][18:36:19] Since the configuration settings have changed since the last run, the cache will be flushed and rebuilt.
[DEBUG][18:36:19] The file cache was successfully flushed.
Executing section Default...
[DEBUG][18:36:19] coala is run only on changed files, bears' log messages from previous runs may not appear. You may use the `--flush-cache` flag to see them.
[DEBUG][18:36:19] Files that will be checked:
/home/mak/dev/test/test.py
[DEBUG][18:36:19] Running bear DocumentationStyleBear...
[WARNING][18:36:19] Bear DocumentationStyleBear failed to run. Take a look at debug messages (`-L DEBUG`) for further information.
[DEBUG][18:36:19] The bear DocumentationStyleBear raised an exception. If you are the writer of this bear, please make sure to catch all exceptions. If not and this error annoys you, you might want to get in contact with the writer of this bear.

Traceback information is provided below:

Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/coala-0.8.0.dev99999999999999-py3.4.egg/coalib/bears/Bear.py", line 250, in execute
    return [] if result is None else list(result)
  File "/home/mak/dev/coala-bears/bears/documentation/DocumentationStyleBear.py", line 45, in run
    if not stripped_desc[0] == ''
IndexError: list index out of range
mak@hp-elitebook:~/dev/test$ 
```

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,79 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/7sfk3i9oxs1ixg2ncsu3pym0u'
+    CAN_DETECT = {'Documentation'}
+
+    def run(self, filename, file, docstyle: str, language: str):
+        """
+        Checks for certain in-code documentation styles.
+
+        It currently checks for the following style:
```
AbdealiJK: This is generally a nicer way to write this:

```
It currently checks for the following style::

    hello
    ...
```

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,79 @@
...
+        for doc_comment in extract_documentation(file, language, docstyle):
+            metadata = iter(doc_comment.parse())
+
+            # Assuming that the first element is always the only description.
+            new_metadata = [next(metadata)]
+
+            for m in metadata:
+                stripped_desc = list(map(str.lstrip, m.desc.splitlines(True)))
+                if stripped_desc:
+                    stripped_desc[0] = (('\n' + ' ' * 4 + stripped_desc[0])
```
AbdealiJK: Why not make the value 4 into a config? And also allow \t (using a config) for tabs if possible ?
Can be done after merge ...

SanketDG: opened https://github.com/coala-analyzer/coala-bears/issues/705 :+1:

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,79 @@
...
+                                        if not stripped_desc[0] == ''
+                                        else stripped_desc[0])
+
+                new_desc = ''.join(('\n' if not line.endswith('\n') else '') +
+                                   (' ' * 4 if line != '' else '') + line
+                                   for line in stripped_desc)
```
Makman2: not `for line in stripped_desc[1:]`

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,79 @@
...
+                # Something changed, let's apply a result.
+                diff = Diff(file)
+
+                # Delete the old comment
+                diff.delete_lines(doc_comment.range.start.line,
+                                  doc_comment.range.end.line)
+
+                # Apply the new comment
+                assembled_string = new_comment.assemble()
+                assembled = assembled_string.splitlines(True)
```
Makman2: I think you can inline `assembled_string` again :)
-->
```
assembled = new_comment.assemble().splitlines(True)
```

bears/documentation/DocumentationStyleBear.py
Makman2: nvm

SanketDG: no I did still have to take in the first element.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Use >= for development versions so that source builds always work
-coala>=0.8.0
+coala>=0.9.0.dev20160822150444
```
Makman2: do you need this update?

SanketDG: nope not really, does it hurt to have it to the latest version though?

Makman2: then fixup your previous update commit to use this version :+1:

SanketDG: I dont understand :open_mouth:

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,84 @@
...
+                stripped_desc = list(map(str.lstrip, m.desc.splitlines(True)))
+                if stripped_desc:
+                    if len(stripped_desc) == 1:
+                        stripped_desc[0] = ('\n' + ' ' * 4 + stripped_desc[0]
+                                            if not stripped_desc[0] == ''
+                                            else stripped_desc[0])
+                    if len(stripped_desc) == 2:
+                        stripped_desc[0] = ('\n' + ' ' * 4 + stripped_desc[0]
+                                            if stripped_desc[0] == '' and not
+                                            stripped_desc[1].startswith("\n")
```
Makman2: `stripped_desc[1]` can't start with `\n` as this character would appear at the end of `stripped_desc[0]`

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
+from coalib.bearlib.languages.documentation.DocumentationComment import (
+    DocumentationComment)
+from coalib.bearlib.languages.documentation.DocstyleDefinition import (
+    DocstyleDefinition)
+from coalib.bearlib.languages.documentation.DocumentationExtraction import (
```
AbdealiJK: Doing `import coalib.bearlib.languages.documentation as coalib_documentation` and using that from henceforth maybe a good idea to avoid verbosity.
Personal preference though, so not very important

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/7sfk3i9oxs1ixg2ncsu3pym0u'
+    CAN_DETECT = {'Documentation'}
+
+    def run(self, filename, file, docstyle: str, language: str):
+        """
+        Checks for certain in-code documentation styles.
+
+        It currently checks for the following style:::
```
AbdealiJK: I think this should be ending with 2 colons, not 3 ?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+
+            hello
+            :param x:
+                4 space indent
+            :return:
+                also 4 space indent
+                following lines are also 4 space indented
+
+        :param docstyle: The docstyle to use.
```
AbdealiJK: I have no idea what possible docstyles are there ? We should mention them, or atleast tell how to find them

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+        :param language: The language of the file(s).
```
AbdealiJK: Programming language I assume ? Could be confused with normal spoken languages...

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+            metadata = iter(doc_comment.parse())
+
+            # Assuming that the first element is always the only description.
+            new_metadata = [next(metadata)]
+            for m in metadata:
+                stripped_desc = list(map(str.lstrip, m.desc.splitlines(True)))
+                if stripped_desc:
+                    if len(stripped_desc) == 1:
```
AbdealiJK: Consider saving `len_stripped_desc = len(stripped_desc)`
You're looping over the list every time you do `len()`

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+        :param language: The language of the file(s).
+        """
+        for doc_comment in extract_documentation(file, language, docstyle):
+            metadata = iter(doc_comment.parse())
+
+            # Assuming that the first element is always the only description.
+            new_metadata = [next(metadata)]
+            for m in metadata:
+                stripped_desc = list(map(str.lstrip, m.desc.splitlines(True)))
+                if stripped_desc:
```
AbdealiJK: This if condition can be removed if you use `if len==1 ... elif len == 2 ... elif len != 0`
Reduces an indentation level overall

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+                    if len(stripped_desc) == 1:
+                        stripped_desc[0] = ('\n' + ' ' * 4 + stripped_desc[0]
+                                            if not stripped_desc[0] == ''
+                                            else stripped_desc[0])
+                    if len(stripped_desc) == 2:
```
AbdealiJK: elif here ?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+                        stripped_desc[0] = ('\n' + ' ' * 4 + stripped_desc[0]
+                                            if stripped_desc[0] == '' and not
+                                            stripped_desc[1].startswith("\n")
+                                            else stripped_desc[0])
+                        new_desc = stripped_desc[0] + stripped_desc[1]
+                    else:
+                        new_desc = ''.join(('\n' if line == '' else '') +
+                                           (' ' * 4 if line != '' else '') +
```
AbdealiJK: Er, isn't this the same as `'\n' if line == '' else ' ' * 4` but with 2 if conditions which are being added ?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,87 @@
...
+
+                # Apply the new comment
+                assembled = new_comment.assemble().splitlines(True)
+
+                end_line = new_comment.range.end.line
+                end_marker_pos = file[end_line - 1].find(
+                    new_comment.marker[2]) + len(new_comment.marker[2])
+                diff.add_lines(new_comment.range.start.line, assembled[:-1])
+                diff.add_lines(end_line, [assembled[-1] + file[end_line - 1]
+                                          [end_marker_pos:]])
```
AbdealiJK: The splice is split into another line. Kinda weird
```
                diff.add_lines(
                    end_line,
                    [assembled[-1] + file[end_line - 1][end_marker_pos:]])
``` ?

File: tests/documentation/DocumentationStyleBearTest.py [hunk 37]
```diff
@@ -0,0 +1,25 @@
...
+def load_testfile(test, filename):
+    filepath = os.path.join(os.path.dirname(__file__), "test_files",
+                            "DocumentationStyleBear", filename)
+    with open(filepath) as fl:
+        return fl.read()
+
+good_file = load_testfile("DocumentationStyleBear", "good_file.py")
+bad_file = load_testfile("DocumentationStyleBear", "bad_file.py")
+
+file_sameline = load_testfile("DocumentationStyleBear", "bad_file2.py")
```
AbdealiJK: Why declare them in different files ? Why not put them in strings here ?
Adding strings here makes it easier to find and read rather than searching for test_files and switching between editor tabs, etc.

AbdealiJK:
Marking as WIP. I've mentioned a few tweaks which would make things simpler but havent checked the actual functionality. Also, CI fails. Also, @Makman2 's [comment](https://github.com/coala-analyzer/coala-bears/pull/699#discussion_r76517781) hasn't been addressed.

bears/documentation/DocumentationStyleBear.py
SanketDG: Hmm, so for this to work I have to also modify the `__init__` file to import stuff beforehand, would that be okay?

bears/documentation/DocumentationStyleBear.py
SanketDG: This is for the extra colon,

so it appears like

```
    It currently checks for the following style:
```

I will add a extra space I think for readability

bears/documentation/DocumentationStyleBear.py
SanketDG: :+1: this will be done.

bears/documentation/DocumentationStyleBear.py
SanketDG: > Consider saving len_stripped_desc = len(stripped_desc)

:+1:

> You're looping over the list every time you do len()

Not sure what you mean by this, but getting the length of a list is constant time in python.

bears/documentation/DocumentationStyleBear.py
SanketDG: No cant do that, the first if is independent from the second if. Will insert whitespace in between,

tests/documentation/DocumentationStyleBearTest.py
SanketDG: I think using files here makes sense for the sake of clean code. Because I am going to be testing a lot of files of different languages in the future.. 

CC @Makman2, what do you think?

bears/documentation/DocumentationStyleBear.py
Makman2: but you don't modify the list itself, so the length should be preserved^^ then you can make an elif out of it^^

tests/documentation/DocumentationStyleBearTest.py
Makman2: imo files are totally okay, doesn't clutter so much when having many files. And when using an IDE, switching to another file is a piece of cake :)

bears/documentation/DocumentationStyleBear.py
Makman2: you don't need to modify the `__init__` file for that :O
just put @AbdealiJK 's line of code at the top of the file^^

SanketDG: @Makman2 that doesn't work. Can't see what I am doing wrong, Here's the diff

```diff
-from coalib.bearlib.languages.documentation.DocumentationComment import (
+import coalib.bearlib.languages.documentation as coalib_documentation
+from coalib_documentation.DocumentationComment import (
     DocumentationComment)
-from coalib.bearlib.languages.documentation.DocstyleDefinition import (
+from coalib_documentation.DocstyleDefinition import (
     DocstyleDefinition)
-from coalib.bearlib.languages.documentation.DocumentationExtraction import (
+from coalib_documentation.DocumentationExtraction import (
     extract_documentation)
```

Makman2: oh indeed it does not work :O
in fact you need to assign the imports yourself:
```python
import coalib.bearlib.languages.documentation as coaliub_documentation
DocumentationComment = coalib_documentation.DocumentationComment
# ...
```
But I'm not sure if it's that nicely, I would prefer normal imports^^

sils1297:
please rebase

Makman2:
yes I want this merged :D

sils:
@Makman2 this is pending review since more than two weeks now and assigned to you can you please pick it up?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,89 @@
...
+
+                if stripped_desc_length == 2:
+                    stripped_desc[0] = ('\n' + ' ' * 4 + stripped_desc[0]
+                                        if stripped_desc[0] == '' or
+                                        stripped_desc[1].endswith("\n")
+                                        else stripped_desc[0])
+                    new_desc = stripped_desc[0] + stripped_desc[1]
+                elif stripped_desc_length != 0:
+                    new_desc = ''.join(('\n' if line == '' else ' ' * 4) +
+                                       line for line in stripped_desc)
```
Makman2: what about doing maybe this?
```
stripped_desc = map(str.lstrip, m.desc.splitlines())
stripped_desc = ('' if line == '' else ' ' * 4 + line for line in stripped_desc)

new_desc = '\n'.join(stripped_desc)

if m.desc[-1] == '\n':
    new_desc[-1] += '\n'
```
then you can remove this total weird if construct :)

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,89 @@
...
+                # Delete the old comment
+                diff.delete_lines(doc_comment.range.start.line,
+                                  doc_comment.range.end.line)
+
+                # Apply the new comment
+                assembled = new_comment.assemble().splitlines(True)
+
+                end_line = new_comment.range.end.line
+                end_marker_pos = file[end_line - 1].find(
+                    new_comment.marker[2]) + len(new_comment.marker[2])
```
Makman2: can't you use `new_comment.range.end.column - 1` instead of this?

Makman2:
@SanketDG I'm taking over your branch now :)

bears/documentation/DocumentationStyleBear.py
AbdealiJK: Interesting :0 yeah normal imports better

bears/documentation/DocumentationStyleBear.py
AbdealiJK: Sure

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,95 @@
...
+        Checks for certain in-code documentation styles.
+
+        It currently checks for the following style: ::
+
+            hello
+            :param x:
+                4 space indent
+            :return:
+                also 4 space indent
+                following lines are also 4 space indented
```
sils: Would the bear let this pass? Is it smart enough to recognize cpde blocks?

There should be a newline before the first `:param` btw. maybe it could add that as well

Makman2: tests currently pass for this docstring, it doesn't care for the main description. But I can implement that :+1:

bears/documentation/DocumentationStyleBear.py
sils: can't you put a few examples here?

tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
```diff
@@ -0,0 +1,12 @@
+def myfunc(a, b, x):
+    """
+
```
jayvdb: why is this correct?

bears/documentation/DocumentationStyleBear.py
```diff
@@ -0,0 +1,105 @@
...
+    ASCIINEMA_URL = 'https://asciinema.org/a/7sfk3i9oxs1ixg2ncsu3pym0u'
+    CAN_DETECT = {'Documentation'}
+
+    def run(self, filename, file, language: str, docstyle: str='default'):
+        """
+        Checks for certain in-code documentation styles.
+
+        It currently checks for the following style: ::
+
+            The first line needs to have no indentation.
```
jayvdb: I believe this should have a blank line after the first paragraph.

tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test.correct
```diff
@@ -0,0 +1,12 @@
+def hello_there(x):
+    """
```
jayvdb: This is technically the first line.  So where the bear docs says "The first line needs to have no indentation." , it is not happening here.

This is one of the biggest doc style disagreements in python world : should the first text of the multiline docstring appear on the first line, immediately after the `"""`  or on the second line.

`pydocstyle` of course supports both modes.

Makman2: alright I can rephrase a bit clearer :)

But currently I won't support both modes, this bear is currently more or less a demonstration of the inline-documentation capabilities of coala, and as this PR is here for a quite long time I want this merged, such functionality can be added later.

tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
Makman2: Text must begin in the 1st line (technically 2nd line), but it can be empty.

I would file an issue and add new functionality after merge

bears/documentation/DocumentationStyleBear.py
Makman2: The next sentence counts "virtually" to the first paragraph. This is just an example.

Makman2:
ping @jayvdb

Makman2:
Todo list:
- Use annotationbear to get comments, and extract documentation from those (for this purpose create another bear that serves as a dependency)
- Use textwrap?
- `:raises:` statements are not recognized
- for param annotations: Only check the first line for correct indentation, not every following line
- whitespaces before first markers are **required**
- Exctraction with comments having this style:
  ```
  """
 xxx
  """
  ```
  The descriptions shall be aligned correctly, and not cut-off

tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
```diff
@@ -0,0 +1,12 @@
...
+    :param a:
+        slope
+    :param b:
+        offset
+    :param x:
+    :return:
+    """
+
```
jayvdb: Note that this violates [PEP 257](https://github.com/python/peps/blob/master/pep-0257.txt).  "There's no blank line either before or after the docstring."

But as this is not about the contents of the docstring, but is about how the docstring is placed wrt other nodes in the ast, IMO this is definitely beyond scope of the first merge.

tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
Makman2: No haven't seen it either that someone leaves an empty line^^
So yes can be easily added :+1:

tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
Makman2: definitely.

just for idea: it's possible to use the range information of extracted comments to check for this^^
so we get the range and look into the next line

jayvdb:
This working is continuing by @damngamerz on https://github.com/coala/coala-bears/pull/1506

Makman2:
Cool, I think we can close then. @SanketDG let's continue on that PR ;)

Makman2:
(Don't delete the branch yet)

tests/documentation/test_files/DocumentationStyleBear/bad_file4.py.test.correct
damngamerz: @jayvdb Oh it's the other way around :O . What I and Sanket discussed while before the GSoC started that we should have a "blank line before and after" which I guess you wanted to portray as voilation. Tell me if im wrong? Though I will just take a quick look at PEP - 257.

PR #710, opened 2016-08-25
generate_package: Generate ``requirements.txt``

Adrianzatreanu:
Generates ``requirements.txt`` for those bears
that have ``PipRequirements``.

### DIFF
--- a/bears/generate_package.py
+++ b/bears/generate_package.py
@@ -8,6 +8,7 @@ import sys
 import time
 
 from bears import VERSION
+from coalib.bears.requirements.PipRequirement import PipRequirement
 from coalib.collecting.Importers import iimport_objects
 from coalib.parsing.Globbing import glob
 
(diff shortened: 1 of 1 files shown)


bears/generate_package.py
```diff
@@ -120,6 +121,17 @@ def main():
             bear_name, _ = os.path.splitext(os.path.basename(bear_file_name))
             create_file_structure_for_packages(
                 os.path.join('bears', 'upload'), bear_file_name, bear_name)
+            if bear_object.REQUIREMENTS:
+                for requirement in bear_object.REQUIREMENTS:
+                    if isinstance(requirement, PipRequirement):
+                        with open(os.path.join(
+                                    'bears', 'upload',
+                                    bear_name, 'requirements.txt'),
+                                  'a') as reqtxt:
```
sims1253: Can you explain why you use append instead of write?

Asnelchristian:
ack https://github.com/coala-analyzer/coala-bears/pull/710/commits/d046da9267f7faafcc85a87f60bc5b8028693315

bears/generate_package.py
Adrianzatreanu: because else if you have 3 requirements you would overwrite them for every requirement and end up only with the last one?

Adrianzatreanu: requirements.txt need to have all the requirements lined up, not only the last one :)

Asnelchristian: i think it is possible to have a bear with more than one piprequirement, isn't @Adrianzatreanu :sweat_smile:

sims1253: you could first open the file then do the for requirement part.

Adrianzatreanu: no, because i only want the file if there exist pip requirements. if a bear has npm requirements, i dont want to have an empty requirements.txt

Adrianzatreanu: it is possible, @Asnelchristian , yes, thats why appending.

sims1253: Oh ok. Thought it might save some performance but can't think about a nice way...
You could first check if there are pip requirements then open then iterate again. Depends on cost of iterate vs open but probably doesn't matter :)

Asnelchristian: yeah, i know ... it was meant to be an explanation :)

Adrianzatreanu:
@rultor merge

Issue #738, opened 2016-08-30
YapfBear fails on python 2 code


sils1297:
we either should state that it can't deal with python 2 or make it deal with it correctly. Or at least catch the `yapf.yapflib.verifier.InternalError: Missing parentheses in call to 'print' (<string>, line 2)`...

sils1297:
Ideally we would offer an option python 2/3 and run yapf with python 2 or 3 as an external process respectively as yapf will then do the right thing. (Maybe there is a way without switching interpreters? At least official docs say that this is how you should do it which makes sense outside coala.)

PR #748, opened 2016-08-31
YapfBear: Disable syntax verification

underyx:
Syntax verification is supposed to be a private feature in yapf, used for debugging, but they're still defaulting its flag to True on the public FormatFile function. We need to set it to False to disable syntax error checking in yapf (which seems to be an unmaintained feature of it, since it raises exceptions on async/await and py2 print syntax), see discussion at https://github.com/google/yapf/issues/293

Fixes https://github.com/coala-analyzer/coala-bears/issues/738

Also filed an upstream issue here https://github.com/google/yapf/issues/297

### DIFF
--- a/bears/python/YapfBear.py
+++ b/bears/python/YapfBear.py
@@ -142,7 +142,8 @@ space_between_ending_comma_and_closing_bracket= \
         with prepare_file(options.splitlines(keepends=True),
                           None) as (file_, fname):
             corrected = FormatFile(filename,
-                                   style_config=fname)[0].splitlines(True)
+                                   style_config=fname,
+                                   verify=False)[0].splitlines(True)
         diffs = Diff.from_string_arrays(file, corrected).split_diff()
         for diff in diffs:
             yield Result(self,
--- a/tests/python/YapfBearTest.py
+++ b/tests/python/YapfBearTest.py
@@ -1,4 +1,6 @@
+import sys
 from queue import Queue
+from unittest.case import skipIf
 
 from bears.python.YapfBear import YapfBear
 from tests.LocalBearTestHelper import LocalBearTestHelper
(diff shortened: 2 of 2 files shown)


bears/python/YapfBear.py
```diff
@@ -142,7 +142,8 @@ def run(self, filename, file,
         with prepare_file(options.splitlines(keepends=True),
                           None) as (file_, fname):
             corrected = FormatFile(filename,
-                                   style_config=fname)[0].splitlines(True)
+                                   style_config=fname,
+                                   verify=False)[0].splitlines(True)
```
AbdealiJK: If this was a bug, and a major one - shouldn't there be a test case to avoid regression ?

underyx: :+1: on it

tests/python/YapfBearTest.py
```diff
@@ -22,6 +24,15 @@ def test_valid(self):
                             ["x = {  'a':37,'b':42,\n", "'c':927}\n", '\n',
                              "y = 'hello ''world'\n"], valid=False)
 
+    def test_valid_python_2(self):
+        self.check_validity(self.uut, ['print 1\n'], valid=True)
+
+    @skipIf(sys.version_info < (3, 5), "async def can't be parsed without 3.5+")
```
AbdealiJK: Also possibly make a test for python 2 syntax please ?
As that was the case mentioned in the issue and that can also be verified for all python versions

AbdealiJK: Ouch. sorry, I didn't see the test above this. nevermind.

tests/python/YapfBearTest.py
```diff
@@ -22,6 +24,15 @@ def test_valid(self):
                             ["x = {  'a':37,'b':42,\n", "'c':927}\n", '\n',
                              "y = 'hello ''world'\n"], valid=False)
 
+    def test_valid_python_2(self):
+        self.check_validity(self.uut, ['print 1\n'], valid=True)
```
AbdealiJK: How come this works ?
Doesnt the documentation say "if you format Python 3 code with YAPF, run YAPF itself under Python 3 (and similarly for Python 2)." ?

underyx: @AbdealiJK I think it's because of this special handling of the old print syntax in python 3:

```
>>> print 1
  File "<stdin>", line 1
    print 1
          ^
SyntaxError: Missing parentheses in call to 'print'
```

If Python 3 is able to recognize that the syntax is old, then the `ast` module **must** be able to parse this syntax. If the `ast` module is able to parse it, I guess it shouldn't fail.

AbdealiJK: Hm, interesting - alright

AbdealiJK:
ack c02c5d8

AbdealiJK:
ack c02c5d8

AbdealiJK:
@rultor merge

Issue #528, opened 2016-06-15
InvalidLinkBear: Don't consider redirects an error


underyx:
I'm seeing this error:

```
setup.py
|   7| ••••url='https://gitlab.skypicker.com/user/project',
|    | [NORMAL] InvalidLinkBear:
|    | This link redirects to https://gitlab.skypicker.com/users/sign_in
```

That's obviously a false positive :(

We should do something to make sure this sort of server behavior is not considered erroneous.

---

The original suggestion in this issue was:

>Maybe we could check if the new URL matches any of `404`, `error`, `not_found`, `/$`, etc.?

sils1297:
we could just check the exit code of the redirected URL, the idea here was that you can replace your URL with the other one for e.g. a HTTPS redirect

underyx:
Oh, I did not realize that was the goal. I think we also wanted to handle websites that do something like a `HTTP 301 sillywebsite.com/asdfsadfasf` pointing to `HTTP 200 sillywebsite.com/404`. If not, checking the target's status code sounds good to me.

PR #760, opened 2016-09-01
InvalidLinkBear: Add follow_redirects setting

sils1297:
A lot of people aren't using this bear because they don't want redirects
to be corrected. We should build a better heuristic until we make that
functionality default and we need that setting anyway.

Closes https://github.com/coala-analyzer/coala-bears/issues/528


underyx:
ack https://github.com/coala-analyzer/coala-bears/pull/760/commits/a768ac618f1ff6befe751639439b39358e3f7afa

sils1297:
@rultor merge

PR #778, opened 2016-09-02
KeywordBear: Remove TODO

netman92:
Coala now ask us if we want to remove TODO notes.

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -34,6 +34,7 @@ max_lines_per_file = 1000
 [TODOS]
 enabled = False
 bears = KeywordBear
+language = python3
 
 # Note that the ci_keywords and cs_keywords are only used here because coala
 # 0.8.x (current stable release) needs them.
--- a/bears/general/KeywordBear.py
+++ b/bears/general/KeywordBear.py
@@ -2,7 +2,50 @@ import re
 
 from coalib.bearlib import deprecate_settings
 from coalib.bears.LocalBear import LocalBear
+from coalib.results.Diff import Diff
 from coalib.results.Result import RESULT_SEVERITY, Result
+from coalib.results.SourceRange import SourceRange
+
+from bears.general.AnnotationBear import AnnotationBear
+
+
+def _get_comments(dependency_results):
+    if not dependency_results:
+        return
+
+    annotation_bear_results = dependency_results.get('AnnotationBear')
+    if (not annotation_bear_results or
+            not isinstance(annotation_bear_results, list)):
+        return
+
+    for result in annotation_bear_results:
+        yield from result.contents.get('comments', [])
+
+
+def generate_diff(comments, file, filename,
+                  line, line_number, pos):
+    diffs = {}
+    todo_source_range = SourceRange.from_values(filename, line_number + 1,
+                                                pos + 1)
+    for comment_sourcerange in comments:
+        if todo_source_range not in comment_sourcerange:
+            continue
+
+        diff = Diff(file)
+        comment_start = comment_sourcerange.start.column
+        line_before_todo_comment = line[:comment_start - 1].strip()
+
+        if (line_before_todo_comment and
+                line_number + 1 == comment_sourcerange.start.line):
+            diff.change_line(
+                line_number + 1,
+                line,
+                line[:comment_start - 1].rstrip() + '\n')
+        else:
+            diff.delete_line(line_number + 1)
+
+        diffs[filename] = diff
+    return diffs
 
 
 class KeywordBear(LocalBear):
@@ -11,9 +54,14 @@ class KeywordBear(LocalBear):
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Documentation'}
+    BEAR_DEPS = {AnnotationBear}
 
     @deprecate_settings(keywords='ci_keywords')
-    def run(self, filename, file, keywords: list):
+    def run(self,
+            filename,
+            file,
+            keywords: list,
+            dependency_results: dict=None):
         '''
         Checks the code files for given keywords.
 
@@ -25,8 +73,17 @@ class KeywordBear(LocalBear):
             '(' + '|'.join(re.escape(key) for key in keywords) + ')',
             re.IGNORECASE)
 
+        comments = _get_comments(dependency_results)
+
         for line_number, line in enumerate(file):
             for keyword in keywords_regex.finditer(line):
+                diffs = generate_diff(
+                    comments,
+                    file,
+                    filename,
+                    line,
+                    line_number,
+                    keyword.start())
                 yield Result.from_values(
                     origin=self,
                     message="The line contains the keyword '{}'."
@@ -36,4 +93,5 @@ class KeywordBear(LocalBear):
                     column=keyword.start() + 1,
                     end_line=line_number + 1,
                     end_column=keyword.end() + 1,
-                    severity=RESULT_SEVERITY.INFO)
+                    severity=RESULT_SEVERITY.INFO,
+                    diffs=diffs)
--- a/bears/general/KeywordBear.py
+++ b/bears/general/KeywordBear.py
@@ -25,7 +25,7 @@ def _get_comments(dependency_results):
 def generate_diff(comments, file, filename,
                   line, line_number, pos):
     diffs = {}
-    todo_source_range = SourceRange.from_values(filename, line_number + 1,
+    todo_source_range = SourceRange.from_values(filename, line_number,
                                                 pos + 1)
     for comment_sourcerange in comments:
         if todo_source_range not in comment_sourcerange:
@@ -36,13 +36,13 @@ def generate_diff(comments, file, filename,
         line_before_todo_comment = line[:comment_start - 1].strip()
 
         if (line_before_todo_comment and
-                line_number + 1 == comment_sourcerange.start.line):
+                line_number == comment_sourcerange.start.line):
             diff.change_line(
-                line_number + 1,
+                line_number,
                 line,
                 line[:comment_start - 1].rstrip() + '\n')
         else:
-            diff.delete_line(line_number + 1)
+            diff.delete_line(line_number)
 
         diffs[filename] = diff
     return diffs
@@ -75,7 +75,7 @@ class KeywordBear(LocalBear):
 
         comments = _get_comments(dependency_results)
 
-        for line_number, line in enumerate(file):
+        for line_number, line in enumerate(file, start=1):
             for keyword in keywords_regex.finditer(line):
                 diffs = generate_diff(
                     comments,
@@ -89,9 +89,9 @@ class KeywordBear(LocalBear):
                     message="The line contains the keyword '{}'."
                             .format(keyword.group()),
                     file=filename,
-                    line=line_number + 1,
+                    line=line_number,
                     column=keyword.start() + 1,
-                    end_line=line_number + 1,
+                    end_line=line_number,
                     end_column=keyword.end() + 1,
                     severity=RESULT_SEVERITY.INFO,
                     diffs=diffs)
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Use >= for development versions so that source builds always work
-coala>=0.8.1
+coala>=0.9.0.dev20160909114951
 munkres3==1.*
 pylint==1.*
 language-check==0.8.*
--- a/tests/general/KeywordBearTest.py
+++ b/tests/general/KeywordBearTest.py
@@ -1,5 +1,13 @@
+from collections import namedtuple
+from queue import Queue
+import unittest
+
 from bears.general.KeywordBear import KeywordBear
-from tests.LocalBearTestHelper import verify_local_bear
+from coalib.results.HiddenResult import HiddenResult
+from coalib.results.SourceRange import SourceRange
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+from tests.LocalBearTestHelper import verify_local_bear, execute_bear
 
 test_file = """
 test line fix me
@@ -14,3 +22,99 @@ KeywordBearTest = verify_local_bear(
     invalid_files=("test line todo",
                    "test line warNING"),
     settings={"keywords": "todo, warning"})
+
+
+class KeywordBearDiffTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section = Section("")
+        self.section.append(Setting('language', 'python3'))
+        self.section.append(Setting('keywords', '\# TODO, \#todo'))
+        self.python_uut = KeywordBear(self.section, Queue())
+
+        self.annotation_bear_result_type = namedtuple('result', ['contents'])
+        self.text = ['# todo 123']
+        self.dep_results = {
+            'AnnotationBear': [
+                self.annotation_bear_result_type({'comments': ()})
+            ]
+        }
+        self.dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', {'comments': ()})}
+
+    def test_todo(self):
+        text = ['a == b']
+
+        with execute_bear(self.python_uut, "F", text,
+                          dependency_results=self.dep_results) as result:
+            self.assertEqual(result, [])
+
+        dep_results = {
+            'AnnotationBear': {}
+        }
+        with execute_bear(self.python_uut, "F", self.text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs, {})
+            self.assertEqual(result[0].affected_code[0].start.line, 1)
+            self.assertEqual(result[0].affected_code[0].start.column, 1)
+
+        dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', {'comments': 123})}
+        with execute_bear(self.python_uut, "F", self.text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs, {})
+            self.assertEqual(result[0].affected_code[0].start.line, 1)
+            self.assertEqual(result[0].affected_code[0].start.column, 1)
+
+        dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', 123)}
+        with execute_bear(self.python_uut, "F", self.text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs, {})
+            self.assertEqual(result[0].affected_code[0].start.line, 1)
+            self.assertEqual(result[0].affected_code[0].start.column, 1)
+
+        text = ['# todo 123\n']
+        comments = (SourceRange.from_values("F", 1, 1),
+                    SourceRange.from_values("F", 1, 40))
+        dep_results = {
+            'AnnotationBear': [
+                self.annotation_bear_result_type({'comments': comments})
+            ]
+        }
+
+        with execute_bear(self.python_uut, "F", text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +0,0 @@\n'
+                             '-# todo 123\n')
+
+        text = ['test = 55 # todo 123\n']
+        comments = (SourceRange.from_values("F", 1, 11),
+                    SourceRange.from_values("F", 1, 23))
+        dep_results = {
+            'AnnotationBear': [
+                self.annotation_bear_result_type({'comments': comments})
+            ]
+        }
+        with execute_bear(self.python_uut, "F", text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-test = 55 # todo 123\n'
+                             '+test = 55\n')
+
+    def test_todo_outside_of_comment(self):
+        section = Section("")
+        section.append(Setting('language', 'python3'))
+        section.append(Setting('keywords', 'todo'))
+        python_uut = KeywordBear(section, Queue())
+
+        text = ['todo = 123\n']
+        with execute_bear(python_uut, "F", text,
+                          dependency_results=self.dep_results) as result:
+            self.assertEquals(result[0].diffs, {})


sils1297:
@netman92 your tests fail though ;)

.coafile
```diff
@@ -34,6 +34,6 @@ max_lines_per_file = 1000
 [TODOS]
 enabled = False
 bears = KeywordBear
-
+language = python3
```
hypothesist: Unrelated change to what this commit does? If so, this goes into its own commit.

bears/general/KeywordBear.py
```diff
@@ -34,20 +55,57 @@ def run(self,
...
+                comment_start = source_range_comment.start.column
+                line_before_todo_comment = line[:comment_start-1].strip()
+                if (line_before_todo_comment and
+                        line_number + 1 == source_range_comment.start.line):
+                    diff.change_line(
+                        line_number + 1,
+                        line,
+                        line.replace(line[comment_start - 1:],
+                                     '').rstrip() + '\n'
+                    )
```
hypothesist: I'm not sure what standard we follow at coala, but from what I've seen, this bracket goes at the end of the last argument (after the `'\n'` in Line 102). Not sure, others can pip in here.

hypothesist:
ack https://github.com/coala-analyzer/coala-bears/pull/778/commits/deb00d52db441a3d1de8c9b7c652a5464460b11e

.coafile
sils1297: it kind of is related since we want the coafile to work with the dev version of coala bears always. See also https://github.com/coala-analyzer/coala/issues/2709 though :/

bears/general/KeywordBear.py
sils1297: IMO doesn't matter much

netman92:
@sils1297 What's the status here?  Should I rebase/change something?

sils1297:
I like your PR, @Makman2 seems to have doubts about the general thing so we'll want to continue a discussion with him. And then there's coala-analyzer/coala#2709 - I wouldn't like to merge your PR without merging this first because it breaks all user's usages of keywordbear if they don't have a language setting in the section.

sils1297:
@Makman2 can you take a look at this? I think this is great. If a patch is provided it will only remove the comment.

Oh actually @netman92 could you add a test where a TODO appears outside a comment? :)

sils1297:
@netman92 so there was some changes to the keywordbear which conflict with yours, it's unlikely to happen .again because now it's in a way better shape. Can you make a rebase? Really sorry :/

netman92:
@sils1297 Done.

sils1297:
:+1: I'll talk with our devs maybe we can fix that stupid bug and make this happen. In the meantime you can at least already use your branch development version for your purposes.

sils1297:
I'm fixing the issue now so we can merge this :)

sils1297:
see https://github.com/coala-analyzer/coala/pull/2717 when that is merged you can rebase, update coala to the then newly released thing

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Use >= for development versions so that source builds always work
-coala>=0.8.0
+coala>=0.9.0.dev20160902132949
```
sils1297: please upgrade this to the latest which will fix the bug that you hit and it will correctly ask you about missing settings - also if you could try it out that'd be nice as well :)

sils1297:
https://github.com/coala-analyzer/coala-bears/pull/778/commits/55e431ad1d3c08ee68bf48800d5945697637e75d needs work

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+from coalib.results.Diff import Diff
 from coalib.results.Result import RESULT_SEVERITY, Result
 
+from bears.general.AnnotationBear import AnnotationBear
+from coalib.results.SourceRange import SourceRange
+
+
+def _get_comments(dependency_results):
+    if not dependency_results:
+        return []
```
sils1297: can you return a set in all cases?

sils1297: also adding a newline after the return would increase readability very much for me

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+    annotation_bear_results = dependency_results.get('AnnotationBear')
+    if (not annotation_bear_results or
+            not isinstance(annotation_bear_results, list)):
+        return []
+    comments = []
+    for result in annotation_bear_results:
```
sils1297: ah so this is a workaround for the AnnotationBear not necessarily returning only one result? Can you fix that in the AnnotationBear instead, just adding a return or so would probably be sufficient so we don't need such complicated logic here.

bears/general/KeywordBear.py
```diff
@@ -27,6 +80,14 @@ def run(self, filename, file, keywords: list):
 
         for line_number, line in enumerate(file):
             for keyword in keywords_regex.finditer(line):
+                diffs = get_diff_comment_todo(
+                    dependency_results,
+                    file,
+                    filename,
+                    line,
+                    line_number,
+                    keyword.start()
```
sils1297: so I'd either have the closing parenthesis here or a trailing comma for diffs

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+        return []
+    comments = []
+    for result in annotation_bear_results:
+        if not isinstance(result.contents, dict):
+            continue
+        comments += list(result.contents.get('comments', []))
+    return set(comments)
+
+
+def get_diff_comment_todo(dependency_results, file, filename,
```
sils1297: hm that name doesn't really help me understanding what this does, maybe we can find a better one?

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+                          line, line_number, pos):
+    comments = _get_comments(dependency_results)
+    diffs = None
+    for source_range_comment in comments:
```
sils1297: rename to `comment_sourcerange` would be clearer to me

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+        todo_source_range = SourceRange.from_values(
+            filename, line_number + 1, pos + 1)
```
sils1297: can we do this outside the loop? Also I wouldn't name it todo_source_range because this bear isn't necessarily only for TODOs but for anything basically

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+        diff = Diff(file)
+        # remove just part of the line
+        comment_start = source_range_comment.start.column
+        line_before_todo_comment = line[:comment_start - 1].strip()
+        if (line_before_todo_comment and
+                line_number + 1 == source_range_comment.start.line):
+            diff.change_line(
+                line_number + 1,
+                line,
+                line.replace(line[comment_start - 1:],
```
sils1297: can you rather break after the opening parenthesis? I think that might be better readable

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+    comments = _get_comments(dependency_results)
+    diffs = None
+    for source_range_comment in comments:
+        todo_source_range = SourceRange.from_values(
+            filename, line_number + 1, pos + 1)
+        if todo_source_range not in source_range_comment:
+            continue
+
+        diff = Diff(file)
+        # remove just part of the line
```
sils1297: this comment belongs inside the diff huh?

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,96 @@
     invalid_files=("test line todo",
                    "test line warNING"),
     settings={"keywords": "todo, warning"})
+
+
+class KeywordBearTodoTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section1 = Section("")
```
sils1297: can we rename that to section? I know we use section1, 2, ... in other tests but that's actually bad style becuase the names don't convey meaning so if you remember a place where we did that you could help us a lot by filing an issue that our code is bad there ;)

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,96 @@
...
+        self.section1.append(Setting('language', 'python3'))
+        self.section1.append(Setting('keywords', '\# TODO, \#todo'))
+        self.python_uut = KeywordBear(self.section1, Queue())
+
+        self.annotation_bear_result_type = namedtuple('result', ['contents'])
```
sils1297: why not just import HiddenResult?

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
     CAN_DETECT = {'Documentation'}
+    BEAR_DEPS = {AnnotationBear}
 
     @deprecate_settings(keywords='ci_keywords')
-    def run(self, filename, file, keywords: list):
+    def run(self,
+            filename,
+            file,
+            keywords: list,
+            dependency_results: dict = None):
```
sils1297: we usually don't want spaces around the `=`, PEP8Bear doesn't catch that sadly...

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,96 @@
...
+        section1.append(Setting('language', 'python3'))
+        section1.append(Setting('keywords', 'todo'))
+        python_uut = KeywordBear(section1, Queue())
+
+        text = ['todo = 123\n']
+        dep_results = {
+            'AnnotationBear': [
+                self.annotation_bear_result_type({'comments': ()})
+            ]
+        }
```
sils1297: maybe you can store some dep results in self in the setup method so you can reuse them and don't have to duplicate the code?

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,96 @@
...
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-test = 55 # todo 123\n'
+                             '+test = 55\n')
+
+    def test_todo_outside_of_comment(self):
+        section1 = Section("")
+        section1.append(Setting('language', 'python3'))
+        section1.append(Setting('keywords', 'todo'))
```
sils1297: you could just use `self.section1` here and delete those lines, am I wrong? :confused:

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,96 @@
...
+        python_uut = KeywordBear(section1, Queue())
+
+        text = ['todo = 123\n']
```
sils1297: this seems to be reoccurring as well, can we store it in self and not define it multiple times?

sils1297:
ok @netman92 that's a lot of comments, I hope it isn't too intimidating, in the beginning everyone has to adapt to our style and requirements which usually ends like this including experienced developers.

Tell me if you need any help in refactoring it, it's all very little things that should be easy to change :)

netman92:
@sils1297 Fixed

sils1297:
@netman92 your windows builds seem to fail https://ci.appveyor.com/project/coala/coala-bears/build/1.0.2816/job/p0u6wuxtycqv4s80

sils1297:
@netman92 ping us if you want help fixing windows. Maybe @Makman2 has an idea

Makman2:
this looks more like a logic error to me this time:
>```
E           Traceback (most recent call last):
E             File "C:\Python34\lib\site-packages\coalib\bears\Bear.py", line 250, in execute
E               return [] if result is None else list(result)
E             File "C:\projects\coala-bears\bears\general\KeywordBear.py", line 91, in run
E               keyword.start())
E             File "C:\projects\coala-bears\bears\general\KeywordBear.py", line 37, in generate_diff
E               if todo_source_range not in comment_sourcerange:
E           TypeError: argument of type 'SourceRange' is not iterable
```

File: bears/general/KeywordBear.py [hunk 18]
```diff
@@ -2,18 +2,73 @@
...
+def _get_comments(dependency_results):
+    if not dependency_results:
+        return {}
+
+    annotation_bear_results = dependency_results.get('AnnotationBear')
+    if (not annotation_bear_results or
+            not isinstance(annotation_bear_results, list)):
+        return {}
+
+    comments = []
```
Makman2: why not making this a set directly?
`comments = set()`

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,73 @@
...
+    for result in annotation_bear_results:
+        if not isinstance(result.contents, dict):
+            continue
+        comments += list(result.contents.get('comments', []))
```
Makman2: you could do (see also above):
```
try:
    comments.add(result.contents.get('comments'))
except:
    pass
```

Makman2: then you can also just return `comments` itself below^^

Makman2:
hm nvm this is weird...

Makman2:
forgot that I need to modify appveyor.yml for that...

Makman2:
huh now it's also on circle :O
I also can't reproduce on the RDP session on appveyor, KeywordBear passes ...
I don't know what's the cause currently :/

sils1297:
whoops? Somethings wrong with this: https://github.com/coala-analyzer/coala-bears/pull/778/commits/88842e027c06897f3efb76bd30b4ef2cc09077d9 needs work

sils1297:
waiting for you to fix your dependency commit, builds will probably fail without it

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,73 @@
...
+
+    annotation_bear_results = dependency_results.get('AnnotationBear')
+    if (not annotation_bear_results or
+            not isinstance(annotation_bear_results, list)):
+        return {}
+
+    comments = set()
+    for result in annotation_bear_results:
+        if not isinstance(result.contents, dict):
+            continue
```
sils1297: uncovered during tests

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/778/commits/8f95b56c34ba9efdb01819c40fd4eab47b0cae33

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,71 @@
...
+
+    comments = set()
+    for result in annotation_bear_results:
+        comments = comments.union(result.contents.get('comments', []))
+    return comments
+
+
+def generate_diff(dependency_results, file, filename,
+                  line, line_number, pos):
+    comments = _get_comments(dependency_results)
```
sils1297: you're doing this every single loop, might be better to do it once?

sils1297: you can also spot those performance thingies because a function that gennerates a diff shouldn't be responsible for getting te information it needs for it as well

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,67 @@
 
 from coalib.bearlib import deprecate_settings
 from coalib.bears.LocalBear import LocalBear
+from coalib.results.Diff import Diff
 from coalib.results.Result import RESULT_SEVERITY, Result
 
+from bears.general.AnnotationBear import AnnotationBear
+from coalib.results.SourceRange import SourceRange
```
Makman2: please sort imports :)
-->
```
from coalib.bearlib import deprecate_settings
from coalib.results.Diff import Diff
from coalib.results.Result import RESULT_SEVERITY, Result
from coalib.results.SourceRange import SourceRange # <-- belongs here :3

from bears.general.AnnotationBear import AnnotationBear
```

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,67 @@
...
+    if not dependency_results:
+        return {}
+
+    annotation_bear_results = dependency_results.get('AnnotationBear')
+    if (not annotation_bear_results or
+            not isinstance(annotation_bear_results, list)):
+        return {}
+
+    for result in annotation_bear_results:
+        yield from result.contents.get('comments', [])
```
Makman2: oh `_get_comments` is an iterator as you use `yield`. The above returned values have no effect, as these are not returned :)

So either return nothing above and keep `_get_comments` as a generator function (btw you return dicts above, I think you want sets, in this case `{}` --> `set()`), or change the last line to
```
return (result.contents.get('comments', []) for result in annotation_bear_results)
```

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,67 @@
...
+    todo_source_range = SourceRange.from_values(filename, line_number + 1,
+                                                pos + 1)
+    for comment_sourcerange in comments:
+        if todo_source_range not in comment_sourcerange:
+            continue
+
+        diff = Diff(file)
+        comment_start = comment_sourcerange.start.column
+        line_before_todo_comment = line[:comment_start - 1].strip()
+        # remove just part of the line
```
Makman2: redundant comment

Makman2: redundant comment

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,67 @@
...
+                line_number + 1 == comment_sourcerange.start.line):
+            diff.change_line(
+                line_number + 1,
+                line,
+                line.replace(
+                    line[comment_start - 1:], '').rstrip() + '\n')
+        else:
+            diff.delete_line(line_number + 1)
+
+        diffs = {filename: diff}
```
Makman2: what if `comment_sourcerange` contains more than one item? then you override the last calculated diff. Imo you want to initialize above `diff = {}` (instead of setting to None) and then
```
diffs[filename] = diff
```

bears/general/KeywordBear.py
```diff
@@ -36,4 +94,5 @@ def run(self, filename, file, keywords: list):
                     column=keyword.start() + 1,
                     end_line=line_number + 1,
                     end_column=keyword.end() + 1,
-                    severity=RESULT_SEVERITY.INFO)
+                    severity=RESULT_SEVERITY.INFO,
+                    diffs=diffs)
```
Makman2: as everywhere `line_number + 1` is used, you maybe can modify `enumerate` above:
```python
for line_number, line in enumerate(file, start=1)
```
then you can replace every occurrence of `line_number + 1` with `line_number`.

Maybe you can make a new commit before this one which changes the `enumerate` for the old version :+1:

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +23,99 @@
     invalid_files=("test line todo",
                    "test line warNING"),
     settings={"keywords": "todo, warning"})
+
+
+class KeywordBearTodoTest(unittest.TestCase):
```
Makman2: hm I would call this one `KeywordBearDiffTest` or `KeywordBearPatchTest`, as "Todo" is just a keyword itself^^

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,67 @@
...
+        comment_start = comment_sourcerange.start.column
+        line_before_todo_comment = line[:comment_start - 1].strip()
+        # remove just part of the line
+        if (line_before_todo_comment and
+                line_number + 1 == comment_sourcerange.start.line):
+            diff.change_line(
+                line_number + 1,
+                line,
+                line.replace(
+                    line[comment_start - 1:], '').rstrip() + '\n')
```
Makman2: please do `line[:comment_start - 1].rstrip() + '\n'`

bears/general/KeywordBear.py
```diff
@@ -2,18 +2,67 @@
...
+    diffs = None
+    todo_source_range = SourceRange.from_values(filename, line_number + 1,
+                                                pos + 1)
+    for comment_sourcerange in comments:
+        if todo_source_range not in comment_sourcerange:
+            continue
+
+        diff = Diff(file)
+        comment_start = comment_sourcerange.start.column
+        line_before_todo_comment = line[:comment_start - 1].strip()
```
Makman2: not `line[:comment_start]`?

Makman2: not `line[:comment_start]`?

tests/general/KeywordBearTest.py
```diff
@@ -1,5 +1,14 @@
+import unittest
+from collections import namedtuple
+from queue import Queue
```
Makman2: import order:

```python
from collections import namedtuple
from queue import Queue
import unittest
```

bears/general/KeywordBear.py
netman92: Nope, bcs. of indexing

sils1297:
@Makman2 you're assigned, can you please look into it and help him figure out whats wrong with the tests?

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,99 @@
...
                    "test line warNING"),
     settings={"keywords": "todo, warning"})
+
+
+class KeywordBearDiffTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section = Section("")
+        self.section.append(Setting('language', 'python3'))
+        self.section.append(Setting('keywords', '\# TODO, \#todo'))
```
Makman2: if you want to escape `#` so it gets not parsed from `Setting` as a comment, you need to escape it twice, as python already does it once :)
-->
```
'\\# TODO, \\#todo'
```

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,99 @@
...
+        self.python_uut = KeywordBear(self.section, Queue())
```
Makman2: why not just `self.uut`? :O

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,99 @@
...
+                self.annotation_bear_result_type({'comments': ()})
+            ]
+        }
+        self.dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', {'comments': ()})}
+
+    def test_todo(self):
+        text = ['a == b']
+
+        with execute_bear(self.python_uut, "F", text,
```
Makman2: `F` is the filename, right? could you just place `"filename"` or `"Filename"` in there, makes it better understandable on first sight than `F` :)

Makman2: Same below :+1:

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,99 @@
...
+class KeywordBearDiffTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section = Section("")
+        self.section.append(Setting('language', 'python3'))
+        self.section.append(Setting('keywords', '\# TODO, \#todo'))
+        self.python_uut = KeywordBear(self.section, Queue())
+
+        self.annotation_bear_result_type = namedtuple('result', ['contents'])
+        self.text = ['# todo 123']
```
Makman2: as you reset the test-`text` everwhere except in two cases, I would move this out of `setUp()` and just do below
```python
text = ['# todo 123']
# tests...
```

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,99 @@
...
+            self.assertEqual(result[0].affected_code[0].start.line, 1)
+            self.assertEqual(result[0].affected_code[0].start.column, 1)
+
+        dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', 123)}
+        with execute_bear(self.python_uut, "F", self.text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs, {})
+            self.assertEqual(result[0].affected_code[0].start.line, 1)
+            self.assertEqual(result[0].affected_code[0].start.column, 1)
```
Makman2: you should maybe also assert for length of `result`
```
self.assertEqual(len(result), 1)
```
same everywhere :)

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,99 @@
...
+
+        self.annotation_bear_result_type = namedtuple('result', ['contents'])
+        self.text = ['# todo 123']
+        self.dep_results = {
+            'AnnotationBear': [
+                self.annotation_bear_result_type({'comments': ()})
+            ]
+        }
+        self.dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', {'comments': ()})}
```
Makman2: why do you override `dep_results` right again? :O

bears/general/KeywordBear.py
```diff
@@ -25,7 +25,7 @@ def _get_comments(dependency_results):
 def generate_diff(comments, file, filename,
```
Makman2: Commit message:
-->
```
KeywordBear: Change enumerate
```
Uppercase letters after tags :)

Adrianzatreanu:
closed due to huge inactivity.

tests/general/KeywordBearTest.py
netman92: Bcs another test will fail.

netman92:
@Adrianzatreanu can you reopen it? I've already fix it and pushed.

Issue #735, opened 2016-08-30
CPDBear: Increase minimum tokens default to 30


sils1297:
I think we should rather have a lower false positive rate by default even.

difficulty/newcomer

Makman2:
from @ku3o
> After a discussion, we decided not to change minimum-tokens value as a fix of the bug should increase report readability and communicate potential false positive cases better.

Closing this.

Makman2:
See https://github.com/coala-analyzer/coala-bears/pull/805

PR #805, opened 2016-09-09
CPDBear: Increase minimum token to 30

ku3o:
This should lower false positive rate in duplicity detection.

Closes https://github.com/coala-analyzer/coala-bears/issues/735

### DIFF
--- a/bears/xml2/XMLBear.py
+++ b/bears/xml2/XMLBear.py
@@ -26,6 +26,8 @@ def path_or_url(xml_dtd):
 class XMLBear:
     """
     Checks the code with ``xmllint``.
+
+    See http://xmlsoft.org/xmllint.html
     """
     LANGUAGES = {"XML"}
     REQUIREMENTS = {DistributionRequirement(apt_get='libxml2')}


sils1297:
Hi,

your commit message isn't fully compliant to our standards yet. Don't worry, most people don't know about good commit messages when they join us, including very experienced developers :) Standards vary from project to project.

The most important part is the first line of your commit message - maintainers will use it to read through a large number of commits and thus it's important that it's short but most importantly concise. I'd heartily recommend reading <http://coala.io/commit>!

- Please change the `Fixes` to a `Closes` for our bug prediction.
- Remove the period in the shortlog.

ku3o:
Thanks for the comment. The commit message was amended to comply with the standards.

sils1297:
ack https://github.com/coala-analyzer/coala-bears/pull/805/commits/eb8ea2ae8c95aea201dac4463e777dcd7c0cba36 FYI you don't have to change the PR, we don't care a lot ;)

Makman2:
looks like our bad test file slips through analysis now due to the new threshold :3

@sils1297 I consider the failing file also code-clone, so what do we do now?

Makman2:
@ku3o wait wait, don't close so fast :D
we just need to discuss whether we change the test slightly or not :)

ku3o:
:)
It was closed unintentionally (I guess), because I reset the commit. It's no longer needed, at least for now. @sils1297 found a bug inside the `CPDBear` which was causing incorrect printing of duplicate lines.

After a discussion, we decided not to change `minimum-tokens` value as a fix of the bug should increase report readability and communicate potential false positive cases better.

Makman2:
ah alright :)
closed the issue about raising the `minimum-tokens` threshold :+1:

PR #842, opened 2016-09-22
.ci Add shellcheck tests for MacOSX

sepihere:
The .ci folder did not include shellshock test for MacOSX users.
Script checks shellcheck's existence in brew's directory and installs
shellcheck via Homebrew if it is outdated or not installed. Script is
.ci/deps.shellcheck.sh's version for MacOSX.

Fixes https://github.com/coala/coala-bears/issues/153

### DIFF
--- a/.ci/deps.osx.sh
+++ b/.ci/deps.osx.sh
@@ -1,0 +1,22 @@
+VERSION=0.4.4
+BIN_PATH=/usr/local/Cellar/shellcheck/
+
+install_shellcheck() {
+  brew update && brew install shellcheck
+}
+
+currently_installed_shellcheck_version() {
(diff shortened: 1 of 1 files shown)


.ci/deps.osx.sh
```diff
@@ -0,0 +1,22 @@
+VERSION=0.4.4
```
Adrianzatreanu: regarding commit message. There should be a ``:`` between .ci and Add

Adrianzatreanu: also in body, you wrote shellshock. shouldnt that be shellcheck?

sepihere: This word got stuck in my mind from the moment I forked. sorry :)

PR #846, opened 2016-09-28
bears/python: Add Flake8Bear

srisankethu:
Fixes https://github.com/coala/coala-bears/issues/169

### DIFF
--- a/bears/js/JSONFormatBear.py
+++ b/bears/js/JSONFormatBear.py
@@ -1,13 +1,13 @@
 import json
 from collections import OrderedDict
 
+from coala_utils.param_convertion import negate
 from coalib.bearlib import deprecate_settings
 from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
 from coalib.bears.LocalBear import LocalBear
 from coalib.misc.Compatibility import JSONDecodeError
 from coalib.results.Diff import Diff
 from coalib.results.Result import Result
-from coala_utils.param_convertion import negate
 
 
 class JSONFormatBear(LocalBear):


bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,23 @@
...
+                     r'(?P<message>.*)',
+        severity_map={
+            "F": RESULT_SEVERITY.MAJOR,
+            "E": RESULT_SEVERITY.MAJOR,
+            "W": RESULT_SEVERITY.NORMAL,
+            "N8": RESULT_SEVERITY.INFO,
+            "C9": RESULT_SEVERITY.INFO})
+class Flake8Bear:
+    """
+    Checks for styling errors based on pep8, pyflakes, pycodestyle and mccabe
```
Adrianzatreanu: hey, first of all, this needs a period.

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+
```
Adrianzatreanu: could you please fill in the metadata attributes, such as author, requirements, etc? You can check any other bear to see what you have to write :)

bears/python/Flake8Bear.py
```diff
@@ -0,0 +1,23 @@
...
+    @staticmethod
+    def create_arguments(filename, file, config_file):
```
Adrianzatreanu: does this bear have absolutely no arguments?

Adrianzatreanu:
also, shouldnt you add the package to requirements.txt?

bears/python/Flake8Bear.py
srisankethu: It does! http://flake8.pycqa.org/en/latest/user/options.html Can you please comment what options would be helpful?

bears/python/Flake8Bear.py
srisankethu: I did not get you

bears/python/Flake8Bear.py
srisankethu: Oh! I will fill them up. I was worrying why the tests are failing and totally forgot about it.

srisankethu:
Will add it

PR #853, opened 2016-10-01
Test

srisankethu:


tests/LocalBearTestHelper.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 
 from tests.BearTestHelper import generate_skip_decorator
-from coalib.bearlib.abstractions.Lint import Lint
```
sils1297: why not completely remove the import?

Issue #829, opened 2016-09-16
Circle CI trusty image does not pass


jayvdb:
My first CI build of master  is very broken

https://circleci.com/gh/jayvdb/coala-bears/1

SanketDG:
So this was like all things failing at once:

* debc4d2 is the first issue, it is definitely not being done in the right way. CC @underyx 

* `CSSAutoPrefixBear` was failing, primarily because `-webkit-flex` is no more in usage, so there was result generation in the good file testing.

* `DockerfileLintBear`'s executable is actually broken for about, um, 15 days. https://github.com/projectatomic/dockerfile_lint/issues/54

The first two have a way out, I have no idea how to fix the third one. Maybe we will have to use a specific version or skip the test until the upstream bug gets fixed.

sils1297:
Sanket can you look into making the fixes?

SanketDG:
Fix is at https://github.com/coala/coala-bears/pull/832. Apparently debc4d2 works, sorry for all the confusion though

jayvdb:
It is still broken for me. https://circleci.com/gh/jayvdb/coala-bears/4

sils1297:
IIUC dockerfile_lint has fixed their issue and the tests should pass. However caching might hinder circleci from getting the new version, can't we bump the version in our package.json to force it to get the new fixed version?

Also the TSLintBear seems to fail. Maybe for similar reasons?

jayvdb:
On Travis it is easy to clear the cache. Can't that be done on Circle?

SanketDG:
Yeah I have already cleared the cache. But there are other breaking changes in other tools. I am looking into it.

SanketDG:
Fixed in 18defb4 94c7897 3ba5bbd.

If your tests are failing in your PR, please rebase.

Feel free to reopen if the tests fail again after rebasing.

jayvdb:
I did rebuild master and it failed. Trying again with the cache cleared.  https://circleci.com/gh/jayvdb/coala-bears/tree/master (builds 9 & 10)

SanketDG:
https://circleci.com/gh/jayvdb/coala-bears/10 The tests pass for `DockerlintFileBear` but other tests time out. Could you rebuild again?

sils1297:
we should change our config to allow larger runtimes if the timeout is the
problem

jayvdb:
Still failing

SanketDG:
[...]

PR #862, opened 2016-10-03
Upgrade libgmp10 to version 6

jayvdb:
When initialising Julia, without a ~/.julia , the following
warning is shown.

```
WARNING: Error during initialization of module GMP:
ErrorException("The dynamically loaded GMP library
(version 5.1.3 with __gmp_bits_per_limb == 64)
does not correspond to the compile time version
(version 6.1.0 with __gmp_bits_per_limb == 64).
Please rebuild Julia.")
```

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -14,7 +14,7 @@ sudo add-apt-repository -y ppa:staticfloat/juliareleases
 sudo add-apt-repository -y ppa:staticfloat/julia-deps
 sudo add-apt-repository -y ppa:avsm/ppa
 sudo apt-get -y update
-deps="espeak libclang1-3.4 indent mono-mcs chktex hlint r-base julia luarocks verilator cppcheck flawfinder"
+deps="espeak libclang1-3.4 indent mono-mcs chktex hlint r-base julia libgmp10 luarocks verilator cppcheck flawfinder"
 deps_python_dbus="libdbus-glib-1-dev libdbus-1-dev"
 deps_python_gi="glib2.0-dev gobject-introspection libgirepository1.0-dev python3-cairo-dev"
 deps_perl="perl libperl-critic-perl"


jayvdb:
See https://github.com/coala/coala-bears/issues/859#issuecomment-251154687
It does not occur on `coala/coala-bears` which is running with a cache.

jayvdb:
And because forks refuse to build (#829), this is red ;-(

.ci/deps.sh
```diff
@@ -14,7 +14,7 @@ sudo add-apt-repository -y ppa:staticfloat/juliareleases
 sudo add-apt-repository -y ppa:staticfloat/julia-deps
```
Adrianzatreanu: is this fixing an issue? if so, please link it with "Fixes issue_url"

also change commit message please, such as: "ci: Upgrade ..."
and body should explain why: "When initialising Julia bla bla, an error was shown. This helps prevent that error from coming back up."

jayvdb:
This is a fix for using the following on Circle CI trusty images.

```
sudo add-apt-repository -y ppa:staticfloat/juliareleases
sudo add-apt-repository -y ppa:staticfloat/julia-deps
```
Bite that my current trusty strategy actually comments those out due to this bug, replacing them with a different source of julia (xenial).  So this does not currently impact coala-bears, but it does prevent us from using one documented approach to installing Julia'.

I suspect this is better solved upstream, but it seems that launchpad doesnt give me a way to report bugs upstream to that ppa, so pinging @staticfloat here.

The problem can be seen here: https://circleci.com/gh/jayvdb/coala-bears/22

What I've found so far is that http://bazaar.launchpad.net/~staticfloat/julianightlies/trunk/view/head:/debian/control specifies `libgmp10 (>= 5.1.3)` , and `ibgmp10 (>= 5.1.3` is already installed, so the `2:6.1.0+dfsg-2~ubuntu15.10.1` from https://launchpad.net/~staticfloat/+archive/ubuntu/julia-deps isnt being installed.

staticfloat:
Ah, you're not doing an `apt-get upgrade`, so it doesn't automatically give you the latest `gmp`.  That's slightly annoying.  I will fix `debian/control` to require a higher version of `gmp`.

jayvdb:
Ya. The setup time is very large, so we only install the bare minimum to get functional versions of the tools (and this warning appears to be a very very strong hint that something could be broken).
Thanks for looking at this.

jayvdb:
Has this been done?
http://bazaar.launchpad.net/%7Estaticfloat/julianightlies/trunk/view/head:/debian/control still show `libgmp10 (>= 5.1.3)`.

staticfloat:
Yes, it has been done.  That branch you're looking at is the old, bzr branch that I used to use before launchpad allowed for `git` repositories on launchpad.  Note that the bzr branch hasn't been updated in over 6 months.  The `git` repository is viewable ([here's the `debian/control` file](https://git.launchpad.net/~staticfloat/julianightlies/tree/debian/control)), but more directly, on the most recently published binaries you can look at [the `.dsc` file](https://launchpadlibrarian.net/291021392/julia_0.6.0-3000~201610270458~ubuntu15.04.1.dsc) to see what packages are required, and `libgmp-dev >= 6.1.0` is shown.

jayvdb:
> .. https://git.launchpad.net/%7Estaticfloat/julianightlies/tree/debian/control ...

That link doesnt appear to work her, but I see it if I manually retype the `~` in the URL.  A bug in launchpad?

Anyway, it is fixed.  Thanks!

PR #877, opened 2016-10-06
InvalidLinkBear: Allow variable timeouts

meetmangukiya:
Allow variable timeouts for servers that are slow in responding, and
hence increase the accuracy of the number of invalid links reported by
coala.
Closes https://github.com/coala/coala-bears/issues/686

---
This change is [Reviewable](https://reviewable.io/reviews/coala/coala-bears/877)

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -1,5 +1,6 @@
 import re
 import requests
+from urllib.parse import urlparse
 
 from difflib import SequenceMatcher
 
@@ -11,6 +12,7 @@ from coalib.results.Result import Result
 from coalib.bearlib import deprecate_settings
 from coalib.settings.Setting import typed_list
 from coalib.parsing.Globbing import fnmatch
+from coalib.settings.Setting import typed_dict
 
 
 class InvalidLinkBear(LocalBear):
@@ -48,7 +50,8 @@ class InvalidLinkBear(LocalBear):
         return splitted_schema
 
     @staticmethod
-    def find_links_in_file(file, timeout, link_ignore_regex, link_ignore_list):
+    def find_links_in_file(file, network_timeout, link_ignore_regex,
+                           link_ignore_list):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
             r"""
@@ -85,12 +88,15 @@ class InvalidLinkBear(LocalBear):
                         fnmatch(link, link_ignore_list)):
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
-                    code = InvalidLinkBear.get_status_code(link, timeout)
+                    host = urlparse(link).netloc
+                    code = InvalidLinkBear.get_status_code(
+                               link, network_timeout.get(
+                                   host, InvalidLinkBear.DEFAULT_TIMEOUT))
                     yield line_number + 1, link, code
 
     @deprecate_settings(link_ignore_regex='ignore_regex')
     def run(self, filename, file,
-            timeout: int=DEFAULT_TIMEOUT,
+            network_timeout: typed_dict(str, int, DEFAULT_TIMEOUT)=dict(),
             link_ignore_regex: str='([.\/]example\.com|\{|\$)',
             link_ignore_list: typed_list(str)='',
             follow_redirects: bool=False):
@@ -108,13 +114,19 @@ class InvalidLinkBear(LocalBear):
         `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
         all your data.
 
-        :param timeout:          Request timeout period.
+        :param network_timeout:       A dict mapping URLs and timeout to be
+                                      used for that URL. All the URLs that have
+                                      the same host as that of URLs provided
+                                      will be passed that timeout.
         :param link_ignore_regex:     A regex for urls to ignore.
         :param link_ignore_list: Comma separated url globs to ignore
         :param follow_redirects: Set to true to autocorrect redirects.
         """
+        network_timeout = {urlparse(url).netloc: timeout
+                           for url, timeout in network_timeout.items()}
+
         for line_number, link, code in InvalidLinkBear.find_links_in_file(
-                file, timeout, link_ignore_regex, link_ignore_list):
+                file, network_timeout, link_ignore_regex, link_ignore_list):
             if code is None:
                 yield Result.from_values(
                     origin=self,
--- a/tests/general/InvalidLinkBearTest.py
+++ b/tests/general/InvalidLinkBearTest.py
@@ -3,6 +3,7 @@ from queue import Queue
 import requests
 import requests_mock
 import unittest
+import unittest.mock
 
 from bears.general.InvalidLinkBear import InvalidLinkBear
 from coalib.settings.Section import Section
@@ -229,3 +230,35 @@ class InvalidLinkBearTest(unittest.TestCase):
 
         self.assertResult(valid_file=valid_file,
                           settings={'link_ignore_list': link_ignore_list})
+
+    def test_variable_timeouts(self):
+        nt = {
+            'https://google.com/timeout/test/2/3/4/5/something': 10,
+            'https://facebook.com/timeout': 15
+        }
+
+        file_contents = """
+        https://facebook.com/
+        https://google.com/
+        https://coala.io/som/thingg/page/123
+        """.splitlines()
+
+        def response(status_code, *args, **kwargs):
+            res = requests.Response()
+            res.status_code = status_code
+            return res
+
+        with unittest.mock.patch(
+                'tests.general.InvalidLinkBearTest.requests.head',
+                return_value=response(status_code=200)) as mock:
+            uut = InvalidLinkBear(self.section, Queue())
+            self.assertEqual([x.message
+                              for x in list(uut.run('file', file_contents,
+                                                    network_timeout=nt))], [])
+            mock.assert_has_calls([
+                unittest.mock.call('https://facebook.com/', timeout=15,
+                                   allow_redirects=False),
+                unittest.mock.call('https://google.com/',
+                                   timeout=10, allow_redirects=False),
+                unittest.mock.call('https://coala.io/som/thingg/page/123',
+                                   timeout=2, allow_redirects=False)])


bears/general/InvalidLinkBear.py
```diff
@@ -1,7 +1,9 @@
 import re
 import requests
+from urllib.parse import urlparse
```
meetmangukiya: This was added to use `urlparse` to find the host of the given url. And then filter those urls with 
```py
if host in link:
    # the timeout for this link will be the one mentioned in `.coafile`
```

But ATM I'm not able to find how to assert the found URL with a given timeout. For that I'll have to change the function `find_links_in_file`. Suggestions for implementing this are welcome.

bears/general/InvalidLinkBear.py
```diff
@@ -57,6 +59,8 @@ def find_links_in_file(file, timeout, link_ignore_regex):
     def run(self, filename, file,
             timeout: int=DEFAULT_TIMEOUT,
```
meetmangukiya: I see, I forgot to update the `DEFAULT_TIMEOUT` to `DEFAULT_TIMEOUT['anything could go here and it will still return 2, you see thats the beauty of defaultdict']` Thanks for introducing it to me, makman

meetmangukiya:
The tests are yet to be added.

bears/general/InvalidLinkBear.py
```diff
@@ -75,9 +79,16 @@ def run(self, filename, file,
...
+                                 server.
+        :param network_url:      The root url of the server whose timeout has to
+                                 be changed.
         """
+        if network_timeout and network_url:
+            urlparse(network_url).netloc
+
         for line_number, link, code in InvalidLinkBear.find_links_in_file(
-                file, timeout, link_ignore_regex):
+                file, timeout= DEFAULT_TIMEOUT if not network_timeout else network_timeout, link_ignore_regex):
```
meetmangukiya: This is completed. :+1:

bears/general/InvalidLinkBear.py
```diff
@@ -75,9 +79,17 @@ def run(self, filename, file,
...
+                                 server.
+        :param network_url:      The root url of the server whose timeout has
+                                 to be changed.
         """
+        if bool(network_timeout) and bool(network_url):
+            urlparse(network_url).netloc
+
         for line_number, link, code in InvalidLinkBear.find_links_in_file(
-                file, timeout, link_ignore_regex):
+                file, timeout=timeout if not bool(network_timeout) else
```
Makman2: `bool` is implicitly invoked, you don't need to do that yourself ;)

meetmangukiya: I thought it doesn't, so I updated the PR, initially it was just the variable. But then i thought it'd give TypeError

meetmangukiya:
Need ideas on how to test the custom timeout :confused:

Makman2:
@meetmangukiya you could try to mock the functions that are responsible for connecting to the network/internet and replace them with a sleep function I think^^

Makman2:
sorry for responding so late, haven't seen your post :/

bears/general/InvalidLinkBear.py
```diff
@@ -72,12 +81,20 @@ def run(self, filename, file,
         `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
         all your data.
 
-        :param timeout:          Request timeout period.
+        :param network_timeout:       A dict containing the url and the timeout
         :param link_ignore_regex:     A regex for urls to ignore.
         :param follow_redirects: Set to true to autocorrect redirects.
         """
+        host_dict = dict()
```
Makman2: you can immediately use a `defaultdict`, then you don't need to create another one below which imports all keys and values^^

Makman2: Then you can also make the above `network_timeout=defaultdict(lambda: DEFAULT_TIMEOUT)` to a normal dict: `network_timeout=dict`, as the default value doesn't matter any more since you use the defaultdict later^^

bears/general/InvalidLinkBear.py
```diff
@@ -40,22 +43,28 @@ def get_status_code(url, timeout):
...
 
     @staticmethod
-    def find_links_in_file(file, timeout, link_ignore_regex):
+    def find_links_in_file(file, network_timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'((ftp|http)s?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'((ftp|http)s?://[^.:\s_/?#[\]@\\]+\.(?:[^\s()\'"`<>|\\]+|'
+            r'\([^\s()\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
```
Makman2: seems you have rebased but kept the old changes :3
Please remove this change, we want the new regex :+1:

bears/general/InvalidLinkBear.py
```diff
@@ -1,5 +1,7 @@
 import re
 import requests
+from urllib.parse import urlparse
+from collections import defaultdict
```
Makman2: import order: put this one at top :)

bears/general/InvalidLinkBear.py
```diff
@@ -40,22 +43,28 @@ def get_status_code(url, timeout):
...
         for line_number, line in enumerate(file):
             match = regex.search(line)
             if match:
                 link = match.group()
                 if not link_ignore_regex.search(link):
-                    code = InvalidLinkBear.get_status_code(link, timeout)
+                    host = urlparse(link).netloc
+                    code = InvalidLinkBear.get_status_code(
+                                                        link,
+                                                        network_timeout[host])
```
Makman2: looks a bit weird, wanna do
````python
code = InvalidLinkBear.get_status_code(
    link, network_timeout[host])
```
? :)

Makman2:
for the tests: I'm confused why they fail, the general result-yielding-facility is not touched by you. I hope it's no weird side-effect... makes debugging quite hard...
Anyway: As the tests do fail also locally, this makes stuff a bit easier: You can try to insert print statements at some points to see how the algorithms flows. The outputs should be shown inside the test output. If you need help with that, tell me :)

bears/general/InvalidLinkBear.py
meetmangukiya: alphabetical order, right? 

We should have a bear do that IMO

bears/general/InvalidLinkBear.py
meetmangukiya: Sure, anyways gitmate is complaining about the style already :P

bears/general/InvalidLinkBear.py
meetmangukiya: yeah.. I can.. will do it by tomorrow evening IST/GMT +5:30 

It didn't get merged by tonight! :(

bears/general/InvalidLinkBear.py
Makman2: no but that's no problem, just remove your change :)

meetmangukiya: Yeah, I have noticed you work on many PRs parallel-*ly*

bears/general/InvalidLinkBear.py
meetmangukiya: Done in latest push.

bears/general/InvalidLinkBear.py
meetmangukiya: Done in latest push.

bears/general/InvalidLinkBear.py
meetmangukiya: Done in latest push as well!

meetmangukiya:
Almost a month of inactivity. 
God give me time!

bears/general/InvalidLinkBear.py
```diff
@@ -45,17 +48,25 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
                 if not link_ignore_regex.search(link):
-                    code = InvalidLinkBear.get_status_code(link, timeout)
+                    host = urlparse(link).netloc
+                    code = InvalidLinkBear.get_status_code(
+                        link, network_timeout[host])
                     yield line_number + 1, link, code
 
+    def execute(self, *args, **kwargs):
+        return super().execute(*args, **kwargs)
+
```
meetmangukiya: @Makman2 do you remember any reason why I'd add this ?

meetmangukiya:
I'll get this PR going again :tada:

tests/general/InvalidLinkBearTest.py
```diff
@@ -229,3 +230,31 @@ def test_links_to_ignore(self):
...
+        def response(*args, **kwargs):
+            res = requests.Response()
+            res.status_code = 200
+            return res
+
+        with patch('tests.general.InvalidLinkBearTest.requests.head',
+                   side_effect=response ) as mock:
+            uut = InvalidLinkBear(self.section, Queue())
+            self.assertEqual(list(uut.run('file', file_contents,
+                                 network_timeout=network_timeout)), [])
```
meetmangukiya: How is this greater then 80 while below lines are not.? Gitmate is drunk :P

meetmangukiya:
I've corrected in the latest push :+1: 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/coala/coala-bears/877#-:-KZLnls3vaK0VDbsd81L:b-qb97f0)*

bears/general/InvalidLinkBear.py
```diff
@@ -85,12 +89,18 @@ def find_links_in_file(file, timeout, link_ignore_regex, link_ignore_list):
...
+                                                               host])
                     yield line_number + 1, link, code
 
+    return_default_timeout = lambda: DEFAULT_TIMEOUT  # pragma: no cover
+
     @deprecate_settings(link_ignore_regex='ignore_regex')
     def run(self, filename, file,
-            timeout: int=DEFAULT_TIMEOUT,
+            network_timeout: typed_dict(str, int, DEFAULT_TIMEOUT)=
+            defaultdict(return_default_timeout),
```
meetmangukiya: Explanation for not using `lambda` directly here: 
- Because the coverage wasn't counted correctly, since the `lambda` was **never** called according to `coverage`, so I moved it out and marked it a `no cover` outside, because marking it here would lead to marking the whole function as `no cover` while it is completely covered

bears/general/InvalidLinkBear.py
```diff
@@ -85,12 +89,18 @@ def find_links_in_file(file, timeout, link_ignore_regex, link_ignore_list):
...
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
-                    code = InvalidLinkBear.get_status_code(link, timeout)
+                    host = urlparse(link).netloc
+                    code = InvalidLinkBear.get_status_code(link,
+                                                           network_timeout[
+                                                               host])
                     yield line_number + 1, link, code
 
+    return_default_timeout = lambda: DEFAULT_TIMEOUT  # pragma: no cover
```
meetmangukiya: Explanation for using `no cover` given below :point_down: 

@Makman2 you are on!
Any suggestions for getting the lambda covered, while declaring it in the run function itself, this was the workaround I found :wink:

bears/general/InvalidLinkBear.py
Makman2: you sure it wasn't counted correctly? did you try to insert print stmts to check that?

meetmangukiya: Yeah, I checked the html file.. It showed all were covered except the lambda.

meetmangukiya:
Before: 
coverage for bears-general-invalidlinkbear py 99- - chromium_016
It says *line 101 didn't finish the lambda on line 101*
While as you can see the lines below it are covered.

After : 
coverage for bears-general-invalidlinkbear py 100- - chromium_017
coverage for bears-general-invalidlinkbear py 100- - chromium_018

Mixih:
the --doctest-modules augment used in the ci tests is causing an OOM error.. I wonder why

tests/general/InvalidLinkBearTest.py
```diff
@@ -3,6 +3,7 @@
 import requests
 import requests_mock
 import unittest
+from unittest.mock import patch, call
```
Mixih: This `call` import is doing bad things when py.test is run with --doctest-modules. Either import straight unittest.mock (probably due to it overriding sys.call() or something like that), or find a way to namespace it differently

bears/general/InvalidLinkBear.py
```diff
@@ -108,13 +115,22 @@ def run(self, filename, file,
         `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
         all your data.
 
-        :param timeout:          Request timeout period.
+        :param network_timeout:       A dict containing the url and the timeout
```
Makman2: Missing period + I would add a bit more description, why and when this is useful, and how it does work exactly (extracts hostname from given address, etc.)

bears/general/InvalidLinkBear.py
```diff
@@ -108,13 +115,22 @@ def run(self, filename, file,
...
         :param link_ignore_regex:     A regex for urls to ignore.
         :param link_ignore_list: Comma separated url globs to ignore
         :param follow_redirects: Set to true to autocorrect redirects.
         """
+
+        host_dict = dict()
+        for url in network_timeout:
+            host = urlparse(url).netloc
+            timeout = network_timeout[url]
+            host_dict[host] = timeout
```
Makman2: boom:  :D
```
host_dict = {urlparse(url).netloc: timeout for url, timeout in network_timeout.items()}
```
5 loc -> 1 loc :)

Makman2: oh and you could also remove the stmt below when assigning to `network_timeout` directly^^

bears/general/InvalidLinkBear.py
```diff
@@ -85,12 +89,15 @@ def find_links_in_file(file, timeout, link_ignore_regex, link_ignore_list):
                         fnmatch(link, link_ignore_list)):
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
-                    code = InvalidLinkBear.get_status_code(link, timeout)
+                    host = urlparse(link).netloc
+                    code = InvalidLinkBear.get_status_code(
+                                link, network_timeout.get(
+                                        host, InvalidLinkBear.DEFAULT_TIMEOUT))
```
meetmangukiya: I assume you meant to use 4 spaces for hanging indents

bears/general/InvalidLinkBear.py
```diff
@@ -1,5 +1,7 @@
 import re
 import requests
+from collections import defaultdict
```
Mixih: defaultdict is now an unused import

meetmangukiya: Sorry for that, hadn't had time for self review.. I'll be doing it now. Thanks for looking out!

tests/general/InvalidLinkBearTest.py
```diff
@@ -229,3 +230,35 @@ def test_links_to_ignore(self):
...
+        """.splitlines()
+
+        def response(status_code, *args, **kwargs):
+            res = requests.Response()
+            res.status_code = status_code
+            return res
+
+        with unittest.mock.patch(
+                   'tests.general.InvalidLinkBearTest.requests.head',
+                   return_value=response(status_code=200)) as mock:
```
Makman2: indentation is off

tests/general/InvalidLinkBearTest.py
```diff
@@ -229,3 +230,35 @@ def test_links_to_ignore(self):
...
+            self.assertEqual([x.message
+                              for x in list(uut.run('file', file_contents,
+                                                    network_timeout=nt))], [])
+            mock.assert_has_calls([
+                unittest.mock.call('https://facebook.com/', timeout=15,
+                                   allow_redirects=False),
+                unittest.mock.call('https://google.com/',
+                                   timeout=10, allow_redirects=False),
+                unittest.mock.call('https://coala.io/',
+                                   timeout=2, allow_redirects=False)])
```
Makman2: would you test a bigger url too to test whether their host names are properly used? Like `https://google.com/testpage/x/y/z`?

tests/general/InvalidLinkBearTest.py
meetmangukiya: Why doesn't coala catch this?

meetmangukiya: What is the preferred indentation in this case?
*two tabs or three tabs* relative to the `with` block?

meetmangukiya: I am making it *one tab*... as suggested by pep8. where tab_width=4

tests/general/InvalidLinkBearTest.py
meetmangukiya: Introduced in network dict as well as in the URL :+1:

Makman2:
ack f72c246

Makman2:
@rultor merge

tests/general/InvalidLinkBearTest.py
```diff
@@ -229,3 +230,35 @@ def test_links_to_ignore(self):
...
+
+        with unittest.mock.patch(
+                'tests.general.InvalidLinkBearTest.requests.head',
+                return_value=response(status_code=200)) as mock:
+            uut = InvalidLinkBear(self.section, Queue())
+            self.assertEqual([x.message
+                              for x in list(uut.run('file', file_contents,
+                                                    network_timeout=nt))], [])
+            mock.assert_has_calls([
+                unittest.mock.call('https://facebook.com/', timeout=15,
```
jayvdb: `timeout`?? -- the setting name is `network_timeout`.  See https://github.com/coala/coala-bears/issues/1240

sils: but the mocked method from requests is called like this: https://github.com/coala/coala-bears/pull/877/commits/f72c246f9967d498660668512cd8f99eae7589fe#diff-62d2f8ee0cfde374ac098c94b75e3a44R41

jayvdb: ok, sorry, it is testing the setting correctly.

sils:
@Makman2 @meetmangukiya be more careful with renaming settings please, see #1244 we gotta be more mindful about our API.

Makman2:
ah shit sorry have totally forgotten...

PR #878, opened 2016-10-07
Adding C to list of languages supported

jjhonson:
Updating the Supported Languages table to include 'C' in the first postion and shifted the rest of the table down - https://github.com/coala/coala-bears/issues/481


sils:
Hi,

thanks for your interest! You'll see the reason for your build failure down in the bottom of the log (gitmate doesn't pick it up because it doesn't use the development version of coala, otherwise it'd have left a comment for you): Your commit message doesn't comply to our standards.

- Adding context would be helpful (but optional)
- Tense needs to be corrected
- A "Closes" note needs to be added before the issue URL (we use that in bug prediction, that's why our bot gave this PR a "review carefully" label, it uses the git history)

The code itself looks good.

Here's a quick tutorial on getting it right:

---

it seems that your commit message doesn't contain enough information. Don't worry, most people don't know about good commit messages when they join us, including very experienced developers :) Standards vary from project to project.

The most important part is the first line of your commit message - maintainers will use it to read through a large number of commits and thus it's important that it's short but most importantly concise. I'd heartily recommend reading <http://coala.io/commit>!

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix up the message for you - just tell us :).

Cheers!

sils:
Let us know on coala.io/chat when you need any help!

Adrianzatreanu:
theres a mess between the commits, like one adds an unnecessary space added by the other, please squash your commits and have one commit only :( ask on coala.io/chat if you need help

PR #882, opened 2016-10-10
README.rst: Adding 'C' to list of languages supported

jjhonson:
Updated the Supported Languages table to include 'C'

Fixes https://github.com/coala/coala-bears/issues/481


sils:
Hey,

we use some bug prediction in coala which relies on the `Fixes` keyword in commit messages. To get good results from that we need to use `Closes` for normal issues instead of `Fixes` which should only be used for real bugs. (See also [the commit message docs](http://coala.io/commit).)

Please also address the issues pointed out by GitMate.

Cheers!

jjhonson:
@sils - I didn't realize I had done that ... I feel like I'm digging myself deeper though.  I tried to update this pull request and I seem to have created yet another one.  In looking at that I see that there is indentation on that one too - and I'm sure I didn't have that when I was in the text editor so I'm not sure what I did.  I'd like to figure out now how to get rid of the old PRs so they aren't cluttering things up.

sils:
Sure you can close and open your PRs as you please. We recommend sticking to one though ;P

PR #884, opened 2016-10-10
KeywordBear: Remove TODO

netman92:
Coala now ask us if we want to remove TODO notes.

See also https://github.com/coala/coala-bears/pull/778

### DIFF
--- a/bears/general/KeywordBear.py
+++ b/bears/general/KeywordBear.py
@@ -2,7 +2,68 @@ import re
 
 from coalib.bearlib import deprecate_settings
 from coalib.bears.LocalBear import LocalBear
+from coalib.results.Diff import Diff
 from coalib.results.Result import RESULT_SEVERITY, Result
+from coalib.results.SourceRange import SourceRange
+
+from bears.general.AnnotationBear import AnnotationBear
+
+
+def _get_comments(dependency_results):
+    if not dependency_results:
+        return
+
+    annotation_bear_results = dependency_results.get('AnnotationBear')
+    if (not annotation_bear_results or
+            not isinstance(annotation_bear_results, list)):
+        return
+
+    for result in annotation_bear_results:
+        yield from result.contents.get('comments', [])
+
+
+def generate_diff(comments, file, filename,
+                  line, line_number, pos):
+    todo_source_range = SourceRange.from_values(filename, line_number,
+                                                pos + 1)
+    affected_comment_sourcerange = [
+        c for c in comments if todo_source_range in c]
+
+    affected_len = len(affected_comment_sourcerange)
+
+    if affected_len == 0:
+        return {}
+    assert affected_len == 1, "More than 1 affected comment source ranges"
+
+    comment_sourcerange = affected_comment_sourcerange[0]
+
+    comment_start = comment_sourcerange.start.column
+    comment_end = comment_sourcerange.end.column
+    in_multi_line_comment = (
+        comment_sourcerange.start.line != comment_sourcerange.end.line
+    )
+
+    line_before_todo_comment = line[:comment_start - 1].rstrip()
+    line_behind_todo_comment = line[comment_end:].rstrip()
+
+    line_replacement = line_before_todo_comment + line_behind_todo_comment
+
+    diff = Diff(file)
+    if line_replacement and not in_multi_line_comment:
+        diff.change_line(
+            line_number,
+            line,
+            line_replacement + '\n')
+    elif line_replacement and in_multi_line_comment:
+        text_replacement = line[pos:]
+        diff.change_line(
+            line_number,
+            line,
+            line.replace(text_replacement, '').rstrip() + '\n')
+    else:
+        diff.delete_line(line_number)
+
+    return {filename: diff}
 
 
 class KeywordBear(LocalBear):
@@ -11,9 +72,14 @@ class KeywordBear(LocalBear):
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Documentation'}
+    BEAR_DEPS = {AnnotationBear}
 
     @deprecate_settings(keywords='ci_keywords')
-    def run(self, filename, file, keywords: list):
+    def run(self,
+            filename,
+            file,
+            keywords: list,
+            dependency_results: dict=None):
         '''
         Checks the code files for given keywords.
 
@@ -25,15 +91,25 @@ class KeywordBear(LocalBear):
             '(' + '|'.join(re.escape(key) for key in keywords) + ')',
             re.IGNORECASE)
 
-        for line_number, line in enumerate(file):
+        comments = _get_comments(dependency_results)
+
+        for line_number, line in enumerate(file, start=1):
             for keyword in keywords_regex.finditer(line):
+                diffs = generate_diff(
+                    comments,
+                    file,
+                    filename,
+                    line,
+                    line_number,
+                    keyword.start())
                 yield Result.from_values(
                     origin=self,
                     message="The line contains the keyword '{}'."
                             .format(keyword.group()),
                     file=filename,
-                    line=line_number + 1,
+                    line=line_number,
                     column=keyword.start() + 1,
-                    end_line=line_number + 1,
+                    end_line=line_number,
                     end_column=keyword.end() + 1,
-                    severity=RESULT_SEVERITY.INFO)
+                    severity=RESULT_SEVERITY.INFO,
+                    diffs=diffs)
--- a/tests/general/KeywordBearTest.py
+++ b/tests/general/KeywordBearTest.py
@@ -1,5 +1,13 @@
+from collections import namedtuple
+from queue import Queue
+import unittest
+
 from bears.general.KeywordBear import KeywordBear
-from tests.LocalBearTestHelper import verify_local_bear
+from coalib.results.HiddenResult import HiddenResult
+from coalib.results.SourceRange import SourceRange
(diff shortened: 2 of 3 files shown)


sils:
ack https://github.com/coala/coala-bears/pull/884/commits/02090f6c72d55faa743b85efc88cf40f1a3d8b05

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,50 @@
...
+    diffs = {}
+    todo_source_range = SourceRange.from_values(filename, line_number + 1,
+                                                pos + 1)
+    for comment_sourcerange in comments:
+        if todo_source_range not in comment_sourcerange:
+            continue
+
+        diff = Diff(file)
+        comment_start = comment_sourcerange.start.column
+        line_before_todo_comment = line[:comment_start - 1].strip()
```
sils: this code fails in those cases:

```C
int a = 0; /* TODO test
another test
*/
```

will render the comment invalid.

Also in this case:

```C
int a = 1; /* TODO test */ int b=2;
```

it will remove source code. This should be fixable.

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,139 @@
...
+        with execute_bear(uut, "F", text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-int a = 0; /* TODO test\n'
+                             'another test\n'
+                             '*/\n'
+                             '+int a = 0; /* \n'
```
sils: trailing whitespace should be removed

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,70 @@
...
+        if todo_source_range not in comment_sourcerange:
+            continue
+
+        comment_start = comment_sourcerange.start.column
+        comment_end = comment_sourcerange.end.column
+        in_multi_line_comment = (
+            comment_sourcerange.start.line != comment_sourcerange.end.line
+        )
+
+        line_before_todo_comment = line[:comment_start - 1].strip()
```
sils: not strip, rstrip! We shouldn't strip away indentation.

sils: at least for the stuff before the commet

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,70 @@
...
+        line_behind_todo_comment = line[comment_end:].strip()
+
+        line_replacement = line_before_todo_comment + line_behind_todo_comment
+
+        diff = Diff(file)
+        if (line_replacement and
+            line_number == comment_sourcerange.start.line and
+                line_number == comment_sourcerange.end.line):
```
sils: aren't line 47 and 48 equiv. to `not in_multi_line_comment` as per line 31?

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,68 @@
...
+        diff = Diff(file)
+        if line_replacement and not in_multi_line_comment:
+
+            diff.change_line(
+                line_number,
+                line,
+                line_replacement + '\n')
+
+        elif (in_multi_line_comment and
+              line_before_todo_comment):
```
sils: I think this fits in one line

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,68 @@
...
+        line_replacement = line_before_todo_comment + line_behind_todo_comment
+
+        diff = Diff(file)
+        if line_replacement and not in_multi_line_comment:
+
+            diff.change_line(
+                line_number,
+                line,
+                line_replacement + '\n')
+
```
sils: I'd remove this one too

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,68 @@
...
+              line_before_todo_comment):
+            # remove just keyword, not comment
+            text_replacement = line[pos:]
+
+            diff.change_line(
+                line_number,
+                line,
+                line.replace(text_replacement, '').rstrip() + '\n')
+        else:
+            diff.delete_line(line_number)
```
sils: hm, `if in_multi_line_comment and not line_before_todo_comment` this branch will be run and might destroy the comment, right?

sils: e.g. 

```C
/* TODO
test
*/
```

it would remove the first line

netman92: Ou, good point, fixed.

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,170 @@
...
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-test = 55 # todo 123\n'
+                             '+test = 55\n')
+
+    def test_keyword_outside_of_comment(self):
+        section = Section("")
+        section.append(Setting('language', 'python3'))
+        section.append(Setting('keywords', 'todo'))
```
sils: why not just use self.section?

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,170 @@
...
+        self.section = Section("")
+        self.section.append(Setting('language', 'python3'))
+        self.section.append(Setting('keywords', 'TODO'))
+        self.uut = KeywordBear(self.section, Queue())
+
+        self.annotation_bear_result_type = namedtuple('result', ['contents'])
+        self.dep_results = {'AnnotationBear': HiddenResult(
+            'AnnotationBear', {'comments': ()})}
+
+    def test_keyword(self):
```
sils: every test has a purpose, right? Ideally I'd rather split them up, just making everyone of them an own function

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,63 @@
...
+        elif line_replacement and in_multi_line_comment:
+            text_replacement = line[pos:]
+            diff.change_line(
+                line_number,
+                line,
+                line.replace(text_replacement, '').rstrip() + '\n')
+        else:
+            diff.delete_line(line_number)
+
+        diffs[filename] = diff
```
sils: actually you could do a `return {filename: diff}` right here and a `return {}` in line 61 which would give a small performance benefit and simplifies the code (we can remove line 27). There should never be more than one source range touching the keyword, right? And if there is more than one (I could think of some nasty corner cases) this loop would yield conflicting patches.

We can also remove the for loop in favour of a filter: do a filter, assert the length of 1 or 0 (exception otherwise), use the first element if it's there, return `{}` otherwise.

This should be a very simple&quick change

sils: the filter will also be faster than the loop

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,170 @@
...
+        uut = KeywordBear(section, Queue())
+
+        text = ['todo = 123\n']
+        with execute_bear(uut, "F", text,
+                          dependency_results=self.dep_results) as result:
+            self.assertEquals(result[0].diffs, {})
+
+    def test_keyword_between_code(self):
+        section = Section("")
+        section.append(Setting('language', 'c'))
```
sils: same here, use self.section and override the language setting instead of making a new one

tests/general/KeywordBearTest.py
```diff
@@ -14,3 +22,170 @@
...
+        }
+
+        with execute_bear(uut, "F", text,
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-int a=0; /* TODO: Test */ int b=1;\n'
+                             '+int a=0; int b=1;\n')
```
sils: this is so super awesome

sils:
unack https://github.com/coala/coala-bears/pull/884/commits/210d281bf43820c29c34f702e65080191d8128c3

A few simplifications, codestyle and performance suggestions.

sils:
ack https://github.com/coala/coala-bears/pull/884/commits/643f5a1cfb151284ccf6d1029d2cd8a833fc75cf

bears/general/KeywordBear.py
```diff
@@ -2,7 +2,66 @@
...
+
+def generate_diff(comments, file, filename,
+                  line, line_number, pos):
+    todo_source_range = SourceRange.from_values(filename, line_number,
+                                                pos + 1)
+    affected_comment_sourcerange = [
+        c for c in comments if todo_source_range in c]
+    if len(affected_comment_sourcerange) == 0:
+        return {}
+    elif len(affected_comment_sourcerange) > 1:
```
sils: ah you need to put a `# pragma: no cover` command here or just replace the elif by an `assert len..., "More than 1 ..."` which is probably simpler

sils:
ack https://github.com/coala/coala-bears/pull/884/commits/1da70cdea23bfea603deb39fd6459773785e3017

sils:
@rultor merge

jayvdb:
I believe this added `language` as a mandatory missing parameter for KeywordBear , breaking 0.8 CI instances.  See https://github.com/coala/coala-bears/issues/1012 .
But then I do wonder how [this test](https://github.com/coala/coala-bears/commit/1da70cdea23bfea603deb39fd6459773785e3017#diff-23343822d18eeb3fabbd71823cc3643fR19) passes without a `language` setting.

sils:
yes it did

sils:
the test passes because the test doesn't use coala to run the bear, the dependencymechanism is never triggered

PR #912, opened 2016-10-18
requirements: Use compatability operator

sils:

PR #915, opened 2016-10-18
Upgrade yapf

sils:


Asnelchristian:
ack https://github.com/coala/coala-bears/pull/915/commits/8c9fda584e06c15f41d3298685c327cdc7a8f8a1

Asnelchristian:
ack
https://github.com/coala/coala-bears/pull/915/commits/8c9fda584e06c15f41d3298685c327cdc7a8f8a1

2016-10-18 22:13 GMT+01:00 GitMate <notifications@github.com>:

sils:
Yeah thing is, that somehow tests fail when upgrading yapf. The bear seems to behave pretty sanely when running it locally.

sils:
This is the reason why #912 fails

Adrianzatreanu:
It appears this breaks the tests, which probably comes because some settings were renamed or deleted in the latest yapf version. 

CC  @Asnelchristian as i cannot prove you more information on this, no idea whats the matter with it

sils:
if anyone has any idea on this that would be very helpful

Asnelchristian:
nailed it

Asnelchristian:
you need to update the ``allow_multiline_lambdas`` tests

sils:
hm doesn't seem to work for me :/ just adding the `\n` @Asnelchristian ?

sils:
got it!

Asnelchristian:
try this instead ``222222222\n', ')\n'],``

sils:
why does it fail on windows but not on linux? Maybe some dependency caching thing? @Makman2 ?

Makman2:
it's because of carriage returns ... thanks windows again for making stuff different, for the sake of being different...

Makman2:
so yes seems Yapf has changed behaviour for this, this is what I get when hijacking the test:
```
['print 1\n'] #<-- file before analysis
['print 1\r\n'] # <-- after analysis
```

Makman2:
seems this one may have solved this issue:
https://github.com/google/yapf/issues/288

jayvdb:
That fix has been released as 0.13.2, on 2016-10-22, so if it is still broken something else is the problem.

Makman2:
maybe the module just doesn't handle it like providing `--in-place` in the `yapf` executable^^ Possibly there's a parameter for that for `FormatFile`

Makman2:
btw @sils: if that's the case, we need to have 0.13.2 as a minimum version^^

jayvdb:
The build log should show whether 0.13.2 is being installed; it probably is.
In which case, something else is broken.

sils:
assigning @Makman2 so he can reproduce and debug this ;)

jayvdb:
I've updated this PR branch "sils/yapf" with the two relevant commits from https://github.com/coala/coala-bears/pull/1042 (omnibus) , so that those two commits are together and we can close this PR.
(@sils: the previous branch commits are now found at sils/yapf-old)

Mixih:
ack ea3b387 956840f

jayvdb:
@rultor merge

PR #930, opened 2016-10-24
Correction to improve the UX

Att4ck3rS3cur1ty:


Nosferatul:
Hello there, you should write a commit message, check please https://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html. Also could you please what issue this PR is for?

sils:
uhm @Att4ck3rS3cur1ty you can't just change LANGUAGE to LANGUAGES, this thing is used in multiple places. Is there a particular reason for this change? Marking WIP and needsinfo.

sils:
(Thanks for your contribution and interest anyway, didn't try to be harsh)

AbdealiJK:
So, yeah. As @sils said, LANGUAGES is called so because it does normally have more than 1 language and having 2 different keywords is annoying to handle in the code.

If you'd like to discuss the usability of it, please create an issue or drop in on gitter. But at the moment that variable is used in every bear and it tells coala and the user which languages can be handled by this bear. Hence is very important and can't be changed impromptu.

Closing the PR. Please open an issue to discuss further

PR #932, opened 2016-10-24
InvalidLinkBear.py: Amend regex, ignore ``%s``,etc 

amrandahvur:
Modify regex to include links with `%`
in between and to ignore placeholders
like `%s`,`%c`,`%f`, etc. and add corresponding
tests to InvalidLinkBearTest.py

Closes https://github.com/coala/coala-bears/issues/758

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -68,12 +68,48 @@ class InvalidLinkBear(LocalBear):
                                         # set of parenthesis.
                                         # An example can be:
                                         # http://wik.org/Hello_(Adele_song)/200
-            )
-            *)
-                                        # Thus, the whole part above
+            )                           # Thus, the whole part above
                                         # prevents matching of
                                         # Unbalanced parenthesis
+            (
+                (%?[0-9]+)              # A ``%`` sign (0 or 1 times)
+                                        # followed by any number
+                                        # Restricts URLs to end with
+                                        # a number (HTTP Status Code)
+                [^\s()%\'"`<>|\\]+      # Same regex for path name
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Same regex for path name
+                                        # contained within ()
+                ([^\s()%\'"`<>|\\]*)    # Same regex for path name
+                                        # for including HTTP Status Codes
+            )
             (?<!\.)(?<!,)               # Exclude trailing `.` or `,` from URL
+            )                           # First part of the regex ends here
+                                        # Covers URLs with or without
+                                        # placeholders but with necessary
+                                        # HTTP Status Codes at the end
+            |                           # OR (Second part of regex begins here)
+                                        # Covers URLs without any
+                                        # numbers at the end
+                                        # Essential for URLs with placeholders
+                                        # but without HTTP Status Codes
+            (((https?://                # http:// or https:// as only these
+                                        # are supported by the ``requests``
+                                        # library
+            [^.:%\s_/?#[\]@\\]+         # Initial part of domain
+            \.                          # A required dot `.`
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Same regex for path name
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Same regex for path name
+                                        # contained within ()
+            )
+            ))                          # Thus, the whole part above
+                                        # prevents matching of
+                                        # Unbalanced parenthesis
+            (?!(.))                     # Negative lookahead to match URLs not
+                                        # followed by any character
+            (?<!\.)(?<!,))              # Exclude trailing `.` or `,` from URL
             """, re.VERBOSE)
         for line_number, line in enumerate(file):
             match = regex.search(line)
--- a/tests/general/InvalidLinkBearTest.py
+++ b/tests/general/InvalidLinkBearTest.py
@@ -81,6 +81,8 @@ class InvalidLinkBearTest(unittest.TestCase):
         http://httpbin.org/status/200
         http://httpbin.org/status/201
         http://httpbin.org/status/401  # Unauthorized
+        http://httpbin.org/status%20is/200
+        https://encrypted.google.com/search?hl=en&q=cpp+%25s+error/200
 
         # Parentheses
         https://en.wikipedia.org/wiki/Hello_(Adele_song)/200
@@ -108,6 +110,9 @@ class InvalidLinkBearTest(unittest.TestCase):
         http://www.%s.com
         http://www.%d.com
         http://www.%f.com
+        http://httpbin.org/%s/status/520
+        https://encrypted.google.com/search?hl=en&q=cpp+%s+error
+        http://www.%20.com
 
         # Redirect
         http://httpbin.org/status/301


amrandahvur:
@sils I have incorporated the following changes:
1) Add ``http://www.google.com/hello%20world/200`` and ``http://httpbin.org/status%20is/200`` to the list of valid URLs
2) Removed ``http://www.google.com/hello%20world`` from the list of Invalid URLs since it has no reason to be invalid except for ``/200`` not being there.
3) Modified Regex so that tests pass

tests/general/InvalidLinkBearTest.py
```diff
@@ -82,6 +82,8 @@ def test_run(self):
         valid_file = """
         http://httpbin.org/status/200
         http://httpbin.org/status/201
+        http://httpbin.org/status%20is/200
+        http://www.google.com/hello%20world/200
```
sils: this line is redundant, also you should add a test for the invalid ones

amrandahvur: Which line specifically ? In the review, you seem to have highlighted both the lines I have added i.e. ``http://httpbin.org/status%20is/200`` and ``http://www.google.com/hello%20world/200``.
Ok, adding a test for the invalid URLs as well :)

sils: line86

sils: the mocking doesn't care bout it being google or httpbin

sils:
the commit messag doesn't describe your change

tests/general/InvalidLinkBearTest.py
amrandahvur: Oh, Okay :D Updating the PR now

amrandahvur:
@sils Updated, Please Check ?

sils:
please check your commit shortlog, "amend regex" doesn't really tell me what you did. What did you do?

bears/general/InvalidLinkBear.py
```diff
@@ -44,7 +44,8 @@ def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
             r'((ftp|http)s?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'\\]+|\([^\s()\'\"`<>|\\]*\))%?[0-9]*(?:[^\s()%\'\"`<>|\\]+|'
+            r'\([^\s()\'\"`<>|\\]*\))*)(?<!\.)(?<!,)')
```
sils: why are you almost doubling the size of the regex? So much complication?

sils:
Hi,

please don't ping me personally for such things. Everyone can do them - and almost everyone pings me. If you don't address a person specifically you'll likely get help faster from other people!

(Don't worry, this is not a problem but keeping things general is usually a good policy :))

bears/general/InvalidLinkBear.py
amrandahvur: I tried different things and only this seemed to work, Looks like I will have to look for another solution :/

sils: the thing is that if something just seems to work in the given testcases than it'll likely be buggy unless you understand fully *why* it works

amrandahvur:
Okay, Give me some time to figure out some optimal solution for regex :)

sils:
sure

sils:
sure!

amrandahvur:
I need some more time, There are a lot of Deadlines I am currently having in my Institute. Please give me some days if possible :)

Makman2:
@amrandahvur sure no problem :)

amrandahvur:
Logistics:
1) Modified Regex to exclude links with placeholders like ``%s`` in URLs
The solution is a bit complex and messy as there are lot of combinations. Couldn't come up with a simpler one :(
2) Added an Invalid Test Case too
3) Changed Commit Shortlog to be as appropriate as possible (limit of 50 characters)

Adrianzatreanu:
apparently we have merged something on this regex just now (removed the ftp, it doesnt affect you) but you need to rebase and fix the conflict :)

amrandahvur:
@Adrianzatreanu Done :)

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,9 @@ def get_status_code(url, timeout):
     def find_links_in_file(file, timeout, link_ignore_regex):
```
Adrianzatreanu: better shortlog: "Edit regex to ignore placeholders"

tests/general/InvalidLinkBearTest.py
```diff
@@ -123,7 +124,7 @@ def test_run(self):
         http://httpbin.org/status/410
         http://httpbin.org/status/500
         http://httpbin.org/status/503
-        http://www.google.com/hello%20world"""
```
Adrianzatreanu: is this considered to be a valid link actually? or why do you remove it?

bears/general/InvalidLinkBear.py
amrandahvur: This cannot be done in 50 characters. (Because the filename occupies some space too) What should I do ? 
``InvalidLinkBear.py: Edit regex ignore placeholders``
Is this fine ?

AbdealiJK: InvalidLinkBear: Allow placeholders in regex

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,9 @@ def get_status_code(url, timeout):
...
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'\"`<>|\\]+|'
```
AbdealiJK: Wow, this is complicated.
Can we please split this into sections and explain each section ?
```python
            r'(https?://'  # http or https
            r'[^.:%\s_/?#[\]@\\]+'  # Initial part of the domain
            r'\.'  # A required dot (.)
            r'(?:[^\s()%\'\"`<>|\\]+|\([^\s()%\'\"`<>|\\]*\))' # whatsit ?
            r'%?[0-9]*(?:[^\s()%\'\"`<>|\\]+|' # eh ?
            r'\([^\s()%\'\"`<>|\\]*\))*)(?<!\.)(?<!,)') # ooh
```

jayvdb:
Maybe switch to regex whitespace mode so it can be presented logically.

amrandahvur:
Updating the PR, give me some time :)

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,19 @@ def get_status_code(url, timeout):
...
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'(https?://'  # http:// or https://
+            r'[^.:%\s_/?#[\]@\\]+'  # Initial Part of Domain
+            r'\.'  # A required dot (.)
+            r'(?:[^\s()%\'\"`<>|\\]+'  # Domain / Subdomain Name
+            r'|'  # OR
+            r'\([^\s()%\'\"`<>|\\]*\))'
+            # Subdomain Name that is contained within ()
+            r'%?[0-9]*'  # A % sign (0 or 1 times) followed by any number
+                        # Thus %20 is allowed and %s isn't allowed
```
AbdealiJK: Sorry for being pedantic - an extra space needed here

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,19 @@ def get_status_code(url, timeout):
...
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'(https?://'  # http:// or https://
+            r'[^.:%\s_/?#[\]@\\]+'  # Initial Part of Domain
+            r'\.'  # A required dot (.)
+            r'(?:[^\s()%\'\"`<>|\\]+'  # Domain / Subdomain Name
+            r'|'  # OR
+            r'\([^\s()%\'\"`<>|\\]*\))'
+            # Subdomain Name that is contained within ()
+            r'%?[0-9]*'  # A % sign (0 or 1 times) followed by any number
```
AbdealiJK: So, if no letters are allowed, are we assuming a google search like "cpp %s error" is not a valid link that we will check ?
Example: https://encrypted.google.com/search?hl=en&q=cpp+%s+error

If this is expected, can this be added as a test case with a comment stating this ?

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,19 @@ def get_status_code(url, timeout):
...
+                        # Thus %20 is allowed and %s isn't allowed
+            r'(?:[^\s()%\'\"`<>|\\]+'  # Same Regex for Domain Name
+            r'|'  # OR
+            r'\([^\s()%\'\"`<>|\\]*\))*)(?<!\.)(?<!,)')  # Same Regex for
```
AbdealiJK: Thanks a lot for the break up.

This now makes it clear that we're using `[^\s()%\'\"`<>|\\]` 4 times !
Can we store it in a variable like `domain_re = r'[^\s()%\'\"`<>|\\]'` and use it to avoid 4 changes if we want to change that regex ?

Note that this is not in the scope of this PR (as you solve the issue). So, preferably, just open a new github issue and mark it as newcomer so that a newcomer can solve it!

tests/general/InvalidLinkBearTest.py
```diff
@@ -82,6 +82,7 @@ def test_run(self):
         valid_file = """
         http://httpbin.org/status/200
         http://httpbin.org/status/201
+        http://httpbin.org/status%20is/200
```
AbdealiJK: Please add test cases along with the commit that needs these test cases :)
Squish both the commits together ?

bears/general/InvalidLinkBear.py
Makman2: if he already did this he can merge this in a row with this PR imo (in a separate commit obviously :smile:)

bears/general/InvalidLinkBear.py
amrandahvur: Hey, So, I guess I shall ask the same, Would you want me to consider ``https://encrypted.google.com/search?hl=en&q=cpp+%s+error`` as valid or invalid ? What should it be ideally ?

bears/general/InvalidLinkBear.py
amrandahvur: Extra space where specifically ? Line Number ? I guess its 54 right ? For the alignment of ``#`` ?

bears/general/InvalidLinkBear.py
amrandahvur: Because according to what you guys want, I shall try to modify the regex although its getting complicated already.

AbdealiJK: Hm. I think @sils needs to comment here

bears/general/InvalidLinkBear.py
AbdealiJK: Yes, for the alignment of the # - sorry about the ambiguity :D

bears/general/InvalidLinkBear.py
sils: `%` is `%25` so your link should be https://encrypted.google.com/search?hl=en&q=cpp+%25s+error if you want it to be recognized

sils: because if it's a %s the user might want to actually replace it there as well and do a format on that

amrandahvur: Cool, Thank you, I will be updating the PR soon :)

amrandahvur:
Okay, So, I have figured out a regex that works for all possible combinations.
Here is the link https://regex101.com/r/MrDBzP/8
But when I run the tests, they raise an error.


Why is this an error ? ``AssertionError: [ ] == [ ]`` ?
After all, two empty lists are always the same right ?

sils:
it's an assert *not* equal

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

amrandahvur:
Yeah, I figured that out, thus deleted my previous comment. This is taking time although, Its really complicated :(

Makman2:
don't give up, if something's unclear just ping us :) :+1:

amrandahvur:
@Makman2 Thanks for sticking with me :)
Okay, So here is the regex ``https://regex101.com/r/MrDBzP/13`` and it also contains all the testcases for the ``test_run`` function and it works well :) as the tests pass.
But the problem occurs with ``test_link_ignore_regex`` function in InvalidLinkBearTest.py
Here is the error:

How can I possibly solve this now ?

Makman2:
honestly I don't understand the meaning of this particular test, the mocks made there are too weird for me :3
@sils why the heck some of those links shall be ignored?

sils:
@Makman2 usually they're on the ignore list because they have placeholders or it's likely that something's replaced and the URL isn't valid on it's own

amrandahvur:
So what shall be done finally (after using non-capturing groups) ? The issue with the ``test_link_ignore_regex`` function has to be resolved ultimately.
Also, The URLs listed in this function namely:
``
http://sub.example.com
http://sub.example.com/something
``
AND
``
http://www.notexample.com
http://exampe.com
http://example.co.in
``
Do they need to pass the regex (a successful match)?
Also, should I add a status code at the end of each URL?
Example: ``http://sub.example.com/202``

Makman2:
@sils have you looked at the test links? they are completely valid imo^^

Makman2:
@sils ping

amrandahvur:
The Tests pass on my local machine, So they should pass on CircleCI too right ??

Makman2:
@amrandahvur your coverage is not 100%, maybe this is the fault?
https://github.com/coala/coala-bears/pull/932/commits/6db43de3a5ea7e42f9c2ce1d6817e09917b59cb9#diff-c4521389f8a26d35c040551df14b3557L129

Makman2:
@amrandahvur btw you can see coverage report locally when appending `--cov` to your test execution command :)

amrandahvur:
@Makman2 Well, Is this because I added the status code ``500`` to ``http://coalaisthebest.com/`` ?
Although the regex makes links with placeholders (present anywhere in the URL) invalid, but it also demands that the URL ends with a status code. Thus I had to add it :/
How do I solve this issue now pheww :/

Makman2:
@amrandahvur just remove the `500` again and run tests locally with coverage report and see whether coverage is complete. I'm not sure if this is the problem, was just a first guess^^

amrandahvur:
@Makman2 So here are the results 
1) With ``500``, Coverage is 97% and tests pass
2) Without ``500``, Coverage is 95% and tests fail

Makman2:
okay then this is not the cause. Then try to add another test case which covers the missing lines :+1:

sils:
btw. please consult me before merging this PR, I don't have time to watch
over this the whole time but I want to see to it that it's done properly

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

amrandahvur:
@Makman2 Fixed all errors :)
The coverage is also 100% now :)

jayvdb:
Also it would be good if https://github.com/coala/coala-bears/issues/1011 was fixed first.

Makman2:
@jayvdb @amrandahvur you don't have to fix #1011, this is just overhead work. Tests do work, beautification can be done later^^

jayvdb:
@Makman2 , it is not overhead. This patch adds comments to the regex. Those comments were wrong, and the existing regex wasnt understood. Rather than drown *this* PR with review comments about what the regex is currently doing, that processis being done on the other PR, and it will mean this PR is easier to review.

Makman2:
@jayvdb #1011 is about improving the mocking stuff of the current `InvalidLinkBearTest`. I think you are talking about something else :)

amrandahvur:
@jayvdb Since https://github.com/coala/coala-bears/issues/1018 has been closed now, Could you please remove ``status/blocked`` tag ? I would like to resume my work on this :D

sils:
sure

amrandahvur:
All the tests pass with 100% coverage, What can possibly be the error for failure of CircleCi ? The log shows it has to do with something about ``pyflakes`` and stuff

aptrishu:
@amrandahvur try `pip install "pyflakes!=1.2.0,!=1.2.1,!=1.2.2,<1.3.0,>=0.8.1"`

aptrishu:
rebase and push again..

amrandahvur:
@aptrishu Done, Still the same :/

aptrishu:
still the pyflakes version you have is not up to what is required.. search for the version in requirements.txt.

jayvdb:
(rebuilding without cache)

amrandahvur:
I can't understand :/
``requirements.txt`` tells me to install ``pyflakes~=1.3``
while the failed tests of CircleCi tells me to install using ``pip install "pyflakes!=1.2.0,!=1.2.1,!=1.2.2,<1.3.0,>=0.8.1"``
What should be done ?
Is there an escape to this ? :(

bears/general/InvalidLinkBear.py
```diff
@@ -57,24 +57,64 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
             \.                          # A required dot `.`
             (
-                (?:[^\s()%\'"`<>|\\]+)  # Path name
+                (?:[^\s()%\'\"`<>|\\]+) # Path name
                                         # This part does not allow
                                         # any parenthesis: balanced or
                                         # unbalanced.
             |                           # OR
-                \([^\s()%\'"`<>|\\]*\)  # Path name contained within ()
+                \([^\s()%\'\"`<>|\\]*\) # Path name contained within ()
```
jayvdb: the commit summary doesnt explain this change.
I believe it is a separate problem?  If so, it should be a separate PR, with a separate commit message.

jayvdb:
Re pyflakes, somehow we have 'flake8' in the CircleCI cache, and it is causing the pyflakes problems.  We dont need flake8, but it is being installed anyway ;-(

A fresh build without cache and it works : https://circleci.com/gh/coala/coala-bears/4176

bears/general/InvalidLinkBear.py
amrandahvur: Well, This isn't a major issue. We are just escaping ``"`` with a backslash (``\``). Is it still necessary for a separate PR ?

sils:
Can be in the same pr as long as it's an own commit

amrandahvur:
@sils @jayvdb Please have a look

bears/general/InvalidLinkBear.py
amrandahvur: It's just that when I tested the regex on ``regex101.com``, It shows an error unless I escape it, Should I still remove the commit ?

Makman2:
```
Python 3.5.2 (default, Sep 14 2016, 11:28:32) 
[GCC 6.2.1 20160901 (Red Hat 6.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> re.compile('"')
re.compile('"')
>>> 
```
no errors. I think this is because regex101 assumes that you wrap your regex string inside double quotes.

amrandahvur: Cool, Removing it :D

bears/general/InvalidLinkBear.py
```diff
@@ -68,13 +68,53 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
                                         # prevents matching of
                                         # Unbalanced parenthesis
+            (
+                (%?[0-9]+)              # A ``%`` sign (0 or 1 times)
+                                        # followed by any number
+                                        # Thus ``%25`` is allowed
+                                        # and ``%s`` isn't allowed
+                                        # This also puts a restriction
+                                        # that the URLs have to end with
+                                        # a number (HTTP Status Code)
```
Mixih: These examples could be removed to thin down this  comment string leaving
````
# A ``%`` sign (0 or 1 times)
# followed by any number
````

File: bears/general/InvalidLinkBear.py [hunk 12]
```diff
@@ -68,13 +68,53 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
+            )                           # This is the first part of the regex
+                                        # that covers URLs with or without
+                                        # placeholders but with necessary
+                                        # HTTP Status Codes at the end
+            |                           # OR (Second part of regex begins here)
+                                        # This covers URLs without any
+                                        # numbers at the end
+                                        # This is essential for URLs
+                                        # with placeholders but
+                                        # without HTTP Status Codes
```
Mixih: see if you can thin down line 90-99 if possible

bears/general/InvalidLinkBear.py
amrandahvur: Ok :+1: Removing it :smile:

bears/general/InvalidLinkBear.py
amrandahvur: But I should let the meaning be intact right ? I will try to re-frame these phrases :+1:

bears/general/InvalidLinkBear.py
Mixih: Wait you might want to keep 
````
#This also puts a restriction
# that the URLs have to end with
# a number (HTTP Status Code)
````
in some shorter 2 line form...

**Comment by sils** [comment]
@amrandahvur this seems to be good work but TBH the regex seems to grow way too complex, I'm having problems understanding it and that (usually) means others will as well. That will lead to people changing it and breaking it and also makes it hard to determins if it's already broken.

So I really think we gotta make this better maintainable. Maybe we can put parts of the regex in variables to reduce the duplications and give them names==meanings?

**Comment by amrandahvur** [comment]
@sils Yes, there are many parts of the regex that are repeated. We can store them in variables and reuse them which will make it more maintainable and also reduce duplications. The question is, Should I do it or perhaps I could create a new issue out of it ?

File: bears/general/InvalidLinkBear.py [hunk 13]
```diff
@@ -68,13 +68,49 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Same Regex for Path Name
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Same Regex for Path name
+                                        # contained within ()
+            )
+            ))                          # Thus, the whole part above
+                                        # prevents matching of
+                                        # Unbalanced parenthesis
+            (?!(.))(?<!\.)(?<!,))"""    # Negative lookahead to match URLs not
```
jayvdb: please move the closing `"""` to the same position as the previous regex, before `, re.VERBOSE)`

bears/general/InvalidLinkBear.py
```diff
@@ -68,13 +68,49 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
-                                        # Thus, the whole part above
+            )                           # Thus, the whole part above
                                         # prevents matching of
                                         # Unbalanced parenthesis
+            (
+                (%?[0-9]+)              # A ``%`` sign (0 or 1 times)
+                                        # followed by any number
+                                        # Restricts URLs to end with
+                                        # a number (HTTP Status Code)
+                [^\s()%\'"`<>|\\]+      # Same Regex for Path Name
```
jayvdb: Regex and Path Name should be all lower case (here and elsewhere, only use capitals for start of sentence , or proper nouns , or abbreviations).

amrandahvur:
@jayvdb Updated :)

Mixih:
I'd suggest you make variables now, at least for the parts you annotated.

Nosferatul:
closing due to high inactivity

Issue #928, opened 2016-10-23
YapfBear.check_prerequisites isnt reachable


jayvdb:
YapfBear.check_prerequisites emits a warning that `Yapf only supports Python 2.7 and Python 3.4+`.

Except it isnt possible, as coala supports only python 3.4 or later, and wont install on lower versions.

aptrishu:
This issue is solved I think.
https://github.com/ragnerok/coala-bears/pull/1

jayvdb:
> This issue is solved I think. ragnerok#1

No quite.  Please read *this* issue carefully.  It is broader than the other issue.

ragnerok:
Yeah so I only have to change this
return 'Yapf only supports Python 2.7 and Python 3.4+'

to 
return 'Yapf only supports Python 3.4+' and change the if statement too right
and could you please also tell me the mistake in my previous commit message . 
PS I am a noob so any help will be appreciated ! Thanks

jayvdb:
@ragnerok , please try to run this bear on Python 3.3.

sils:
according https://github.com/google/yapf#python-versions they don't support python 3.3 @jayvdb

jayvdb:
@sils : from the issue description ... "Except it isnt possible, as coala supports only python 3.4 or later, and wont install on lower versions."

`YapfBear.check_prerequisites` is *unreachable*.  It needs to be burnt with fire, not fiddled with.

sils:
Right, only check if yapf is available and be done.

jayvdb:
> Right, only check if yapf is available and be done.

@sils, does `PipRequirement.is_installed` need to be called, or is that checked automatically?

sils:
I actually wanted that to happen automatically. There's an issue somewhere - but @fneu was working on a refactoring of the requirement stuff so that was postponed a bit. He's rather busy with lots of stuff...

nikhilsingh291:
Is this issue still available?  I'm a newcomer.

jayvdb:
Previous attempts were https://github.com/coala/coala-bears/pull/936 and https://github.com/coala/coala-bears/pull/937 - repository deleted.

@nikhilsingh291 , make sure you follow http://api.coala.io/en/latest/Developers/Git_Basics.html , especially creating a branch : do not use "master" for your development - that is the most common mistake .  And then when you are ready to commit, also read http://api.coala.io/en/latest/Developers/Writing_Good_Commits.html .

[...]

PR #937, opened 2016-10-26
YapfBear.py: Reduce if-condition

ragnerok:
if-condition was simplified as in #928
Fixes https://github.com/coala/coala-bears/issues/928


Makman2:
@ragnerok you sure the referenced issue is addressed with that commit? I think it's about something else^^ Could you maybe ask @jayvdb again what he exactly means?

Nosferatul:
Also please do not refer to an issue 2 times `Fixes: #928` is fine

Nosferatul:
Also you have 2 PRs for the same issue. please close one of them

jayvdb:
@ragnerok , I already told you that this patch wasn't sufficient to solve the problem, and 15 mins later you submit this Pull Request with the same patch.  If you need clarification, please discuss this at https://github.com/coala/coala-bears/issues/928 .

PR #951, opened 2016-10-28
Issue896

mdorner:
Addresses issue 896. Throws ValueError, if an unsupported option (use_spaces=False or indent_size != 2) is passed. Behavior is consistent with java/CheckstyleBear.


Adrianzatreanu:
im sorry but this PR makes no sense to me, you have commits that have a message and do a totally  different thing, you also have a merge commit, and have totally different authorships on each different commit. could you please fix the stuff, because in this state it is unreviewable.

PR #954, opened 2016-10-29
Bears: Add CsvLintBear

stefanbalas:
Closes https://github.com/coala/coala-bears/issues/927

### DIFF
--- a/bears/csv/CsvLintBear.py
+++ b/bears/csv/CsvLintBear.py
@@ -1,0 +1,22 @@
+from coalib.bearlib.abstractions.Linter import linter
+
+
+@linter(executable='csvlint',
+        output_format='regex',
+        output_regex=r'\d\. (?P<message>.+)\. Row:'
+                     r' (?P<row>\d+)\. (?P<information>.+)',
+        result_message='This ``csv`` file is invalid.')
+class CsvLintBear:
+    """
+    Verifies using ``csvlint`` if ``.csv`` files are valid csv or not.
+    """
+
+    LANGUAGES = {"csv"}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Syntax'}
+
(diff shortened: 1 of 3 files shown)


bears/csv/CsvLintBear.py
```diff
@@ -0,0 +1,10 @@
+from coalib.bearlib.abstractions.Linter import linter
+
+@linter(executable='csvlint',
+        output_format='regex',
+        output_regex=r'INVALID',
+        result_message = 'Your csv file is INVALID')
```
Adrianzatreanu: add period at the end.

Adrianzatreanu: also, no reason to capitalize "invalid"

bears/csv/CsvLintBear.py
```diff
@@ -0,0 +1,10 @@
+from coalib.bearlib.abstractions.Linter import linter
+
+@linter(executable='csvlint',
+        output_format='regex',
+        output_regex=r'INVALID',
```
Adrianzatreanu: because the output the tool gives is like:
``..............................................................( * 50 lines) VALID``

Adrianzatreanu: what other regex would you get? i agree that invalid/valid is a regex enough xD

sils:
so we can't get better results than wether or not the CSV is syntactically valid!? Not even where the problem is?

Adrianzatreanu:
nope

Adrianzatreanu:
but its the best we can get for csv within the next 30mins

sils:
why not just parse CSV with any python lib, that gives you more info
without needing to wrap a linter and the error message of the exception at
least contains info

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

Adrianzatreanu:
ah apparently it does check for stuff, but when stefan was running it on his laptop it wasnt showing shit (using the tool) im testing the tool on my pc now

Adrianzatreanu:
ack b14c440

Adrianzatreanu:
@rultor merge

PR #969, opened 2016-11-01
New Bear: Jinja2Bear

NiklasMM:
This Bear is a native Bear which processes [Jinja2](http://jinja.pocoo.org/docs/dev/) template files.

There is no real global style guide for Jinja2. The one by the [Chromium Project](https://www.chromium.org/developers/jinja), which I referenced in the proposal, turned out to be ~~crap~~ suboptimal, because in many cases it collides with the official documentation and as a heavy user of Jinja2 I just don't like it.

So I chose the defaults of this Bear by looking at the official documentation, my own code and other sources on the Web, I also made it configurable, so that even the Chromium folks could still use it.

The bear can now detect:

* **Unmatched opening or closing control blocks:**  In Jinja2 each `if` and `for` construct must be matched by a corresponding closing tag. If this is not the case this is a syntax error and the file can't be processed.

The bear can detect and fix:

* **Variable and control tag spacing issues**: Variables `{{ var_name }}` and control tags `{% for x in y %}` should be padded with a space on each side for better readability. This is both detected and can be automatically fixed by the Bear. The number of spaces can be changed via settings.

* **Missing/Wrong control end labels** Since there can be much code in between the beginning and end of a control block, it can be hard to know which for-loop of if-block is being closed, when encountering an `{% endfor %}` or `{% endif %}` tag, especially when many of them are nested. Therefore it is good practice to add a comment indicating which block is closed, which is equal to the start tag, e.g.: `{% endfor %}{# for x in y #}`. Missing and differing control end labels can be detected and automatically fixed by the Bear.

This is AFAIK the only linter for Jinja2. There is a couple of more things that it could be checking for, but I think this is already a good set of features. I have already tried it out on some of my templates from work and found it very useful.

*Please review carefully, as this is my first Bear!*

Closes https://github.com/coala/coala-bears/issues/959

### DIFF
--- a/bears/jinja2/Jinja2Bear.py
+++ b/bears/jinja2/Jinja2Bear.py
@@ -1,0 +1,370 @@
+import re
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result, RESULT_SEVERITY
+from coalib.results.Diff import Diff
+
+
+def generate_spacing_diff(file, filename, line, line_number,
+                          match_object, required_spacing):
+    """
+    Generate a diff for incorrectly spaced control or variable tags.
+
+    :param match_object:     A Match object containing the groups ``open``,
+                             ``close`` containing the opening and closing
+                             delimiters of a Jinja2 tag and ``content``
+                             containing everything in between.
+    :param required_spacing: The number of spaces expected after the ``open``
+                             delimiter and before the ``close`` delimiter
+    """
+    diff = Diff(file)
+
+    content_before = line[:match_object.start("open")]
+    content_after = line[match_object.end("close"):]
+
+    spacing = " " * required_spacing
+    replacement = (
+        "{before}{open}{spacing}{content}{spacing}{close}{after}".format(
+            before=content_before,
+            spacing=spacing,
+            after=content_after,
+            content=match_object.group("content").strip(),
+            open=match_object.group("open"),
+            close=match_object.group("close")))
+    diff.change_line(
+        line_number,
+        line,
+        replacement)
+    return {filename: diff}
+
+
+def generate_label_diff(file, filename, line, line_number,
+                        match_object, expected_label):
+    """
+    Generates a diff for a missing or wrong control loop end label.
+    Missing labels will be added, wrong ones replaced, content after the
+    label will be left untouched.
+
+
+    :param match_object:   A Match object containing the groups ``close``
+                           containing the closing delimiters of a control
+                           end tag and the optional group ``label``
+                           containing the end tag label.
+    :param expected_label: The expected label for that control block.
+
+    """
+    diff = Diff(file)
+
+    content_before = line[:match_object.end("close")]
+
+    suffix_start = match_object.end("close")
+    # if a label is there, we cut it out, by starting the kept suffix after it
+    if match_object.group("label") is not None:
+        suffix_start = match_object.end("label")
+
+    content_after = line[suffix_start:]
+
+    replacement = "{before}{label}{after}".format(
+        before=content_before,
+        after=content_after,
+        label=expected_label
+    )
+    diff.change_line(
+        line_number,
+        line,
+        replacement)
+    return {filename: diff}
+
+
+def has_required_spacing(string, required_spacing):
+    """
+    Checks if a string has the required amount of spaces
+    on each side.
+
+    :param string:           The string which is to be checked.
+    :param required_spacing: The number of spaces expected on both
+                             sides of the string.
+    :return:                 True if the string has the required
+                             number of spaces on both sides, False
+                             otherwise.
+
+    >>> has_required_spacing("  foo  ", 2)
+    True
+    >>> has_required_spacing(" foo", 1)
+    False
+    """
+    leading_spaces = len(string) - len(string.lstrip(' '))
+    trailing_spaces = len(string) - len(string.rstrip(' '))
+
+    return (
+        leading_spaces == required_spacing
+        and trailing_spaces == required_spacing)
+
+
+class Jinja2Bear(LocalBear):
+    LANGUAGES = {"Jinja2"}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/azi6u1gcxutoxn0l7xpu4pljp'
+    CAN_DETECT = {'Syntax'}
+    CAN_FIX = {'Formatting', 'Documentation'}
+
+    VARIABLE_REGEX = re.compile(
+        r"(?P<open>{{)(?P<content>.*?)(?P<close>}})")
+    CONTROL_START_REGEX = re.compile(
+        r"(?P<open>{%)(?P<content>\s*(for|if).*?)(?P<close>%})")
+    CONTROL_END_REGEX = re.compile(
+        r"(?P<open>{%)(?P<content>\s*end(for|if)\s*?)"
+        r"(?P<close>%})(?P<label>{#.*?#})?")
+
+    def handle_control_spacing_issue(self, file, filename, line, line_number,
+                                     control_spacing, match_object):
+        """
+        Handles incorrectly spaced control tags (if and for),
+        both start and end tags
+
+        :param control_spacing: The number of spaces placed around control tags
+                                as required by the settings.
+        :param match_object:    A Match object containing the groups ``open``,
+                                ``close`` containing the opening and closing
+                                delimiters of a Jinja2 tag and ``content``
+                                containing everything in between.
+        """
+        diff = generate_spacing_diff(
+            file, filename, line, line_number, match_object, control_spacing)
+        return Result.from_values(
+            origin=self,
+            message="Control blocks should be spaced with "
+                    "`{0}` spaces on each side.".format(
+                        control_spacing),
+            file=filename,
+            line=line_number,
+            column=match_object.start(0) + 1,
+            end_line=line_number,
+            end_column=match_object.end(0) + 1,
+            diffs=diff)
+
+    def check_for_variable_spacing_issues(self,
+                                          file,
+                                          filename,
+                                          line,
+                                          line_number,
+                                          variable_spacing):
+        """
+        Checks any variable in the given line for spacing issues.
+        Yields a Result for each issue found.
+
+        :param file:
+            The content of the file currently being inspected.
+        :param filename:
+            The name of the file currently being inspected.
+        :param line:
+            The content of the line currently being inspected.
+        :param line_number:
+            The current line number.
+        :param variable_spacing:
+            The number of spaces required on each side of a variable tag.
+        """
+        for m in self.VARIABLE_REGEX.finditer(line):
+            match = m.group("content")
+            if not has_required_spacing(match, variable_spacing):
+                diff = generate_spacing_diff(
+                    file, filename, line, line_number, m,
+                    variable_spacing)
+                yield Result.from_values(
+                    origin=self,
+                    message="Variable blocks should be spaced with "
+                            "`{0}` spaces on each side.".format(
+                                variable_spacing),
+                    file=filename,
+                    line=line_number,
+                    column=m.start(0) + 1,
+                    end_line=line_number,
+                    end_column=m.end(0) + 1,
+                    diffs=diff)
+
+    def check_control_start_tags(self,
+                                 file,
+                                 filename,
+                                 line,
+                                 line_number,
+                                 control_spacing):
+        """
+        Checks any control start tag in the given line for spacing issues
+        and puts the expected label on the ``control_stack``
+        Yields a Result for each issue found.
+
+        :param file:
+            The content of the file currently being inspected.
+        :param filename:
+            The name of the file currently being inspected.
+        :param line:
+            The content of the line currently being inspected.
+        :param line_number:
+            The current line number.
+        :param control_spacing:
+            The number of spaces required on each side of a control tag.
+        """
+        for m in self.CONTROL_START_REGEX.finditer(line):
+            # build the label which is expected at the end of this block
+            end_label = "{{#{spacing}{content}{spacing}#}}".format(
+                content=m.group("content").strip(),
+                spacing=" " * control_spacing)
+            self.control_stack.append((end_label, line_number))
+
+            # check for spacing issues with this tag
+            if not has_required_spacing(m.group("content"), control_spacing):
+                yield self.handle_control_spacing_issue(
+                    file, filename, line, line_number, control_spacing, m)
+
+    def check_control_end_tags(self,
+                               file,
+                               filename,
+                               line,
+                               line_number,
+                               control_spacing):
+        """
+        Checks any control end tag in the given line for spacing issues,
+        missing/wrong labels or missing corresponding opening tag.
+        Yields a Result for each issue found.
+
+        :param file:
+            The content of the file currently being inspected.
+        :param filename:
+            The name of the file currently being inspected.
+        :param line:
+            The content of the line currently being inspected.
+        :param line_number:
+            The current line number.
+        :param control_spacing:
+            The number of spaces required on each side of a control tag.
+        """
+        for m in self.CONTROL_END_REGEX.finditer(line):
+            label = m.group("label")
+            try:
+                expected_label, start_in_line = self.control_stack.pop()
+            except IndexError:
+                # an end tag without an opening tag
+                yield Result.from_values(
+                    origin=self,
+                    message="Control end tag has no corresponding start",
+                    file=filename,
+                    line=line_number,
+                    column=m.start(0) + 1,
+                    end_line=line_number,
+                    end_column=m.end(0) + 1,
+                    severity=RESULT_SEVERITY.MAJOR)
+                # Continue to next line as we've detected a syntax error
+                # and label checking can give false results
+                continue
+
+            # check for spacing issues with this tag
+            if not has_required_spacing(m.group("content"), control_spacing):
+                yield self.handle_control_spacing_issue(
+                    file, filename, line, line_number, control_spacing, m)
+
+            # yield results for incorrect or missing end labels
+            if label is None and line_number != start_in_line:
+                diff = generate_label_diff(
+                    file, filename, line, line_number, m, expected_label)
+                yield Result.from_values(
+                    origin=self,
+                    message="Unlabeled control end tag",
+                    file=filename,
+                    line=line_number,
+                    column=m.start(0) + 1,
+                    end_line=line_number,
+                    end_column=m.end(0) + 1,
+                    diffs=diff)
+            elif label != expected_label and line_number != start_in_line:
+                diff = generate_label_diff(
+                    file, filename, line, line_number, m, expected_label)
+                yield Result.from_values(
+                    origin=self,
+                    message="End tag label does not match expected label",
+                    file=filename,
+                    line=line_number,
+                    column=m.start("label") + 1,
+                    end_line=line_number,
+                    end_column=m.end("label") + 1,
+                    diffs=diff)
+
+    def check_for_empty_control_stack(self, filename):
+        """
+        Checks if the control stack is empty and yields a Result for each
+        item left on the stack.
+
+        :param filename: The name of the file currently being inspected.
+        """
+        if len(self.control_stack) > 0:
+            for (_, line_number) in self.control_stack:
+                yield Result.from_values(
+                    origin=self,
+                    message="Control start tag has no corresponding end",
+                    file=filename,
+                    line=line_number,
+                    severity=RESULT_SEVERITY.MAJOR)
+
+    def run(self,
+            filename,
+            file,
+            variable_spacing: int=1,
+            control_spacing: int=1):
+        """
+        Check Jinja2 templates for syntax, formatting and documentation issues.
+        The following aspects are being looked at:
+
+        * Variable spacing:
+            Variable tags should be padded with one space on each side, like
+            this: ``{{ var_name }}``. This can be set to any number of spaces
+            via the setting variable_spacing. Malformatted variable tags are
+            detected and fixes suggested.
+        * Control spacing:
+            Like variable spacing, but for control blocks, i.e. if and for
+            constructs. Looks at both start and end block.
+        * Control labels:
+            It is good practice to label the end of an ``if`` or ``for``
+            construct with a comment equal to the content of the start,
+            like so::
+
+                {% for x in y %}
+                  do something
+                {% endfor %}{# for x in y %}
+
+            Mising or differing labels are detected and fixes suggested.
+            Constructs with start and end on the same line are being ignored.
+        * unbalanced blocks:
+            Each opening tag for a ``for`` or ``if`` construct must be closed
+            by a corresponding end tag. An unbalanced number of opening and
+            closing tags is invalid syntax and will be reported with
+            MAJOR severity by the bear.
+
+
+        :param variable_spacing:
+            The number of spaces a variable block should be spaced with.
+            Default is 1.
+        :param control_spacing:
+            The number of spaces a control block should be spaced with.
+            Default is 1.
+        """
+
+        # Whenever a control construct starts a tuple of the
+        # expected end label and line number are put on this stack.
+        # Whenever an end tag is encountered the last item added is popped.
+        self.control_stack = []
+
+        for line_number, line in enumerate(file, start=1):
+
+            yield from self.check_for_variable_spacing_issues(
+                file, filename, line, line_number, variable_spacing)
+
+            yield from self.check_control_start_tags(
+                file, filename, line, line_number, control_spacing)
+
+            yield from self.check_control_end_tags(
+                file, filename, line, line_number, control_spacing)
+
+        # We've reached the end of the file.
+        # Check if all control blocks have been closed
+        yield from self.check_for_empty_control_stack(filename)
--- a/tests/jinja2/Jinja2BearTest.py
+++ b/tests/jinja2/Jinja2BearTest.py
@@ -1,0 +1,148 @@
+import unittest
+from queue import Queue
+from coalib.settings.Section import Section
+from bears.jinja2.Jinja2Bear import Jinja2Bear
+from tests.LocalBearTestHelper import verify_local_bear, execute_bear
+
+
+Jinja2BearVariableSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r"foo {{ var }}",
+                 r'foo {{ var1|filter }} bar {{ var2 }}'),
+    invalid_files=(r"foo {{var }}",
+                   r"foo {{ var}} bar",
+                   r"{{  var }}",
+                   r"{{ good_var }} foo {{bad_var }}"))
+
+Jinja2BearCustomVariableSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r"foo {{var}}",
+                 r'foo {{var1|filter}} bar {{var2}}'),
+    invalid_files=(r"foo {{ var }}",
+                   r"foo {{ var}} bar",
+                   r"{{  var }}"),
+    settings={"variable_spacing": "0"})
+
+
+class Jinja2BearSpacingDiffTest(unittest.TestCase):
+
+    def setUp(self):
+        self.uut = Jinja2Bear(Section(""), Queue())
+
+    def test_variable_spacing(self):
+        content = (r"foo {{var }} bar",)
+        with execute_bear(self.uut, "F", content) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-foo {{var }} bar'
+                             '+foo {{ var }} bar')
+
+    def test_control_spacing(self):
+        content = [
+            "{% for x in y %}\n",
+            "rendering stuff\n",
+            "{% endfor%}{# for x in y #}\n"
+        ]
+        with execute_bear(self.uut, "F", content) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1,3 +1,3 @@\n'
+                             ' {% for x in y %}\n'
+                             ' rendering stuff\n'
+                             '-{% endfor%}{# for x in y #}\n'
+                             '+{% endfor %}{# for x in y #}\n')
+
+Jinja2BearControlSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r"foo {% if something %} bar {% endif %}",),
+    invalid_files=(r"foo {% if var%} bar {% endif %}",
+                   r"{% if something %} foo {%endif%}{# if something #}",
+                   r"{%for a in var %} {% endfor %}"))
+
+Jinja2BearCustomControlSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r"foo {%if something%} bar {%endif%}",),
+    invalid_files=(r"foo {% if var%} bar {%endif%}",
+                   r"foo {%if foo%} bar {% endif%}",
+                   r"{%for a in var %} {%endfor%}"),
+    settings={"control_spacing": "0"})
+
+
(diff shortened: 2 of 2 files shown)


sils:
/me awards the cool and awesome label

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+            filename,
+            file,
+            variable_spacing: int=1,
+            control_spacing: int=1):
+        '''
+        Check Jinja2 templates for syntax, formatting and documentation issues.
+        The following aspects are being looked at:
+
+        * Variable spacing: Variable tags should be padded with one space on
+                            each side, like this: `{{ var_name }}`. This can
```
AbdealiJK: We use RST for the documentation (i.e. we use sphinx). So, code needs to have 2 backticks.

Also, I'm not sure whether `*` is a valid list item in RST

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                            Malformatted variable tags are detected and fixes
+                            suggested.
+        * Control spacing:  Like variable spacing, but for control blocks,
+                            i.e. if and for constructs. Looks at both start
+                            and end block.
+        * Control labels:   It is good practise to label the end of an if or
+                            for construct with a comment equal to the content
+                            of the start, like so:
+
+                            ```
```
AbdealiJK: I think in RST This is done using `::` and indents.
Could you make the documentation locally and check if formatting is fine ?

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                               invalid syntax and will be reported with MAJOR
+                               severity by the bear.
+
+
+        :param variable_spacing: The number of spaces a variable block
+                                 should be spaced with. Default is 1.
+        :param control_spacing:  The number of spaces a control block
+                                 should be spaced with. Default is 1.
+        '''
+        variable_regex = re.compile(
```
AbdealiJK: I think these should be class variables as the `re.compile` will only happen when the class is being defined and not at each run. Also, these are constants so capitals like `Jinja2Bear.VARIABLE_RE` would be appropriate ?

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                # check for spacing issues with this tag
+                if not is_spacing_valid(m.group("content"), control_spacing):
+                    yield self.handle_control_spacing_issue(
+                        file, filename, line, line_number, control_spacing, m)
+
+            # check for control end tags
+            for m in control_end_regex.finditer(line):
+                label = m.group("label")
+                try:
+                    (expected_label, start_in_line) = control_stack.pop()
```
AbdealiJK: Minor: These curved brackets are redundant

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+        '''
+        variable_regex = re.compile(
+            r"(?P<open>{{)(?P<content>.*?)(?P<close>}})")
+        control_start_regex = re.compile(
+            r"(?P<open>{%)(?P<content>\s*(for|if).*?)(?P<close>%})")
+        control_end_regex = re.compile(
+            r"(?P<open>{%)(?P<content>\s*end(for|if)\s*?)"
+            r"(?P<close>%})(?P<label>{#.*?#})?")
+
+        # Whenever a control construct starts a tuple of the
```
AbdealiJK: Minor modifications for easier reading:

Whenever a control construct starts**,** a tuple of the expected end label and line number are put on this stack. Whenever an end tag is encounter**ed** the **last item added** is popped.

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def generate_spacing_diff(file, filename, line, line_number,
+                          match_object, required_spacing):
+    '''
+    Generate a diff for incorrectly spaced control or variable tags.
+
+    :param match_object:     the match object generated by a match on the
+                             regular expression control_start_regex,
+                             control_end_regex or variable_regex in the
```
AbdealiJK: As control_end_regex, control_start_regex, variable_regex are variables using backticks for code block here would be easier to understand

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Whenever a control construct starts a tuple of the
+        # expected end label and line number are put on this stack.
+        # Whenever an end tag is encountert the item on top is popped.
+        control_stack = []
+
+        for line_number, line in enumerate(file, start=1):
+
+            # check for variable spacing issues
+            for m in variable_regex.finditer(line):
```
AbdealiJK: I'm assuming a variable cannot be defined over multiple lines ? Something like:

    {{ name |
       striptags |
       title }}

bears/jinja2/Jinja2Bear.py
AbdealiJK: Actually, rather than saying it's from those regex, better is to mention it's a Match object with the named groups: start, content, end as that's what matters here.
Otherwise, a user would have to go back, read the regex and see what all can be queried

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+    diff.change_line(
+        line_number,
+        line,
+        replacement)
+    return {filename: diff}
+
+
+def is_spacing_valid(string, required_spacing):
+    '''
+    Returns True iff `string` has exactly <required_spacing>
```
AbdealiJK: Needs `:param:` and `:return:` documentation

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+    spaces at the start and at the end
+    '''
+    leading_spaces = len(string) - len(string.lstrip())
```
AbdealiJK: Note: `lstrip()` searches for any whitespace.
So, if I used 1 tab instead of 1 space it would think that's ok

AbdealiJK: Trying to do complexity analysis here:
Assume length of string is L. Assume number of whitespaces is W.
Your current code does L + W + (L-W) = (for first length) + (for the lstrip) + (for the second length)
This can ideally be done in just W by a for loop:
    
    count = 0
    for s in string:
        count += 1
        if s != ' ':
            break

OR:

    for (i, s) in enumerate(string):
        if s != " ":
            count = i
            break

If you're interested, a quick benchmark would help in understanding this, otherwise it's not that important.

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                        line=line_number,
+                        column=m.start(0) + 1,
+                        end_line=line_number + 1,
+                        end_column=m.end(0) + 1,
+                        diffs=diff)
+
+            # check for control start tags
+            for m in control_start_regex.finditer(line):
+                # build the label which is expected at the end of this block
+                end_label = "{{#{content}#}}".format(
```
AbdealiJK: So, if there's a whitespace issue in the tag name, the whitespace issue is expected in the label also.
Although this is a minor thing, it is a false positive every time there is a wrongly whitespaced control tag.

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                    yield self.handle_control_spacing_issue(
+                        file, filename, line, line_number, control_spacing, m)
+
+            # check for control end tags
+            for m in control_end_regex.finditer(line):
+                label = m.group("label")
+                try:
+                    (expected_label, start_in_line) = control_stack.pop()
+                except IndexError:
+                    # and end tag without an opening tag
```
AbdealiJK: and = an ?

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                    yield Result.from_values(
+                        origin=self,
+                        message="Control end tag has no corresponding start",
+                        file=filename,
+                        line=line_number,
+                        column=m.start(0) + 1,
+                        end_line=line_number + 1,
+                        end_column=m.end(0) + 1,
+                        severity=RESULT_SEVERITY.MAJOR)
+                    continue
```
AbdealiJK: Can you put a comment here saying "Continue to next line as label checking can give false results" because if someone wants to add another feature here later, they need to be aware to put it on top

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                    yield self.handle_control_spacing_issue(
+                        file, filename, line, line_number, control_spacing, m)
+
+                # yield results for incorrect or missing end labels
+                if label is None and line_number != start_in_line:
+                    diff = generate_label_diff(
+                        file, filename, line, line_number, m, expected_label)
+                    yield Result.from_values(
+                        origin=self,
+                        message="Untagged control end tag",
```
AbdealiJK: Un-labeled ? tags and labels being interchanged is confusing X-(

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                    diff = generate_spacing_diff(
+                        file, filename, line, line_number, m, variable_spacing)
+                    yield Result.from_values(
+                        origin=self,
+                        message="Variable blocks should be spaced with "
+                                "`{0}` spaces on each side.".format(
+                                    variable_spacing),
+                        file=filename,
+                        line=line_number,
+                        column=m.start(0) + 1,
```
AbdealiJK: Is `m.start("open")` allowed ? It is easier to understand

AbdealiJK: And the end column should be `m.end("close")` or `m.end(2)` If I understand this correct

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                        diffs=diff)
+                elif label != expected_label and line_number != start_in_line:
+                    diff = generate_label_diff(
+                        file, filename, line, line_number, m, expected_label)
+                    yield Result.from_values(
+                        origin=self,
+                        message="Tag does not match expected label",
+                        file=filename,
+                        line=line_number,
+                        column=m.start(0) + 1,
```
AbdealiJK: This one should have the columns of the label rather than the tag.
So, `m.start("label")`/`m.end("label")` or `m.start(3)`, ..

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                        end_column=m.end(0) + 1,
+                        diffs=diff)
+
+        # We've reached the end of the file.
+        # Check if all control blocks have been closed
+        if len(control_stack) > 0:
+            for (_, line_number) in control_stack:
+                yield Result.from_values(
+                    origin=self,
+                    message="Control start tag has no corresponding end",
```
AbdealiJK: As we already have the label name, can we do "Control start tag 'tag_content' has no corresponding end tag" ? Does that make sense ?

tests/jinja2/Jinja2BearTest.py
```diff
@@ -0,0 +1,147 @@
+import unittest
+from queue import Queue
+from coalib.settings.Section import Section
+from bears.jinja2.Jinja2Bear import Jinja2Bear
+from tests.LocalBearTestHelper import verify_local_bear, execute_bear
+
+
+Jinja2BearVariableSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r"foo {{ var }}",),
```
AbdealiJK: Add one more test case for a valid file with 2 variables just to be sure :)

tests/jinja2/Jinja2BearTest.py
```diff
@@ -0,0 +1,147 @@
...
+                   r"foo {{ var}} bar",
+                   r"{{  var }}"),
+    settings={"variable_spacing": "0"})
+
+
+class Jinja2BearSpacingDiffTest(unittest.TestCase):
+
+    def setUp(self):
+        self.section = Section("")
+        self.uut = Jinja2Bear(self.section, Queue())
```
AbdealiJK: `self.uut = Jinja2Bear(Section(), Queue())` as section isnt really being used

tests/jinja2/Jinja2BearTest.py
AbdealiJK: Also, add a test case with filters

bears/jinja2/Jinja2Bear.py
sils: yeah I should be able to understand this comment without having to read anything else. thats not true atm

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+
+        :param control_spacing: The number of spaces placed around control tags
+                                as required by the settings.
+        :param match_object:    the match object generated by a match on the
+                                regular expression control_start_regex or
+                                control_end_regex in the run method.
+        '''
+        diff = generate_spacing_diff(
+            file, filename, line, line_number, match_object, control_spacing)
+        return Result.from_values(
```
sils: btw. you can also use `self.new_result` and not have to pass `origin=self` as it's set implicitly  (it's a prop that binds the param to the returning closure)

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+                               be closed by a corresponding end tag. An
+                               unbalanced number of opening and closing tags is
+                               invalid syntax and will be reported with MAJOR
+                               severity by the bear.
+
+
+        :param variable_spacing: The number of spaces a variable block
+                                 should be spaced with. Default is 1.
+        :param control_spacing:  The number of spaces a control block
+                                 should be spaced with. Default is 1.
```
sils: either all comment contents should be column aligned or use this:

```
:param ...:
    Content ...
    further content
```

I think that'll make things way easier to read

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+        # We've reached the end of the file.
+        # Check if all control blocks have been closed
+        if len(control_stack) > 0:
+            for (_, line_number) in control_stack:
+                yield Result.from_values(
+                    origin=self,
+                    message="Control start tag has no corresponding end",
+                    file=filename,
+                    line=line_number,
+                    severity=RESULT_SEVERITY.MAJOR)
```
sils: I feel that this function is too long, coulld we check every of the aspects that this bear can do in different functions maybe? That'll make the transition to cEP-0005 easier as well later I think.

bears/jinja2/Jinja2Bear.py
sils: actually I'm not fully sure if iff is commonly known at all IT people, over 70% of senior sw devs have only a bachelor and a rather high percentage doesn't have a university degree at all. That doesn't usually affect the quality of their service but that term definitely comes from the university world.

sils: or is it just a typo? :package:

bears/jinja2/Jinja2Bear.py
NiklasMM: [Asterisk should be fine](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bullet-lists), I'll double check in the generated docs though.

bears/jinja2/Jinja2Bear.py
NiklasMM: Yes, `m.start('open')` and `m.end('close')` would also work, but group `0` is always the whole match and imho the correct thing to use here.

bears/jinja2/Jinja2Bear.py
NiklasMM: Nope it's short short for "if and only if" as you noticed. But I've rephrased it anyway ;)

bears/jinja2/Jinja2Bear.py
NiklasMM: Ok, it seems they are allowed. But I'd leave dealing with them for a future iteration of the Bear.

bears/jinja2/Jinja2Bear.py
NiklasMM: Hmm yeah, I have the expected end label, which is almost the name of the tag and I'm highlighting the line in questions. That should be enough, don't you think?

NiklasMM:
@sils @AbdealiJK Thanks for you thorough review (and the awesome tag)! I've implemented most requested changes and added comments to the others.

bears/jinja2/Jinja2Bear.py
AbdealiJK: Huh, wow. I could have sworn that https://wiki.python.org/moin/TimeComplexity used to show Len() to be O(n). But now it seems to show O(1), so it's cool.
Thanks! This conversation updated my knowledge on python :)

bears/jinja2/Jinja2Bear.py
AbdealiJK: Then a comment is needed that it breaks. And ideally a unit test for that case with elaborate commenya

bears/jinja2/Jinja2Bear.py
AbdealiJK: Ah, I see. Could you comment `0 => the whole match` or so, was a bit confused in the start (Probably because of my lack of understanding on re though... But a comment doesn't hurt)

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,271 @@
...
+        and trailing_spaces == required_spacing)
+
+
+class Jinja2Bear(LocalBear):
+    LANGUAGES = {"Jinja2"}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Syntax'}
+    CAN_FIX = {'Formatting', 'Documentation'}
```
AbdealiJK: Don't do this! Make it a gci!

bears/jinja2/Jinja2Bear.py
NiklasMM: Thanks for the link! I've been looking for something like that. I think the way it's implemented is that the object just has a member variable with it's length, which is updated each time the object is changed.

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+def generate_spacing_diff(file, filename, line, line_number,
+                          match_object, required_spacing):
+    '''
+    Generate a diff for incorrectly spaced control or variable tags.
+
+    :param match_object:     A Match object containing the groups ``open``,
+                             ``close`` containing the opening and closing
+                             delimiters of a Jinja2 tag and ``content``
+                             containing everything in between.
+    :param required_spacing: the number of spaces expected after the open
```
Adrianzatreanu: consistency: start this with an uppercase letter

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+                             delimiter and before the close delimiter
```
Adrianzatreanu: should enclose in double backticks "open" and "close" as you did with the other param

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+    diff.change_line(
+        line_number,
+        line,
+        replacement)
+    return {filename: diff}
+
+
+def generate_label_diff(file, filename, line, line_number,
+                        match_object, expected_label):
+    '''
```
Adrianzatreanu: same """ issue as earlier

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+    Generates a diff for a missing or wrong control loop end label.
+    Missing labels will be added, wrong ones replaced, content after the
+    label will be left untouched.
+
+
+    :param match_object:   the match object generated by a match on the regular
```
Adrianzatreanu: start with uppercase

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+                           expression CONTROL_END_REGEX in the run method
```
Adrianzatreanu: end with period

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+        line,
+        replacement)
+    return {filename: diff}
+
+
+def is_spacing_valid(string, required_spacing):
+    '''
+    Checks if a string is properly spaced.
+
+    :param string:           The string which is to be checked
```
Adrianzatreanu: period

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+    :param required_spacing: The number of spaces expected on both
+                             sides of the string
```
Adrianzatreanu: period

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+    :return:                 True, if the string has the required
```
Adrianzatreanu: redundant comma

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+            end_column=match_object.end(1) + 1,
+            diffs=diff)
+
+    def check_for_variable_spacing_issues(self,
+                                          file,
+                                          filename,
+                                          line,
+                                          line_number):
+        '''
+        Checks any variablein the given line for spacing issues.
```
Adrianzatreanu: space after "variable"

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+        * Variable spacing:    Variable tags should be padded with one space on
+                               each side, like this: ``{{ var_name }}``. This
+                               can be set to any number of spaces via the
+                               setting variable_spacing.
+                               Malformatted variable tags are detected and
+                               fixes suggested.
+        * Control spacing:     Like variable spacing, but for control blocks,
+                               i.e. if and for constructs. Looks at both start
+                               and end block.
+        * Control labels:      It is good practise to label the end of an if or
```
Adrianzatreanu: practise is a verb, practice is the noun

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+                               for construct with a comment equal to the
+                               content of the start, like so::
```
Adrianzatreanu: i think redundant double colon

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+
+                                    {% for x in y %}
+                                      do something
+                                    {% endfor %}{# for x in y %}
+
+                               Mising or differing labels are detected and
+                               fixes suggested.
+                               Constructs with start and end on the same line
+                               are being ignored.
+        * unbalanced blocks:   Each opening tag for a for or if construct must
```
Adrianzatreanu: for & if need to be enclosed in double backticks

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
...
+                               be closed by a corresponding end tag. An
+                               unbalanced number of opening and closing tags is
+                               invalid syntax and will be reported with MAJOR
+                               severity by the bear.
+
+
+        :param variable_spacing: The number of spaces a variable block
+                                 should be spaced with. Default is 1.
+        :param control_spacing:  The number of spaces a control block
+                                 should be spaced with. Default is 1.
```
Adrianzatreanu: bump, please use the one lasse mentioned :D 

check example here to see better http://docs.coala.io/en/latest/Developers/Codestyle.html

bears/jinja2/Jinja2Bear.py
NiklasMM: You mean like a test, which fails and then mark it as skipped?

sils: like it, xfail

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,335 @@
+import re
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result, RESULT_SEVERITY
+from coalib.results.Diff import Diff
+
+
+def generate_spacing_diff(file, filename, line, line_number,
+                          match_object, required_spacing):
+    '''
```
NiklasMM: It's not consistent throughout the codebase. Both can be found in other Bears.

bears/jinja2/Jinja2Bear.py
NiklasMM: No, it denotes the start of a [literal block](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#literal-blocks)

bears/jinja2/Jinja2Bear.py
sils: we prefer `"""`, we did use `'''` as a "marking for translation" back in the old times when we attempted to make all those docs translatable. Nowadays we only use `'''` for legacy code and when we use `"""` inside the doc comment.

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+    Generate a diff for incorrectly spaced control or variable tags.
```
sils: I feel like this function would really like a doctest example! Makes it way better understandable and it's easily viewable :)

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+            before=content_before,
+            spacing=spacing,
+            after=content_after,
+            content=match_object.group("content").strip(),
+            open=match_object.group("open"),
+            close=match_object.group("close")))
+    diff.change_line(
+        line_number,
+        line,
+        replacement)
```
sils: you should check if replacement != line shouldn't you?

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+    return {filename: diff}
+
+
+def generate_label_diff(file, filename, line, line_number,
+                        match_object, expected_label):
+    '''
+    Generates a diff for a missing or wrong control loop end label.
+    Missing labels will be added, wrong ones replaced, content after the
+    label will be left untouched.
+
```
sils: same here, doctest so I have an example. Way easier to understand.

sils:
sorry for being so incomplete, out of time :/

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+    replacement = "{before}{label}{after}".format(
+        before=content_before,
+        after=content_after,
+        label=expected_label
+    )
+    diff.change_line(
+        line_number,
+        line,
+        replacement)
```
sils: same here, if the replacement doesn't differ, we shouldn't have a diff at all, right?

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+    return {filename: diff}
+
+
+def is_spacing_valid(string, required_spacing):
+    '''
+    Checks if a string is properly spaced.
```
sils: doctest please

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+        leading_spaces == required_spacing
+        and trailing_spaces == required_spacing)
+
+
+class Jinja2Bear(LocalBear):
+    LANGUAGES = {"Jinja2"}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    ASCIINEMA_URL = 'https://asciinema.org/a/7i5xxgwf20hkb3gz8jvdxe1hu'
```
sils: nice, I'd use a more common aspect ratio (less height) so it can be properly viewed on all devices. This is super cool.

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,337 @@
...
+            The number of spaces a control block should be spaced with.
+            Default is 1.
+        '''
+
+        # Whenever a control construct starts a tuple of the
+        # expected end label and line number are put on this stack.
+        # Whenever an end tag is encountered the last item added is popped.
+        self.control_stack = []
+        self.variable_spacing = variable_spacing
+        self.control_spacing = control_spacing
```
sils: I'm not a fan of making those class variables because explicit (passing as parameters) is better than implicit according to PEP 20.

sils:
@NiklasMM leaving it to you again then :) I hope the comments are helpful, Ithink introducing a few doctests will help understanding things quicker and easier.

bears/jinja2/Jinja2Bear.py
NiklasMM: The function will only be called if there is an issue, i.e. the line is not as expected. Therefore replacement == line can't happen

bears/jinja2/Jinja2Bear.py
NiklasMM: Ah, shoot, this thing records the aspect ratio? Alright, I'll do it again...

bears/jinja2/Jinja2Bear.py
NiklasMM: If the replacement does not differ the function will not be called

bears/jinja2/Jinja2Bear.py
NiklasMM: I'm sorry, I fail to see how this could be done in a readable way. Also, doctest aren't being run by the testrunner, so what's the point?

Or are you talking about like a psudo-doctest example?

bears/jinja2/Jinja2Bear.py
NiklasMM: See my [other comment](https://github.com/coala/coala-bears/pull/969#discussion_r87249162)

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,370 @@
...
+            file, filename, line, line_number, match_object, control_spacing)
+        return Result.from_values(
+            origin=self,
+            message="Control blocks should be spaced with "
+                    "`{0}` spaces on each side.".format(
+                        control_spacing),
+            file=filename,
+            line=line_number,
+            column=match_object.start(1) + 1,
+            end_line=line_number + 1,
```
sils: why `line_number + 1` and not just `line_number`? The match is all in the smae line right?

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,370 @@
...
+                    variable_spacing)
+                yield Result.from_values(
+                    origin=self,
+                    message="Variable blocks should be spaced with "
+                            "`{0}` spaces on each side.".format(
+                                variable_spacing),
+                    file=filename,
+                    line=line_number,
+                    column=m.start(0) + 1,
+                    end_line=line_number + 1,
```
sils: same here

bears/jinja2/Jinja2Bear.py
```diff
@@ -0,0 +1,370 @@
...
+        Checks if the control stack is empty and yields a Result for each
+        item left on the stack.
+
+        :param filename: The name of the file currently being inspected.
+        """
+        if len(self.control_stack) > 0:
+            for (_, line_number) in self.control_stack:
+                yield Result.from_values(
+                    origin=self,
+                    message="Control start tag has no corresponding end",
```
sils: maybe name the tag here? You have it anyway

sils:
the line number thing is the only thing that seems "wrong", I think if this is fixed we can merge it.

bears/jinja2/Jinja2Bear.py
NiklasMM: As discussed [here](https://github.com/coala/coala-bears/pull/969#discussion_r86066166), I do have the expected label, which is not exactly equal to the tag and the tag in question is highlighted by the Result anyway.

sils:
ack https://github.com/coala/coala-bears/pull/969/commits/885fa9b075878ca99a0d319fff5beb5951341113

sils:
@rultor merge

PR #950, opened 2016-10-28
DartLintBear: check settings validity

NiklasMM:
As pointed out here https://github.com/dart-lang/dart_style/issues/261
dartanalyzer only supports indentation using 2 spaces. This commit
changes DartLintBear to check that use_spaces is True and
indent_size is 2 and raises a ValueError if not.

Fixes https://github.com/coala/coala-bears/issues/897

PR #974, opened 2016-11-04
Replace pull request https://github.com/coala/coala-bears/pull/951, which had a history beyond saving, because it was done during the pycon sprint.

mdorner:
Similar to java's LanguageCheck, CPPLint only supports certain kinds of
settings (use_spaces can only be true, indent_size has to be 2).

Consistent with what LanguageCheck does, this commit makes
CPPLintBear throw ValueError in case unsupported options are passed.

Extends test-case to test for exceptions being thrown for unsupported
options.

### DIFF
--- a/bears/c_languages/CPPLintBear.py
+++ b/bears/c_languages/CPPLintBear.py
@@ -25,6 +25,8 @@ class CPPLintBear:
     @staticmethod
     def create_arguments(filename, file, config_file,
                          max_line_length: int=79,
+                         use_spaces: bool=True,
+                         indent_size: int=2,
                          cpplint_ignore: typed_list(str)=(),
                          cpplint_include: typed_list(str)=()):
         """
@@ -32,6 +34,10 @@ class CPPLintBear:
         :param cpplint_ignore:  List of checkers to ignore.
         :param cpplint_include: List of checkers to explicitly enable.
         """
+        if (not use_spaces):
+            raise ValueError("CPPLint requires use_spaces=True")
+        if (indent_size != 2):
+            raise ValueError("CPPLint requires indent_size=2")
         ignore = ','.join('-'+part.strip() for part in cpplint_ignore)
         include = ','.join('+'+part.strip() for part in cpplint_include)
         return ('--filter=' + ignore + ',' + include,
--- a/tests/c_languages/CPPLintBearTest.py
+++ b/tests/c_languages/CPPLintBearTest.py
@@ -1,5 +1,9 @@
 from bears.c_languages.CPPLintBear import CPPLintBear
-from tests.LocalBearTestHelper import verify_local_bear
+from tests.LocalBearTestHelper import LocalBearTestHelper, verify_local_bear
+from queue import Queue
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+from tests.BearTestHelper import generate_skip_decorator
 
 test_file = """
 int main() {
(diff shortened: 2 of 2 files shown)


bears/c_languages/CPPLintBear.py
```diff
@@ -25,13 +25,19 @@ class CPPLintBear:
...
+                         use_spaces: bool=True,
+                         indent_size: int=2,
                          cpplint_ignore: typed_list(str)=(),
                          cpplint_include: typed_list(str)=()):
         """
         :param max_line_length: Maximum number of characters for a line.
         :param cpplint_ignore:  List of checkers to ignore.
         :param cpplint_include: List of checkers to explicitly enable.
         """
+        if (not use_spaces):
```
NiklasMM: These parentheses are redundant.

bears/c_languages/CPPLintBear.py
```diff
@@ -25,13 +25,19 @@ class CPPLintBear:
...
+            raise ValueError("CPPLint requires use_spaces=True")
+        if (indent_size != 2):
```
NiklasMM: dito. Redundant parentheses.

sils:
hey, which issue is that related to? Can we have a `Closes <full issue url>` in the message? Also your author information seems not to be linked to your github profile - we don't care but if you want it to appear on your stats you have to set the author properly.

tests/c_languages/CPPLintBearTest.py
```diff
@@ -13,16 +17,42 @@
...
-    CPPLintBear,
-    valid_files=(test_file,),
-    invalid_files=(),
-    settings={'cpplint_ignore': 'legal'},
-    tempfile_kwargs={"suffix": ".cpp"})
+        CPPLintBear,
+        valid_files=(test_file,),
+        invalid_files=(),
+        settings={'cpplint_ignore': 'legal'},
+        tempfile_kwargs={"suffix": ".cpp"})
```
sils: why change the indentation here? This is unrelated and we always use four spaces in those cases

bears/c_languages/CPPLintBear.py
```diff
@@ -25,13 +25,19 @@ class CPPLintBear:
...
+                         indent_size: int=2,
                          cpplint_ignore: typed_list(str)=(),
                          cpplint_include: typed_list(str)=()):
         """
         :param max_line_length: Maximum number of characters for a line.
         :param cpplint_ignore:  List of checkers to ignore.
         :param cpplint_include: List of checkers to explicitly enable.
         """
+        if (not use_spaces):
+            raise ValueError("CPPLint requires use_spaces=True")
```
NiklasMM: I recently did a similar thing for `DartLintBear` (#950), there we agreed on the error message saying that the *Bear* does not support/require certain settings rather than the wrapped tool, as the user might not even know which tool they are using. So please change this to something line `CPPLintBear requires...`

tests/c_languages/CPPLintBearTest.py
NiklasMM: Ẁhy did you change the indentation here?

tests/c_languages/CPPLintBearTest.py
```diff
@@ -13,16 +17,42 @@
...
+                  'max_line_length': '13'},
+        tempfile_kwargs={"suffix": ".cpp"})
+
+
+@generate_skip_decorator(CPPLintBear)
+class CPPLintBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section("test section")
+        # uut = unit under test
```
sils: this is usually considered a common abbreviation. If you argue that it's not common at all we might want to change the name because we use uut in a lot of tests

tests/c_languages/CPPLintBearTest.py
```diff
@@ -13,16 +17,42 @@
...
+        self.uut = CPPLintBear(self.section, Queue())
+
+    def test_config_failure_use_spaces(self):
+        self.section["use_spaces"] = "False"
+        self.section.append(Setting('use_spaces', "False"))
```
NiklasMM: If I'm not mistaken, you just neet the `self.section.append(Setting('use_spaces', "False"))`, not the `self.section["use_spaces"] = "False"`.
This goes for the other test cases as well

sils:
the changes look very good and well tested!

Mixih:
@mdorner Status? Have you made all the changes requested? if so, please rebase and update your branch so we know you are ready for re-review.

Nosferatul:
closing due to high inactivity

PR #981, opened 2016-11-07
MarkdownBear: verify config_file

Nosferatul:
If the config_file is provided, all other options are ignored.

Closes https://github.com/coala/coala-bears/issues/925


bears/markdown/MarkdownBear.py
```diff
@@ -101,7 +101,10 @@ def create_arguments(filename, file, config_file,
             "ruleRepetition": markdown_horizontal_rule_repeat,  # int
         }
 
-        config_json = json.dumps(remark_configs)
-        # Remove { and } as remark adds them on its own
-        settings = config_json[1:-1]
-        return '--no-color', '--quiet', '--setting', settings
```
sils: wait - did this actually work? At least this commit seems to be not helpful, config_file should be None all the time

sils:
https://github.com/coala/coala-bears/pull/981/commits/770cb24f2f15a990276987210530466b4b045dee will introduce definite bugs, needs work

bears/markdown/MarkdownBear.py
Nosferatul: I don't understand the purpose of the issue if the `config_file` should be `None` all the time.

Nosferatul: Thx 👍

bears/markdown/MarkdownBear.py
```diff
@@ -1,13 +1,18 @@
...
-from coalib.bearlib.abstractions.Linter import linter
+from coalib.bearlib.abstractions.Linter import linter, _create_linter
 from coalib.bears.requirements.NpmRequirement import NpmRequirement
 
 
 @linter(executable='remark',
         use_stdin=True,
         output_format='corrected',
         result_message='The text does not comply to the set style.')
+@staticmethod
```
AbdealiJK: Um by putting this function here, the linter decorator goes to this function and not the class...

bears/markdown/MarkdownBear.py
```diff
@@ -1,13 +1,18 @@
 import json
 
-from coalib.bearlib.abstractions.Linter import linter
+from coalib.bearlib.abstractions.Linter import linter, _create_linter
```
AbdealiJK: Why is create linter needed?

bears/markdown/MarkdownBear.py
AbdealiJK: I'm fairly confused by this. Where does config_file get set? Where does the value come from?

bears/markdown/MarkdownBear.py
Nosferatul: oh... i didn't saw that...

sils:
please read the docs about generating a config, this is unreviewable.

sils:
not speaking of it being obviously broken

Nosferatul:
My bad, I'll do that

bears/markdown/MarkdownBear.py
```diff
@@ -101,7 +100,11 @@ def create_arguments(filename, file, config_file,
...
-        settings = config_json[1:-1]
-        return '--no-color', '--quiet', '--setting', settings
+        return json.dumps(remark_configs)
+
+    @staticmethod
+    def create_arguments(filename, file, config_file):
+        if config_file:
+            return '--no-color', '--quiet', '--setting', config_file
+        else:
+            return (filename,)
```
sils: shouldn't we rather throw like an exception in this case? There should always be a config generated

sils:
hey this looks good, let's rather do an `assert` on having a config since you generate it and be done with it? Also, are there tests that the configuration works generally? Then we need a rebase and it looks like this can be merged :)

sils:
@Nosferatul your tests fail

Nosferatul:
Yes, I'll fix it when I can

PR #983, opened 2016-11-10
CI: Update bear-docs with every merge

adtac:


.ci/update_docs.sh
```diff
@@ -0,0 +1,22 @@
...
+
+cd bear-docs
+git checkout pre
+
+python3 ./generate local
+
+git add -A
+git commit -m "Docs Update"
+
+git remote add github "https://$GITHUB_USERNAME:$GITHUB_PASSWORD@github.com/coala/bear-docs.git"
```
sils: rather use a token than usename and password

.ci/update_docs.sh
```diff
@@ -0,0 +1,22 @@
...
+
+cd ~
+rm -rf bear-docs
+
+git clone https://github.com/coala/bear-docs.git
+
+cd bear-docs
+git checkout pre
+
+python3 ./generate local
```
sils: you should probably check here if anything actually changed

.ci/update_docs.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+source ../rultor_secrets.sh
+
+set -x
+set -e
```
sils: the sets are to prevent the source from showing the passwords, right? So they should be *before* the sourcing, right?

sils:
oh right

adtac:
ack https://github.com/coala/coala-bears/pull/983/commits/c0bd42418781157f0888874dd5552f640eccf294

sils:
ack https://github.com/coala/coala-bears/pull/983/commits/c1cd3815ccd0e33efe31d7231c1c4110eda12b08

sils:
@rultor merge

PR #987, opened 2016-11-13
YAMLLintBear.py: Correct argument "--config" to "--config-file"

meetarora29:
Corrected the argument at line 53 to ``--config-file``

Fixes https://github.com/coala/coala-bears/issues/979


bears/yml/YAMLLintBear.py
```diff
@@ -50,7 +50,7 @@ def create_arguments(filename, file, config_file, yamllint_config: str=''):
         """
```
Nosferatul: Check you commit shortlog. It should be 50 chars maximum.

PR #996, opened 2016-11-14
AnnotationBear: Use unescaped_search_for

abhsag24:
Fixes https://github.com/coala/coala-bears/issues/993

### DIFF
--- a/bears/general/AnnotationBear.py
+++ b/bears/general/AnnotationBear.py
@@ -4,6 +4,7 @@ from coalib.results.HiddenResult import HiddenResult
 from coalib.results.Result import Result, RESULT_SEVERITY
 from coalib.results.SourceRange import SourceRange
 from coalib.results.AbsolutePosition import AbsolutePosition
+from coala_utils.string_processing.Core import unescaped_search_for
 
 
 class AnnotationBear(LocalBear):
@@ -199,10 +200,8 @@ class AnnotationBear(LocalBear):
             A SourceRange object holding the range of the multi-line annotation
             and the end_position of the annotation as an integer.
         """
-        end_start = text.find(annotation_end,
-                              position + 1)
-        end_end = end_start + len(annotation_end) - 1
-        if end_start == -1:
+        end_end = get_end_position(annotation_end, text, position)
+        if end_end == -1:
             _range = SourceRange.from_absolute_position(
                 filename,
                 AbsolutePosition(file, position))
@@ -238,9 +237,8 @@ class AnnotationBear(LocalBear):
             A SourceRange object identifying the range of the single-line
             string and the end_position of the string as an integer.
         """
-        end_position = (text.find(string_end, position + 1)
-                        + len(string_end) - 1)
-        newline = text.find("\n", position + 1)
+        end_position = get_end_position(string_end, text, position)
+        newline = get_end_position("\n", text, position)
         if newline == -1:
             newline = len(text)
         if end_position == -1:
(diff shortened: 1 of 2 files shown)


aptrishu:
Your commit structure is not correct. see https:coala.io/commit.

bears/general/AnnotationBear.py
```diff
@@ -4,6 +4,7 @@
 from coalib.results.Result import Result, RESULT_SEVERITY
```
Nosferatul: `find` -> `Find`

Nosferatul: Blank line before `Fixes..`

bears/general/AnnotationBear.py
```diff
@@ -199,10 +200,14 @@ def get_multiline(file,
...
-        end_start = text.find(annotation_end,
-                              position + 1)
-        end_end = end_start + len(annotation_end) - 1
-        if end_start == -1:
+        try:
+            end_match = next(unescaped_search_for(annotation_end,
+                                                  text[position + 1:]))
+        except StopIteration:
+            end_match = None
+        print(end_match)
```
Nosferatul: We need this? I don't know what should be printed when the bear is used.

abhsag24: no we dont. Thanks!

bears/general/AnnotationBear.py
Nosferatul: I saw many times review like this

bears/general/AnnotationBear.py
```diff
@@ -199,10 +200,13 @@ def get_multiline(file,
...
-        end_start = text.find(annotation_end,
-                              position + 1)
-        end_end = end_start + len(annotation_end) - 1
-        if end_start == -1:
+        try:
+            end_match = next(unescaped_search_for(annotation_end,
+                                                  text[position + 1:]))
+        except StopIteration:
+            end_match = None
+        end_end = position + end_match.span()[1] if end_match else -1
```
AbdealiJK: If end match is none, then this will give an error saying span() not found in None

AbdealiJK: Sorry, nevermind. Just saw the if

AbdealiJK: It would be easier to just put the end_end calculation inside the try/cath. Why do another if?

bears/general/AnnotationBear.py
```diff
@@ -238,9 +242,19 @@ def get_singleline_strings(file,
...
+        except StopIteration:
+            end_match = None
+        end_position = (position + end_match.span()[1] if end_match
+                        else -1)
+        try:
+            newline_match = next(
+                unescaped_search_for("\n", text[position + 1:]))
+        except StopIteration:
+            newline_match = None
+        newline = position + newline_match.span()[1] if newline_match else -1
```
AbdealiJK: Same end position and newline can be inside the try/catch

abhsag24: they can't be right? they'll raise error at different times and different behavior is expected at those times

bears/general/AnnotationBear.py
```diff
@@ -238,9 +242,18 @@ def get_singleline_strings(file,
...
+                string_end, text[position + 1:]))
+            end_position = (position + end_match.span()[1])
+        except StopIteration:
+            end_position = -1
+        try:
+            newline_match = next(
+                unescaped_search_for("\n", text[position + 1:]))
+            newline = position + newline_match.span()[1]
+        except StopIteration:
+            newline = -1
```
sils: I think you should put this try except construction into a function!

sils:
ok small change, big merge?

sils:
ack https://github.com/coala/coala-bears/pull/996/commits/e68e893aa4693c5ebe56082fe81281fbcfd0a8d0

sils:
@rultor merge

PR #997, opened 2016-11-15
Do not require language_check for coala-bears

sils:
A lot of people are struggling because we actually require java to
install coala-bears. That sucks. Let's rather have this one bear not
work by default than all of those users giving up on the whole tool.

We can even have a nice message.


srisankethu:
Looks good to me :+1:

srisankethu:
ack https://github.com/coala/coala-bears/pull/997/commits/8dd45e3561a0841877611a0392711a2b9e0edcca

sils:
@rultor merge

PR #1002, opened 2016-11-16
New Bear: PycodestyleBear

NiklasMM:
A wrapper for `pycodestyle` formerly known
as `pep8`.

Closes https://github.com/coala/coala-bears/issues/986

### DIFF
--- a/bears/python/PycodestyleBear.py
+++ b/bears/python/PycodestyleBear.py
@@ -1,0 +1,50 @@
+from coalib.bearlib.abstractions.Linter import linter
+
+from coalib.bears.requirements.PipRequirement import PipRequirement
+
+
+@linter(executable='pycodestyle',
+        output_format='regex',
+        output_regex=r'(?P<line>\d+) (?P<column>\d+) '
+                     r'(?P<message>(?P<origin>\S+).*)')
+class PycodestyleBear:
+    """
+    A wrapper for the tool ``pycodestyle`` formerly known as ``pep8``.
+    """
+    LANGUAGES = {"Python", "Python 2", "Python 3"}
+    REQUIREMENTS = {PipRequirement('pycodestyle')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(
+            filename, file, config_file,
+            pycodestyle_ignore: str="",
+            pycodestyle_select: str="",
+            max_line_length: int=79):
+        """
+        :param pycodestyle_ignore:
+            Comma separated list of errors to ignore.
+            See ``pydocstyle`` documentation for a complete list of errors.
+        :param pycodestyle_select:
+            Comma separated list of errors to detect. If given only
+            these errors are going to be detected.
+            See ``pydocstyle`` documentation for a complete list of errors.
+        :param max_line_length:
+            Limit lines to this length.
+        """
+        arguments = [r"--format='%(row)d %(col)d %(code)s %(text)s'"]
+
+        if pycodestyle_ignore:
+            arguments.append("--ignore=" + pycodestyle_ignore)
+
+        if pycodestyle_select:
+            arguments.append("--select=" + pycodestyle_select)
+
+        arguments.append("--max-line-length=" + str(max_line_length))
+
+        arguments.append(filename)
+
+        return arguments
--- a/tests/python/PycodestyleBearTest.py
+++ b/tests/python/PycodestyleBearTest.py
@@ -1,0 +1,50 @@
+from bears.python.PycodestyleBear import PycodestyleBear
+from tests.LocalBearTestHelper import verify_local_bear
+
+
+good_file = '''
+def hello():
+    print("hello world")
+'''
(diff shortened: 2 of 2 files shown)


bears/python/PycodestyleBear.py
```diff
@@ -0,0 +1,48 @@
...
+    REQUIREMENTS = {PipRequirement('pycodestyle')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(
+            filename, file, config_file,
+            ignore: str="",
```
sils: ignore is a coala builtin, rather use pep_ignore or pycodestyle_ignore?

bears/python/PycodestyleBear.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@linter(executable='pycodestyle',
+        output_format='regex',
+        output_regex=r'(?P<line>\d+) (?P<column>\d+) (?P<message>.*)')
+class PycodestyleBear:
+    """
+    A wrapper for the tool ``pycodestyle`` formerly known as ``pep8``.
+    """
+    LANGUAGES = {"Python", "Python 2", "Python 3"}
```
sils: do we need to run pycodestyle with the python 2/3 binary to get python 2 or 3 support?

NiklasMM: I could not find anything that states this explicitly. But I believe it does not matter, because `pycodestyle` only checks the style conventions of PEP8, which apply to both versions. Syntax errors are **not** detected, e.g. a print statement yields no error if run with python3. I briefly scanned the source code and they do at some point query the python version, but it's for some file encoding related issue. So I think were good.

bears/python/PycodestyleBear.py
NiklasMM: done. Although it's kinda long, I think `pycodestyle_X` is the most explicit (hence pythonic) choice.

bears/python/PycodestyleBear.py
```diff
@@ -0,0 +1,48 @@
+from coalib.bearlib.abstractions.Linter import linter
+
+from coalib.bears.requirements.PipRequirement import PipRequirement
+
+
+@linter(executable='pycodestyle',
+        output_format='regex',
+        output_regex=r'(?P<line>\d+) (?P<column>\d+) (?P<message>.*)')
```
sils: if you want, you can put the code into the `origin` capture group, the result will then appear as `PycodestyleBear (E123)`

bears/python/PycodestyleBear.py
```diff
@@ -0,0 +1,48 @@
...
+            pycodestyle_select: str="",
+            max_line_length: int=None):
+        """
+        :param pycodestyle_ignore:
+            Comma separated list of errors to ignore.
+            See ``pydocstyle`` documentation for a complete list of errors.
+        :param pycodestyle_select:
+            Comma separated list of errors to detect. If given only
+            these errors are going to be detected.
+            See ``pydocstyle`` documentation for a complete list of errors.
```
sils: doc for max_line_length missing

bears/python/PycodestyleBear.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        arguments = [r"--format='%(row)d %(col)d %(code)s %(text)s'"]
+
+        if pycodestyle_ignore:
+            arguments.append("--ignore=" + pycodestyle_ignore)
+
+        if pycodestyle_select:
+            arguments.append("--select=" + pycodestyle_select)
+
+        if max_line_length:
```
sils: I'd remove this if and give it a proper default on top, this way people can also see in the doc what's done by default and if they want to modify it

NiklasMM: I initially wanted to set it to 79, but then in the *hugely improbable* case of `pycodestyle` changing it's default users who leave this to the default would get different results when running `pycodestyle` directly and through the Bear. But yeah that's theoretical so I'll change it :D

sils:
maybe even go for 120 which is more popular nowadays? We want the bear to
do the right thing for those users, not annoy them. And setting it is easy

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

NiklasMM:
@sils I addressed the comments. I'd rather leave the default to be the same as in the underlying tool.

bears/python/PycodestyleBear.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        :param pycodestyle_ignore:
+            Comma separated list of errors to ignore.
+            See ``pydocstyle`` documentation for a complete list of errors.
+        :param pycodestyle_select:
+            Comma separated list of errors to detect. If given only
+            these errors are going to be detected.
+            See ``pydocstyle`` documentation for a complete list of errors.
+        :param max_line_length:
+            Limit lines to this length. Defaults to 79 according to PEP8.
```
sils: the `Defaults to` will be automatically added from your default value, please remove that here

sils:
ack https://github.com/coala/coala-bears/pull/1002/commits/900438064956bbd5167eb83aec89966a7929c8cb

sils:
@rultor merge

PR #1004, opened 2016-11-16
Sils/quot

sils:


bears/general/QuotesBear.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def run(self, filename, file, dependency_results,
+            preferred_quotation: str='"'):
+        """
+        Checks and corrects your quotation style.
+
+        For all single line strings, this bear will correct the quotation to
+        your preferred quotation style if that kind of quote is not included
+        within the string.
+
```
Asnelchristian: is  ``"""``->``'''`` possible ?

bears/general/AnnotationBear.py
```diff
@@ -18,14 +18,17 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
...
+        :param language:
+            Language to be whose annotations are to be searched.
+        :param coalang_dir:
+            External directory for coalang file.
+        :return:
+            One HiddenResult containing a dictionary with keys being 'strings'
+            or 'comments' and values being a tuple of SourceRanges pointing to
+            the strings and a tuple of SourceRanges pointing to all comments
+            respectively. The ranges do include string quotes or the comment
+            starting separator but not anything before (e.g. a ``u`` when using
```
Asnelchristian: this means that ``u`` is not included right ? in that case i fel like you need to say it explictly (this kind of suggests that ``u`` is included)

tests/general/QuotesBearTest.py
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+    def test_error_handling(self):
+        dep_results = {'AnnotationBear': [Result("test", "test")]}
+        with execute_bear(self.uut, self.filename, self.double_quote_file,
+                          dependency_results=dep_results) as results:
+            self.assertEqual(len(results), 0)
+
+        dep_results = {'AnnotationBear': [HiddenResult('a', 'error!')]}
+        with execute_bear(self.uut, self.filename, self.double_quote_file,
```
Asnelchristian: this is the cse where there is no string in the file right ? maybe it should go in a different test function (this is just a preference :) )

bears/general/QuotesBear.py
sils: no

sils: "For all single line strings"

tests/general/QuotesBearTest.py
sils: no, then there won't be any sourceranges, this is the case where something totally wrong happened in the annotationbear

bears/general/AnnotationBear.py
sils: the question came up, I think this states it rather clearly

bears/general/AnnotationBear.py
```diff
@@ -18,14 +18,17 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
...
-        :param language:        Language to be whose annotations are to be
-                                searched.
-        :param coalang_dir:     external directory for coalang file.
-        :return:                HiddenResults containing a dictionary with
-                                keys as 'strings' or 'comments' and values as
-                                a tuple of SourceRanges of strings and
-                                a tuple of SourceRanges of comments
-                                respectively.
+        :param language:
+            Language to be whose annotations are to be searched.
```
NiklasMM: Either my english module is malfunctioning or this sentence is broken.
How about `Determines the languages (and therefore comment styles) to expect  .` or something?

bears/general/QuotesBear.py
```diff
@@ -0,0 +1,66 @@
...
+class QuotesBear(LocalBear):
+
+    BEAR_DEPS = {AnnotationBear}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+
+    def correct_single_line_str(self, filename, file, sourcerange,
+                                preferred_quotation):
```
NiklasMM: Please add a docstring to this method.

bears/general/QuotesBear.py
```diff
@@ -0,0 +1,66 @@
...
+        str_contents = file[sourcerange.start.line - 1][
+                       sourcerange.start.column:sourcerange.end.column-1]
+
+        if preferred_quotation in str_contents:
```
NiklasMM: If my preferred_quotation is `'` and the tested string is `"Man, I love McDonald's burgers"` this would yield no result although it's incorrectly quoted, right?

sils: yes, in this version. We could add a `force_preferred_quotation` setting at a later stage but this isthe most common usecase I think

bears/general/QuotesBear.py
```diff
@@ -0,0 +1,82 @@
...
+            The file contents as list of lines.
+        :param sourcerange:
+            The sourcerange indicating where to find the string.
+        :param preferred_quotation:
+            ``'`` or ``"`` respectively.
+        """
+        str_contents = file[sourcerange.start.line - 1][
+                       sourcerange.start.column:sourcerange.end.column-1]
+
+        if preferred_quotation in str_contents:
```
NiklasMM: Why do you even have to check for this? The method is not entered if the string is correctly quoted.

bears/general/QuotesBear.py
```diff
@@ -0,0 +1,82 @@
...
+        string, no correction will be made.
+
+        This function will yield one or no Result objects.
+
+        :param filename:
+            The filename of the file to correct the line in.
+        :param file:
+            The file contents as list of lines.
+        :param sourcerange:
+            The sourcerange indicating where to find the string.
```
NiklasMM: I'd expect the type of the expected value here or at least a hint on where to find it. e.g. "is an object with the keys..."

bears/general/QuotesBear.py
sils: if I prefer double quotes and I have a string `' hey " there '` the method will be entered

NiklasMM: ohhh I get it now.. This is here so it does not break stuff.... yeah nvm

bears/general/QuotesBear.py
NiklasMM: Is that a stdlib thing? I'm not familiar.

bears/general/QuotesBear.py
NiklasMM: Wait, but would you not want to still yield a result, but without a diff?

bears/general/QuotesBear.py
sils: it's a coalib thing, all results have SourceRanges actually

NiklasMM: OK, so I would add this or at least write it in CamelCase, but I'll leave it to your discretion.

bears/general/QuotesBear.py
sils: it's not only that it doesn't break stuff. Most people allow using the not-preferred-quote for avoiding having to escape the quote inside the string. At a later point I'll add a `force_preferred_quotation` option that will autoescape that

NiklasMM: :+1:

NiklasMM:
ack f62d76b a430608

sils:
yayyyy!

sils:
@rultor merge

PR #1019, opened 2016-11-19
InvalidLinkBear: Add pip vcs url parser

rafidaslam:
Add pip vcs url parser in InvalidLinkBear, so InvalidLinkBear
now can detects some pip vcs urls without 'Broken link' warning.

Closes https://github.com/coala/coala-bears/issues/1009

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -40,16 +40,24 @@ class InvalidLinkBear(LocalBear):
             pass
 
     @staticmethod
+    def parse_pip_vcs_url(link):
+        splitted_at = link.split('@')[0]
+        splitted_schema = splitted_at[splitted_at.index('+') + 1:]
+        return splitted_schema
+
+    @staticmethod
     def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'((git\+|bzr\+|svn\+|hg\+|)https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s('
+            r')%\'"`<>|\\]+|\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
         for line_number, line in enumerate(file):
             match = regex.search(line)
             if match:
                 link = match.group()
                 if not link_ignore_regex.search(link):
+                    if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
+                        link = InvalidLinkBear.parse_pip_vcs_url(link)
                     code = InvalidLinkBear.get_status_code(link, timeout)
                     yield line_number + 1, link, code
 
--- a/tests/general/InvalidLinkBearTest.py
+++ b/tests/general/InvalidLinkBearTest.py
@@ -151,3 +151,58 @@ class InvalidLinkBearTest(unittest.TestCase):
         self.assertResult(valid_file=long_url_redirect,
                           invalid_file=short_url_redirect,
                           settings={'follow_redirects': 'yeah'})
+
+    def test_pip_vcs_url(self):
+        with_at = """
+        git+http://httpbin.org/status/200@master
+        svn+http://httpbin.org/status/200@master
+        hg+http://httpbin.org/status/200@master
+        bzr+http://httpbin.org/status/200@master
+        """.splitlines()
+
+        self.assertResult(valid_file=with_at)
+
+        with_hash = """
+        git+http://httpbin.org/status/200#egg=coala-bears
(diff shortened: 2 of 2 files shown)


bears/general/InvalidLinkBear.py
```diff
@@ -40,18 +40,32 @@ def get_status_code(url, timeout):
...
+            r')%\'"`<>|\\]+|\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
         for line_number, line in enumerate(file):
             match = regex.search(line)
             if match:
                 link = match.group()
                 if not link_ignore_regex.search(link):
-                    code = InvalidLinkBear.get_status_code(link, timeout)
-                    yield line_number + 1, link, code
+                    if ('hg+' in link) or ('bzr+' in link) or \
+                       ('git+' in link) or ('svn+' in link):
```
sils: rather than `in` I'd use `startswith`

bears/general/InvalidLinkBear.py
```diff
@@ -40,18 +40,32 @@ def get_status_code(url, timeout):
...
+                        parsed_link = InvalidLinkBear.parse_pip_vcs_url(link)
```
sils: I'd do `link = Inva...` then you can unindent line 63ff again and remove the else clause, less code duplication

bears/general/InvalidLinkBear.py
```diff
@@ -40,18 +40,32 @@ def get_status_code(url, timeout):
...
+        return splitted_schema
+
+    @staticmethod
     def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'((git\+|bzr\+|svn\+|hg\+)?https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s('
+            r')%\'"`<>|\\]+|\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
```
sils: maybe instead of adding the `git+` etc. in front of the regex rather forbit an `@` in the later part of the regex to cut it off correctly? @jayvdb makes sense? Less code, then we could just change the regex and go with that

sils:
testcase missing

bears/general/InvalidLinkBear.py
rafidaslam: good idea

bears/general/InvalidLinkBear.py
rafidaslam: I doubt this will break url that have '@' symbol

bears/general/InvalidLinkBear.py
```diff
@@ -40,16 +40,24 @@ def get_status_code(url, timeout):
...
+        return splitted_schema
+
+    @staticmethod
     def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r'((git\+|bzr\+|svn\+|hg\+|)https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s('
+            r')%\'"`<>|\\]+|\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
```
Mixih: Don't we need to escape the literal `/`s?

rafidaslam: Yes, I think we don't need to escape that

Mixih: ok

Mixih:
ack 2a2be49

Mixih:
wait... unack 2a2be49

Mixih:
unack 2a2be49

Mixih:
ack 2a2be49

Mixih:
@rultor merge

tests/general/InvalidLinkBearTest.py
```diff
@@ -151,3 +151,58 @@ def test_redirect_threshold(self):
...
+        bzr+http://httpbin.org/status/404#egg=coala-bears
+        """.splitlines()
+
+        self.assertResult(invalid_file=brokenlink_hash)
+
+        brokenlink_at_hash = """
+        git+http://httpbin.org/status/404@master#egg=coala-bears
+        svn+http://httpbin.org/status/404@master#egg=coala-bears
+        hg+http://httpbin.org/status/404@master#egg=coala-bears
+        bzr+http://httpbin.org/status/404@master#egg=coala-bears
```
sils: @rafidaslam @Mixih those tests are 75% useless: the assertResult will only make sure that *at least one* result is raised, not necessarily one for each line. That's why we have a loop in the big testcase. This test needs fixing.

Mixih: Ok, I'll add a loop

PR #1021, opened 2016-11-19
InvalidLinkBear.py: Document regex in Verbose

amrandahvur:
Document complex regex using ``re.X`` mode

Closes https://github.com/coala/coala-bears/issues/1018

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -49,8 +49,32 @@ class InvalidLinkBear(LocalBear):
     def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'((git\+|bzr\+|svn\+|hg\+|)https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s('
-            r')%\'"`<>|\\]+|\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r"""
+            ((git\+|bzr\+|svn\+|hg\+|)  # For VCS URLs
+            https?://                   # http:// or https:// as only these
+                                        # are supported by the ``requests``
+                                        # library
+            [^.:%\s_/?#[\]@\\]+         # Initial part of domain
+            \.                          # A required dot `.`
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Path name
+                                        # This part does not allow
+                                        # any parenthesis: balanced or
+                                        # unbalanced.
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Path name contained within ()
+                                        # This part allows path names that
+                                        # are explicitly enclosed within one
+                                        # set of parenthesis.
+                                        # An example can be:
+                                        # http://wik.org/Hello_(Adele_song)/200
+            )
+            *)
+                                        # Thus, the whole part above
+                                        # prevents matching of
+                                        # Unbalanced parenthesis
+            (?<!\.)(?<!,)               # Exclude trailing `.` or `,` from URL
+            """, re.VERBOSE)
         for line_number, line in enumerate(file):
             match = regex.search(line)
             if match:


Asnelchristian:
can you change you commit's shortlog to ``Document regex`` ( i don't the use of the ``to Verbose`` part )

amrandahvur:
@jayvdb Please Check ? I guess these are the changes right ?

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,16 @@ def get_status_code(url, timeout):
     def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r"""
+            (https?://                     # http:// or https://
```
jayvdb: Only these two because requests only supports those two.
This should also be mentioned in the bear docs.

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,16 @@ def get_status_code(url, timeout):
...
+            [^.:%\s_/?#[\]@\\]+            # Initial Part of Domain
+            \.                             # A required dot (.)
+            (?:[^\s()%\'"`<>|\\]+          # Domain / Subdomain Name
```
jayvdb: Indent the two halves of the OR.

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,16 @@ def get_status_code(url, timeout):
...
+            |                              # OR
+            \([^\s()%\'"`<>|\\]*\))*)      # Subdomain Name contained within ()
```
jayvdb: This isnt not explained very well. More detail needed

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,16 @@ def get_status_code(url, timeout):
...
+            (?<!\.)(?<!,)                  # URL shouldn't end with (.) or (,)
```
jayvdb: I think this should be: Exclude trailing dot or comma from the URL.

Or does it reject URLSLs ending in dot or comma?

amrandahvur: It doesn't reject the whole URL as such, just doesn't include the trailing ``commas`` or ``periods``.

bears/general/InvalidLinkBear.py
amrandahvur: Added an example, should that suffice ?

bears/general/InvalidLinkBear.py
amrandahvur: Done

bears/general/InvalidLinkBear.py
amrandahvur: Done

amrandahvur:
@jayvdb Should this be fine ?

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,23 @@ def get_status_code(url, timeout):
...
+                                        # 2 are supported by the ``requests``
+                                        # library
+            [^.:%\s_/?#[\]@\\]+         # Initial Part of Domain
+            \.                          # A required dot (.)
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Domain / Subdomain Name
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Subdomain Name contained within ()
+                                        # An example can be:
+                                        # http://wik.org/Hello_(Adele_song)/200
```
amrandahvur: Okay :/ What should I write them as then ? They are essentially the parts after the domain. Can I call them as ``parameters`` or something ?

amrandahvur:
I just changed a name in comments and the ``appveyor`` tests fail! But they do pass on my local machine with 100% coverage. Has something broken since yesterday ?

jayvdb:
ya, we are almost done fixing the builds

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,23 @@ def get_status_code(url, timeout):
...
+            r"""
+            (https?://                  # http:// or https:// as only these
+                                        # 2 are supported by the ``requests``
+                                        # library
+            [^.:%\s_/?#[\]@\\]+         # Initial Part of Domain
+            \.                          # A required dot (.)
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Path Name
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Path Name contained within ()
```
jayvdb: This is not an acceptable comment yet - it only explains what is extremely obvious, which are not appropriate comments.
We need the comments to explain the complex aspects.

why contained within () ?

My understanding is that that both parts of this `OR` is to prevent matching a trailing `)` after the URL, such as `(e.g. http://foo.bar/path)`

This part of the regex ensures that it only matches balanced round brackets within the URL. If the URL actually should contain unbalanced round brackets, this regex will skip the URL, because it cant reliably find the end of the URL.
This part of the regex is sub-optimal, but is 'good enough' and doesnt have false positives.

amrandahvur: Aah yes, I now see it clearly, Thank you for pointing it out. Also, I think this has been done because both the parts of the ``OR`` include ``[^\s()%\'"`<>|\\]`` wherein ``(`` and ``)`` are also present. This makes the regex ``not`` match the URLs containing unbalanced round brackets. And thus the second part of ``OR`` is required to match ``path names within ()``. So, a quick question, 
1) Should I document it as you stated ?
2) Or Should I actually remove ``(`` and ``)`` from ``[^\s()%\'"`<>|\\]``. By doing this, we won't need the second part of ``OR`` and also, maybe the unbalanced round brackets can also be matched.

amrandahvur:
@jayvdb Is this fine ?

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,31 @@ def get_status_code(url, timeout):
     def find_links_in_file(file, timeout, link_ignore_regex):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
-            r'(https?://[^.:%\s_/?#[\]@\\]+\.(?:[^\s()%\'"`<>|\\]+|'
-            r'\([^\s()%\'"`<>|\\]*\))*)(?<!\.)(?<!,)')
+            r"""
+            (https?://                  # http:// or https:// as only these
+                                        # 2 are supported by the ``requests``
```
jayvdb: "2" is not necessary; it is implicit.

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,31 @@ def get_status_code(url, timeout):
...
+                                        # library
+            [^.:%\s_/?#[\]@\\]+         # Initial Part of Domain
```
jayvdb: Please use uppercase only on first word

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,31 @@ def get_status_code(url, timeout):
...
+            \.                          # A required dot (.)
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Path Name
+                                        # This Part Does not allow
+                                        # Unbalanced Parenthesis to be present
```
jayvdb: This part does not allow any parenthesis: balanced or unbalanced.

bears/general/InvalidLinkBear.py
```diff
@@ -43,8 +43,31 @@ def get_status_code(url, timeout):
...
+                                        # are explicitly enclosed within one
+                                        # set of Parenthesis.
+                                        # An example can be:
+                                        # http://wik.org/Hello_(Adele_song)/200
+            )
+            *)
+                                        # Thus, the whole part above
+                                        # prevents matching of
+                                        # Unbalanced Parenthesis
+            (?<!\.)(?<!,)               # Exclude trailing (.) or (,) from URL
```
jayvdb: > trailing (.)

Please dont use round brackets to indicate a literal. Use backticks.

amrandahvur:
I do not understand why the ``appveyor`` tests fail again. A change in documentation shouldn't make this happen. Wasn't this thing fixed earlier ?

amrandahvur:
Fixed Merge Conflicts

amrandahvur:
@jayvdb Please see, This should have been merged by now :/

bears/general/InvalidLinkBear.py
```diff
@@ -49,8 +49,33 @@ def parse_pip_vcs_url(link):
...
+                                        # library
+            [^.:%\s_/?#[\]@\\]+         # Initial part of domain
+            \.                          # A required dot `.`
+            (
+                (?:[^\s()%\'"`<>|\\]+)  # Path name
+                                        # This part does not allow
+                                        # any parenthesis: balanced or
+                                        # unbalanced.
+            |                           # OR
+                \([^\s()%\'"`<>|\\]*\)  # Path name contained within ()
```
AbdealiJK: Question here: Why do we assume there can't be unbalanced paranthesis in a URL ?

AbdealiJK:
Documentation looks good. Question about the regex though (now that I understand what the regex actually does)

ack 67c5dad

bears/general/InvalidLinkBear.py
jayvdb: There can be, but we cant detect it with this regex. I.e. the regex is sub-optimal, but improving it is hard without introducing false positives.

jayvdb:
unack 67c5dad

bears/general/InvalidLinkBear.py
```diff
@@ -49,8 +49,33 @@ def parse_pip_vcs_url(link):
...
+                                        # set of parenthesis.
+                                        # An example can be:
+                                        # http://wik.org/Hello_(Adele_song)/200
+            )
+            *)
+                                        # Thus, the whole part above
+                                        # prevents matching of
+                                        # Unbalanced parenthesis
+            (?<!\.)(?<!,)               # Exclude trailing `.` or `,` from URL
+            """
```
amrandahvur: Okay, give me a minute

jayvdb:
ack cdb153c

jayvdb:
@rultor merge

bears/general/InvalidLinkBear.py
sils: > There can be, but we cant detect it with this regex. I.e. the regex is sub-optimal, but improving it is hard without introducing false positives.

Actually WRT this I disagree and think the behaviour is already pretty optimal: unbalanced parentheses are likely part of markup around the URL. The whole truth is that the whole URL detection is a heuristic because otherwise we'd have to understand the markup language which we ATM don't. So it does things like this and so far this has yielded a very very low false positive rate.

(I wrote the regex mostly I think FWIW.)

sils:
:tada:

PR #1041, opened 2016-11-22
GitCommitBear: Test for duplicate shortlogs

meetmangukiya:
Fixes https://github.com/coala/coala-bears/issues/861

### DIFF
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -61,25 +61,31 @@ class GitCommitBear(GlobalBear):
                                            allowed or not.
         """
         with change_directory(self.get_config_dir() or os.getcwd()):
-            stdout, stderr = run_shell_command('git log -1 --pretty=%B')
+            stdout_head, stderr_head = run_shell_command(
+                "git log -1 --pretty=%B")
+            stdout_shortlogs, stderr_shortlogs = run_shell_command(
+                "git log --pretty=%s")
 
-        if stderr:
-            self.err('git:', repr(stderr))
+        if stderr_head or stderr_shortlogs:
+            self.err("git:", repr(stderr_head),
+                     "\ngit:", repr(stderr_shortlogs))
             return
 
-        stdout = stdout.rstrip('\n').splitlines()
+        stdout_head = stdout_head.rstrip("\n").splitlines()
+        stdout_shortlogs = stdout_shortlogs.rstrip("\n").splitlines()
 
-        if len(stdout) == 0:
+        if len(stdout_head) == 0:
             if not allow_empty_commit_message:
-                yield Result(self, 'HEAD commit has no message.')
+                yield Result(self, "HEAD commit has no message.")
             return
 
         yield from self.check_shortlog(
-            stdout[0],
+            stdout_head[0],
             **self.get_shortlog_checks_metadata().filter_parameters(kwargs))
         yield from self.check_body(
-            stdout[1:],
+            stdout_head[1:],
             **self.get_body_checks_metadata().filter_parameters(kwargs))
+        yield from self.check_duplicate_shortlogs(stdout_shortlogs)
 
     def check_shortlog(self, shortlog,
                        shortlog_length: int=50,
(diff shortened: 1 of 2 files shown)


meetmangukiya:
It'd be real nice if I can get some advice on how can I fix this tiny bit of missing coverage :disappointed: Those are just two lines.. :(

bears/vcs/git/GitCommitBear.py
```diff
@@ -204,3 +213,17 @@ def check_body(self, body,
...
                                'Commit body lines should not exceed {} '
                                'characters.'.format(body_line_length))
+
+    def check_duplicate_shortlogs(self, shortlogs: list):
+        shortlogs_set = set(shortlogs)
+        if len(shortlogs_set) == len(shortlogs):
+            return False
+        else:
+            for log in shortlogs_set:
+                if log in shortlogs:
```
aptrishu: `if log in shortlogs:`
why is this check necessary ?
A set can not have such elements which are not present in the parent list from which it is formed.

Nosferatul:
@meetmangukiya  try to debug your tests using `pdb`, search on google for this.

meetmangukiya:
> @meetmangukiya try to debug your tests using pdb, search on google for this.

@Nosferatul thanks :+1: seems interesting because I never used a debugger earlier :smile:

meetmangukiya:
@underyx take a look, does it meet the requirement that you though of ? Do you think we need some kind of modification?

bears/vcs/git/GitCommitBear.py
```diff
@@ -61,25 +61,31 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
                                            allowed or not.
```
sils: no option for this?

sils:
this issue needs some design thinking. We can't just implement it like this.

bears/vcs/git/GitCommitBear.py
```diff
@@ -61,25 +61,31 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
...
         """
         with change_directory(self.get_config_dir() or os.getcwd()):
-            stdout, stderr = run_shell_command('git log -1 --pretty=%B')
-
-        if stderr:
-            self.err('git:', repr(stderr))
+            stdout_head, stderr_head = run_shell_command(
+                "git log -1 --pretty=%B")
+            stdout_shortlogs, stderr_shortlogs = run_shell_command(
+                "git log --pretty=%s")
```
meetmangukiya: Oh, I assumed that just splitting at new lines wouldn't be a problem even if it were say 10k ...

I'll try to come up with something better

aptrishu:
Marking it wip. You need to reinvent the idea.

bears/vcs/git/GitCommitBear.py
```diff
@@ -64,7 +66,8 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
             stdout_head, stderr_head = run_shell_command(
                 "git log -1 --pretty=%B")
             stdout_shortlogs, stderr_shortlogs = run_shell_command(
-                "git log --pretty=%s")
+                "git log -{} --pretty=%s".format(
```
jayvdb: please use `--max-count=`

meetmangukiya: done

meetmangukiya:
`git coala` ? Passing `coala` as argument to `git`  ? :thinking:

meetmangukiya:
And I guess that means close this PR? What about another related issue about checking commits passed on by CI, should that be closed too?

jayvdb:
> git coala ? Passing coala as argument to git ?

As I understand it, interacting with multiple commits should be achieved by creating a tool which uses coala.  `git coala` would be one approach, and would work sort-of line [`git review`](https://github.com/openstack-infra/git-review), checking each commit before submitting it to be reviewed. And if someone doesnt use `git coala`, they will submit and any problems would be caught by `gitmate`.

I havent closed the PR, as you may want to discuss it further.

meetmangukiya:
I thought that was like raising a PR at git to get coala command. But I guess we can mimic it without any permission. I think we should let gitmate handle this issue :+1:

PR #228, opened 2016-03-12
TSLintBear: Fix column and end_column arguments

arafsheikh:

PR #1042, opened 2016-11-22
requirements: Upgrade coala

sils:


Mixih:
ack 0ee7565

requirements.txt
```diff
@@ -1,33 +1,33 @@
...
-eradicate==0.1.*
-autoflake==0.6.*
-restructuredtext_lint==0.14.*
-proselint==0.3.*
-cpplint==1.*
-yapf==0.11.*
-isort==4.*
-html-linter==0.3.*
+coala~=0.9
+munkres3~=1
```
Mixih: pip install is failing for `munkres3~=1`

sils: I know, on it. Also more commits coming

sils:
marking wip

underyx:
unack https://github.com/coala/coala-bears/pull/1042/commits/4fc6a64075803380e0d70e438e2e8caaa2f9e816

compatability -> compatibility

sils:
YapfBear problem

Mixih:
From https://github.com/google/yapf CHANGELOG, here are some things they changed since v11

## [0.14.0] 2016-11-21
### Fixed
- "not in" and "is not" should be subtyped as binary operators.
- A non-Node dictionary value may have a comment before it. In those cases, we
  want to avoid encompassing only the comment in pseudo parens. So we include
  the actual value as well.
- Adjust calculation so that pseudo-parentheses don't count towards the total
  line length.
- Don't count a dictionary entry as not fitting on a single line in a
  dictionary.
- Don't count pseudo-parentheses in the length of the line.

## [0.13.2] 2016-10-22
### Fixed
- REGRESSION: A comment may have a prefix with newlines in it. When calculating
  the prefix indent, we cannot take the newlines into account. Otherwise, the
  comment will be misplaced causing the code to fail.

## [0.13.1] 2016-10-17
### Fixed
- Correct emitting a diff that was accidentally removed.

## [0.13.0] 2016-10-16
### Fixed
- Functions or classes with comments before them were reformatting the comments
  even if the code was supposed to be ignored by the formatter. We now don't
  adjust the whitespace before a function's comment if the comment is a
  "disabled" line. We also don't count "# yapf: {disable|enable}" as a disabled
  line, which seems logical.
- It's not really more readable to split before a dictionary value if it's part
  of a dictionary comprehension.
- Enforce two blank lines after a function or class definition, even before a
  comment. (But not between a decorator and a comment.) This is related to PEP8
  error E305.
- Remove O(n^2) algorithm from the line disabling logic.

## [0.12.2] 2016-10-09
### Fixed
- If `style.SetGlobalStyle(<create pre-defined style>)` was called and then
  `yapf_api.FormatCode` was called, the style set by the first call would be
  lost, because it would return the style created by `DEFAULT_STYLE_FACTORY`,
  which is set to PEP8 by default. Fix this by making the first call set which
  factory we call as the "default" style.
- Don't force a split before non-function call arguments.
- A dictionary being used as an argument to a function call and which can exist
  on a single line shouldn't be split.
- Don't rely upon the original line break to determine if we should split
  before the elements in a container. Especially split if there's a comment in
  the container.
- Don't add spaces between star and args in a lambda expression.
- If a nested data structure terminates in a comma, then split before the first
  element, but only if there's more than one element in the list.

## [0.12.1] 2016-10-02
### Changed
- Dictionary values will be placed on the same line as the key if *all* of the
  elements in the dictionary can be placed on one line. Otherwise, the
  dictionary values will be placed on the next line.

### Fixed
- Prefer to split before a terminating r-paren in an argument list if the line
  would otherwise go over the column limit.
- Split before the first key in a dictionary if the dictionary cannot fit on a
  single line.
- Don't count "pylint" comments when determining if the line goes over the
  column limit.
- Don't count the argument list of a lambda as a named assign in a function
  call.

## [0.12.0] 2016-09-25

## Changed
- Turn "verification" off by default for external APIs.
- If a function call in an argument list won't fit on the current line but will
  fit on a line by itself, then split before the call so that it won't be split
  up unnecessarily.

## Fixed
- Don't add space after power operator if the next operator's a unary operator.

## [0.11.1] 2016-08-17
### Changed
- Issue #228: Return exit code 0 on success, regardless of whether files were
  changed.  (Previously, 0 meant success with no files
  modified, and 2 meant success with at least one file modified.)

### Fixed
- Enforce splitting each element in a dictionary if comma terminated.
- It's okay to split in the middle of a dotted name if the whole expression is
  going to go over the column limit.
- Asynchronous functions were going missing if they were preceded by a comment
  (a what? exactly). The asynchronous function processing wasn't taking the
  comment into account and thus skipping the whole function.
- The splitting of arguments when comma terminated had a conflict. The split
  penalty of the closing bracket was set to the maximum, but it shouldn't be if
  the closing bracket is preceded by a comma.

sils:
@Makman2 had a newline issue on windows, tests seem to only fail on windows... did we fix that @Makman2?

Mixih:
test function `test_allow_multiline_lambdas` is failing.... still debugging

Mixih:
ack d0b08e 0ee756 bd753b

underyx:
unack https://github.com/coala/coala-bears/pull/1042/commits/bd753b325f9a835bd86fe1d4b8e77b73cdc533c9

compatability -> compatibility

sils:
unack bd753b3

sils:
indeed, will fix that little typo

sils:
first fixing vulturebear though based on @Adrianzatreanu s work

sils:
reack https://github.com/coala/coala-bears/pull/1042/commits/a6bcb34d3822c076f98ba1381f8dd264a3854c03

underyx:
ack https://github.com/coala/coala-bears/pull/1042/commits/100fb1ace1e67caaeca850124be094d4f8e60d81

Mixih:
ack 603454

bears/python/PyUnusedCodeBear.py
```diff
@@ -8,7 +8,7 @@
 
 class PyUnusedCodeBear(LocalBear):
     LANGUAGES = {'Python', 'Python 2', 'Python 3'}
-    REQUIREMENTS = {PipRequirement('autoflake', '0.6.*')}
+    REQUIREMENTS = {PipRequirement('autoflake', '0.6XML')}
```
jayvdb: as best I can tell, this is wrong.

Mixih:
unack f432ce

Mixih:
ack 
f31db6 - requirements: Upgrade restructuredtext_lint 
d17c89 - requirements: Upgrade proselint 
b400cb - requirements: Upgrade cppclean

Mixih:
ack acdfd7

Mixih:
ack e01f2d - YapfBear: Use FormatCode 
Code looks good. My localenv is acting up so couldn't do manual lint tests.

bears/c_languages/codeclone_detection/ClangFunctionDifferenceBear.py
```diff
@@ -83,7 +84,7 @@ def get_difference(function_pair,
 class ClangFunctionDifferenceBear(GlobalBear):
     check_prerequisites = classmethod(clang_available)
     LANGUAGES = ClangBear.LANGUAGES
-    REQUIREMENTS = ClangBear.REQUIREMENTS
+    REQUIREMENTS = ClangBear.REQUIREMENTS | {PipRequirement('munkres', '1.0')}
```
Mixih: Why do we need munkres? I'm not familiar with this bear.

Mixih: Why do we need munkres for this bear? I'm not familiar with this bear.

jayvdb: I believe this should be `munkres3` , per requirements;

```
$ git grep munkres
...
requirements.txt:munkres3==1.*
```

https://pypi.python.org/pypi/munkres (1.0.8) vs https://pypi.python.org/pypi/munkres3 (1.0.5.5)

Mixih:
unack d17c89 - should be `munkres3` not `munkres`

Mixih:
ack 30d328

Mixih:
reack d17c895

Mixih:
unack c57a190 - should be `munkres3` not `munkres`

Mixih:
ack d17c895

Mixih:
@sils This is NOT ready for merge. Somehow gitmate bugged out and went green. Everything has been reviewed. Two commits need clarification, minor editing.

jayvdb:
I have the vulture fixes as https://github.com/jayvdb/coala-bears/tree/gh_817 , split into two commits, if you're interested in getting those in.

sils:
> This is NOT ready for merge. Somehow gitmate bugged out and went green.

@Mixih known issue of gitmate, this has just not been implemented yet ;)

sils:
reack
c6dcdcb VultureBear: Convert to global bear
CHANGED AS PER REVIEW Unify bear requirement style
ab23980 requirements: Upgrade safety
934466c requirements: Upgrade bandit
f407faf requirements: Upgrade cppclean
0d9cb0e requirements: Upgrade proselint
d776777 requirements: Upgrade restructuredtext_lint
CHANGED AS PER REVIEW ClangFunctionDifferenceBear: Add munkres3 req
CONFLICT DURING REBASE requirements: Use compatibility operator
b7707a9 Use latest coala release

jayvdb:
ack 6dee81a 7e9c19be3

jayvdb:
ack 6c4f2674c47

jayvdb:
ack 6a051d1bd 7b48519e
(the appsdir thing should have been a separate commit, but .. meh;-)

sils:
reack 7b48519

sils:
@rultor merge

requirements.txt
```diff
@@ -1,33 +1,32 @@
...
+yamllint~=1.5
+cppclean~=0.9
+pydocstyle~=1.1
+cmakelint~=1.3
+vim-vint~=0.3
+nltk~=3.2
+pyyaml~=3.12
+vulture~=0.10
+bandit~=1.1
+nbformat>=4.1
```
jayvdb: See https://github.com/coala/coala-bears/issues/1212

PR #1066, opened 2016-11-27
bears/general: Add CheckmanifestBear

aptrishu:
Adds CheckmanifestBear for https://pypi.python.org/pypi/check-manifest
to check MANIFEST.in for completeness.

Closes https://github.com/coala/coala-bears/issues/798


jayvdb:
Work continues at https://github.com/coala/coala-bears/issues/1067

PR #1094, opened 2016-12-05
CheckstyleBear.py : modified create_argument() return

SarthakNijhawan:
The commit demonstrates issue regarding the calling of
checkstyle_jar_file var as a static function , earlier
returned as static var.

Fixes https://github.com/coala/coala-bears/issues/1084

### DIFF
--- a/bears/c_languages/CPPLintBear.py
+++ b/bears/c_languages/CPPLintBear.py
@@ -3,6 +3,12 @@ from coalib.bears.requirements.PipRequirement import PipRequirement
 from coalib.settings.Setting import typed_list
 
 
+def check_invalid_configuration(use_spaces, indent_size):
+    if (not use_spaces or indent_size != 2):
+        return False
+    return True
+
+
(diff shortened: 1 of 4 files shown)


bears/c_languages/CPPLintBear.py
```diff
@@ -26,14 +32,22 @@ class CPPLintBear:
     def create_arguments(filename, file, config_file,
                          max_line_length: int=79,
                          cpplint_ignore: typed_list(str)=(),
-                         cpplint_include: typed_list(str)=()):
+                         cpplint_include: typed_list(str)=(),
+                         use_spaces: bool=True,
+                         indent_size: int=2):
```
meetmangukiya: We use a default indent of `4` in space consistency IIRC, maybe we should go with `4` here as well

jayvdb:
This is no where near the problem, and on the issue you will see that we dont actually know what the problem is.  More analysis needed before coding.

PR #1100, opened 2016-12-06
AnnotationBear: Allow missing language setting

jayvdb:
A missing language setting is not any different from
a language name which has a Language definition that
does not contain the necessary information to correctly
parse the file.

Use a special 'UnknownLanguage' in this scenario.

Also upgrades AnnotationBear from LanguageDefinition
to the new Language.

Fixes https://github.com/coala/coala-bears/issues/1012
Closes https://github.com/coala/coala-bears/issues/1095


bears/general/AnnotationBear.py
```diff
@@ -7,12 +7,17 @@
 from coala_utils.string_processing.Core import unescaped_search_for
 
 
+@Language
+class UnknownLanguage:
+    pass
```
sils: this shouldn't be here but rather with the others

jayvdb: I assume you mean in coala.  Do you also want it appearing in `coalib/bearlib/languages/__init__.py` /  `LanguageUberMeta.all` ?

If we include it here, at least temporarily, we don't need to do a release of `coala` to fix this bug.  It could just be part of coala-bears 0.9.2 .

sils:
I don't see a problem in doing a bugfix release for coala

jayvdb:
Do you also want it appearing in `coalib/bearlib/languages/__init__.py` / `LanguageUberMeta.all` ?

sils:
yeah

jayvdb:
@sils, so it has been merged, and now you need to do a coala 0.9.1 release.

sils:
Hm this renders the bears using the AnnotationBear in most cases useless by default, doesn't it? KeywordBear is an exception because it has only a soft dependency on AnnotationBear but usually it's a hard one. Maybe we should rather allow bears to spec a soft dependency and if the settings aren't given we just don't run the dependent bear and inform the depending bear about it?

sils:
I think this is not the solution. Marking WIP while waiting for response.

jayvdb:
I'll add some more tests to address that query.

jayvdb:
@sils, I have added `IndentationBearTest.MissingLanguageTest` to show you what happens if `language` is omitted.
The "UnknownLanguage" doesnt have the information that `IndentationBear` needs.  We would need a better way of reporting this type of error, but the fundamental assumption should be that not every language definition will support every feature.

ergo, every bear using languages will need to correctly handle the worst case: i.e. `UnknownLanguage` which has no attributes.

Gotta say I am a bit miffed that this fix for a breaking change in 0.9 was completed four weeks ago, and you are only now raising design issues.  Earlier you told me to move `UnknownLanguage` into `coala`, and thus had to wait for the coala 0.9.1 release.  Now there is an objection to the concept.

sils:
I'm sorry, I didn't think this through properly in the first place. I can assure you that I did not mean to waste your time. Thanks for speaking this out openly and please continue to do so.

I'll try to more consistently think through my design review before going to implementation specific suggestions in the future.

> Earlier you told me to move UnknownLanguage into coala

At the danger of repeating something you take for granted: Please note that you don't have to follow what I say blindly. As you and I know I have my flaws, if you disagree thats ok and even if we don't come to an agreement we can pull in other maintainers for a vote. I do respect your opinion a lot.

> ergo, every bear using languages will need to correctly handle the worst case: i.e. UnknownLanguage which has no attributes.

That is true indeed, what I'm worried about is that users of e.g. QuotesBear or others won't figure out why it's not working when they just have to add a `language` setting because it's nowhere documented. Given this example:

[QuotesBear](https://github.com/coala/bear-docs/blob/master/docs/QuotesBear.rst#settings) doesn't list the `language` setting because our inheritance doesn't look up inherited settings yet (I think we have an issue somewhere but I couldn't find it right now :/). By default coala will ask the user for the `language` right now in the interactive mode and fail in the CI mode. This change is indeed breaking the API of the KeywordBear (from 0.8 to 0.9) which should be allowed since we're pre 1.0. This (current) behaviour means that the user gets the best possible behaviour (indeed the worst case still applies but the average case gets better) but he has to give the language setting.

This change means that if I run the QuotesBear without knowing coala already it won't do anything and I have no clue on how to make it work.

Proposed workaround: for now we can add a non optional `language` setting to QuotesBear and all others that are completely useless without the `language` given. We should fix that bug related to dependent parameters and add "soft bear dependencies" on long term so we can deduplicate the parameters in the end.

Does that sound sane?

tests/general/IndentationBearTest.py
```diff
@@ -19,25 +19,72 @@ class TestLanguage:
...
+    comment_delimiter = {}
+
+
+class IndentationBearTestBase(unittest.TestCase):
 
     def setUp(self):
         self.section = Section('')
-        self.section.append(Setting('language', 'TestLanguage'))
+        if self.language:
+            self.section.append(Setting('language', self.language))
```
arush0311: where is self.language initialized?

sils:
@jayvdb I'm waiting for your response here, pinged you twice or so on gitter about this, I think we wanna have this done for this release? Let me/us know if you need any help.

PR #1107, opened 2016-12-08
XMLBear.py: Add severity to XMLBear

yukiisbored:
This commit adds severity which are provided from the underlying
linter's output

Fixes https://github.com/coala/coala-bears/issues/1099


jayvdb:
That didnt work.

PR #1111, opened 2016-12-09
Remove libclang-py3 from requirements.txt

jayvdb:
This dependency is already managed by the coala package.
Due to continuous integration coala and coala-bears installing the
latest master of the other package, it can not be modified if it
appears in both otherwise a pkg_resources.ContextualVersionConflict
is raised when it is used.

Fixes https://github.com/coala/coala-bears/issues/1110

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,8 @@
 # Use >= for development versions so that source builds always work
 coala>=0.9
+# Dependencies inherited from coala
+# libclang-py3
+# coala_utils
 munkres3~=1.0
 pylint~=1.6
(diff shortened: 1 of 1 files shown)


jayvdb:
This is needed for https://github.com/coala/coala/issues/3187 , which is sort of needed for Travis CI builds to work; see https://travis-ci.org/jayvdb/coala/builds/182586208 (i.e. not have very ugly hacks there).

aptrishu:
ack b1b66cc

aptrishu:
@rultor merge

PR #1116, opened 2016-12-11
Ignore multiple urls

dracarys09:
Ensures that multiple urls can be ignored according to
links_to_ignore setting (comma separated urls).

Closes https://github.com/coala/coala-bears/issues/1106

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -9,6 +9,8 @@ from coalib.bears.requirements.PipRequirement import PipRequirement
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.results.Result import Result
 from coalib.bearlib import deprecate_settings
+from coalib.settings.Setting import typed_list
+from coalib.parsing.Globbing import fnmatch
 
 
 class InvalidLinkBear(LocalBear):
@@ -46,7 +48,7 @@ class InvalidLinkBear(LocalBear):
         return splitted_schema
 
     @staticmethod
-    def find_links_in_file(file, timeout, link_ignore_regex):
+    def find_links_in_file(file, timeout, link_ignore_regex, link_ignore_list):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
             r"""
@@ -79,7 +81,8 @@ class InvalidLinkBear(LocalBear):
             match = regex.search(line)
             if match:
                 link = match.group()
-                if not link_ignore_regex.search(link):
+                if not (link_ignore_regex.search(link) or
+                        fnmatch(link, link_ignore_list)):
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
                     code = InvalidLinkBear.get_status_code(link, timeout)
@@ -89,6 +92,7 @@ class InvalidLinkBear(LocalBear):
     def run(self, filename, file,
             timeout: int=DEFAULT_TIMEOUT,
             link_ignore_regex: str='([.\/]example\.com|\{|\$)',
+            link_ignore_list: typed_list(str)='',
             follow_redirects: bool=False):
         """
         Find links in any text file and check if they are valid.
@@ -106,10 +110,11 @@ class InvalidLinkBear(LocalBear):
 
         :param timeout:          Request timeout period.
         :param link_ignore_regex:     A regex for urls to ignore.
+        :param link_ignore_list: Comma separated url globs to ignore
         :param follow_redirects: Set to true to autocorrect redirects.
         """
         for line_number, link, code in InvalidLinkBear.find_links_in_file(
-                file, timeout, link_ignore_regex):
+                file, timeout, link_ignore_regex, link_ignore_list):
             if code is None:
                 yield Result.from_values(
                     origin=self,
--- a/tests/general/InvalidLinkBearTest.py
+++ b/tests/general/InvalidLinkBearTest.py
@@ -203,3 +203,29 @@ class InvalidLinkBearTest(unittest.TestCase):
 
         for line in brokenlink_at_hash.splitlines():
             self.assertResult(invalid_file=[line])
+
+    def test_links_to_ignore(self):
+        valid_file = """http://httpbin.org/status/200
+        http://httpbin.org/status/201
+        http://coalaisthebest.com/
+        http://httpbin.org/status/404
+        http://httpbin.org/status/410
+        http://httpbin.org/status/500
+        http://httpbin.org/status/503
+        http://www.notexample.com/404
+        http://exampe.com/404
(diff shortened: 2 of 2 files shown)


bears/general/InvalidLinkBear.py
```diff
@@ -79,14 +81,14 @@ def find_links_in_file(file, timeout, link_ignore_regex):
             match = regex.search(line)
             if match:
                 link = match.group()
-                if not link_ignore_regex.search(link):
+                if not (link_ignore_regex.search(link) or (fnmatch(link, links_to_ignore))):
```
nemaniarjun: Go through `https://coala.io/commit`.
Always check the things in gitmate-bot's comment.

bears/general/InvalidLinkBear.py
```diff
@@ -79,14 +81,15 @@ def find_links_in_file(file, timeout, link_ignore_regex):
             match = regex.search(line)
             if match:
                 link = match.group()
-                if not link_ignore_regex.search(link):
+                if not (link_ignore_regex.search(link) or \
```
Makman2: obsolete `\`

dracarys09: thanks for pointing it out... I modified it... still struggling with appveyor though...

sils:
can you add a testcase please?

Mixih:
WIP because no testcase/test failures

dracarys09:
@sils  @Mixih I've added a test case for this feature. Now all previous test cases are passing as well.

tests/go/GoReturnsBearTest.py
```diff
@@ -29,6 +29,5 @@
 
 GoReturnsBearTest = verify_local_bear(
     GoReturnsBear,
-    (good_file1,
-     good_file2,),
+    (good_file1, good_file2,),
```
jayvdb: what is this doing in your patch? ;-)

dracarys09: This was the first newcomer issue I solved...and I did all the stuff in master branch...since then this commit is there in this PR as well...any suggestions how can I remove this commit from this PR ?

dracarys09: @jayvdb removed that commit from this patch...

dracarys09:
@jayvdb I've squashed my commits as you suggested.

bears/general/InvalidLinkBear.py
```diff
@@ -79,14 +81,15 @@ def find_links_in_file(file, timeout, link_ignore_regex):
...
+                if not (link_ignore_regex.search(link) or
+                        fnmatch(link, links_to_ignore)):
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
                     code = InvalidLinkBear.get_status_code(link, timeout)
                     yield line_number + 1, link, code
 
     @deprecate_settings(link_ignore_regex='ignore_regex')
-    def run(self, filename, file,
+    def run(self, filename, file, links_to_ignore: typed_list(str)='',
```
jayvdb: `links_to_ignore` should be placed after `link_ignore_regex`, to be consistent with the order in the documentation, and also to avoid modifying this line unnecessarily.

Also, can the default be changed to `None`.

dracarys09: @jayvdb I've made changes as you suggested (Changed ```links_to_ignore``` to ```link_ignore_list```). However I think using ```''``` instead of ```None``` is correct because ```fnmatch``` function expects a glob string which can be ```''``` by default.

bears/general/InvalidLinkBear.py
```diff
@@ -106,10 +110,11 @@ def run(self, filename, file,
 
         :param timeout:          Request timeout period.
         :param link_ignore_regex:     A regex for urls to ignore.
+        :param link_ignore_list: comma separated urls to ignore
```
sils: url *globs*, also start it upper case like the others

bears/general/InvalidLinkBear.py
```diff
@@ -79,7 +81,8 @@ def find_links_in_file(file, timeout, link_ignore_regex):
             match = regex.search(line)
             if match:
                 link = match.group()
-                if not link_ignore_regex.search(link):
+                if not (link_ignore_regex.search(link) or
```
sils: the or should be and not shouldn't it?

tests/general/InvalidLinkBearTest.py
```diff
@@ -203,3 +203,27 @@ def test_pip_vcs_url(self):
...
+                           'localhost:8080',
+                           'http://httpbin.org/status/500',
+                           'http://www.notexample.com/404',
+                           'http://example.com/404',
+                           'http://example.co.in/404'
+                          ]
+
+        self.assertResult(valid_file=valid_file,
+                          invalid_file=invalid_file,
+                          settings={'link_ignore_list': link_ignore_list})
```
sils: this test doesn't test properly. Read the other tests please and understand how this is done.

bears/general/InvalidLinkBear.py
dracarys09: the 'or' as well as 'not' both should be there because we want to display info about invalid links which are 'not' present in ```link_ignore_regex``` or 'not' present in ```link_ignore_list``` that is why outer ```not``` and inner ```or``` is present.

bears/general/InvalidLinkBear.py
dracarys09: I have changed it and this will be reflected in the next commit.

tests/general/InvalidLinkBearTest.py
dracarys09: @sils yeah... thanks for pointing out... I made changes in the testcase...
Now ```valid_file``` contains valid as well as invalid urls and ```link_ignore_list``` contains a list of invalid urls... the intended behaviour should be to ignore invalid urls and produce no warnings in this case. This is what the test case is checking now.

tests/general/InvalidLinkBearTest.py
```diff
@@ -203,3 +203,29 @@ def test_pip_vcs_url(self):
...
+        http://www.notexample.com/404
+        http://exampe.com/404
+        http://example.co.in/404""".splitlines()
+
+        link_ignore_list = [
+                           'http://coalaisthebest.com/',
+                           'http://httpbin.org/status/404',
+                           'http://httpbin.org/status/410',
+                           'http://httpbin.org/status/500',
+                           'http://httpbin.org/status/503',
```
meetmangukiya: Can replace with a glob... to test if the **globbing** is working.
`http://httpbin.org/status/4[0-9][0-9]`
Similarly for `50x`..

Other then that **LGTM** :+1:

meetmangukiya:
ack a14f6c69

sils:
your commit was cherry picked into https://github.com/coala/coala-bears/pull/1173 and will be merged for you

PR #1130, opened 2016-12-14
CircleCI: Do not install precise hlint

jayvdb:
precise hlint is 1.8.24, which isnt usable as --json was added
in version 1.9.1.

The build script currently installs precise hlint, and then replaces
it with a seemingly custom .deb of 1.9.26.

The trusty hlint is 1.8.54, however xenial hlint is 1.9.26,
so it can be installed using apt-get on the trusty image.

Also move the .deb save location to ~/.apt-cache, with the
other cached .debs.

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -11,7 +11,7 @@ esac
 # apt-get commands
 export DEBIAN_FRONTEND=noninteractive
 
-deps="espeak libclang1-3.4 indent mono-mcs chktex hlint r-base julia golang luarocks verilator cppcheck flawfinder"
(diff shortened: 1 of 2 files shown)


jayvdb:
Just noticed, currently ghc isnt being upgraded at all.  This is presumably upgrading it to 7.4.1-1ubuntu2_amd64.deb , which will negate any benefits of not installing hlint twice.

jayvdb:
This is blocking the addition of Travis builds.

Mixih:
ack  	87e865f

Mixih:
ack 87e865f

jayvdb:
@rultor merge

PR #1132, opened 2016-12-14
CircleCI: Move R cache to be same as Travis

jayvdb:
Travis CI has native R caching in ~/R/Library.
coala-bears has custom R caching in ~/.RLibrary.
Move the custom R caching to be consistent with Travis CI,
while maintaining the cache for existing branches.

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -66,8 +66,13 @@ sudo rm -rf /opt/alex # Delete ghc-alex as it clashes with npm deps
 npm install
 
 # R commands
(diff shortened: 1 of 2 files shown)


jayvdb:
Turns out the R caching is only possible if it is the language chosen, if [" arbitrary directories"](https://docs.travis-ci.com/user/caching/#Enabling-multiple-caching-features) are used, which we do.

jayvdb:
Not worth doing atm, and unlikely that Travis will fix this any time soon.

PR #1141, opened 2016-12-15
.coafile: add PycodestyleBear

andreidinu:
Add PycodestyleBear in the autopep8 section
of coafile.

Closes https://github.com/coala/coala-bears/issues/1122

### DIFF
--- a/bears/natural_language/WriteGoodLintBear.py
+++ b/bears/natural_language/WriteGoodLintBear.py
@@ -71,7 +71,7 @@ class WriteGoodLintBear:
             'allow_extra_words': '--tooWordy',
             'allow_cliche_phrases': '--cliches'
         }
-        l = locals()
+        arg_values = locals()
         args = tuple(arg for key, arg in arg_map.items()
-                     if not l[key])
+                     if not arg_values[key])
         return args + (filename,)
--- a/bears/ruby/RuboCopBear.py
+++ b/bears/ruby/RuboCopBear.py
@@ -176,62 +176,81 @@ class RuboCopBear:
         """
         naming_convention = {'camel': 'camelCase', 'snake': 'snake_case'}
         options = {
-                  'Style/AccessModifierIndentation':
-                      {'EnforcedStyle': access_modifier_indentation},
-                  'Style/Alias':
-                      {'EnforcedStyle': preferred_alias},
-                  'Style/AlignHash':
-                      {'EnforcedHashRocketStyle': align_hash_rocket_by,
-                       'EnforcedColonStyle': align_colon_by,
-                       'EnforcedLastArgumentHashStyle':
-                           inspect_last_argument_hash},
-                  'Style/AlignParameters':
-                      {'EnforcedStyle': align_parameters},
-                  'Style/ClassCheck':
-                      {'EnforcedStyle': class_check},
-                  'Style/CommentAnnotation':
-                      {'Keywords': comment_keywords},
-                  'Style/GuardClause':
-                      {'MinBodyLength': min_if_unless_guard},
-                  'Style/IndentationWidth':
-                      {'Width': indent_size},
-                  'Style/MethodName':
-                      {'EnforcedStyle': naming_convention.get(
-                         method_naming_convention, method_naming_convention)},
-                  'Style/StringLiterals':
-                      {'EnforcedStyle': string_literals},
-                  'Style/VariableName':
-                      {'EnforcedStyle': naming_convention.get(
-                         variable_naming_convention,
-                         variable_naming_convention)},
-                  'Metrics/ClassLength':
-                      {'Max': max_class_length,
-                       'CountComments': class_length_count_comments},
-                  'Metrics/ModuleLength':
-                      {'CountComments': module_length_count_comments,
-                       'Max': max_module_length},
-                  'Metrics/CyclomaticComplexity':
-                      {'Max': cyclomatic_complexity},
-                  'Metrics/LineLength':
-                      {'Max': max_line_length,
-                       'AllowHeredoc': line_length_allow_here_doc,
-                       'AllowURI': line_length_allow_uri},
-                  'Metrics/MethodLength':
-                      {'CountComments':  method_length_count_comments,
-                       'Max': max_method_length},
-                  'Metrics/ParameterLists':
-                      {'Max': max_parameters,
-                       'CountKeywordArgs': count_keyword_args},
-                  'Lint/UnusedBlockArgument':
-                      {'IgnoreEmptyBlocks': ignore_unused_block_args_if_empty,
-                       'AllowUnusedKeywordArguments':
-                           allow_unused_block_keyword_arguments},
-                  'Lint/UnusedMethodArgument':
-                      {'AllowUnusedKeywordArguments':
-                           allow_unused_method_keyword_args,
-                       'IgnoreEmptyMethods':
-                           ignore_unused_method_args_if_empty},
-                  }
+            'Style/AccessModifierIndentation': {
+                'EnforcedStyle': access_modifier_indentation
+            },
+            'Style/Alias': {
+                'EnforcedStyle': preferred_alias
+            },
+            'Style/AlignHash': {
+                'EnforcedHashRocketStyle': align_hash_rocket_by,
+                'EnforcedColonStyle': align_colon_by,
+                'EnforcedLastArgumentHashStyle': inspect_last_argument_hash
+            },
+            'Style/AlignParameters': {
+                'EnforcedStyle': align_parameters
+            },
+            'Style/ClassCheck': {
+                'EnforcedStyle': class_check
+            },
+            'Style/CommentAnnotation': {
+                'Keywords': comment_keywords
+            },
+            'Style/GuardClause': {
+                'MinBodyLength': min_if_unless_guard
+            },
+            'Style/IndentationWidth': {
+                'Width': indent_size
+            },
+            'Style/MethodName': {
+                'EnforcedStyle': naming_convention.get(
+                                 method_naming_convention,
+                                 method_naming_convention)
+            },
+            'Style/StringLiterals': {
+                 'EnforcedStyle': string_literals
+            },
+            'Style/VariableName': {
+                'EnforcedStyle': naming_convention.get(
+                                 variable_naming_convention,
+                                 variable_naming_convention)
+            },
+            'Metrics/ClassLength': {
+                'Max': max_class_length,
+                'CountComments': class_length_count_comments
+            },
+            'Metrics/ModuleLength': {
+                'CountComments': module_length_count_comments,
+                'Max': max_module_length
+            },
+            'Metrics/CyclomaticComplexity': {
+                'Max': cyclomatic_complexity
+            },
+            'Metrics/LineLength': {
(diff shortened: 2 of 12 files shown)


jayvdb:
Hi @andreidinu , you also need to fix the problems identified.  Discuss any complicated problems on the issue #1122 and on gitter

setup.py
```diff
@@ -1,18 +1,17 @@
 #!/usr/bin/env python3
-
-# Start ignoring PyImportSortBear as imports below may yield syntax errors
-from bears import assert_supported_version
```
Abhisek-: In [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) it's mentioned that your import order should be

* standard library imports
* related third party imports
* local application/library specific imports

and also it would be even cleaner if you group the `from` to one block and `import` to other

andreidinu: Can you explain more please? In my new code `import` are first and `from` comes after.

Abhisek-: I would do something like the following

    from subprocess import call
    import locale
    import sys

    from setuptools import find_packages, setup
    from setuptools.command.test import test as TestCommand
    import setuptools.command.build_py

    from bears import assert_supported_version
    from bears import Constants

As `subprocess, sys, locale` are `python's` own library they come first. then `setuptools`, which is third party library and that is nether `python's` nor local to our repo and then comes our local imports.

Also do not forget to check the comments appear in the right place

setup.py
```diff
@@ -1,18 +1,19 @@
 #!/usr/bin/env python3
+from subprocess import call
+import locale
+import sys
 
+from setuptools import find_packages, setup
+from setuptools.command.test import test as TestCommand
+import setuptools.command.build_py
+
+from bears import assert_supported_version
```
Abhisek-: Why two times importing the same function.

bears/ruby/RuboCopBear.py
```diff
@@ -176,61 +176,61 @@ class members.
...
+            'Lint/UnusedBlockArgument':
+                {'IgnoreEmptyBlocks': ignore_unused_block_args_if_empty,
+                    'AllowUnusedKeywordArguments':
+                    allow_unused_block_keyword_arguments},
+            'Lint/UnusedMethodArgument':
+                {'AllowUnusedKeywordArguments':
+                    allow_unused_method_keyword_args,
+                    'IgnoreEmptyMethods':
+                    ignore_unused_method_args_if_empty},
                   }
```
sils: indentation of ths one seems wrong

setup.py
```diff
@@ -1,18 +1,19 @@
...
+from bears import Constants
```
sils: MOVING THOSE IMPORTS UP IS DANGEOUS! Rather put an ignore statement there.

tests/natural_language/WriteGoodLintBearTest.py
```diff
@@ -141,8 +141,9 @@
                                                       invalid_files=(
                                                           illusion_file,),
                                                       settings={
-                                                        'allow_repeated_words':
-                                                            False})
+                                                          'allow_' +
+                                                          'repeated_words':
```
sils: this seriously doesn't fit in one line? I find this formatting worse than before

setup.py
Abhisek-: It's dangerous because the imports may fail(because of version incompatibility) or any other reason ?

sils:
yes, those imports will actually import almost all coala code and we have
to check the version before because otherwise it'll error out

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

bears/scss/SCSSLintBear.py
```diff
@@ -268,9 +268,9 @@ def generate_config(filename, file,
                                    'width': indent_size},
                    'LeadingZero': {
                                    'enabled': True,
-                                   'style': (
-                                       'exclude_zero' if exclude_leading_zero
-                                        else 'include_zero')
```
jayvdb: this should be fixed by removing one space from before the `else`.  No need to modify the line above it.

setup.py
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
 
-# Start ignoring PyImportSortBear as imports below may yield syntax errors
+# Start ignoring PyImportSortBear and PycodestyleBear* as imports below may
+# yield syntax errors
 from bears import assert_supported_version
 assert_supported_version()
```
jayvdb: this should be moved to after the imports.
then all the problems disappear

bears/natural_language/WriteGoodLintBear.py
```diff
@@ -71,7 +71,7 @@ def create_arguments(filename, file, config_file,
             'allow_extra_words': '--tooWordy',
             'allow_cliche_phrases': '--cliches'
         }
-        l = locals()
+        L = locals()
```
jayvdb: this bypasses the error, but doesnt fix the problem.  The variable needs a meaningful name.  `arg_values` is my suggestion.

jayvdb:
Then IMO the correct way to do that is have an import that has side effects.

i.e. create a new module `bears.check_supported_version` which invokes `bears.assert_supported_version()`.

Then setup.py only needs to import `bears.check_supported_version` and the assert is done automatically.
That import can even be in a `try:` block, as pyflakes and others expect and allow code to appear in the `except:` clause.

I can do this in about an hour.

setup.py
andreidinu: @Abhisek- you're right. If we keep it like this in case of an older version of python the code won't break. It will only show a warning.

jayvdb: Actually, there are a lot of assumptions here.
Firstly, the test wasnt even working: https://github.com/coala/coala/issues/3310
Secondly, I've checked the package `__init__.py` for coala and -bears, and neither drags in lots of dependencies, so we should be ok to import the package early in the process.  I've not finished analysis of this part though.

sils:
How is an import with side effects better than a function?

sils:
IMO we should avoid an import with side effects whereever possible. Explicit is better than implicit and I'd rather explicitly run the version check *earliest possible*, then do the import so people see explicitly why this is happening. I think we also need to improve the comment that is explaining why this happens as it doesn't seem to be too obvious.

jayvdb:
If explicit is the goal, then this tiny function can be moved into setup.py, so it is clear what the rule is.

However, as I noted before, that is an unnecessary discussion.
https://github.com/coala/coala/pull/3328 shows that this assert function call can be moved down to the correct place in `coala`, and I have also checked that the same can be done for `coala-bears`.

sils:
https://github.com/coala/coala/pull/3328#issuecomment-267829370 let's discuss over there

Mixih:
imports still under discussion. WIP

Mixih:
rebase please... I also don't know why you modified travis, should be unrelated.

andreidinu:
@Mixih there was a discussion above about imports. There is also a new topic where `setup.py` was merged for `coala/coala` with some changes in travis so I thought I have to do something similar for `coala-bears`.

jayvdb:
blocked by https://github.com/coala/coala-bears/pull/1200 .

bears/ruby/RuboCopBear.py
```diff
@@ -176,62 +176,62 @@ class members.
...
-                       'IgnoreEmptyMethods':
-                           ignore_unused_method_args_if_empty},
-                  }
+            'Style/AccessModifierIndentation':
+                {'EnforcedStyle': access_modifier_indentation},
+            'Style/Alias':
+                {'EnforcedStyle': preferred_alias},
+            'Style/AlignHash':
+                {'EnforcedHashRocketStyle': align_hash_rocket_by,
+                    'EnforcedColonStyle': align_colon_by,
```
jayvdb: this is aligned badly :/
use

```
            'Style/AlignHash': {
                'EnforcedHashRocketStyle': align_hash_rocket_by,
                'EnforcedColonStyle': align_colon_by,
                ...
            },
```

?

bears/ruby/RuboCopBear.py
```diff
@@ -176,62 +176,62 @@ class members.
...
+                {'EnforcedStyle': class_check},
+            'Style/CommentAnnotation':
+                {'Keywords': comment_keywords},
+            'Style/GuardClause':
+                {'MinBodyLength': min_if_unless_guard},
+            'Style/IndentationWidth':
+                {'Width': indent_size},
+            'Style/MethodName':
+                {'EnforcedStyle': naming_convention.get(
+                    method_naming_convention, method_naming_convention)},
```
jayvdb: as above

bears/ruby/RuboCopBear.py
```diff
@@ -176,62 +176,62 @@ class members.
...
+            'Style/StringLiterals':
+                {'EnforcedStyle': string_literals},
+            'Style/VariableName':
+                {'EnforcedStyle': naming_convention.get(
```
jayvdb: as above, etc, etc below

tests/natural_language/WriteGoodLintBearTest.py
```diff
@@ -127,22 +127,23 @@
...
-                                                          passive_file,
-                                                          bad_file,
-                                                          so_file,),
-                                                      invalid_files=(
-                                                          illusion_file,),
-                                                      settings={
-                                                        'allow_repeated_words':
-                                                            False})
+WriteGoodLintBearWithIllusionTest = verify_local_bear(
+                                                WriteGoodLintBear,
```
jayvdb: (as explained above, you do not need to do this, and shouldnt do this... but ...)

**If** you are going to revise all these lines, you should try to avoid future line length problems by de-denting everything as much as possible.

This line only needs to be indented by four spaces, and the other lines below indented by four or eight spaces.  i.e.

```
WriteGoodLintBearWithIllusionTest = verify_local_bear(
    WriteGoodLintBear,
    valid_files=(
    ....
```

tests/natural_language/WriteGoodLintBearTest.py
```diff
@@ -127,22 +127,23 @@
...
-                                                          cliche_file,
-                                                          adverb_file,
-                                                          passive_file,
-                                                          bad_file,
-                                                          so_file,),
-                                                      invalid_files=(
-                                                          illusion_file,),
-                                                      settings={
-                                                        'allow_repeated_words':
-                                                            False})
```
jayvdb: I believe this is the only line with an error for `WriteGoodLintBearWithIllusionTest`.

The easiest fix is to de-dent `False})` by four spaces.

Another way to solve it, without re-indenting everything, by creating a dict called `setting1` and using it here.

```
setting1 = {
    'allow_repeated_words': False,
}
...

                                                      settings=setting1)
```

tests/python/BanditBearTest.py
```diff
@@ -80,7 +80,11 @@ class BanditBearTest(LocalBearTestHelper):
 
 # The following test will ignore some error codes, so "good" and "bad" doesn't
 # reflect the actual code quality.
+
+
 good_files = ('good_file.py', 'assert.py')
+
+
```
jayvdb: only one space needed here.

jayvdb:
coala/coala-bears#1200 was merged; your .travis.yml and setup.py changes can now be removed.

NiklasMM:
ack 008e8df

sils:
@rultor merge

PR #1142, opened 2016-12-16
deps.sh: Cache installed tailor

jayvdb:
The install.sh provided by tailor installs into /usr/local,
which is not cachable and requires sudo.

Revise the install.sh to install with a PREFIX of ~/.local/,
which places everything into ~/.local/tailor/, with the
version specific components in ~/.local/tailor/tailor-1.9.x/

---
This change is [Reviewable](https://reviewable.io/reviews/coala/coala-bears/1142)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ cache:
     - ~/pmd-bin-5.4.1
     - ~/bakalint-0.4.0
     - ~/dart-sdk/bin
+    - ~/.local/tailor/
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -53,6 +53,7 @@ cache:
     - ~/pmd-bin-5.4.1
     - ~/bakalint-0.4.0
     - ~/dart-sdk/bin
+    - ~/.local/tailor/
```
sils: don't you need to add this to PATH in 64?

jayvdb: ya. good catch.  it worked *once* without it because ~/.local/bin is in the PATH , and tailor's install.sh drops a sym link in there.

sils:
cn you rebase? LGTM. https://github.com/coala/coala-bears/pull/1142/commits/b5c64b3d7023ee0b48ec2a351632d1a6af43235f ack

jayvdb:
reack e16a04a

jayvdb:
@rultor merge

PR #1156, opened 2016-12-17
WIP: InvalidLinkBear: Add more error codes

Mixih:
Made 405, 406, and 408 http response codes become invalid links

Closes https://github.com/coala/coala-bears/issues/1155


jayvdb:
I've left a longer note on the issue, but ..  if we are going to increase the scope of the link checker, IMO we need to be using severity to indicate which errors are 100% errors (MAJOR), and which are typically errors (NORMAL) and those that should be recorded but may not be an error (MINOR).

Mixih:
Marking WIP as there is active functionality being discussed.

Mixih:
Issue is won't fix now

PR #1164, opened 2016-12-18
.ci/dep.sh;.travis.yml;requirements.txt: Added pip install

DanRadulescu962:
pip install git+https://github.com/willkg/dennis was added to .travis.yml and
.ci/deps.sh. Also added dennis~=0.7 in requirements.
It fixes setup.py requirement fails.

Fixes https://github.com/coala/coala-bears/issues/1161

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -100,6 +100,7 @@ for dep_version in "${dep_versions[@]}" ; do
   pip install pip -U
   pip install -U setuptools
   pip install -r test-requirements.txt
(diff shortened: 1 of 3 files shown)


DanRadulescu962:
The solution posted was for something else. I messed up those things a little. But I am able to reproduce the error. I will try to fix it.

jayvdb:
As explained at https://github.com/coala/coala-bears/issues/1161#issuecomment-267815369 , this issue is *not* requesting that someone installs dennis .

PR #1183, opened 2016-12-20
.coafile: Enable reSTCheckBEar on all rst files

Mixih:
This forces developers to write proper rst files

Closes https://github.com/coala/coala-bears/issues/1182
Related to https://github.com/coala/coala/issues/3044

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -39,7 +39,7 @@ max_lines_per_file = 1000
 
 [rst]
 bear = reSTLintBear
-files = README.rst
+files = **.rst
 
 [TODOS]
 enabled = False


.coafile
```diff
@@ -39,7 +39,7 @@ max_lines_per_file = 1000
 
 [rst]
 bear = reSTCheckBear
-files = README.rst
+files = **/*.rst
```
jayvdb: Doesnt this exclude rst in root directory?

Mixih: Seems ok... Here is my coala run output(I think the double star also matches none)
````
Executing section rst...
[DEBUG][19:14:03] coala is run only on changed files, bears' log messages from previous runs may not appear. You may use the `--flush-cache` flag to see them.
[DEBUG][19:14:03] Files that will be checked:
/home/dev/projects/dev/coala-bears/RELEASE_NOTES.rst
/home/dev/projects/dev/coala-bears/README.rst
/home/dev/projects/dev/coala-bears/docs/index.rst
````

Comment (jayvdb) [inline .coafile]: Isnt `**/*.rst` collecting root directory members a bug?

My guess is the glob is being run on absolute-path filenames, instead of project-root-relative filenames.

Comment (Mixih) [inline .coafile]: I think it is....(though to only collect subdirs, wouldn't `*/**.rst` work?)

Comment (jayvdb) [inline .coafile]: to *only* collect subdirs, `*/**.rst` is correct.

But... I cant reproduce this... but I can raise a bug: https://github.com/coala/coala/issues/3369

File: .coafile [hunk 2]
```diff
@@ -39,7 +39,7 @@ max_lines_per_file = 1000
 
 [rst]
 bear = reSTCheckBear
```
jayvdb: this isnt a real bear name .  https://github.com/coala/coala-bears/pull/1189

Mixih:
The bear name has been repaired, not sure if we still want this patch

jayvdb:
ack 4138053

jayvdb:
> not sure if we still want this patch

I am not sure either, but I dont want to hold it up.  IMO, deleting docs, and the CI for it, is a better strategy.
Then we can mark all bears internals as private, so we dont need them to maintain backwards compatibility at the dev API level.
And then we don't have docs PRs appearing here, constantly conflicting other changes.

Mixih:
rebased

Mixih:
Reack a54cfde

Mixih:
reack a54cf9de

Mixih:
@rultor merge

PR #1187, opened 2016-12-21
WIP: XMLBear: Support for style specifications

namanyadav12:
A format can be specified in the arguments or in the configuration.
The valid args are c14n, c14n11, exc-c14n and oldxml10.
If a style is specified, then the style messages from the bear have MAJOR severity. 

Closes https://github.com/coala/coala-bears/issues/1098

### DIFF
--- a/bears/xml2/XMLBear.py
+++ b/bears/xml2/XMLBear.py
@@ -1,5 +1,6 @@
 import itertools
 import re
+import logging
 
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.DistributionRequirement import (
@@ -11,7 +12,6 @@ from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 def path_or_url(xml_dtd):
     """
     Converts the setting value to url or path.
-
     :param xml_dtd: Setting key.
     :return:        Returns a converted setting value.
     """
@@ -21,13 +21,27 @@ def path_or_url(xml_dtd):
         return path(xml_dtd)
 
 
+def xml_style_valid(xml_style):
+    """
+    Checks if xml_style is valid.
+    :param xml_style: Setting key.
+    :return:          Returns the value if valid else returns none.
+    """
+    style = str(xml_style)
+    if style not in XMLBear._styles:
+        logging.warn('Unrecognised style ' + style + '. Valid xml'
+                     ' styles are c14n, c14n11, exc-c14n and oldxml10.'
+                     ' Running XMLBear without any xml_style argument.')
+        return None
+    return '--' + style
+
+
 @linter(executable='xmllint',
         use_stdout=True,
         use_stderr=True)
 class XMLBear:
     """
     Checks the code with ``xmllint``.
-
     See http://xmlsoft.org/xmllint.html
     """
     LANGUAGES = {'XML'}
@@ -40,21 +54,30 @@ class XMLBear:
     _output_regex = re.compile(
         r'.*:(?P<line>\d+):.*(?P<severity>error|warning)\s?: '
         r'(?P<message>.*)\n.*\n.*')
+    _diff_severity = RESULT_SEVERITY.INFO
+    _styles = ('c14n', 'c14n11', 'exc-c14n', 'oldxml10')
 
-    @staticmethod
-    def create_arguments(filename, file, config_file,
+    def create_arguments(self, filename, file, config_file,
                          xml_schema: path='',
-                         xml_dtd: path_or_url=''):
+                         xml_dtd: path_or_url='',
+                         xml_style: xml_style_valid=None):
         """
         :param xml_schema: ``W3C XML Schema`` file used for validation.
         :param xml_dtd:    ``Document type Definition (DTD)`` file or
                            url used for validation.
+        :param xml_style:  ``XML Style Specification`` Relevant args are
+                           c14n, c14n11, exc-c14n and oldxml10.
+                           Find out more about the formats at
+                           https://www.w3.org/TR/#tr_XML_Canonicalization
         """
         args = (filename,)
         if xml_schema:
             args += ('-schema', xml_schema)
         if xml_dtd:
             args += ('-dtdvalid', xml_dtd)
+        if xml_style:
+            args += (xml_style,)
+            self._diff_severity = RESULT_SEVERITY.MAJOR
 
         return args
 
@@ -68,7 +91,7 @@ class XMLBear:
                     output_regex=self._output_regex),
                 self.process_output_corrected(
                     stdout, filename, file,
-                    diff_severity=RESULT_SEVERITY.INFO,
+                    diff_severity=self._diff_severity,
                     result_message='XML can be formatted better.'))
         else:
             # Return issues from stderr if stdout is empty
--- a/tests/xml2/XMLBearTest.py
+++ b/tests/xml2/XMLBearTest.py
@@ -8,6 +8,7 @@ from coalib.testing.LocalBearTestHelper import verify_local_bear, execute_bear
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.settings.Section import Section
 from coala_utils.ContextManagers import prepare_file
+from coalib.settings.Setting import Setting
 
 
 def load_testdata(filename):
@@ -34,6 +35,10 @@ invalid_xml_chars = """<?xml version="1.0"?>
 
 valid_xml_chars = """<?xml version="1.0"?>
 <a>hey and hi</a>
+"""
+
+invalid_xml_file_c14n = """<?xml version="1.0"?>
+<a/>
 """
 
 dtd_file = os.path.join(os.path.dirname(__file__),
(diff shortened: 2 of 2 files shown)


bears/xml2/XMLBear.py
```diff
@@ -40,21 +40,32 @@ class XMLBear:
...
         r'(?P<message>.*)\n.*\n.*')
+    _diff_severity = RESULT_SEVERITY.INFO
 
-    @staticmethod
-    def create_arguments(filename, file, config_file,
+    def create_arguments(self, filename, file, config_file,
                          xml_schema: path='',
-                         xml_dtd: path_or_url=''):
+                         xml_dtd: path_or_url='',
+                         xml_style: str=''):
```
jayvdb: instead of `str` it should be a restricted list of values checked with a function like `path_or_url`.

bears/xml2/XMLBear.py
```diff
@@ -40,21 +40,32 @@ class XMLBear:
...
                            url used for validation.
+        :param xml_style:  ``XML Style Specification`` Relevant args are
+                           c14n, c14n11, exc-c14n and oldxml10.
         """
         args = (filename,)
         if xml_schema:
             args += ('-schema', xml_schema)
         if xml_dtd:
             args += ('-dtdvalid', xml_dtd)
+        styles = ('c14n', 'c14n11', 'exc-c14n', 'oldxml10')
```
jayvdb: CONSTANT list, should be a module or class attribute

tests/xml2/XMLBearTest.py
```diff
@@ -79,6 +79,20 @@ def load_testdata(filename):
     settings={'xml_dtd': dtd_url},
     tempfile_kwargs={'suffix': '.xml'})
 
+XMLBearStyleTest = verify_local_bear(
+    XMLBear,
+    valid_files=(valid_xml_file,),
+    invalid_files=(invalid_xml_chars,),
+    settings={'xml_style': 'oldxml10'},
```
jayvdb: This style is not being tested.  This test class is identical to `XMLBearCorrectedTest` , etc.
You need to find an XML style issue, and ensure that the linter bear reports the error that `xmllint` emits and that its severity is correct.

tests/xml2/XMLBearTest.py
```diff
@@ -79,6 +79,20 @@ def load_testdata(filename):
...
+    tempfile_kwargs={'suffix': '.xml'})
+
+XMLBearInvalidStyleTest = verify_local_bear(
```
jayvdb: XMLBearInvalidStyleTest -> XMLBearUnrecognisedStyleNameTest

bears/xml2/XMLBear.py
```diff
@@ -40,21 +40,32 @@ class XMLBear:
...
+        if xml_style in styles:
+            args += ('--' + xml_style,)
+            self._diff_severity = RESULT_SEVERITY.MAJOR
+        elif xml_style != '':
+            self.warn('Invalid xml_style argument ' + xml_style +
```
jayvdb: this should be an error -- the user has mistyped the style.  coala should not allow invalid configuration.

tests/xml2/XMLBearTest.py
```diff
@@ -79,6 +79,20 @@ def load_testdata(filename):
...
+    XMLBear,
+    valid_files=(valid_xml_file,),
+    invalid_files=(invalid_xml_chars,),
+    settings={'xml_style': 'wrong args'},
```
jayvdb: need to confirm that this error is the cause of the failure.
otherwise, some other error could occur and this test look like it has been effective.

bears/xml2/XMLBear.py
```diff
@@ -21,6 +22,21 @@ def path_or_url(xml_dtd):
...
+def xml_style_valid(xml_style):
+    """
+    Checks if xml_style is valid.
+
+    :param xml_style: Setting key.
+    :return:          Returns the value if valid else returns none.
+    """
+    style = str(xml_style)
+    if style not in XMLBear._styles:
+        logging.error('Unrecognised style ' + style + '. Valid xml'
```
sils: I think this should be a warning and you should mention that you just ignore the argument because you're not erroring out, you're just ignoring that part right?

bears/xml2/XMLBear.py
```diff
@@ -40,21 +56,28 @@ class XMLBear:
...
+                           c14n, c14n11, exc-c14n and oldxml10.
         """
         args = (filename,)
         if xml_schema:
             args += ('-schema', xml_schema)
         if xml_dtd:
             args += ('-dtdvalid', xml_dtd)
+        if xml_style:
+            args += (xml_style,)
+            self._diff_severity = RESULT_SEVERITY.MAJOR
```
sils: why is this major? Also if you do this, all the other errors in the diff will have a major severity as well, right? That sounds bad.

namanyadav12: This was given in the issue by @jayvdb, you can check it out here https://github.com/coala/coala-bears/issues/1098

namanyadav12:
@sils you can find out more about the formats at
https://www.w3.org/TR/2001/REC-xml-c14n-20010315
https://www.w3.org/TR/xml-c14n11/
https://www.w3.org/TR/xml-exc-c14n/

namanyadav12:
@sils you can find out more about the formats at
https://www.w3.org/TR/2001/REC-xml-c14n-20010315
https://www.w3.org/TR/xml-c14n11/
https://www.w3.org/TR/xml-exc-c14n/

namanyadav12:
Also @sils, self._diff_severity = RESULT_SEVERITY.MAJOR is to make sure that If the user has chosen a style in the configuration, the style messages from the bear are treated as "MAJOR" severity errors.
This was given in the issue by @javydb

bears/xml2/XMLBear.py
```diff
@@ -40,21 +56,28 @@ class XMLBear:
...
                          xml_schema: path='',
-                         xml_dtd: path_or_url=''):
+                         xml_dtd: path_or_url='',
+                         xml_style: xml_style_valid=None):
         """
         :param xml_schema: ``W3C XML Schema`` file used for validation.
         :param xml_dtd:    ``Document type Definition (DTD)`` file or
                            url used for validation.
+        :param xml_style:  ``XML Style Specification`` Relevant args are
+                           c14n, c14n11, exc-c14n and oldxml10.
```
sils: you need to include that info here so the user can find out more about the specs, not me

NiklasMM:
Please include the links that you gave in the comment, as sils pointet out [here](https://github.com/coala/coala-bears/pull/1187#discussion_r94385139)

NiklasMM:
54f9232 needs work

bears/xml2/XMLBear.py
namanyadav12: oh, ok !! I misunderstood you earlier

namanyadav12:
Done

tests/xml2/XMLBearTest.py
```diff
@@ -36,6 +37,10 @@ def load_testdata(filename):
 <a>hey and hi</a>
 """
 
+invalid_xml_file_c14n = """<?xml version="1.0"?>
```
NiklasMM: It's not clear without knowing the c14n style, why this is invalid. Could you please add a comment explaining why this fails under this style?

bears/xml2/XMLBear.py
```diff
@@ -21,13 +21,27 @@ def path_or_url(xml_dtd):
...
+    return '--' + style
+
+
 @linter(executable='xmllint',
         use_stdout=True,
         use_stderr=True)
 class XMLBear:
     """
     Checks the code with ``xmllint``.
-
```
sils: why do you delete this, unrelated and I don't think that should happen anyway

bears/xml2/XMLBear.py
```diff
@@ -40,21 +54,30 @@ class XMLBear:
...
+                         xml_dtd: path_or_url='',
+                         xml_style: xml_style_valid=None):
         """
         :param xml_schema: ``W3C XML Schema`` file used for validation.
         :param xml_dtd:    ``Document type Definition (DTD)`` file or
                            url used for validation.
+        :param xml_style:  ``XML Style Specification`` Relevant args are
+                           c14n, c14n11, exc-c14n and oldxml10.
+                           Find out more about the formats at
+                           https://www.w3.org/TR/#tr_XML_Canonicalization
```
sils: can you rather give a link where I can *immediately* see what those things mean for my code?  This is not helpful (maybe a bit) but almost nobody will actually browse that website since it doesn't seem to contain the desired info easily available.

bears/xml2/XMLBear.py
```diff
@@ -21,13 +21,27 @@ def path_or_url(xml_dtd):
...
+def xml_style_valid(xml_style):
+    """
+    Checks if xml_style is valid.
+    :param xml_style: Setting key.
+    :return:          Returns the value if valid else returns none.
+    """
+    style = str(xml_style)
+    if style not in XMLBear._styles:
+        logging.warn('Unrecognised style ' + style + '. Valid xml'
+                     ' styles are c14n, c14n11, exc-c14n and oldxml10.'
```
sils: you should rather do something like `"Unrecognised style {}. Valid styles are {}. ......".format(...)` we have a `join_names` function in coala (which should be moved to coala-utils: coala/coala#3486) that can concatenate them properly

sils: ideally coala/coala#3486 should be solved first, that'll save us some hassle later

tests/xml2/XMLBearTest.py
```diff
@@ -98,3 +103,29 @@ def test_errors(self):
...
+                self.assertEqual(results[0].severity, RESULT_SEVERITY.MAJOR)
+
+    def test_unrecognised_xml_style_name(self):
+        self.section.append(Setting('xml_style', 'wrong-args'))
+        content = invalid_xml_file.splitlines()
+        with prepare_file(content, None) as (file, fname):
+            with execute_bear(self.uut, fname, file) as results:
+                self.assertEqual(results[0].message,
+                                 'XML can be formatted better.')
+                self.assertEqual(results[0].severity, RESULT_SEVERITY.INFO)
```
sils: can we have a valid file too?

bears/xml2/XMLBear.py
namanyadav12: @sils can i just copy the information in (xmllint --help) like
--c14n : save in W3C canonical format v1.0 (with comments)

bears/xml2/XMLBear.py
jayvdb: This was merged, and is ready to use.

jayvdb: Why not an error, or even exception?  Do we habe guidance on this?

bears/xml2/XMLBear.py
```diff
@@ -68,7 +91,7 @@ def process_output(self, output, filename, file):
                     output_regex=self._output_regex),
                 self.process_output_corrected(
                     stdout, filename, file,
-                    diff_severity=RESULT_SEVERITY.INFO,
+                    diff_severity=self._diff_severity,
```
jayvdb: Isnt this a thread unsafe operation?
Is that a problem? cc: @Makman2

jayvdb:
We've all invested a lot of time into this PR.
Please come back and complete this PR, when you have time.

bears/xml2/XMLBear.py
Makman2: in praxis it's not (due to python executing everything in several processes), though this is a local variable and shall be used as such. So replace `self._diff_severity` with just `diff_severity` :+1:

PR #1200, opened 2016-12-24
setup.py: Delegate Python version check to coala

jayvdb:
Depends on https://github.com/coala/coala/pull/3384

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ sudo: required
 
 language: python
 python:
+  # 2.7 is rejected by dependency mypy-lang
+  - 3.3
   - 3.4
   - 3.5
   - 3.6-dev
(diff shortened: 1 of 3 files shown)


jayvdb:
crap. also depends on https://gitlab.com/coala/package_manager/issues/2 , sort of.

jayvdb:
(Circle is failing because it is a Trusty build; I forgot to switch.)
Commit message and other stuff is needing improvement.
I would like a review of the concept before tidying it up.

.travis.yml
```diff
@@ -74,25 +79,43 @@ before_install:
...
   - echo 'language_check==0.8.*' >> requirements.txt
+  # On unsupported versions, coala will be installed in the `script` block
+  - >
+    if [[ $TRAVIS_PYTHON_VERSION_MINOR < 4 || $TRAVIS_PYTHON_VERSION_MAJOR == 2 ]]; then
+      sed -e '/^coala/d' requirements.txt > requirements.new
+      mv requirements.new requirements.txt
+    fi
 
 before_script:
+  - mv requirements.orig requirements.txt
```
Mixih: isn't this going to overwrite the custom requirements.txt made in line 90?

jayvdb: setup.py reads `requirements.txt` , and needs the original, without all of the fiddles done above.

sils:
@jayvdb I wonder if we need an own test build for that, is there any danger of regressions here? This should happen by default if setuptools doesn't mess up, right? So that's upstream functionality that we don't necessarily want to waste a lot of build time on.

Mixih:
@sils can't we rely on the regression testing built into coala now?

sils:
yep coala has tests for it so we don't need them in coala-bears

Mixih:
ack dee94fd

Mixih:
Ready to go after a rebase....

jayvdb:
heavily revised.

Mixih:
ack a0f258e

jayvdb:
@rultor merge

PR #1236, opened 2017-01-01
KeywordBear.py: Add optional regex_keyword param

dracarys09:
Ensures that keywords within a file can be matched against
a user defined regex. A testcase is also added in KeywordBearTest.py
for this feature.

Closes https://github.com/coala/coala-bears/issues/311

### DIFF
--- a/bears/general/KeywordBear.py
+++ b/bears/general/KeywordBear.py
@@ -79,6 +79,7 @@ class KeywordBear(LocalBear):
             filename,
             file,
             keywords: list=['todo', 'fixme'],
+            regex_keyword: str='',
             dependency_results: dict=None):
         '''
         Checks the code files for given keywords.
@@ -86,15 +87,46 @@ class KeywordBear(LocalBear):
         :param keywords:
             A list of keywords to search for (case insensitive).
             Default are TODO and FIXME.
+        :param regex_keyword:
+            A regular expression to search for matching keywords in a file.
         '''
-        keywords_regex = re.compile(
+        comments = _get_comments(dependency_results)
+
+        simple_keywords_regex = re.compile(
             '(' + '|'.join(re.escape(key) for key in keywords) + ')',
             re.IGNORECASE)
 
-        comments = _get_comments(dependency_results)
+        message = "The line contains the keyword '{}'."
+        yield from self.check_keywords(filename, file, comments,
+                                       simple_keywords_regex, message)
+
+        if regex_keyword is not '':
+            regex = re.compile(regex_keyword)
+            message = ("The line contains the keyword '{}' which "
+                       'resulted in a match with given regex.')
+            yield from self.check_keywords(filename, file, comments, regex,
+                                           message)
+
+    def check_keywords(self,
+                       filename,
+                       file,
+                       comments,
+                       regex,
+                       message):
+        '''
+        Checks for the presence of keywords according to regex in a given file.
+
+        :param regex:
+           A regular expression which is used to search matching
+           keywords in a file.
+        :param message:
+           A message to be displayed to the user when a keyword in a given file
+           results in a match. It may have an unnamed placeholder for the
+           keyword.
+        '''
 
         for line_number, line in enumerate(file, start=1):
-            for keyword in keywords_regex.finditer(line):
+            for keyword in regex.finditer(line):
                 diffs = generate_diff(
                     comments,
                     file,
@@ -104,8 +136,7 @@ class KeywordBear(LocalBear):
                     keyword.start())
                 yield Result.from_values(
                     origin=self,
-                    message="The line contains the keyword '{}'."
-                            .format(keyword.group()),
+                    message=message.format(keyword.group()),
                     file=filename,
                     line=line_number,
                     column=keyword.start() + 1,
--- a/tests/general/KeywordBearTest.py
+++ b/tests/general/KeywordBearTest.py
@@ -185,3 +185,32 @@ class KeywordBearDiffTest(unittest.TestCase):
                              '+/*\n'
                              ' test\n'
                              ' */\n')
+
+    def test_keyword_regex(self):
+        text = ['# add two given values and result the result\n',
+                'def add(a, b):',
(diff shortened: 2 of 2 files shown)


bears/general/KeywordBear.py
```diff
@@ -113,3 +117,25 @@ def run(self,
                     end_column=keyword.end() + 1,
                     severity=RESULT_SEVERITY.INFO,
                     diffs=diffs)
+
+            if regex_keyword is not '':
```
sils: I'd make two for loops actually so we can move the if outside the loop:

```python
keywords_regex = ...
if keywords:
    for ...:
        ...

if regex_keyword:
    for ...:
        ...
```

Note that I also added the `if keywords` checking the normal keywords for emptiness!

And if we have that we see, that we have a huge chunk of duplication here, so you can actually move the whole if and for loop outside into a function that takes a regex and a custom message with one format argument, saving like half of the code here.

sils:
Otherwise good! :)

bears/general/KeywordBear.py
```diff
@@ -79,22 +79,54 @@ def run(self,
...
+                       message):
+        '''
+        Checks for the presence of keywords according to regex in a given file.
+
+        :param regex:
+           A regular expression which is used to search matching
+           keywords in a file.
+        :param message:
+           A message to be displayed to the user when a keyword in a given file
+           results in a match.
```
sils: you have to mention that it may have an unnamed placeholder for the keyword

dracarys09: @sils I didn't get it... are you referring to '{}' in message text which gets replaced by actual keyword ?
the modification that is needed is to modify :param message: so that it also has the line 'it may have an unnamed placeholder for the keyword' 
Is that all  or something else is needed ?

sils:
yes you need to tell the user of the function that he can use that
placeholder

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

dracarys09:
Modified the param doc as you suggested

bears/general/KeywordBear.py
```diff
@@ -79,22 +79,55 @@ def run(self,
...
         '''
         Checks the code files for given keywords.
 
         :param keywords:
             A list of keywords to search for (case insensitive).
             Default are TODO and FIXME.
+        :param regex_keyword:
+            A regular expression to search for matching keywords in a file.
         '''
         keywords_regex = re.compile(
```
sils: you gotta change this to somethin' like `simple_keywords_regex` otherwise keywords regex and regex keywords is super confusing

bears/general/KeywordBear.py
```diff
@@ -79,22 +79,55 @@ def run(self,
...
             '(' + '|'.join(re.escape(key) for key in keywords) + ')',
             re.IGNORECASE)
 
+        regex = re.compile(regex_keyword)
```
sils: also I'd compile this in line 104 or so so it's compiled right before it's used - and only *if* it's used!

bears/general/KeywordBear.py
```diff
@@ -79,22 +79,55 @@ def run(self,
...
+
         comments = _get_comments(dependency_results)
 
+        message = "The line contains the keyword '{}'."
+        yield from self.check_keywords(filename, file, comments,
+                                       keywords_regex, message)
```
sils: IMO there should be an if around this one as well, the keywords setting may be given empty as well

sils:
This is gett'n' good, we can merge it soon :)

bears/general/KeywordBear.py
dracarys09: @sils but that is already handled... even if the keywords setting is empty by default we are checking for TODO|FIXME (case insensitive). IMO adding an ```if``` won't make sense because keywords setting is already getting converted to to regex for checking todo and fixme.

However this is not the case we are calling ```check_keywords``` for the second time. so ```if```` condition there makes sense.

bears/general/KeywordBear.py
dracarys09: how about we keep it as ```regex_for_keywords``` or just ```regex``` ?

bears/general/KeywordBear.py
dracarys09: made this change. :)

bears/general/KeywordBear.py
dracarys09: changed keywords_regex to simple_keywords_regex
Now ```regex``` is for ```regex_keyword``` setting and ```simple_keywords_regex``` is for ```keywords``` setting

tests/general/KeywordBearTest.py
```diff
@@ -185,3 +185,32 @@ def test_keyword_between_code(self):
...
+                '               \n',
+                'print(add(2, 3))\n']
+
+        regex_keyword = 'r.s.l.'
+
+        with execute_bear(self.uut, filename='F', file=text,
+                          regex_keyword=regex_keyword,
+                          dependency_results=self.dep_results) as result:
+            self.assertEqual(result[0].message, 'The line contains the keyword'
+                                                " 'result' which resulted in a"
```
nemaniarjun: Not sure it its okay, but maybe change all 3 lines to use `"` instead of `'`

sils: it's not gitmate will catch you for it

nemaniarjun: Okay :) Otherwise it looks pretty awesome :D

dracarys09: @nemaniarjun if we do that... Gitmate will mark it as wrong... I did that before but changed it after gitmate's warning...  :smile:

dracarys09: any other changes needed ?

fneu: No. We talked about the quoting style in the maintainers channel and apparently this is the preferred way.

fneu:
ack https://github.com/coala/coala-bears/pull/1236/commits/353bce2435373f7a9ec8499afd996ac327f54121

PR #1242, opened 2017-01-02
RadonBear: Pin to version 1.4.0

jayvdb:
radon 1.4.2 added a dependency on flake8-polyfill, which depends
on flake8 without any version specification, and flake8 versions
have various version dependencies on pyflakes and pycodestyle.
As a result, it is not possible to know what versions of pyflakes
and pycodestyle that will be required to be compatible with
whatever flake8 version happens to be installed.

Fixes https://github.com/coala/coala-bears/issues/1228


sils:
ack https://github.com/coala/coala-bears/pull/1242/commits/00103fed78a4bd0988f2cde8504c6b836dea553b

sils:
@rultor merge

PR #1246, opened 2017-01-03
InvalidLinkBear: Deprecate timeout setting

meetmangukiya:
Closes https://github.com/coala/coala-bears/issues/1244

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -90,11 +90,16 @@ class InvalidLinkBear(LocalBear):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
                     host = urlparse(link).netloc
                     code = InvalidLinkBear.get_status_code(
-                               link, network_timeout.get(
-                                   host, InvalidLinkBear.DEFAULT_TIMEOUT))
+                               link,
+                               network_timeout.get(host)
+                               if host in network_timeout
+                               else network_timeout.get('*')
+                               if '*' in network_timeout
+                               else InvalidLinkBear.DEFAULT_TIMEOUT)
                     yield line_number + 1, link, code
 
-    @deprecate_settings(link_ignore_regex='ignore_regex')
+    @deprecate_settings(link_ignore_regex='ignore_regex',
+                        network_timeout=('timeout', lambda t: {'*': t}))
     def run(self, filename, file,
             network_timeout: typed_dict(str, int, DEFAULT_TIMEOUT)=dict(),
             link_ignore_regex: str='([.\/]example\.com|\{|\$)',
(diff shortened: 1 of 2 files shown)


bears/general/InvalidLinkBear.py
```diff
@@ -95,6 +95,7 @@ def find_links_in_file(file, network_timeout, link_ignore_regex,
                     yield line_number + 1, link, code
 
     @deprecate_settings(link_ignore_regex='ignore_regex')
+    @deprecate_settings(network_timeout='timeout')
```
sils: just add it to the existing decorator, don't make two please

sils:
Closes -> Fixes

meetmangukiya:
But for a note, network_timeout isn't exactly timeout, it is a `dict`,
while timeout was `int`, so deprecation will not work actually...
Since the decorator passes the `timeout` setting as `network_timeout`,
which will make it malfunction, we need an error in this case and not
a deprecation warning.

sils:
uhm deprecate_settings allows you to provide a conversion function, just read its docs, in this case you'll want a testcase for that as well. Maybe you want the dict to support something like `'*': 15` to match all URLs and then translate `15` to `{'*': 15}`

sils:
we definitely need a testcase, 7902eb9 needs work

meetmangukiya:
ahh,, My very first PR at coala was renaming `ignore_regex` to `link_ignore_regex`, but that was just a rename that's why the conversion didn't come into picture I guess, I made the assumption on that basis, I'll be looking into it now.

tests/general/InvalidLinkBearTest.py
```diff
@@ -231,7 +231,7 @@ def test_links_to_ignore(self):
         self.assertResult(valid_file=valid_file,
                           settings={'link_ignore_list': link_ignore_list})
 
-    def test_variable_timeouts(self):
+    def test_variable_timeouts_and_deprecation(self):
```
Asnelchristian: `test_variable_timeouts` was okay  i think :sweat_smile:

Asnelchristian:
this looks good to me 
ack https://github.com/coala/coala-bears/pull/1246/commits/c19e136b91849852228875b646338c6f12b50007

tests/general/InvalidLinkBearTest.py
Mixih: agreed

tests/general/InvalidLinkBearTest.py
```diff
@@ -255,10 +255,19 @@ def response(status_code, *args, **kwargs):
             self.assertEqual([x.message
                               for x in list(uut.run('file', file_contents,
                                                     network_timeout=nt))], [])
+            self.assertEqual([x.message
+                              for x in list(uut.run('file', file_contents,
+                                                    timeout=20))], [])
```
Mixih: well while we are at it, might as well add a test for the star syntax too.

meetmangukiya:
@sils now does it look like what you requested earlier ?

sils:
ack https://github.com/coala/coala-bears/pull/1246/commits/32a7d83177cbb98478b329d8b424d76030fff5ec

sils:
remove WIP and rebase?

tests/general/InvalidLinkBearTest.py
```diff
@@ -255,10 +257,26 @@ def response(status_code, *args, **kwargs):
             self.assertEqual([x.message
                               for x in list(uut.run('file', file_contents,
                                                     network_timeout=nt))], [])
+
+            with self.assertLogs(logging.getLogger()) as cm:
+                self.assertEqual([x.message
+                                  for x in list(uut.run('file', file_contents,
+                                                        timeout=20))], [])
+            self.assertEqual(cm.output,
```
Mixih: hmm indent this please to line up so it's in the the with block. It's generally a bad idea to use a cm var outside of the with. (let me know if it's jsut github being stupid)

Mixih: Also I'm not really a fan of the var name cm as it's not very descriptive. How about log?

Mixih:
ack e6f8d32

Mixih:
unack e6f8d32

Mixih:
ack e20e679

Mixih:
@rultor merge

PR #1264, opened 2017-01-05
AStyleLintBear.py: Add a linter bear for AStyle

ivange94:
Add a linter bear for the Artistic Style linter

Closes https://github.com/coala/coala-bears/issues/388

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ addons:
       - opam
       - php-codesniffer
       - verilator
+      - astyle
 
 cache:
   pip: true
--- a/bears/c_languages/AStyleLintBear.py
+++ b/bears/c_languages/AStyleLintBear.py
@@ -1,0 +1,56 @@
+from coalib.bearlib.abstractions.Linter import linter
+from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='astyle',
+        output_format='corrected',
+        use_stdin=True)
+class AStyleLintBear:
+    """
+    Artistic Style is a source code indenter, formatter,
+    and beautifier for the C, C++, C++/CLI, Objective‑C,
+    C# and Java programming languages.
+
+    For more information, consult <http://astyle.sourceforge.net/astyle.html>.
+    """
+
+    LANGUAGES = {'C', 'C++', 'Objective-C', 'C#', 'Java'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='astyle')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         astyle_bracket_style: str='',
+                         use_spaces: bool=None,
+                         indent_size: int=0):
+        """
+        :param astyle_bracket_style:
+            Defines the bracket style to use. If no bracket style is
+            specified, the default 'allman' is used.
+        :param use_spaces:
+            If None, no indentation is specified and astyle uses a default.
+            If True, astyle uses spaces or if explicitly set to False, astyle
+            uses tabs for indentation.
+        :param indent_size:
+            Number of spaces per indentation level.
+        """
+        args = ['--dry-run', '--suffix=none']
+        if astyle_bracket_style:
+            args.append('--style=' + astyle_bracket_style)
+        if use_spaces is None:
+            indent = None
+        elif use_spaces:
+            indent = '--indent=spaces'
+        else:
+            indent = '--indent=tab'
+        if indent:
+            if indent_size:
+                args.append(indent + '=' + str(indent_size))
+            else:
+                args.append(indent)
+        return args
--- a/tests/c_languages/AStyleLintBearTest.py
+++ b/tests/c_languages/AStyleLintBearTest.py
@@ -1,0 +1,72 @@
+import os
+
+from bears.c_languages.AStyleLintBear import AStyleLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+test_file1 = """
+int main()
+{
+    int a = 19;
+    return 0;
+}\n"""
+
+test_file2 = """
+int main() {
+    int a = 19;
+    return 0;
+}\n"""
+
+test_file3 = """
+int main() {
+\tint a = 12;
+\treturn 0;
+}\n"""
+
+test_file4 = """
+int main()
+{
+        int a=23;
+        return 0;
+}\n"""
+
+
+def get_testfile_path(name):
+    return os.path.join(os.path.dirname(__file__), 'astyle_test_files',  name)
+
+
+AStyleLintBearTestUseJavaStyle = verify_local_bear(
+                     AStyleLintBear,
+                     valid_files=(test_file1,),
+                     invalid_files=(test_file2,),
+                     settings={'astyle_style': 'allman'}
+                 )
+
+AStyleLintBearTestUseSpaces = verify_local_bear(
+                    AStyleLintBear,
+                    valid_files=(test_file1,),
+                    invalid_files=(test_file3,),
+                    settings={'use_spaces': True}
+                )
+
+AStyleLintBearTestUseTabs = verify_local_bear(
+                    AStyleLintBear,
+                    valid_files=(test_file3,),
+                    invalid_files=(test_file2,),
+                    settings={'use_spaces': False}
+                )
+
+AStyleLintBearTestIndentFourSpaces = verify_local_bear(
+                    AStyleLintBear,
+                    valid_files=(test_file1,),
+                    invalid_files=(test_file4,),
+                    settings={'use_spaces': True,
+                              'indent_size': 4}
+                )
+
+AStyleLintBearTestUseOptionsFile = verify_local_bear(
+                    AStyleLintBear,
+                    valid_files=(test_file1,),
+                    invalid_files=(test_file2,),
+                    settings={'astyle_config': get_testfile_path(
+                        'astylerc_file')}
+                )
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -13,7 +13,8 @@ esac
 # apt-get commands
 export DEBIAN_FRONTEND=noninteractive
 
-deps="libclang1-3.4 indent mono-mcs chktex r-base julia golang-go luarocks verilator cppcheck flawfinder"
+deps="libclang1-3.4 indent mono-mcs chktex r-base julia golang-go luarocks verilator cppcheck flawfinder astyle"
+
 deps_infer="m4 opam"
 
 case $CIRCLE_BUILD_IMAGE in
--- a/tests/c_languages/astyle_test_files/astylerc_file
+++ b/tests/c_languages/astyle_test_files/astylerc_file
@@ -1,0 +1,1 @@
+--style=allman


bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,50 @@
...
+
+    For more information, consult <http://astyle.sourceforge.net/astyle.html>.
+    """
+
+    LANGUAGES = {'C, C++, Objective-C, C#, Java'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+
```
vijeth-aradhya: Add `REQUIREMENTS` here
And also `ASCIINEMA_URL`

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        :param astyle_style:
+             Defines the bracket style to use. If no bracket style is
+             specified, the default 'allman' is used.
+        :param astyle_indent:
+             The indentation type to use. It can either be spaces or tabs.
+        :param astyle_astylerc:
+             Optional configuration file for astyle to read configs.
+        :param astyle_indent_size:
+            Number of spaces per indentation level.
```
nemaniarjun: Indentation error

tests/c_languages/AStyleLintBearTest.py
```diff
@@ -0,0 +1,71 @@
+from bears.c_languages.AStyleLintBear import AStyleLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+import os
```
nemaniarjun: Google  PEP8 imports.

tests/c_languages/AStyleLintBearTest.py
```diff
@@ -0,0 +1,71 @@
...
+test_file4 = """
+int main()
+{
+        int a=23;
+        return 0;
+}"""
+
+
+def get_testfile_path(name):
+    return os.path.join(os.path.dirname(__file__), 'astyle_test_files',  name)
```
nemaniarjun: No need for a function

bears/c_languages/AStyleLintBear.py
nemaniarjun: Also doesn't this correct the file? If so, use can fix also.

nemaniarjun:
Since it also supports Java, we might not want to keep it in C_languages. Also change your commit shortlog to `C_languages: Add AstyleLintBear`

bears/c_languages/AStyleLintBear.py
ivange94: I thought running 'coala' would fix something like that.

tests/c_languages/AStyleLintBearTest.py
ivange94: Is there a problem with having that function?

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,53 @@
+from coalib.bearlib.abstractions.Linter import linter
+from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
```
ivange94: Not sure what the problem here is.

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+        :param astyle_indent_size:
+            Number of spaces per indentation level.
+        """
+        args = ('--dry-run', '--suffix=none')
+        if astyle_style:
+            args += ('--style=' + astyle_style,)
+        if astyle_indent:
+            indent = '--indent=' + astyle_indent
+            indent += '=' + \
+                str(astyle_indent_size) if astyle_indent_size else ''
```
Makman2: please use braces to wrap your expression:
```
indent += ('=' +
    str(astyle.....)... )
```

Makman2: btw if you have no indent size given, is `--indent=style=` a valid option?

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+            args += (indent,)
+        if astyle_astylerc:
+            args += ('--options=' + astyle_astylerc,)
```
Makman2: actually you can use a list instead of a tuple for `args`, linter doesn't alter the args returned + you don't need them anyway after that call. Makes stuff slightly faster, as appending to tuples copies the old tuple every time^^

tests/c_languages/AStyleLintBearTest.py
Makman2: it's used, so it has definitely a reason^^

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+class AStyleLintBear:
+    """
+    Artistic Style is a source code indenter, formatter,
+    and beautifier for the C, C++, C++/CLI, Objective‑C,
+    C# and Java programming languages.
+
+    For more information, consult <http://astyle.sourceforge.net/astyle.html>.
+    """
+
+    LANGUAGES = {'C, C++, Objective-C, C#, Java'}
```
Makman2: naah that's not how sets work :D -->
```
LANGUAGES = {'C', 'C++', 'Objective-C', 'C#', 'Java'}
```

nemaniarjun:
Astlye can be installed using apt-get. Confirmed. Downloads the version 2.05, latest is 2.06.
Lets use apt for travis.

jayvdb:
@ivange94 , could you please rebase.  I suspect the main errors in your builds are caused by a problem which has been fixed.

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+    For more information, consult <http://astyle.sourceforge.net/astyle.html>.
+    """
+
+    LANGUAGES = {'C', 'C++', 'Objective-C', 'C#', 'Java'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='astyle')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+    CAN_FIX = {'Formatting'}
```
Asnelchristian: CAN_FIX is enough

Asnelchristian: what i mean is, there is no need for ``CAN_DETECT``

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         astyle_style: str='',
+                         astyle_indent: str='',
+                         astyle_astylerc: str='',
+                         astyle_indent_size: int=0):
```
Asnelchristian: use ``indent_style`` for consistency

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+    CAN_FIX = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         astyle_style: str='',
+                         astyle_indent: str='',
+                         astyle_astylerc: str='',
```
Asnelchristian: maybe ``astyle_config`` for consistency, like ``yamllint_config``, ``tslint_config``, ``scalalint_config`` etc....

bears/c_languages/AStyleLintBear.py
Asnelchristian: the same goes for the other settings, you should rename them with consistency in mind

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,54 @@
...
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Formatting'}
+    CAN_FIX = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         astyle_style: str='',
+                         astyle_indent: str='',
```
Asnelchristian: maybe ``use_spaces: bool=None``

bears/c_languages/AStyleLintBear.py
ivange94: All settings had consistency in mind. The question is consistency with what. I designed them to be consistent with the underlying tool they are wrapping(i.e astyle) so users of these tools will not have to spend time learning new names. But you suggestions mean consistency with other coala bears. Is that preferable?

Makman2: consistency with our other bears is preferred :+1: :) Then you can easily re-use settings from a coafile :+1:

Asnelchristian:
wrt to the other settings,  settings of the other bears e.g ``use_spaces``,
several bears use this setting ( which simply require spaces to be used for
indentation ) so in you case you might wanna use the same name for
``astyle_indent``

2017-02-01 11:44 GMT+01:00 Ivange Larry <notifications@github.com>:

.travis.yml
```diff
@@ -93,6 +94,9 @@ before_install:
       sed -e '/^coala/d' requirements.txt > requirements.new
       mv requirements.new requirements.txt
     fi
+  # Install astyle linter
+  - sudo apt-get -qq update
+  - sudo apt-get install -y astyle
```
jayvdb: why is this necessary?

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,62 @@
...
+    CAN_FIX = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         astyle_style: str='',
+                         use_spaces: bool=None,
+                         astyle_config: str='',
+                         indent_size: int=0):
+        """
+        :param astyle_style:
```
jayvdb: does it only apply to brackets?  if so `astyle_bracket_style` would be better ?

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,62 @@
...
+            Defines the bracket style to use. If no bracket style is
+            specified, the default 'allman' is used.
+        :param use_spaces:
+            If None, no indentation is specified and astyle uses a default.
+            If True, astyle uses spaces or if explicitly set to False, astyle
+            uses tabs for indentation.
+        :param astyle_config:
+            Optional configuration file for astyle to read configs.
```
jayvdb: I think this can be achieved using `config_file`

tests/c_languages/AStyleLintBearTest.py
```diff
@@ -0,0 +1,72 @@
...
+
+from bears.c_languages.AStyleLintBear import AStyleLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+test_file1 = """
+int main()
+{
+    int a = 19;
+    return 0;
+}"""
```
jayvdb: please add a line feed before `"""` so the outputted file ends with a line feed.
(Otherwise the linter may create an error about no EOL and EOF.

.travis.yml
ivange94: I just googled how to install dependencies on travis and I got that. But still didn't work. My builds are failing.

ivange94: It was suggested that my builds are failing on travis because it does not have astyle installed

bears/c_languages/AStyleLintBear.py
ivange94: I think its just the bracket style. At least thats what the documentation says. But astyle itself just call it `style` and the command line parameter is `style`

bears/c_languages/AStyleLintBear.py
ivange94: a lot of of bears still specify the the linter specific config file. It gets me confused. For example https://github.com/coala/coala-bears/blob/master/bears/python/PyLintBear.py#L38

Not sure though but I think the coala config file is called using `--config-file` or `--config` which is different from the option astyle uses to call it's configuration file. So I don't understand how the two relate.

Makman2: nope you define the flag in `create_arguments` :)

The `config_file` parameter contains the full path to the generated config file in `generate_config`.

Anyway coala still supports setting configuration files specific to that tool, so people can migrate easier. You are right, ideally we don't want to allow those, but often those contain way more options than coala currently supports.

This changes with "aspects", where we want to streamline settings and options, and there we won't allow passing some `astyle_config` files (or config files used by other tools).

Makman2:
@ivange94 wanna continue? I think there's not much left for merge. I hope we don't have over-reviewed you, we just try to make the best bears we can :D

ivange94:
@Makman2 sorry about the delay. Will look at this in the morning. Its already 11 pm here and I've been working all day. Been really busy with other stuffs and still haven't figured out why travis and cirlce ci are failing on github. It doesn't locally.

Makman2:
@ivange94 ah sure take your time, just wanted to ask :)

bears/c_languages/AStyleLintBear.py
```diff
@@ -0,0 +1,56 @@
...
+    LANGUAGES = {'C', 'C++', 'Objective-C', 'C#', 'Java'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='astyle')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Formatting'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         astyle_bracket_style: str='',
```
jayvdb: The default here should be 'allman', and then it is self-documenting, and the note below isnt necessary.
Also that means that if astyle releases a version with the default changed, coala users are not immediately impacted.

tests/c_languages/AStyleLintBearTest.py
jayvdb: I see you have added a `\n` before the `"""`.  What I meant, and sorry for not being more clear, was to use
```
"""
...
}
"""
```

tests/c_languages/AStyleLintBearTest.py
```diff
@@ -0,0 +1,72 @@
...
+
+from bears.c_languages.AStyleLintBear import AStyleLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+test_file1 = """
+int main()
+{
+    int a = 19;
+    return 0;
+}\n"""
```
jayvdb: Please use
use
```
"""
...
}
"""
```

tests/c_languages/AStyleLintBearTest.py
```diff
@@ -0,0 +1,72 @@
+import os
+
+from bears.c_languages.AStyleLintBear import AStyleLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+test_file1 = """
```
jayvdb: could you create more descriptive variable names; instead of `test_file1`, maybe `test_good_file_with_assignment`

tests/c_languages/AStyleLintBearTest.py
```diff
@@ -0,0 +1,72 @@
...
+                    invalid_files=(test_file4,),
+                    settings={'use_spaces': True,
+                              'indent_size': 4}
+                )
+
+AStyleLintBearTestUseOptionsFile = verify_local_bear(
+                    AStyleLintBear,
+                    valid_files=(test_file1,),
+                    invalid_files=(test_file2,),
+                    settings={'astyle_config': get_testfile_path(
```
jayvdb: optional, but preferred, this config file could be generated at runtime.

bears/c_languages/AStyleLintBear.py
ivange94: I don't think that is right. If I set the default to allman it means the linter will always be run with command line option --style=allman. Users may not want that. Command line options take precedence over config file options. So If a user had something like --stye=java inside his config file, it will not work because the linter is always run with --style=allman which takes precedence over the config file.

ivange94: I am very certain of this because when I was implementing this long time ago, I set the default to allman and my tests that used config files kept failing. Then I realized that this happened because the config file options were been ignored since they were specified from the command line.

tests/c_languages/AStyleLintBearTest.py
ivange94: Can you please direct me to a linter that does it the way you suggest? If you noticed in my bear I removed the astyle_config as you commented. I was following the linter tutorial and they had that option but you said its not necessary. But I looked at a lot of linters and they still do that. So can you please direct me to a linter does it the way you want?

tests/c_languages/AStyleLintBearTest.py
ivange94: ok

tests/c_languages/AStyleLintBearTest.py
ivange94: ok

tests/c_languages/AStyleLintBearTest.py
jayvdb: Sorry ... I mean that instead of adding "tests/c_languages/astyle_test_files/astylerc_file" to the repo, this could be a generated temporary file. `CheckstyleBearTest.known_path_test` has an example of using `NamedTemporaryFile` to avoid a separate file for the config.

ivange94:
I'm not working on this again. Can't find the time to complete it.

PR #1271, opened 2017-01-07
KeywordBear.py: Output appropriate message

Techievena:
Output appropriate message if the language given in input is not valid/not supported for KeywordBear

Fixes https://github.com/coala/coala-bears/issues/1256

### DIFF
--- a/bears/general/KeywordBear.py
+++ b/bears/general/KeywordBear.py
@@ -1,4 +1,5 @@
 import re
+import logging
 
 from coalib.bearlib import deprecate_settings
 from coalib.bears.LocalBear import LocalBear
@@ -19,7 +20,10 @@ def _get_comments(dependency_results):
         return
 
     for result in annotation_bear_results:
-        yield from result.contents.get('comments', [])
+        if isinstance(result.contents, str):
+            logging.error(result.contents)
+        else:
+            yield from result.contents.get('comments', [])
 
 
 def generate_diff(comments, file, filename,
--- a/tests/general/KeywordBearTest.py
+++ b/tests/general/KeywordBearTest.py
@@ -1,6 +1,7 @@
 from collections import namedtuple
 from queue import Queue
 import unittest
+import logging
 
 from bears.general.KeywordBear import KeywordBear
 from coalib.results.HiddenResult import HiddenResult
@@ -214,3 +215,27 @@ class KeywordBearDiffTest(unittest.TestCase):
             self.assertEqual(result[0].message, 'The line contains the keyword'
                                                 " 'Issue #123' which resulted "
                                                 'in a match with given regex.')
+
+    def test_wrong_language(self):
+        self.section.append(Setting('language', 'anything'))
+        logger = logging.getLogger()
+        annotation_bear_result_type = namedtuple('result', 'contents')
+        dep_results = {
+            'AnnotationBear': [
+                annotation_bear_result_type(
+                  'coalang specification for anything not found.')
+            ]
+        }
+
+        text = ['# todo 123']
+
+        with self.assertLogs(logger, 'ERROR') as log:
+            with execute_bear(self.uut, filename='F', file=text,
+                              dependency_results=dep_results) as result:
+                self.assertEqual(len(result), 1)
+                self.assertEqual(result[0].diffs, {})
+                self.assertEqual(result[0].affected_code[0].start.line, 1)
+                self.assertEqual(len(log.output), 1)
+                self.assertIn(log.output[0],
+                              'ERROR:root:coalang specification'
+                              ' for anything not found.')


Shade5:
KeywordBear.py is not fully covered by the tests. You'll have to add tests for 100% coverage

Shade5:
Run `py.test --cov` to check coverage. You have get 100% coverage for KeywordBear.py

bears/general/KeywordBear.py
```diff
@@ -18,8 +18,12 @@ def _get_comments(dependency_results):
             not isinstance(annotation_bear_results, list)):
         return
 
-    for result in annotation_bear_results:
-        yield from result.contents.get('comments', [])
+    try:
+        for result in annotation_bear_results:
+            yield from result.contents.get('comments', [])
+    except AttributeError:
```
aptrishu: check [January 7, 2017 7:17 PM](https://gitter.im/coala/coala?at=5870f1855ffdeea7231a007e)

sils: @aptrishu I think it's a lot simpler to use github as the main review UI, just copying the text from gitter makes it way easier for other reviewers to see what's going on here

bears/general/KeywordBear.py
```diff
@@ -92,20 +97,21 @@ def run(self,
...
-        yield from self.check_keywords(filename, file, comments,
-                                       simple_keywords_regex, message)
-
-        if regex_keyword is not '':
-            regex = re.compile(regex_keyword)
-            message = ("The line contains the keyword '{}' which "
-                       'resulted in a match with given regex.')
-            yield from self.check_keywords(filename, file, comments, regex,
-                                           message)
+        if comments:
```
sils: I'd just do `if not comments: return` so you don't have to indent everything below

sils: well actually we still want the keywordbear to run even if it doesn't know where the comments are! Knowing the comments is optional and if they're not there the bear  should still work!

tests/general/KeywordBearTest.py
```diff
@@ -214,3 +214,20 @@ def test_keyword_regex(self):
...
+        section.append(Setting('language', 'anything'))
+        section.append(Setting('keywords', 'TODO'))
+        uut = KeywordBear(section, Queue())
+        dep_results = {
+            'AnnotationBear': HiddenResult(
+                'AnnotationBear',
+                'coalang specification for anything not found.')
+        }
+
+        text = ['a==b']
```
sils: put a todo in here and make asserts that the bear still works even if the language is unknown, we just want the warning right?

Techievena: Thanks @sils for your valuable suggestions.  Actually I had wrong concepts about working of Keyword Bear. I was therefore completely pissed off. Well a question for now. Does Keyword Bear have to return the position of  keywords in comments only or it returns position of keywords in the code too?

sils:
all keywords, also if its in the code

> I was therefore completely pissed off

did we do something wrong? If so please tell us so we know what we can do
better next time!

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

bears/general/KeywordBear.py
```diff
@@ -92,6 +97,9 @@ def run(self,
         '''
         comments = _get_comments(dependency_results)
 
+        if not comments:
+            return
```
sils: sorry but this means that if the language isn't known the bear won't even look for keywords

Techievena: Well it is looking for them inspite of these statements.
image

Techievena: I am removing them as they are unnecessary.

bears/general/KeywordBear.py
```diff
@@ -18,8 +19,12 @@ def _get_comments(dependency_results):
             not isinstance(annotation_bear_results, list)):
         return
 
-    for result in annotation_bear_results:
-        yield from result.contents.get('comments', [])
+    try:
```
sils: why is this needed? If the comments are unavailable that should be atched by the ifs above

sils: does this even have any effect?

Techievena:
What am I missing?

Shade5:
Did you figure out how to write the tests or do you need help with that?

Techievena:
@Shade5 Need help in writing test case

Shade5:
Does `py.test` work fine without your changes?
Show me the output of `py.test -k KeywordBearTest --cov`

Techievena:
Here is it @Shade5 
image

Shade5:
So you need to write two tests. One that satisties the `if` at 23 and another that does not. Add those tests to KeywordBearTest.py

Techievena:
Yes I am trying to cover the log statements using assertLogs but they are failing. Can you have a look on test case I have wrriten

Shade5:
How are you calling `_get_comments()` in your test?

Techievena:
I think `execute_bear` is calling them.

Shade5:
Any errors with `py.test -k KeywordBearTest --cov` ?

Shade5:
I'm geting `logging not definied`

Techievena:
I've imported logging in both Keyword.py and KeywordBearTest.py and `py.test -k KeywordBearTest --cov` is returning the same as before

Techievena:
image

tests/general/KeywordBearTest.py
```diff
@@ -214,3 +214,22 @@ def test_keyword_regex(self):
...
                                                 'in a match with given regex.')
+
+    def test_wrong_language(self):
+        self.section.append(Setting('language', 'anything'))
+        annotation_bear_result_type = namedtuple('result', 'contents')
+        dep_results = {
+            'AnnotationBear': [
+                annotation_bear_result_type(
+                  'coalang specification for anything not found.'
+                  )
```
adtac: We end the brackets/parantheses in the same line as the last element. See line 207 of this file.

tests/general/KeywordBearTest.py
```diff
@@ -214,3 +214,22 @@ def test_keyword_regex(self):
...
+            ]
+        }
+
+        text = ['# todo 123']
+
+        with execute_bear(self.uut, 'F', text,
```
adtac: Please use named arguments - they're easier to read. Every other test uses them and consistent is better :)

See line 211 of this file.

tests/general/KeywordBearTest.py
```diff
@@ -214,3 +214,22 @@ def test_keyword_regex(self):
...
+                          dependency_results=dep_results) as result:
+            self.assertEqual(result[0].diffs, {})
+            self.assertEqual(result[0].affected_code[0].start.line, 1)
+            self.assertEqual(len(result), 1)
```
adtac: this should probably be the first assert :3

adtac: And there aren't any checks to make sure the message is logged? I mean, that's what the whole PR is about.

Techievena:
@adtac Done with the changes. Is it mergeable now? :smiley:

bears/general/KeywordBear.py
```diff
@@ -19,7 +20,11 @@ def _get_comments(dependency_results):
         return
 
     for result in annotation_bear_results:
-        yield from result.contents.get('comments', [])
+        if isinstance(result.contents, str):
+            logging.error(result.contents)
+            yield []
```
sils: why do you yield an empty list? I'd just remove it.

adtac: what if I'm expecting a list? not yielding a list might trigger an exception

Techievena: As it was yeilding an empty list when `comments` key was not found in request.contents I thought I had to do the same in case when this execption arises.
Fine I'll remove it :+1:

sils: uhm yield is not a return @adtac

sils:
ack https://github.com/coala/coala-bears/pull/1271/commits/75ae1c9d38ab26533838f852c7d9801e6e6af488

sils:
@rultor merge

bears/general/KeywordBear.py
adtac: ah didn't realize the original was a `yield from`, not a `yield`

PR #1276, opened 2017-01-07
HaskellLintBear.py: Add multiple line diff message workaround

### LINKED ISSUE #1079: The bear HaskellLintBear raised an exception
I've used HaskellLintBear to linting https://github.com/wisn/elm-reactor/

Here is the log
https://travis-ci.org/wisn/elm-reactor/builds/180417562

The build result is green, but the bear HaskellLintBear raised an exception.

It seems HaskellLintBear have a problem
```
[WARNING][14:56:00] Bear HaskellLintBear failed to run. Take a look at debug messages (`-V`) for further information.
```

I've collected the traceback information:
```
Traceback (most recent call last):
  File "/coala-bears/bears/haskell/HaskellLintBear.py", line 41, in process_output
    assert issue['startLine'] == issue['endLine']
  AssertionError

  File "/coala-bears/bears/haskell/HaskellLintBear.py", line 45, in process_output
    newline = line_to_change.replace(issue['from'], issue['to'])
  TypeError: Can't convert 'NoneType' object to str implicitly
```

I think `TypeError: Can't convert 'NoneType' object to str implicitly` is the main problem.
Then, followed by `AssertionError`.

Unfortunately, I can't trace manually with `hlint` because my PC freezes when compiling (in installing) it. Hope this information will be helpful. Thanks and sorry for my bad English...

vijeth-aradhya:
Add multiple line diff message to be handled properly, atleast
when only the first line of `hlint` output is different (which is the
case in most cases)

Closes https://github.com/coala/coala-bears/pull/1276

### DIFF
--- a/bears/haskell/HaskellLintBear.py
+++ b/bears/haskell/HaskellLintBear.py
@@ -38,11 +38,15 @@ class HaskellLintBear:
         output = json.loads(output)
 
         for issue in output:
-            assert issue['startLine'] == issue['endLine']
             diff = Diff(file)
+            from_lines = issue['from'].splitlines()
+            to_lines = issue['to'].splitlines()
+            assert len(from_lines) == len(to_lines)
+            for other_lines in range(1, len(from_lines)):
+                assert from_lines[other_lines] == to_lines[other_lines]
             line_nr = issue['startLine']
             line_to_change = file[line_nr-1]
-            newline = line_to_change.replace(issue['from'], issue['to'])
+            newline = line_to_change.replace(from_lines[0], to_lines[0])
             diff.change_line(line_nr, line_to_change, newline)
 
             yield Result.from_values(
@@ -51,4 +55,7 @@ class HaskellLintBear:
                 file=filename,
                 severity=self.severity_map[issue['severity']],
                 line=issue['startLine'],
+                column=issue['startColumn'],
+                end_line=issue['endLine'],
+                end_column=issue['endColumn'],
                 diffs={filename: diff})
--- a/tests/haskell/HaskellLintBearTest.py
+++ b/tests/haskell/HaskellLintBearTest.py
@@ -1,15 +1,71 @@
+from queue import Queue
+
 from bears.haskell.HaskellLintBear import HaskellLintBear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 from coalib.testing.LocalBearTestHelper import verify_local_bear
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
 
-good_file = """
+good_single_line_file = """
 myconcat = (++)
-"""
+""".splitlines()
 
-bad_file = """
+bad_single_line_file = """
 myconcat a b = ((++) a b)
-"""
+""".splitlines()
 
-HaskellLintBearTest = verify_local_bear(HaskellLintBear,
-                                        valid_files=(good_file,),
-                                        invalid_files=(bad_file,),
-                                        tempfile_kwargs={'suffix': '.hs'})
+good_multiple_line_file = """
+import qualified Data.ByteString.Char8 as BS
+
+
+main :: IO()
+main =
+    return $ BS.concat
+        [ BS.pack "I am being tested by hlint!"
+        , "String dummy"
+        , "Another String dummy"
+        ]
+""".splitlines()
+
+bad_multiple_line_file = """
+import qualified Data.ByteString.Char8 as BS
+
+
+main :: IO()
+main =
+    return $ BS.concat $
+        [ BS.pack $ "I am being tested by hlint!"
+        , "String dummy"
+        , "Another String dummy"
+        ]
+""".splitlines()
+
+
+@generate_skip_decorator(HaskellLintBear)
+class HaskellLintBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('name')
+        self.uut = HaskellLintBear(self.section, Queue())
+
+    def test_valid(self):
+        self.check_validity(self.uut, good_single_line_file, valid=True,
+                            tempfile_kwargs={'suffix': '.hs'})
+        self.check_validity(self.uut, good_multiple_line_file, valid=True,
+                            tempfile_kwargs={'suffix': '.hs'})
+
+    def test_invalid(self):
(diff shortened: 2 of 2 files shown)


vijeth-aradhya:
@Makman2 I see you were also a part of the issue discussions :) I have successfully tested against his repository where it was previously failing. I think it should work fine now. Would you mind having a look at it? :)
Thanks!

bears/haskell/HaskellLintBear.py
```diff
@@ -38,11 +38,15 @@ def process_output(self, output, filename, file):
         output = json.loads(output)
 
         for issue in output:
-            assert issue['startLine'] == issue['endLine']
             diff = Diff(file)
+            from_lines = issue['from'].splitlines()
+            to_lines = issue['to'].splitlines()
+            assert len(from_lines) == len(to_lines)
+            for other_lines in range(1, len(from_lines)):
+                assert from_lines[other_lines] == to_lines[other_lines]
```
Makman2: you don't use "`other_lines`" laer, do you? You have `from_lines[0]` only, so I wouldn't `assert` here, is not necessary imo :)

bears/haskell/HaskellLintBear.py
```diff
@@ -38,11 +38,15 @@ def process_output(self, output, filename, file):
         output = json.loads(output)
 
         for issue in output:
-            assert issue['startLine'] == issue['endLine']
             diff = Diff(file)
+            from_lines = issue['from'].splitlines()
+            to_lines = issue['to'].splitlines()
```
Makman2: can't those fail? because when `issue['to']` or `issue['from']` is `None` like in the exception in the issue, it will fail because there's no `splitlines()` in `None`^^

vijeth-aradhya: If there is an `issue` in `output`, it will at least one line ^ ( IMHO it can't be `None`) - I checked with a file with no lint errors.

bears/haskell/HaskellLintBear.py
vijeth-aradhya: If _ONLY_ all the other lines in the error message are equal (which is the case most of the times, I haven't come across a case where it is NOT equal till now, but it is better to keep the `assert` just in case), then we can output the change which is just the first line ^^
Then, we can apply the patch and change only the first line :smile: 
Otherwise, if the other lines lines are different, then we _CANNOT_ apply the patch

vijeth-aradhya:
@sils Haven't received a lot of reviews here :stuck_out_tongue: This has been pending since a week :open_mouth: 
Does this look good though?

sils:
@vijeth-aradhya sadly a lot of PRs are pending :/ 31 ATM and it's getting worse every day.

Please do use the full URL in the commit message.

tests/haskell/HaskellLintBearTest.py
```diff
@@ -1,15 +1,46 @@
...
+
+SingleLineLintTest = verify_local_bear(HaskellLintBear,
+                                       valid_files=(good_single_line_file,),
+                                       invalid_files=(bad_single_line_file,),
+                                       tempfile_kwargs={'suffix': '.hs'})
+
+MultipleLinesTest = verify_local_bear(HaskellLintBear,
+                                      valid_files=(good_multiple_line_file,),
+                                      invalid_files=(bad_multiple_line_file,),
+                                      tempfile_kwargs={'suffix': '.hs'})
```
sils: can we have a proper test here that tests how the patch looks like?

bears/haskell/HaskellLintBear.py
```diff
@@ -38,11 +38,15 @@ def process_output(self, output, filename, file):
...
+            assert len(from_lines) == len(to_lines)
```
sils: why does the len have to be equal? I don't fully get this

vijeth-aradhya: The `patch` that we apply for correcting the error is only the for the first line of the whole multi line message. From what I have seen, the whole body ( that is message structure ) is the same except the first line where we actually make changes.

These assert statements are required to check that the whole message body ( structure ) is same, and that only the first line needs a patch ( diff ).

Otherwise the patch cannot be applied if there are multiple differences in the `from` and `to`, i.e actual code and linter suggested code.

I guess it'll be better understood once I add a proper testcase?!

tests/haskell/HaskellLintBearTest.py
vijeth-aradhya: Yes, I could do that :+1: :smile:

vijeth-aradhya:
@sils I'll give the testcase as the example, and try to explain.
The reason why there was a bug initially was because the diffs (patch) were always a **single line linting patch**, but in the testcase example I have just given ^, the linting message and the suggested fix by the linter was **multiple lines**
_Linter error code_
```
return $ BS.concat $
        [ BS.pack $ "I am being tested by hlint!"
        , "String dummy" , "Another String dummy" ]
```
_Suggested fix for that ^_
```
return BS.concat $
        [ BS.pack $ "I am being tested by hlint!"
        , "String dummy" , "Another String dummy" ]
```
_Actual way the code was written_
```
return $ BS.concat $
        [ BS.pack $ "I am being tested by hlint!"
        , "String dummy" 
        , "Another String dummy" 
        ]
```
Since there was ambiguity in the linter representation of the code, you cannot apply the suggested fix (patch) there ^ (since for the patch, exact lines were taken into consideration) - that is why only one line was actually tested before.
But, what I noticed is that, only the first line actually has the fix, and the other lines are actually the same ( _suggested fix_ and _linter error code_ ), in that case we can still apply the patch for that line, and it solves the problem!! :smile: 
That is why just to make sure, I have added assert statements to make sure the length of the _linter error code_ and _suggested fixes_ is same, and all the other lines are also same!
I checked with multiple files in his repository, and this workaround works! :smile: It's at least better than the previous version, I can assure you that :laughing: 

And, really sorry for such a long comment :P

bears/haskell/HaskellLintBear.py
vijeth-aradhya: I have made the tests better, can anything else be added to the tests?

vijeth-aradhya:
@Makman2 @sils Any updates on this? :)

vijeth-aradhya:
@Makman2 Reviews please! :)

Makman2:
ack 46ed356

vijeth-aradhya:
reack e4daf7e39fc5d2db6d3139351f1174fb44fa34ff

sils:
@rultor merge

PR #1277, opened 2017-01-08
MarkdownBear: Add remark-validate-links plugin

namanyadav12:
Remark plugin to validate if links to headings and files in markdown point to existing things.

Closes https://github.com/coala/coala-bears/issues/924

### DIFF
--- a/tests/markdown/MarkdownBearTest.py
+++ b/tests/markdown/MarkdownBearTest.py
@@ -1,5 +1,11 @@
+import unittest
+from queue import Queue
+
 from bears.markdown.MarkdownBear import MarkdownBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
+from coalib.testing.BearTestHelper import generate_skip_decorator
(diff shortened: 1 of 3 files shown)


tests/markdown/MarkdownBearTest.py
```diff
@@ -29,3 +45,22 @@
...
+
+
+@generate_skip_decorator(MarkdownBear)
+class MarkdownBearValidateLinksMessageTest(unittest.TestCase):
+
+    def setUp(self):
+        self.uut = MarkdownBear(Section('name'), Queue())
+
+    def test_validate_links_message(self):
+        content = test_file4.splitlines()
```
aptrishu: check test_file5 also..

namanyadav12: I am already checking it in the MarkdownValidateLinksTest, this is to check if the message of the result is correct.

jayvdb:
The bear is broken.  Enhancing the bear now is *not* an option.  It needs to be fixed first: https://github.com/coala/coala-bears/issues/1235

PR #1286, opened 2017-01-10
bear-requirements: Do not sort the file

adtac:
This way, we can preserve the order in which a bear's requirements
are given.


Mariatta:
Any idea why travis build failed?

adtac:
Because the generated bear-requirements.txt file is different from the given one. And that happens because it's no longer sorted and the order is non-deterministic.

And I have no idea what I was thinking - we can't actually have order in a bear's requirements because it's a set. Unless we change it to a list/tuple, we can't really do anything :/

@sils maybe adding click explicitly in the requirement.txt (instead of bear-requirements.txt) is the way to go. This is a temporary workaround, so I guess we can go with it for a while.

adtac:
So this PR is kinda invalid. I'll close it.

PR #1288, opened 2017-01-10
WIP ImageDimensionBear: Add ImageDimensionBear to coala-bears

rhemon:
I have made the bear and written unittest for the bear as well. I am doing this PR to confirm what I have done till now is fine. 
Another reason for making this WIP is because I still have to handle files for the bear. This is a GlobalBear and doesnt need any file, however I am unable to run the bear without a file. Presently .coafile has to look something like this to run ImageDimensionBear: 
```
[default]
bears = ImageDimensionBear
files = .coafile
img_config = src/*.png 240 240, img/*.jpg 400 400
```
now here I think `files` is unnecessary and want to remove this but not sure how to. 

Herer's how the it looks when I run the bear: http://pastebin.com/UebXyuiE
& the Test: 
```
(venv) MBPs-MacBook-Pro:coala-bears rhemon$  /Users/mbp13inch/Documents/Workspace/venv/bin/py.test tests/general/ImageDimensionBearTest.py
================================================== test session starts ===================================================
platform darwin -- Python 3.5.2, pytest-3.0.5, py-1.4.32, pluggy-0.4.0
rootdir: /Users/mbp13inch/Documents/Workspace/coala-bears, inifile: setup.cfg
plugins: cov-2.4.0, env-0.6.0, timeout-1.2.0, xdist-1.15.0
timeout: 35.0s method: signal
collected 4 items 

tests/general/ImageDimensionBearTest.py ....

=============================================== 4 passed in 29.30 seconds ================================================
```


jayvdb:
Regarding your general problem of `files`, we still do want to use `files` here.
Currently you have one `.coafile` setting `image_config`, which include multiple pieces of information, including filenames.

So the first step is splitting `image_config` into multiple settings.

```diff
-    def run(self, image_config):
+    def run(self, files, image_width, image_height):
```

The benefit is that the user only needs to set `files` and two numerical values.  nice and simple to configure.
(Of course this means your job is harder, as you need to convert `files` into the linters yaml, but .. that is the job of a bear -- making linting easy)

This approach has one weakness compared to your `image_config` strategy, which is that each `.coafile` section can only have one width & height.

The result is that the user needs to create multiple `.coafile` sections for different sets of images.
The 'cost' is multiple invocations of the ruby interpreter; one for each section.

Gemfile
```diff
@@ -7,3 +7,5 @@ gem 'scss_lint', require: false# require flag is necessary https://github.com/br
 gem "reek"
 gem "puppet-lint"
 gem "csvlint"
+gem "img_checker"
+gem "fastimage"
```
jayvdb: why do we need this?  doesnt img_checker declare a dependency on fastimage?

rhemon: not sure, but I was having problem in running it, so I added it, and then it started working

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,53 @@
...
+
+        with open('img_config.yml', 'w') as yaml_file:
+            for each in image_config:
+                each = each.split(' ')
+                config = [{'directory': each[0],
+                           'width': int(each[1]),
+                           'height': int(each[2])}]
+                yaml.dump(config, yaml_file, default_flow_style=False)
+
+        cmd = 'ruby -r "img_checker.rb" -e '
```
jayvdb: no need for double quotes, as there are no space characters in the filename.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,53 @@
...
+class ImageDimensionBear(GlobalBear):
+    """
+    Checks the dimension of an image.
+
+    More information is available at <github.com/Abhi2424shek/img_checker>
+    """
+
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    REQUIREMENTS = {GemRequirement(' img_checker'),
```
jayvdb: This workaround is OK for a WIP, but you need to solve the bug in `GemRequirement` before this bear will be merge-able.

rhemon: ok, I have submitted a PR for that https://gitlab.com/coala/package_manager/merge_requests/10

robbyoconnor:
FYI I made changes in https://github.com/Abhi2424shek/img_checker/pull/22 -- soon as @Abhi2424shek merges and it deploys. 

It added an executable. so you type `img_checker` and it will just work(tm).

Gemfile
robbyoconnor: You don't need to add `fast_image`.

bears/general/ImageDimensionBear.py
robbyoconnor: you should just be able to use the ruby gem and run `img_checker`

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,53 @@
...
+        with open('img_config.yml', 'w') as yaml_file:
+            for each in image_config:
+                each = each.split(' ')
+                config = [{'directory': each[0],
+                           'width': int(each[1]),
+                           'height': int(each[2])}]
+                yaml.dump(config, yaml_file, default_flow_style=False)
+
+        cmd = 'ruby -r "img_checker.rb" -e '
+        cmd += '"ImgChecker.new.initialize img_config.yml"'
```
robbyoconnor: Not needed since soon as https://github.com/Abhi2424shek/img_checker/pull/22 is merged -- it should "just work"

rhemon: is that also updated in the gem? it would be  easier if so. will do the changes once its merged. thanks for informing :)

robbyoconnor: I didn't up the version number, that's for them to decide. I just cleaned things up.

rhemon:
@sils @jayvdb all test except the appveyor one (test in windows) didn't pass. my guess is that this is happening because of the space that was missing in dependency_management, but the fix is done and the PR was merged. now it only needs to have a release. I hope that will have that test pass as well.

rhemon:
the documentation goes to [bear-docs](https://github.com/coala/bear-docs/) right?

sils:
no, put it in the documentation comment, bear docs are generated from that

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

rhemon:
oh ok cool, thanks for informing, didint know that @sils

jayvdb:
What will happen if the repository already has a `img_config.yml` ?  Does the linter have a command line flag for specifying the config file?

jayvdb:
(sorry for closing; wrong button)

rhemon:
did not think of that @jayvdb 
is there any way of doing that?

rhemon:
```
(venv) MBPs-MacBook-Pro:coala-bears rhemon$ ruby -r "img_checker.rb" -e "ImgChecker.new.initialize just-a-file.yml"
Critical Error: img_config.yml not found. Please add to the current directory.
(venv) MBPs-MacBook-Pro:coala-bears rhemon$ img_checker img_config.yml
Critical Error: img_config.yml not found. Please add to the current directory.
```
presently it doesn't seem like its accepting other file, should I ask the authors of gemfile if they can add this functionality?

jayvdb:
> should I ask the authors of gemfile if they can add this functionality?

Yes, it is a good idea, but not a high priority.  Then we can create the config file with a random name which doesnt exist, to avoid conflicts.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,70 @@
...
+    CAN_FIX = {'Image Dimension'}
+
+    @classmethod
+    def check_prerequisites(cls):
+        if shutil.which('img_checker') is None:
+            return 'img_checker is not installed.'
+        return True
+
+    def run(self,
+            image_file,
```
jayvdb: you need to define the types of these parameters using Python 3 syntax.
See https://github.com/coala/coala-bears/blob/master/bears/coffee_script/CoffeeLintBear.py#L63 for an example

rhemon: oh ok, thanks

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,68 @@
...
+                             'img_checker is not installed.')
+
+            shutil.which = lambda *args, **kwargs: 'path/to/bundle'
+            self.assertTrue(ImageDimensionBear.check_prerequisites())
+        finally:
+            shutil.which = _shutil_which
+
+    def test_run_with_png(self):
+        message = "message='The image ./test-img/img.png is larger"
+        message += " than 240px x 240px [w x h]'"
```
yukiisbored: You can use parantheses to combine two strings while breaking them into two lines, like this:
```python
message=('Hello'
                  'World')
```

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,68 @@
...
+                               image_file='./test-img/*.jpg',
+                               width=50,
+                               height=50))))
+
+    def test_run_with_both(self):
+        message_png = "message='The image ./test-img/img.png is larger"
+        message_png += " than 50px x 50px [w x h]'"
+        message_jpg = "message='The image ./test-img/images.jpg is larger"
+        message_jpg += " than 50px x 50px [w x h]'"
+        output = str(list(self.idb.run(image_file='./test-img/*.*',
```
yukiisbored: Why not have a regex that eliminates the differences and decrease code duplication? You can use `assertRegex` to check the results.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,70 @@
...
+                       'width': width,
+                       'height': height}]
+            yaml.dump(config, yaml_file, default_flow_style=False)
+        cmd = 'img_checker'
+        output = run(cmd, stdout=Capture(), stderr=Capture())
+        if (output.returncode):
+            lines = output.stdout.text.split('\n')[1:-2]
+            for line in lines:
+                if '.png' in line:
+                    fileName = line[10:line.index('.png')+4]
```
yukiisbored: I think it's preferable to use `filename` instead of `fileName` because the code style we're using is `snake_case` not `camelCase`.

rhemon: didn't know this but will do the change, first let me fix the problem with the test fail in appveyor. also if possible please help me understand the error,

rhemon: i expected the problem to be fixed with the new release of dependency_management but surprisingly it didn't

tests/general/ImageDimensionBearTest.py
yukiisbored: Also use single quote `'` instead of a double `"`

bears/general/ImageDimensionBear.py
Abhi2424shek: Yep as @yukiisbored has suggested pls implement snake_case for variables, filenames etc and camelCase for class names.

rhemon:
@yukiisbored @Abhi2424shek thanks for the review, will do those changes, but can you help find the reason behind the build fail in appveyour please? why is it saying img_checker not installed? is it not installing everything from Gemfile properly?

bears/general/ImageDimensionBear.py
yukiisbored: @Abhi2424shek for class names, use `PascalCase`

yukiisbored:
@rhemon I don't have a Windows machine. @Abhi2424shek Can you try it?

Abhi2424shek:
@yukiisbored i am on heavy loads multitasking, give me half an hour pls

Abhi2424shek:
It says `img_checker is not installed.` I did'nt go into deep, can you check as if you had added it properly.

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,68 @@
...
+        message += " than 240px x 240px [w x h]'"
+        self.assertIn(message,
+                      str(list(self.idb.run(
+                               image_file='./test-img/*.png',
+                               width=240,
+                               height=240))))
+
+    def test_run_with_jpg(self):
+        self.assertEqual('[]',
+                         str(list(self.idb.run(
```
yukiisbored: I think you don't need `str(..)` for this one

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,68 @@
...
+                                  image_file='./test-img/*.jpg',
+                                  width=240,
+                                  height=240))))
+
+    def test_run_with_jpg_fail(self):
+        message = "message='The image ./test-img/images.jpg is larger"
+        message += " than 50px x 50px [w x h]'"
+        self.assertIn(message,
+                      str(list(self.idb.run(
```
yukiisbored: Isn't better to just grab the value from Result instead of converting it into a string?

rhemon:
@Abhi2424shek I am sure its added properly, otherwise it wouldn't have passed the travis test, did you test it locally in your machine and got that? or are you saying from the log of appveyor?

rhemon:
I know very little about gems, windows and cmd so not sure https://github.com/coala/coala-bears/blob/master/.ci/appveyor.yml but is this even running command to install the gems ? I tried doing this 
```
(venv) MBPs-MacBook-Pro:coala-bears rhemon$ git grep "Gemfile" ./
.gitignore:Gemfile.lock
.travis.yml:  # Remove Ruby directive from Gemfile as this image has 2.2.5
.travis.yml:  - sed -i '/^ruby/d' Gemfile
```
doesn't seem like its being used for appveyor, but then again it might be different so I am not sure

yukiisbored:
@rhemon Use `generate_skip_decorator ` to fix it. http://api.coala.io/en/latest/_modules/coalib/testing/BearTestHelper.html?highlight=skip%20decorator

rhemon:
@yukiisbored thanks a lot. you are awesome. you are the best. thanks a lot. (although I wish it was a littler earlier, could have get this done and take another task :P ). still thanks a lot. you have no idea the amount of stress you removed from me. seeing the appveyour pass is making me feel so good.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,70 @@
...
+    CAN_FIX = {'Image Dimension'}
+
+    @classmethod
+    def check_prerequisites(cls):
+        if shutil.which('img_checker') is None:
+            return 'img_checker is not installed.'
+        return True
+
+    def run(self,
+            image_file: str='',
```
jayvdb: what will you do if you receive the default value of ''?
If you specify a default, it is an optional setting.  If you do not specify a default, the setting is mandatory - the user must set it.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,70 @@
...
+            width: int=0,
```
jayvdb: what is the use of default `0`?  wont that mean every image is an error?

And what happens if the user sets this to `0`, or `-1`?
Look at other bears, and how they validate the user settings to reject bad settings.
Also if the docs are not good enough, create bugs to improve the docs.

bears/general/ImageDimensionBear.py
rhemon: ok didnt think of that, will remove the default value.

rhemon:
Added warning about img_config.yml file as well :
```
(venv) MBPs-MacBook-Pro:coala-bears rhemon$ coala -c ./test-img/.coafile --ci
Executing section Default...
[WARNING][10:13:44] WARNING: If there is any img_config.yml file in your root directory, this bear will remove it
[WARNING][10:13:47] ImageDimensionBear: This result has no patch attached.

Project wide:
|    | [MAJOR] ImageDimensionBear:
|    | The image ./test-img/img.png is larger than 240px x 240px [w x h]
```

yukiisbored:
@rhemon I think it shouldn't remove the existing configuration file, coala isn't supposed to modify the existing files.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,87 @@
...
+                           for images
+        """
+
+        warning_message = ('If there is any img_config.yml file'
+                           ' in your root directory, this bear'
+                           ' will remove it')
+
+        self.warn(warning_message)
+
+        if (glob(image_file) == []):
```
jayvdb: These validators should be separate functions.  See `known_checkstyle_or_path` in `CheckstyleBear`.

bears/general/ImageDimensionBear.py
```diff
@@ -0,0 +1,87 @@
...
+
+        :param image_file: The file/directory that bear
+                           will look for images
+        :param width:      The maximum width value allowed
+                           for images
+        :param height:     The maximum height value allowed
+                           for images
+        """
+
+        warning_message = ('If there is any img_config.yml file'
```
jayvdb: This warning should only appear *when* there is a `img_config.yml`.

But ... more importantly ... you can work around this problem in `img_checker` by renaming `img_config.yml` here and then restoring it after `img_checker` has finished.

rhemon:
ok so same error in appveyor and travis. and it seems like previously few appveyor builds had the same error, so i guess another issue which needs to be fixed.

robbyoconnor:
@rhemon There is a conflict you need to resolve :)

Abhi2424shek:
It's been sometime @rhemon any updates bro

rhemon:
@Abhi2424shek well this pr is kinda done but in halt because of the problem in appveyor (it needs to be fixed so that it can be confirmed this bear works fine in windows as well). I am trying to make it install ruby so that it can bear tests that use ruby. https://github.com/coala/coala-bears/pull/1312. But I am running into lots of problems.

Abhi2424shek:
I can see a tick in appveyor @rhemon

test-img/.coafile
```diff
@@ -0,0 +1,6 @@
+[Default]
+bears = ImageDimensionBear
+files = .coafile
+image_file = ./test-img/*.*
+width = 240
+height = 240
```
yukiisbored: missing `\n` on EOL

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,91 @@
+import unittest
+from unittest.case import skip, skipIf
+from queue import Queue
+import shutil
+
+from bears.general.ImageDimensionBear import *
```
yukiisbored: Please don't use `import *`

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    def setUp(self):
+        self.section = Section('')
+        self.queue = Queue()
+        self.file_dict = {}
+        self.idb = ImageDimensionBear(None, self.section, self.queue)
+
+    def test_check_width(self):
+        with self.assertRaisesRegexp(ValueError, 'Width'):
```
yukiisbored: Please use the full error message `Width has to be greater than` to actually check if it's the actual error.

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,91 @@
...
+            check_width('-12')
+        self.assertEqual(check_width('120'), 120)
+
+    def test_check_height(self):
+        with self.assertRaisesRegexp(ValueError, 'Height'):
```
yukiisbored: See https://github.com/coala/coala-bears/pull/1288/files#r99728966

tests/general/ImageDimensionBearTest.py
```diff
@@ -0,0 +1,91 @@
...
+            check_height('-12')
+        self.assertEqual(check_height('120'), 120)
+
+    def test_check_path(self):
+        with self.assertRaisesRegexp(ValueError, 'Provide'):
```
yukiisbored: See https://github.com/coala/coala-bears/pull/1288/files#r99728966

tests/general/ImageDimensionBearTest.py
yukiisbored: Please don't use `import *`

Abhi2424shek:
@jayvdb as known from slack @rhemon is on a tuzzle, so pls give him more time :) I hope you'd be cool with it.

tests/general/ImageDimensionBearTest.py
robbyoconnor: Yeah don't do this @rhemon

Abhi2424shek:
@rhemon  any updates it's been long.

yukiisbored:
@Abhi2424shek People are busy with life, career, family, exams, social, and some other human stuff. Please don't ping them especially when the fact we already have gitmate reminding the author. I personally don't mind having a long time just to implement this new bear into the coala-bears lifecycle because it's done when it's ready, we're not a corporation with fixed and small deadlines.

rhemon:
yup as yuki said i have AS exams (ial), so my parents banned me from coding
(my fault cuz i did BAD in the mocks) and freedom :p  till exams are over
which will be a long time (till june). but i promise to get back at it
after exams, so please allow this and the other prs that i was working on
to be halt till then if its ok

yukiisbored:
Sure, just do it when you have the time. We don't want your grades to be bad just because of this ;). Just prioritize school for the time being.

jayvdb:
https://github.com/coala/coala/pull/3560 is merged.  This bear can now proceed.

yukiisbored:
Please make sure the images used for testing aren't copyrighted. I recommend creating your own images, I think an blank image is enough for this bear.

PR #1297, opened 2017-01-12
InvalidLinkBear: Reduce severity

satwikkansal:
Reduces the severity of the XML namespaces which are
present in xmlns attributes of the file. If the
same namespace link occurs more than once in the file,
it should not be reported.

Closes https://github.com/coala/coala-bears/issues/1239

### DIFF
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -80,17 +80,25 @@ class InvalidLinkBear(LocalBear):
                                         # Unbalanced parenthesis
             (?<!\.)(?<!,)               # Exclude trailing `.` or `,` from URL
             """, re.VERBOSE)
-
+        file_context = {}
         for line_number, line in enumerate(file):
+            xmlns_regex = re.compile(r'xmlns:?\w*="(.*)"')
             for match in re.findall(regex, line):
                 link = match[0]
-                context = enum(
-                    pip_vcs_url=False)
-                if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
-                    context.pip_vcs_url = True
+                link_context = file_context.get(link)
+                if not link_context:
+                    link_context = enum(
+                        xml_namespace=False,
+                        pip_vcs_url=False)
+                    xmlns_match = xmlns_regex.search(line)
+                    if xmlns_match and link in xmlns_match.groups():
+                        link_context.xml_namespace = True
+                    if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
+                        link_context.pip_vcs_url = True
+                    file_context[link] = link_context
                 if not (link_ignore_regex.search(link) or
                         fnmatch(link, link_ignore_list)):
-                    yield link, line_number, context
+                    yield link, line_number, link_context
 
     def analyze_links_in_file(self, file, network_timeout, link_ignore_regex,
                               link_ignore_list):
@@ -149,7 +157,18 @@ class InvalidLinkBear(LocalBear):
 
         for line_number, link, code, context in self.analyze_links_in_file(
                 file, network_timeout, link_ignore_regex, link_ignore_list):
-            if code is None:
+            if context.xml_namespace:
+                if code and 200 <= code < 300:
+                    pass
+                else:
+                    yield Result.from_values(
+                        origin=self,
+                        message=('XML Namespace - '
+                                 '{url}').format(url=link),
+                        file=filename,
+                        line=line_number,
+                        severity=RESULT_SEVERITY.INFO)
+            elif code is None:
                 yield Result.from_values(
                     origin=self,
                     message=('Broken link - unable to connect to '
--- a/bears/general/InvalidLinkBear.py
+++ b/bears/general/InvalidLinkBear.py
@@ -6,6 +6,7 @@ from difflib import SequenceMatcher
 
 from coalib.results.Diff import Diff
 from coalib.bears.LocalBear import LocalBear
+from coalib.misc.Enum import enum
 from dependency_management.requirements.PipRequirement import PipRequirement
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.results.Result import Result
@@ -50,8 +51,7 @@ class InvalidLinkBear(LocalBear):
         return splitted_schema
 
     @staticmethod
-    def find_links_in_file(file, network_timeout, link_ignore_regex,
-                           link_ignore_list):
+    def extract_links_from_file(file, link_ignore_regex, link_ignore_list):
         link_ignore_regex = re.compile(link_ignore_regex)
         regex = re.compile(
             r"""
@@ -80,22 +80,35 @@ class InvalidLinkBear(LocalBear):
                                         # Unbalanced parenthesis
             (?<!\.)(?<!,)               # Exclude trailing `.` or `,` from URL
             """, re.VERBOSE)
+
         for line_number, line in enumerate(file):
             for match in re.findall(regex, line):
                 link = match[0]
+                context = enum(
+                    pip_vcs_url=False)
+                if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
+                    context.pip_vcs_url = True
                 if not (link_ignore_regex.search(link) or
                         fnmatch(link, link_ignore_list)):
-                    if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
-                        link = InvalidLinkBear.parse_pip_vcs_url(link)
-                    host = urlparse(link).netloc
-                    code = InvalidLinkBear.get_status_code(
-                               link,
-                               network_timeout.get(host)
-                               if host in network_timeout
-                               else network_timeout.get('*')
-                               if '*' in network_timeout
-                               else InvalidLinkBear.DEFAULT_TIMEOUT)
-                    yield line_number + 1, link, code
+                    yield link, line_number, context
+
+    def analyze_links_in_file(self, file, network_timeout, link_ignore_regex,
+                              link_ignore_list):
+        for link, line_number, link_context in self.extract_links_from_file(
+                file, link_ignore_regex, link_ignore_list):
+
+            if link_context.pip_vcs_url:
+                link = InvalidLinkBear.parse_pip_vcs_url(link)
+
+            host = urlparse(link).netloc
+            code = InvalidLinkBear.get_status_code(
+                link,
+                network_timeout.get(host)
+                if host in network_timeout
+                else network_timeout.get('*')
+                if '*' in network_timeout
+                else InvalidLinkBear.DEFAULT_TIMEOUT)
+            yield line_number + 1, link, code, link_context
 
     @deprecate_settings(link_ignore_regex='ignore_regex',
                         network_timeout=('timeout', lambda t: {'*': t}))
@@ -134,7 +147,7 @@ class InvalidLinkBear(LocalBear):
                            if not url == '*' else '*': timeout
                            for url, timeout in network_timeout.items()}
 
-        for line_number, link, code in InvalidLinkBear.find_links_in_file(
+        for line_number, link, code, context in self.analyze_links_in_file(
                 file, network_timeout, link_ignore_regex, link_ignore_list):
             if code is None:
                 yield Result.from_values(
--- a/tests/general/InvalidLinkBearTest.py
+++ b/tests/general/InvalidLinkBearTest.py
@@ -7,6 +7,7 @@ import unittest
 import unittest.mock
 
 from bears.general.InvalidLinkBear import InvalidLinkBear
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.settings.Section import Section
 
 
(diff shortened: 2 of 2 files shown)


tests/general/InvalidLinkBearTest.py
```diff
@@ -128,7 +128,11 @@ def test_run(self):
         http://httpbin.org/status/503
         http://www.notexample.com/404
         http://exampe.com/404
-        http://example.co.in/404"""
+        http://example.co.in/404
+        <ruleset name="test" xmlns="http://this.isa.namespace/ruleset/7.0.0"
+        xmlns:xsi="http://this.is.another/kindof/namespace"
+        xsi:schemaLocation="http://this.namespace.dosent/exists/7.0.0"
```
pratyushprakash: IMO there should be a check for the severity

satwikkansal: If that's the case, I guess we should create a separate issue for that because none of the tests are written yet to check the severity.

pratyushprakash: It should be fairly simple. I don't think there is a need to create an issue

tests/general/InvalidLinkBearTest.py
```diff
@@ -128,7 +128,11 @@ def test_run(self):
...
+        xsi:schemaLocation="http://httpbin.com/404">"""
```
pratyushprakash: This is just ensuring there is 100% coverage. Since the PR deals with the severity of the result I feel it should be tested.

satwikkansal:
I always forget to amend my commit after fixing PEP8 issues :(

pratyushprakash:
ack fcd02b1373fae90a7cf9131ceb07c2a418de52bd

bears/general/InvalidLinkBear.py
```diff
@@ -80,19 +81,25 @@ def find_links_in_file(file, network_timeout, link_ignore_regex,
                                         # Unbalanced parenthesis
             (?<!\.)(?<!,)               # Exclude trailing `.` or `,` from URL
             """, re.VERBOSE)
+        xml_namespace_links = set()
```
jayvdb: I'm sure there will be other cases like this. Create an enum for 'first context' of each url, and start it with only 'xml namespace'.

bears/general/InvalidLinkBear.py
```diff
@@ -80,19 +81,25 @@ def find_links_in_file(file, network_timeout, link_ignore_regex,
...
             if match:
                 link = match.group()
+                xmlns_match = xmlns_regex.search(line)
+                if xmlns_match and link in xmlns_match.groups():
+                    xml_namespace_links.add(link)
+                    yield line_number + 1, link, '', True
                 if not (link_ignore_regex.search(link) or
-                        fnmatch(link, link_ignore_list)):
+                        fnmatch(link, link_ignore_list) or
+                        link in xml_namespace_links):
```
jayvdb: argh. This should be happening already for all links. See https://github.com/coala/coala-bears/issues/1319

We should actually report additional occurrences, but re-use the initial context of the url so that extra occurences are a lower severity.

bears/general/InvalidLinkBear.py
```diff
@@ -80,19 +81,25 @@ def find_links_in_file(file, network_timeout, link_ignore_regex,
...
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
                         link = InvalidLinkBear.parse_pip_vcs_url(link)
                     host = urlparse(link).netloc
                     code = InvalidLinkBear.get_status_code(
                                link, network_timeout.get(
                                    host, InvalidLinkBear.DEFAULT_TIMEOUT))
-                    yield line_number + 1, link, code
+                    yield line_number + 1, link, code, False
```
jayvdb: Return the 'context' enum in a dict, instead of a bool, so it is extensible.

pratyushprakash:
unack fcd02b1373fae90a7cf9131ceb07c2a418de52bd

bears/general/InvalidLinkBear.py
```diff
@@ -80,14 +82,22 @@ def find_links_in_file(file, network_timeout, link_ignore_regex,
...
+                if not context.get(link):
+                    context[link] = enum(xml_namespace=False)
+                xmlns_match = xmlns_regex.search(line)
+                if xmlns_match and link in xmlns_match.groups():
+                    context[link].xml_namespace = True
                 if not (link_ignore_regex.search(link) or
                         fnmatch(link, link_ignore_list)):
                     if link.startswith(('hg+', 'bzr+', 'git+', 'svn+')):
-                        link = InvalidLinkBear.parse_pip_vcs_url(link)
+                        parsed_link = InvalidLinkBear.parse_pip_vcs_url(link)
```
jayvdb: just a note : this can become another context , e.g. "pip_vcs", in the future (not this PR).

satwikkansal: I'll create an issue to identify other possible contexts and implement them.

sils:
So I do think we should simplify this code.  Also "Reduce Severity" doesn't really describe it, does it?

> If the
> satme namespace link occurs more than once in the file,
> it should not be reported.

that almost sounds like an own commit.

Can we maybe have like one function that takes all the lines and yields all the links with the protocol and their sourceposition and stuff? Maybe group links belonging together? I'm not fully sure but it seems like this could enefit from *some* simplification.

satwikkansal:
CircleCI is failing due to `cabal`

satwikkansal:
@sils I've simplified the code and split the changes into 2 commits, I guess it's ready to be merged after `CircleCI` stops displaying the strange `cabal` installation error :(

satwikkansal:
All tests are passing :smile:

tests/general/InvalidLinkBearTest.py
satwikkansal: Tests added

tests/general/InvalidLinkBearTest.py
satwikkansal: +1

bears/general/InvalidLinkBear.py
satwikkansal: :+1:

bears/general/InvalidLinkBear.py
satwikkansal: Okay

bears/general/InvalidLinkBear.py
satwikkansal: Done

bears/general/InvalidLinkBear.py
```diff
@@ -80,22 +80,41 @@ def find_links_in_file(file, network_timeout, link_ignore_regex,
...
-                    yield line_number + 1, link, code
+                    yield link, line_number, context
+
+    def analyze_links_in_file(self, file, network_timeout, link_ignore_regex,
+                              link_ignore_list):
+        for link, line_number, link_context in self.extract_links_from_file(
+                file, link_ignore_regex, link_ignore_list):
+
+            if link_context.pip_vcs_url:
+                parsed_link = InvalidLinkBear.parse_pip_vcs_url(link)
```
jayvdb: this should be `link = ...`

tests/general/InvalidLinkBearTest.py
```diff
@@ -139,11 +161,41 @@ def test_run(self):
...
+        self.assertSeverity(info_severity_file, RESULT_SEVERITY.INFO)
+
+        normal_severity_file = '''
+        http://httpbin.org/status/404
+        http://httpbin.org/status/410
+        http://httpbin.org/status/500
+        http://httpbin.org/status/503
+        http://httpbin.org/status/301
+        http://httpbin.org/status/302
+        xsi:schemaLocation="http://httpbin.com/404
```
jayvdb: a closing `"` would be helpful

satwikkansal:
@jayvdb Implemented the changes you suggested. I had a doubt btw, 
If we detect a url as an xml namespace and assign it a context of `xml_namespace`, Now shall the every occurrence of this particular url be yielded as low severity (even the ones which are not part of `xmlns:` or `xsi:` tags)?

tests/general/InvalidLinkBearTest.py
jayvdb: Done. Thanks.

tests/general/InvalidLinkBearTest.py
jayvdb: Resolved. Thanks.

jayvdb:
@satwikkansal, yes ... that context is global (whereas as currently your pip vcs context is not global - that can be fixed).  If a link has been found to be a schema, that automatically says the link may be valid even if it is a 404 - the link doesnt need to have http content for its *intended* use.  It is a declared namespace - that is all the URI is intended to be.  If it is mentioned anywhere in a README or any technical documentation, it is still valid.  This presents a problem for the per-file non-global design of this bear, as it means files processed before the xml namespace is seen will be errors while those after are not.  We can defer that problem to a separate issue, if you like.  The main problem is removing the inter-file errors.

satwikkansal:
@jayvdb Rebased. I guess it's ready to be merged.

Regarding the global context, IMO, we can have a preprocessing stage where the  bear only populates the context of all the links across the files it is run on, and after that, check the status and then generate results based on the status and the context.
Shall this be a part of https://github.com/coala/coala-bears/issues/1319 of shall I create a separate issue for this.

tests/general/InvalidLinkBearTest.py
```diff
@@ -86,6 +87,24 @@ def assertResultCount(self, test_file, expected_num):
...
+        """
+        Test the warnings in each line of the file to match the
+        given severity.
+        :param file: The ``file`` to be checked.
+        :param severity: The severity level of the warnings in each
+                         line of the file.
+        """
+        severity_tag = RESULT_SEVERITY.reverse[severity]
+        with requests_mock.Mocker() as m:
+            InvalidLinkBear.check_prerequisites = lambda *args: True
```
jayvdb: I think this should be removed.
I assume it is for performance reasons.  Create another bug about that ;-)

tests/general/InvalidLinkBearTest.py
```diff
@@ -139,11 +166,44 @@ def test_run(self):
...
+        xsi:schemaLocation="http://httpbin.com/404">"""
 
         for line in invalid_file.splitlines():
             self.assertResult(invalid_file=[line])
 
+    def test_severity(self):
+        info_severity_file = '''
+        <ruleset name="test" xmlns="http://this.isa.namespace/ruleset/7.0.0"
+        xmlns:xsi="http://this.is.another/kindof/namespace">
+        <ruleset name="test" xmlns="http://this.isa.namespace/ruleset/7.0.0"
```
jayvdb: As the parser is context aware, using the same URL multiple times in one test sample input is not as effective as using different URLs.  I think the URL in this `<ruleset>` should be slightly different - maybe different number, or slightly different domain name.

tests/general/InvalidLinkBearTest.py
```diff
@@ -139,11 +166,44 @@ def test_run(self):
...
+        xsi:schemaLocation="http://this.namespace.dosent/exists/7.0.0"
+        xsi:schemaLocation="http://httpbin.com/404">"""
 
         for line in invalid_file.splitlines():
             self.assertResult(invalid_file=[line])
 
+    def test_severity(self):
+        info_severity_file = '''
+        <ruleset name="test" xmlns="http://this.isa.namespace/ruleset/7.0.0"
+        xmlns:xsi="http://this.is.another/kindof/namespace">
```
jayvdb: please close the XML tag using `/>`.
the fragment will still not be a valid document, but at least the fragment is not invalid XML.

tests/general/InvalidLinkBearTest.py
```diff
@@ -139,11 +166,44 @@ def test_run(self):
...
+        http://httpbin.org/status/503
+        http://httpbin.org/status/301
+        http://httpbin.org/status/302
+        xsi:schemaLocation="http://httpbin.com/404"
+        '''.splitlines()
+
+        self.assertSeverity(normal_severity_file, RESULT_SEVERITY.NORMAL)
+
+        major_severity_file = '''
+        http://coalaisthebest.com
```
jayvdb: Using `http://this.isa.namespace/` , and putting it at the top of this test method, would help the reader understand.

tests/general/InvalidLinkBearTest.py
satwikkansal: Do you want me to remove complete `assertSeverity` function or just the single line.

satwikkansal:
@jayvdb Addressed your suggestions. Also, shifted the severity tests to a separate commit as they were getting confusing.

> Also can we have a test case which is exactly the scenario in the issue being fixed, so we dont have a regression somehow.

Done. Added all the relevant test cases in `test_xml_namespaces` method.

> Yes, global scope should be a separate issue.

Okay. I'll create an issue for global scope once this commit related to link contexts https://github.com/coala/coala-bears/pull/1297/commits/54eeec270e31cb0330a082f72494b21aa08b0f51 gets merged.

jayvdb:
ack a5a2fb3

jayvdb:
ack bbd5d33 7f99537 9306520

jayvdb:
@rultor merge

PR #1299, opened 2017-01-13
InvalidLinkBear: Multiple links

satwikkansal:
The InvalidLinkBear should analyze all the links
in a line instead of just analyzing the first one.

Fixes https://github.com/coala/coala-bears/issues/1296

### DIFF
--- a/tests/general/InvalidLinkBearTest.py
+++ b/tests/general/InvalidLinkBearTest.py
@@ -77,6 +77,15 @@ class InvalidLinkBearTest(unittest.TestCase):
                 self.assertNotEqual(out, [])
                 self.assertNotEqual(out, None)
 
+    def assertResultCount(self, test_file, expected_num):
+        with requests_mock.Mocker() as m:
+            InvalidLinkBear.check_prerequisites = lambda *args: True
+            uut = InvalidLinkBear(self.section, Queue())
+            m.add_matcher(custom_matcher)
+            for line, num in zip(test_file, expected_num):
+                outputs = list(uut.run('testline', [line]))
+                self.assertEqual(num, len(outputs))
+
     def test_run(self):
         # Valid Links
         valid_file = """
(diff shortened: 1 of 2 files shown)


Asnelchristian:
needs rebase

satwikkansal:
@Asnelchristian @Nosferatul Rebased!

tests/general/InvalidLinkBearTest.py
```diff
@@ -126,6 +126,7 @@ def test_run(self):
         http://httpbin.org/status/410
         http://httpbin.org/status/500
         http://httpbin.org/status/503
+        http://httpbin.org/status/200 http://httpbin.org/status/404
```
jayvdb: To test properly the second url in the line needs to be a new invalid link,
And that will mean the test will fail until you add the new link in the expected results below.

jayvdb: Ugh, actually the logic of this test wont help verify that there were multiple results. Create a new test method please.

tests/general/InvalidLinkBearTest.py
```diff
@@ -154,6 +171,21 @@ def test_redirect_threshold(self):
                           invalid_file=short_url_redirect,
                           settings={'follow_redirects': 'yeah'})
 
+    def test_multiple_results_per_line(self):
+
+        file = """
```
jayvdb: I am not a fan of using 'file' as a variable name, as it is a [builtin function](https://docs.python.org/2/library/functions.html?highlight=file#file) name in py2 only.

tests/general/InvalidLinkBearTest.py
```diff
@@ -77,6 +77,22 @@ def assertResult(self, valid_file=None, invalid_file=None, settings={}):
                 self.assertNotEqual(out, [])
                 self.assertNotEqual(out, None)
 
+    def assertResultCountPerLine(self, line, expected_num):
```
jayvdb: The name suffix  '....PerLine' does not make sense.
There is only one line.
Also it is strange that `line` is an array.

Either the mentod should only accept a single line, as a `str`, with one expected line count.
Or, accept and process multiple lines, and compare against a list of expected line counts.

tests/general/InvalidLinkBearTest.py
jayvdb: please remove this line. If multi-line breaks, this test method should pass, and `test_multiple_results_per_line` should fail.

sils:
verb missing in commit message -> see coala.io/commit

sils:
@satwikkansal the code is very nice so I don't see being nitpicky about the commit message being of big help - that said: https://github.com/coala/coala-bears/pull/1299/commits/2ef96a9dc35793d0584660a97100b7c40c1d956a is ready

sils:
@rultor merge

PR #1304, opened 2017-01-14
HTMLLintBear: Add settings for individual features

pratyushprakash:
Settings have been added for enabling/disabling indidual
features of the `html_linter`.

Closes https://github.com/coala/coala-bears/issues/1247


tests/hypertext/HTMLLintBearTest.py
```diff
@@ -27,3 +27,10 @@
     invalid_files=(test_file,),
     settings={'htmllint_ignore': 'quotation'},
     tempfile_kwargs={'suffix': '.html'})
+
+HTMLLintBearDisableCheckerTest = verify_local_bear(
+    HTMLLintBear,
+    valid_files=(test_file,),
+    invalid_files=(),
```
namanyadav12: Why is no invalid_file given?

pratyushprakash: The `invalid_files` are tested in a seperate test.

pratyushprakash:
@namanyadav12 What exactly do you want changed ?

tests/hypertext/HTMLLintBearTest.py
Nosferatul: For all tests we should have at least an valid and an invalid file

pratyushprakash: The tests before don't follow that. The `invalid_files` is already covered in the `HTMLLintBearTest` and `HTMLLintBearIgnoreQuotationTest`. I feel since `HTMLLintBearDisableCheckerTest` is supposed to check whether disabling a certain setting takes effect, it would make no sense to include an invalid_file.

namanyadav12:
@pratyushprakash IMO it should include an invalid_file

tests/hypertext/HTMLLintBearTest.py
namanyadav12: @pratyushprakash  How does this test check whether disabling a certain setting takes effect.

pratyushprakash: @namanyadav12 could you please view HTMLLintBearTest.py in its entirety and not just the changes I made

tests/hypertext/HTMLLintBearTest.py
```diff
@@ -27,3 +27,10 @@
...
+    settings={'optional_tag': 'off'},
```
aptrishu: can you add a test with an invalid tag ?
html-lint handles it so we need just to check we are good with that.

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,95 @@ class HTMLLintBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
+                         doctype: str='Enable',
+                         entities: str='Enable',
+                         trailing_whitespace: str='Enable',
```
Asnelchristian: most of the settings start with either ``allow`` or ``enable`` or ``force``... you should consider using that convention, and particularly for this setting we use ``allow_trailing_whitespaces`` [use git grep to check ]

tests/hypertext/HTMLLintBearTest.py
pratyushprakash: Well apparently html-lint does not check for invalid tags.

aptrishu: I think it does see https://github.com/sk-/html-linter/blob/master/html_linter.py#L1083 ..

pratyushprakash: That price of code is actually informing the user that they provided invalid disable arguments.

bears/hypertext/HTMLLintBear.py
jayvdb: And the settings are of type `bool` if they are true/false type flags.

tests/hypertext/HTMLLintBearTest.py
Mixih: Imo testing on this bear is a bit screwy and needs to be improved. Separate issue. Ideally we test most of these options, but not possible right now

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,101 @@ class HTMLLintBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
+                         enable_doctype: bool=True,
+                         enable_entities: bool=True,
+                         enable_trailing_whitespace: bool=True,
```
Asnelchristian: use ``allow_trailing_whitespace`` for consistency

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,101 @@ class HTMLLintBear:
...
+                         enable_tabs: bool=True,
```
Asnelchristian: either ``use_tabs: bool=True`` or ``use_spaces: bool=False``

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,101 @@ class HTMLLintBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
+                         enable_doctype: bool=True,
```
Asnelchristian: what about ``check_doctype`` ? :sweat_smile:

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,101 @@ class HTMLLintBear:
...
+                         enable_invalid_attribute: bool=True,
+                         enable_void_zero: bool=True,
+                         enable_invalid_handler: bool=True,
+                         enable_http_equiv: bool=True,
+                         enable_extra_whitespace: bool=True,
                          htmllint_ignore: typed_list(str)=()):
         """
-        :param htmllint_ignore: List of checkers to ignore.
+        :param enable_doctype:
+            Enable or disable doctype checker.
```
Asnelchristian: maybe you should add an example here, it is not quite clear what this setting does

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,101 @@ class HTMLLintBear:
...
+        :param enable_entities:
+            Enable or disable unicode entity checker.
```
Asnelchristian: here too...

bears/hypertext/HTMLLintBear.py
```diff
@@ -28,9 +28,101 @@ class HTMLLintBear:
...
+        :param enable_void_zero:
+           Enable or disable checker for javascript void(0).
+        :param enable_invalid_handler:
+           Enable or disable checker for Javascript links.
+        :param enable_http_equiv:
+           Enable or disable checker for http-equiv.
+        :param enable_extra_whitespace:
+           Enable or disable checker for extra whitespace.
+        :param htmllint_ignore:
+           List of checkers to ignore.
```
Asnelchristian: i feel like, you should instead tell what the setting do instead of telling what is enabled or disabled

tests/hypertext/HTMLLintBearTest.py
aptrishu: yeah, sorry by tag earlier I was referring to disable argument. damn. >_<

PR #1332, opened 2017-01-20
WIP AnnotationBear: Return precise string/comment info

aptrishu:
Earlier the return used to be either a tuple of
source ranges of strings/comments now it is a
tuple of namedtuples with fields
"start_delimiter_range end_delimiter_range content_range full_range"
of singleline strings/multiline strings/singlieline comments/multiline comments
Also, It modifies KeywordBear,  QuotesBear,
IndentationBear and their tests.


aptrishu:
CC @jayvdb.
Please take a look at error in `coala-ci` log in circleCI.

bears/general/AnnotationBear.py
```diff
@@ -88,66 +96,111 @@ def find_annotation_ranges(self,
...
-                _range, end_position = self.get_range_end_position(
-                    file,
-                    filename,
-                    text,
-                    string_delimiters,
-                    position,
-                    self.get_singleline_strings)
-                if end_position and _range:
-                    strings_range.append(_range)
+                end_position, start_delim, end_delim =\
```
jayvdb: please add a space after the equals, here and below

bears/general/AnnotationBear.py
```diff
@@ -88,66 +96,111 @@ def find_annotation_ranges(self,
...
         text = ''.join(file)
-        strings_range = []
-        comments_range = []
+        singleline_string_range = []
+        multiline_string_range = []
+        singleline_comment_range = []
+        multiline_comment_range = []
+        fields = ('start_delimiter_range end_delimiter_range'
+                  ' content_range full_range')
+        singleline_string = namedtuple('singleline_string', fields)
```
jayvdb: http://stackoverflow.com/questions/16377215/how-to-pickle-a-namedtuple-instance-correctly

aptrishu:
@jayvdb  It's still failing, I have put namedtuples outside the class.
If we use pickle from dill (https://github.com/uqfoundation/dill) it would be easily done without any such issues.

bears/general/AnnotationBear.py
aptrishu: Done with dict..

aptrishu:
All errors are resolved..

sils:
@aptrishu is this still wip? With that in the title you aren't getting any reviews and this is so important :( we need to merge it within *hours*

aptrishu:
Ohh I forgot to remove WIP from the tittle, Though It was marked `pending review`. so, it must be showing in review queue.

bears/general/AnnotationBear.py
```diff
@@ -1,10 +1,13 @@
+from collections import namedtuple
```
abhsag24: do you need this now? since you are using dicts?

bears/general/AnnotationBear.py
```diff
@@ -88,66 +97,130 @@ def find_annotation_ranges(self,
             A dictionary containing the various ways to define multi-line
             comments in a language.
         :return:
-            Two tuples first containing a tuple of strings, the second a tuple
-            of comments.
+            Four tuples containing namedtuples for singleline strings,
```
abhsag24: containing dicts right?

bears/general/IndentationBear.py
```diff
@@ -168,7 +173,7 @@ def get_absolute_indent_of_range(self,
         :param encaps_pos:      A tuple ofSourceRanges of code regions
                                 trapped in between a matching pair of
                                 encapsulators.
-        :param annotation_dict: A dictionary containing sourceranges of all the
+        :param annotation_dict: A dictionary containing namedtuples of all the
```
abhsag24: dicts

bears/general/IndentationBear.py
```diff
@@ -196,7 +201,7 @@ def get_indent_levels(self,
         :param filename:        Name of the file that needs to be checked.
         :param indent_types:    A dictionary with keys as start of indent and
                                 values as their corresponding closing indents.
-        :param annotation_dict: A dictionary containing sourceranges of all the
+        :param annotation_dict: A dictionary containing namedtuples of all the
```
abhsag24: same

bears/general/IndentationBear.py
```diff
@@ -255,7 +260,7 @@ def get_specified_block_range(self,
                                 block has begun.
         :param close_specifier: A character or string indicating that the block
                                 has ended.
-        :param annotation_dict: A dictionary containing sourceranges of all the
+        :param annotation_dict: A dictionary containing namedtuples of all the
```
abhsag24: same

bears/general/IndentationBear.py
```diff
@@ -324,7 +329,7 @@ def get_unspecified_block_range(self,
         :param filename:         Name of the file that needs to be checked.
         :param indent_specifier: A character or string indicating that the
                                  indentation should begin.
-        :param annotation_dict: A dictionary containing sourceranges of all the
+        :param annotation_dict: A dictionary containing namedtuples of all the
```
abhsag24: same

bears/general/IndentationBear.py
```diff
@@ -372,7 +377,7 @@ def get_valid_sequences(file,
         :param file:            File that needs to be checked in the form of
                                 a list of strings.
         :param sequence:        Sequence whose validity is to be checked.
-        :param annotation_dict: A dictionary containing sourceranges of all the
+        :param annotation_dict: A dictionary containing namedtuples of all the
```
abhsag24: same

bears/general/IndentationBear.py
```diff
@@ -394,22 +399,23 @@ def get_valid_sequences(file,
...
+                   lt_eq(sequence_position, comment['full_range'].end)):
                     valid = False
 
-                if(comment.start.line == sequence_position.line and
-                        comment.end.line == sequence_position.line and
-                        check_ending):
-                    sequence_line_text = sequence_line_text[
-                        :comment.start.column - 1] + sequence_line_text[
-                        comment.end.column-1:]
+                if(comment['full_range'].start.line == sequence_position.line
```
abhsag24: this wont work with dicts right?

abhsag24: how are the tests passing? do we nt test this?

abhsag24: same in other bears

aptrishu: why would it not pass with dicts ?
`comment['full_range']` is sourcerange object.
and yes, ofcourse we are testing it.

sils:
you know you have 72 chars in your commit body?

bears/general/AnnotationBear.py
```diff
@@ -1,10 +1,13 @@
+from collections import namedtuple
+
 from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.HiddenResult import HiddenResult
 from coalib.results.Result import Result, RESULT_SEVERITY
 from coalib.results.SourceRange import SourceRange
 from coalib.results.AbsolutePosition import AbsolutePosition
 from coala_utils.string_processing.Core import unescaped_search_for
+import sys
```
sils: put system imports before coalib imports

aptrishu:
Yeah, I'll amend it. I'll remove the last line, its useless.. I just placed some extra info for the ease of reviewing.

bears/general/AnnotationBear.py
```diff
@@ -23,10 +26,13 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
...
-            One HiddenResult containing a dictionary with keys being 'strings'
-            or 'comments' and values being a tuple of SourceRanges pointing to
-            the strings and a tuple of SourceRanges pointing to all comments
-            respectively. The ranges do include string quotes or the comment
+            One HiddenResult containing a dictionary with keys being
+            'singleline strings', 'multiline strings', 'singleline comments'
+            and 'multiline comments' and values being a tuple of
+            namedtuples of the format: ```("CodeAnnotation",
+            "start_delimiter_range end_delimiter_range
+            content_range full_range")```.
```
sils: I don't understand this well, I assume you mean something like `(content_sourcerange, start_delimiter_sourcerange, end_delimiter_sourcerange)`?

aptrishu:
Changes to be done right now,
- Modify docstrings at some places.
- Add tests for `content_range`,`start_delimiter_range`,`end_delimiter_range`
Right now I'm only testing `full_range`.

bears/general/AnnotationBear.py
aptrishu: and `fullcontent_sourcerange` also. I used `range` instead of `sourcerange`, it sounds easy to handle.
should I change it ?

sils:
range instead of sourcerange is good, generally I didn't understand your
version though with the code annotation and the double quotes scattered
seemingly random :P

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net

aptrishu:
Actually that's my fault, I had first used namedtuples here. so after I switched to dicts, I didnt change docstrings at some places.
I'll do all the changes today.

aptrishu:
Updated the tests and docstrings.

tests/general/QuotesBearTest.py
```diff
@@ -32,26 +34,29 @@ def setUp(self):
         """).splitlines(True)
 
         self.filename = 'f'
-
+        x = namedtuple('x', 'full_range')
```
SanketDG: why use a namedtuple when none of its features are being used?

bears/general/QuotesBear.py
```diff
@@ -62,21 +62,29 @@ def run(self, filename, file, dependency_results,
...
                       str):
-            self.err(dependency_results[AnnotationBear.name][0].contents)
+            self.err(dependency_results[AnnotationBear.name].contents)
             return
 
-        ranges = dependency_results[AnnotationBear.name][0].contents['strings']
+        dep_contents = dependency_results[AnnotationBear.name].contents
+        annotation_dict = {}
+        s_a = 'singleline strings'
+        s_b = 'multiline strings'
```
SanketDG: any reason for having spaces?

bears/general/AnnotationBear.py
```diff
@@ -88,66 +95,130 @@ def find_annotation_ranges(self,
...
 
                 return position + 1
 
             position = get_new_position()
 
-        return tuple(strings_range), tuple(comments_range)
+        return (tuple(singleline_string_range),
+                tuple(multiline_string_range),
+                tuple(singleline_comment_range),
+                tuple(multiline_comment_range))
```
SanketDG: if we are going to use each of the tuples here, maybe looking into python namedtuples is worth a shot.

Then we can do:

```
range.singleline_string_range
```

aptrishu: Actually using namedtuple here will create problems with pickling, because we'll have to define them at module level scope and things would look absurd then.
btw that was the first approach.
So, I switched to dicts after a discussion.

bears/general/QuotesBear.py
aptrishu: just readability. :)

bears/general/AnnotationBear.py
```diff
@@ -88,66 +95,130 @@ def find_annotation_ranges(self,
...
-                _range, end_position = self.get_range_end_position(
-                    file,
-                    filename,
-                    text,
-                    string_delimiters,
-                    position,
-                    self.get_singleline_strings)
-                if end_position and _range:
-                    strings_range.append(_range)
+                end_position, start_delim, end_delim = \
```
jayvdb: Please do not use line continuation. `\`. Use round brackets.
Also found below.

bears/general/AnnotationBear.py
jayvdb: Could you add a comment on the PR, not linked to a commit, that explains why you are using tuples instead of named tuples, and link to the discussions involved in the decision.

aptrishu: https://github.com/coala/coala-bears/issues/1352 this is the related issue. The discussion was held in chat so I'll have to dig in a lot. It was also discussed with @abhsag24, when he suggested we can use dicts.

jayvdb: Did you try the solution @Makman2 gave at https://github.com/coala/coala-bears/issues/1352#issuecomment-275846041 ?

aptrishu: yes, It didnt work.

bears/general/AnnotationBear.py
```diff
@@ -88,66 +95,130 @@ def find_annotation_ranges(self,
...
+                                                          end_position)
+                    singleline_comment = {}
+                    singleline_comment[
+                        'start_delimiter_range'] = seperate_ranges[0]
+                    singleline_comment[
+                        'end_delimiter_range'] = seperate_ranges[1]
+                    singleline_comment['content_range'] = seperate_ranges[2]
+                    singleline_comment['full_range'] = seperate_ranges[3]
+                    singleline_comment_range.append(singleline_comment)
                     return end_position + 1
```
userzimmermann: Above is 4 times exactly the same algorithm. Please try to generalize that

bears/general/AnnotationBear.py
```diff
@@ -44,9 +48,9 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
         multiline_comment_delimiters = dict(
             lang_dict['multiline_comment_delimiters'])
         comment_delimiter = dict(lang_dict['comment_delimiter'])
-        string_ranges = comment_ranges = ()
+        ranges = ((), (), (), ())
```
userzimmermann: Why not just creating some `Ranges` class to hold those informations?

bears/general/AnnotationBear.py
```diff
@@ -88,66 +95,130 @@ def find_annotation_ranges(self,
             A dictionary containing the various ways to define multi-line
             comments in a language.
```
userzimmermann: Why not just create some `Delimiters` class to hold that information?

bears/general/AnnotationBear.py
```diff
@@ -58,7 +62,10 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
             yield Result(self, str(e), severity=RESULT_SEVERITY.MAJOR,
                          affected_code=(e.code,))
 
-        content = {'strings': string_ranges, 'comments': comment_ranges}
+        content = {'singleline strings': ranges[0],
+                   'multiline strings': ranges[1],
+                   'singleline comments': ranges[2],
+                   'multiline comments': ranges[3]}
```
userzimmermann: Again :) Why not just create a `Ranges` class to hold that information?

bears/general/AnnotationBear.py
```diff
@@ -88,66 +95,130 @@ def find_annotation_ranges(self,
...
 
                 return position + 1
 
             position = get_new_position()
 
-        return tuple(strings_range), tuple(comments_range)
+        return (tuple(singleline_string_range),
+                tuple(multiline_string_range),
+                tuple(singleline_comment_range),
+                tuple(multiline_comment_range))
```
userzimmermann: And once more ;) Why not a `Ranges` class?

sils:
@aptrishu any news on this?

bears/general/AnnotationBear.py
aptrishu: @userzimmermann what will be the use case of this new `Delimiters` class?

bears/general/AnnotationBear.py
aptrishu: So, I had a thought of creating a `RangeHelper` (like the rangehelper module here https://github.com/coala/coala-bears/pull/1528/files). So are you talking specifically for this module or in general ?

aptrishu:
Hi @sils , so right now there is only one change required, I have asked my doubts to @userzimmermann , I'll update it after that. I was waiting for more reviews thats why I didnt update it earlier.

bears/general/AnnotationBear.py
userzimmermann: @aptrishu Encapsulating data. So that you don't have to always pass several parameters around

bears/general/AnnotationBear.py
userzimmermann: The more general the better ;)

bears/general/AnnotationBear.py
```diff
@@ -88,66 +95,130 @@ def find_annotation_ranges(self,
...
-                    strings_range.append(_range)
+                end_position, start_delim, end_delim = (
+                    self.get_range_end_position(
+                        file,
+                        filename,
+                        text,
+                        multiline_string_delimiters,
+                        position,
+                        self.get_multiline))
+                if end_position:
```
rhemon: why are the separate if blocks when condition is same?

bears/general/AnnotationBear.py
```diff
@@ -157,26 +228,27 @@ def get_range_end_position(file,
...
             if text[position:].startswith(annotation):
+                selected_annotation = annotation
                 if not single_comment:
-                    ret_val = func(file,
-                                   filename,
-                                   text,
-                                   annotation,
-                                   annotations[annotation],
-                                   position)
+                    selected_end_annotation = annotations[selected_annotation]
```
rhemon: why not just directly use `annotations[annotation]` instead of having a separate variable?

tests/general/AnnotationBearTest.py
```diff
@@ -22,47 +22,159 @@ def setUp(self):
...
+            'F',
+            AbsolutePosition(text, 1),
+            AbsolutePosition(text, len(text[0]) - 3))
+
+        compare_start = SourceRange.from_absolute_position(
+            'F',
+            AbsolutePosition(text, 0),
+            AbsolutePosition(text, 0))
+
+        compare_end = SourceRange.from_absolute_position(
```
rhemon: these here seem to be repeated in other functions. would it be good idea to have a common function to do this?

bears/general/AnnotationBear.py
userzimmermann: @rhemon This is exactly what I already mentioned in https://github.com/coala/coala-bears/pull/1332/files#r113981233

cc @aptrishu

tests/general/AnnotationBearTest.py
userzimmermann: @rhemon Better data and algorithm abstractions are a general issue with this PR :/

bears/general/AnnotationBear.py
aptrishu: because `end_position` is being recalculated.

userzimmermann: @aptrishu But you could put everything in a nested loop with changing parameters. Or put the actual logic in a class that encapsulates those different types of single-/multi-line comments/strings with some proper abstraction mechanism. Every redundancy can be solved ;)

bears/general/AnnotationBear.py
```diff
@@ -23,10 +25,12 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
...
             External directory for coalang file.
         :return:
-            One HiddenResult containing a dictionary with keys being 'strings'
-            or 'comments' and values being a tuple of SourceRanges pointing to
-            the strings and a tuple of SourceRanges pointing to all comments
-            respectively. The ranges do include string quotes or the comment
+            One HiddenResult containing a dictionary with keys being
+            'singleline strings', 'multiline strings', 'singleline comments'
+            and 'multiline comments' and values being a dictionary with
+            keys ```'start_delimiter_range', 'end_delimiter_range'
```
sils: can you use single backslaches around every key? Also for the keys above please, be consistent

bears/general/AnnotationBear.py
sils: I'd love to see some classes too, instead of returning this dict rater make it a list of:

```python
@attr.s
class String:
    start = attr.ib(convert=SourcePosition)
    content_start = attr.ib(convert=SourcePosition)
    content_end = attr.ib(convert=SourcePosition)
    end = attr.ib(convert=SourcePosition)
```

plus some convenience properties

sils: that would make a lot of stuff simpler wouldn't it?

sils:
@aptrishu marking WIP, I think having some more explicit data structures would help a lot.

bears/general/AnnotationBear.py
aptrishu: @sils how does this class help?
This is a class for ranges.
While the return value is 
{key:(range1,range2,..),key2:(),..} 
Are you proposing to make a class having attributes {singleline strings, multiline strings,..} ?

aptrishu: @userzimmermann 
1. regarding making new range class - It is being used only once. I don't see the effectiveness of this new range class.
2. regarding creating delimiter class - it is also being passed only once to the `find_annotation_ranges` method.
In other methods, we are sending the delimiter range.
3. regarding using a loop because we are recalculating the same thing 4 times in `find_annotation_ranges` method.
- Actually no, we are not doing the same thing 4 times, we are passing different variables in `get_range_end_position`, and different dicts are being used to store. If we put these things in a loop, It would be like

earlier - 
```
//do stuff1
//do stuff2
//do stuff3
//do stuff4
```

using loop - 

```
for i in range(4):
  if(i==0):
     //do stuff1
  if(i==1):
     //do stuff2
so on..
```

I don't know how does it help.

sils: the way you're using this seems more like an object to me, I feel like it'd be way clearer if we just had a simple 6 line class  like that

aptrishu: We can make a class with attributes storing tuple of ranges, that's why I thought dict would be better.

sils: I'm not sure I get you fully, I'd prefer a class as it's a much clearer and better defined API as "some dict with those keys"

aptrishu:
@Makman2  @userzimmermann review needed. :P

bears/general/AnnotationBear.py
userzimmermann: @aptrishu actually double backticks would be correct here

bears/general/QuotesBear.py
```diff
@@ -62,21 +62,27 @@ def run(self, filename, file, dependency_results,
...
-        ranges = dependency_results[AnnotationBear.name][0].contents['strings']
+        dep_contents = dependency_results[AnnotationBear.name].contents
+        annotation_dict = {}
+        annotation_dict['strings'] = (dep_contents.singleline_strings +
+                                      dep_contents.multiline_strings)
+        ranges = annotation_dict['strings']
 
         for string_range in ranges:
-            if (file[string_range.start.line-1][string_range.start.column-1] ==
+            temp_range = string_range.full_range
```
jayvdb: there is no need to use a different variable here.
Earlier, do
```diff
-    ranges = annotation_dict['strings']
+    ranges = [x.full_range for x in annotation_dict['strings']]
```

PR #1333, opened 2017-01-20
GitCommitBear.py: Enforce fullmatch on body_regex

### LINKED ISSUE #1331: GitCommitBear/body_regex: Use re.fullmatch instead of re.search?
I'm for using `re.fullmatch` to match the whole body of the commit. Currently you can work with anchors, but with the restriction that they work only on a single line, as `re.MULTILINE` is not activated for `body_regex`. Like for `shortlog_regex`, I would suggest to use `fullmatch` instead of `search`.

CC @nkprince007

nkprince007:
Force body to follow a certain multiline pattern

Fixes #1331

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -31,6 +31,10 @@ bears = SpaceConsistencyBear
 bears = GitCommitBear
 shortlog_trailing_period = False
 shortlog_regex = ([^:]*|[^:]+: [A-Z0-9*].*)
+body_close_issue = True
+body_close_issue_full_url = True
+body_close_issue_last_line = True
+body_enforce_issue_reference = False
 
 [LineCounting]
 enabled = False
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -304,7 +304,7 @@ class GitCommitBear(GlobalBear):
                 r'https?://{}\S+/issues/(\S+)'.format(re.escape(host)))
         else:
             result_info = 'issue'
-            issue_ref_regex = r'#(\S+)'
+            issue_ref_regex = r'(?:\w+/\w+)?#(\S+)'
 
         concat_regex = '|'.join(kw for kw in self.CONCATENATION_KEYWORDS)
         compiled_joint_regex = re.compile(
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -241,7 +241,7 @@ class GitCommitBear(GlobalBear):
                                'HEAD commit. Please add one.')
             return
 
-        if body_regex and not re.search(body_regex, body):
+        if body_regex and not re.fullmatch(body_regex, body.strip()):
             yield Result(self, 'No match found in commit message for the '
                                'regular expression provided: %s' % body_regex)
 
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -257,21 +257,21 @@ class GitCommitBear(GlobalBear):
     def check_issue_reference(self, body,
                               body_close_issue: bool=False,
                               body_close_issue_full_url: bool=False,
-                              body_close_issue_on_last_line: bool=False):
+                              body_close_issue_on_last_line: bool=False,
+                              body_enforce_issue_reference: bool=False):
         """
         Check for matching issue related references and URLs.
 
         :param body:
             Body of the commit message of HEAD.
         :param body_close_issue:
-            Whether to check for the presence of issue reference within
-            the commit body by retrieving host information from git
-            configuration. GitHub and GitLab support auto closing issues with
-            commit messages. Checks for matching keywords in the commit body.
-            By default, if none of ``body_close_issue_full_url`` and
-            ``body_close_issue_on_last_line`` are enabled, checks for presence
-            of short references like ``closes #213``. Otherwise behaves
-            according to other chosen flags.
+            GitHub and GitLab support auto closing issues with
+            commit messages. When enabled, this checks for matching keywords
+            in the commit body by retrieving host information from git
+            configuration. By default, if none of ``body_close_issue_full_url``
+            and ``body_close_issue_on_last_line`` are enabled, this checks for
+            presence of short references like ``closes #213``.
+            Otherwise behaves according to other chosen flags.
             More on keywords follows.
             [GitHub](https://help.github.com/articles/closing-issues-via-commit-messages/)
             [GitLab](https://docs.gitlab.com/ce/user/project/issues/automatic_issue_closing.html)
@@ -282,6 +282,9 @@ class GitCommitBear(GlobalBear):
             When enabled, checks for issue close reference presence on the
             last line of the commit body. Works along with
             ``body_close_issue``.
+        :param body_enforce_issue_reference:
+            Whether to enforce presence of issue reference in the body of
+            commit message.
         """
         if not body_close_issue:
             return
@@ -326,7 +329,7 @@ class GitCommitBear(GlobalBear):
 
         matches = compiled_joint_regex.findall(body)
 
-        if len(matches) == 0:
+        if body_enforce_issue_reference and len(matches) == 0:
             yield Result(self, result_message.format(result_info))
             return
 
--- a/tests/vcs/git/GitCommitBearTest.py
+++ b/tests/vcs/git/GitCommitBearTest.py
@@ -269,24 +269,23 @@ class GitCommitBearTest(unittest.TestCase):
                          [])
         self.assertTrue(self.msg_queue.empty())
 
-        # Testing body_regex
+        # body_regex, not fully matched
         self.git_commit('Shortlog\n\n'
                         'First line, blablablablablabla.\n'
                         'Another line, blablablablablabla.\n'
                         'Fix 1112')
         self.assertEqual(self.run_uut(
-                             body_regex=r'Fix\s+[1-9][0-9]*\s*'), [])
+                             body_regex=r'Fix\s+[1-9][0-9]*\s*'),
+                         ['No match found in commit message for the regular '
+                          'expression provided: Fix\s+[1-9][0-9]*\s*'])
         self.assert_no_msgs()
 
-        # Matching with regexp, no match found
+        # Matching with regexp, fully matched
         self.git_commit('Shortlog\n\n'
-                        'First line, blablablablablabla.\n'
-                        'Another line, blablablablablabla.\n'
-                        'Closes #01112')
+                        'TICKER\n'
+                        'CLOSE 2017')
         self.assertEqual(self.run_uut(
-                             body_regex=r'Closes\s+#[1-9][0-9]*\s*'),
-                         ['No match found in commit message for the regular '
-                          'expression provided: Closes\s+#[1-9][0-9]*\s*'])
+                             body_regex=r'TICKER\s*CLOSE\s+[1-9][0-9]*'), [])
         self.assert_no_msgs()
 
     def test_check_issue_reference(self):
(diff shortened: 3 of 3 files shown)


pratyushprakash:
I don't think this needs 4 commits.

nkprince007:
@pratyushprakash Could you be more specific? Because, the commit message specifies respective changes correctly. 

If you insist in squashing them, please specify the commits that could be squashed.

bears/vcs/git/GitCommitBear.py
```diff
@@ -241,8 +241,9 @@ def check_body(self, body,
                                'HEAD commit. Please add one.')
             return
 
-        if body_regex and not re.search(body_regex, body):
-            yield Result(self, 'No match found in commit message for the '
+        if body_regex and not re.fullmatch(
+                body_regex, body.strip(), re.MULTILINE | re.DOTALL):
```
Makman2: I wouldn't enable `re.MULTILINE` and `re.DOTALL`, normally people don't expect this + this is not necessary^^

bears/vcs/git/GitCommitBear.py
```diff
@@ -241,8 +241,9 @@ def check_body(self, body,
...
+            yield Result(self, "HEAD commit's body does not match the "
```
Makman2: messaeg change should go into an own commit ;)

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -271,11 +271,12 @@ def test_body_checks(self):
...
-                        'First line, blablablablablabla.\n'
-                        'Another line, blablablablablabla.\n'
-                        'Fix 1112')
+                        'TICKER---BEGIN\n'
+                        'Bla bla bla\n'
+                        'Fix 1112\n'
+                        'TICKER---END')
         self.assertEqual(self.run_uut(
-                             body_regex=r'Fix\s+[1-9][0-9]*\s*'), [])
+                             body_regex=r'^TICKER---BEGIN.*TICKER---END$'), [])
```
Makman2: you don't need the anchors^^

Makman2: maybe you want another test case, as you change the example completely?

Makman2:
@pratyushprakash commits must only contain a single change, not 10 ;) That makes review easier and modularizes history :+1: So @nkprince007 does exactly right

Makman2:
ack cf9733a

bears/vcs/git/GitCommitBear.py
```diff
@@ -291,6 +291,9 @@ def check_issue_reference(self, body,
         if host not in self.SUPPORTED_HOST_KEYWORD_REGEX:
             return
 
+        if not re.search(self.SUPPORTED_HOST_KEYWORD_REGEX[host], body):
+            return
```
Makman2: not necessary imo, the regex below just won't match and you achieve the same effect^^

.coafile
```diff
@@ -31,6 +31,9 @@ bears = SpaceConsistencyBear
 bears = GitCommitBear
 shortlog_trailing_period = False
 shortlog_regex = ([^:]*|[^:]+: [A-Z0-9*].*)
+body_close_issue = True
+body_close_issue_full_url = True
+body_close_issue_last_line = True
```
Makman2: We don't enforce issue references in our URLs ;)
Like in this commit ;)

bears/vcs/git/GitCommitBear.py
nkprince007: It won't match, deliberately reporting the result that body doesn't contain any issue reference.

```
README: Modify installation procedure.

Add the details about setting up ``git``
```
The above commit, will be reported false positive for errors, which is unexpected, right?

.coafile
nkprince007: ^^Didn't quite get you. I just followed [this.](https://api.coala.io/en/latest/Developers/Writing_Good_Commits.html)

bears/vcs/git/GitCommitBear.py
nkprince007: Sure. :smile:

bears/vcs/git/GitCommitBear.py
Makman2: But if you activate `issue_regex` you want to enforce links, don't you?

nkprince007: Yeah, that's the main issue here.
With coala and other repos, we won't have all the commits fixing some bugs, right?
And if we enforce this in `.coafile`, we would require all commits to have issue references.
That is clearly impossible.
So, a check for presence of the keywords tells us that this isn't a commit that fixes/closes something, right? Hope you get the point.

bears/vcs/git/GitCommitBear.py
nkprince007: Just multiline, then?

bears/vcs/git/GitCommitBear.py
Makman2: actually there are repos which require always an issue reference, or a JIRA ticket reference. So enforcing this on all commits is not completely unuseful.

alright behaviour is okay, but would be good if you implement another flag `body_enforce_issue_reference` to enforce this for all commits. Then you can still keep the "old" behaviour.

Also I think you should modify the code here to implement this behaviour: https://github.com/coala/coala-bears/pull/1333/files/4ce4c848a53d0c9b2024da8b9cea4fedad84ace5#diff-e0a334e6d00dfb6b1fbebb7c515ca4f8R333
In fact you just have to remove this if / extend it by
```python
if body_enforce_issue_reference and if len(results) == 0:
    yield Result(...)
```

.coafile
Makman2: nvm, see above as you change behaviour

bears/vcs/git/GitCommitBear.py
Makman2: no flags

.coafile
nkprince007: So, by implementing the above stuff, we could leave these changes as is?

tests/vcs/git/GitCommitBearTest.py
nkprince007: Actually, I changed this only because the two test cases already in there would be the same, if  I leave the test as is.

bears/vcs/git/GitCommitBear.py
nkprince007: Sounds good. But what if a user has a commit like this.
```
Shortlog
First line, blablablablablabla
Fix http://gitlab.com/user/repo/issues/1112
Another line, blablablablablabla.
```
And he has the flags `body_close_issue`, `body_close_issue_full_url` and `body_close_issue_on_last_line` enabled? 
Shouldn't the expected output be something that states there's no issue reference in last line? Instead it yields no result.

tests/vcs/git/GitCommitBearTest.py
Makman2: if it's the same like before then you don't need to change it, right? :)

bears/vcs/git/GitCommitBear.py
```diff
@@ -241,7 +241,7 @@ def check_body(self, body,
                                'HEAD commit. Please add one.')
             return
 
-        if body_regex and not re.search(body_regex, body):
+        if body_regex and not re.fullmatch(body_regex, body.strip()):
```
Makman2: why `strip()`?

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -269,13 +269,15 @@ def test_body_checks(self):
...
+        # body_regex, not fully matched
         self.git_commit('Shortlog\n\n'
                         'First line, blablablablablabla.\n'
                         'Another line, blablablablablabla.\n'
                         'Fix 1112')
         self.assertEqual(self.run_uut(
-                             body_regex=r'Fix\s+[1-9][0-9]*\s*'), [])
+                             body_regex=r'Fix\s+[1-9][0-9]*\s*'),
+                         ['No match found in commit message for the regular '
+                          'expression provided: Fix\s+[1-9][0-9]*\s*'])
```
Makman2: could you perform also a valid fullmatch?

.coafile
Makman2: Could you add `body_enforce_issue_reference = False` explicitly? We shouldn't rely on bear defaults in this case, as this is definitely the setting we want for coala ;) Better explicit than implicit :)

bears/vcs/git/GitCommitBear.py
nkprince007: To remove any improbable newlines his/her text editor might be adding. :smile:

tests/vcs/git/GitCommitBearTest.py
nkprince007: Oops, totally forgot about the case.

.coafile
nkprince007: Yeah. Sure.

tests/vcs/git/GitCommitBearTest.py
Makman2: ah okay, nvm then ;)

Makman2:
ack 009f46a 2e4e35c 5128b4d 972b979

.coafile
```diff
@@ -31,6 +31,10 @@ bears = SpaceConsistencyBear
...
+body_enforce_issue_reference = False
```
jayvdb: According to :point_up: [January 26, 2017 5:25 PM](https://gitter.im/coala/coala?at=5889cea3074f7be7631e1e30) , I believe we want this to be `True`.  Please correct me if I have misunderstood.
cc @Makman2.

nkprince007: That strictly enforces all commits to *have issue reference*
But not all commits do fix/close something I suppose @jayvdb

nkprince007:
IMO, disabling `body_close_issue_last_line` should be enough for the above mentioned stuff.
@jayvdb

.coafile
jayvdb: Right, but as you can see from the Gitter chat, Lasse is happy for all future commits to have an issue created. It is a good practise anyway.

jayvdb:
Yes, we may disable `body_close_issue_last_line` on the release branches as a temporary fix,

.coafile
Makman2: Let's discuss that in a separate issue, for now we should just follow the current convention^^

nkprince007: @Makman2 @jayvdb please mark the changes, if any, are required.

Else could we get through with the merge?

bears/vcs/git/GitCommitBear.py
jayvdb: There should only be a single trailing `\n`.  Please only remove that, instead of `strip()`, which will mean junky git messages can slip through.

Otherwise the users [git config](https://git-scm.com/docs/git-commit#git-commit---cleanupltmodegt) needs fixing, and it should be an error.

nkprince007:
There's a clear chance when someone could actually place something like 
```
Shortlog: amazing shortlog




Fixes gitterHQ/gitter#32
```
What about this one @jayvdb?
https://github.com/coala/coala-bears/pull/1333#discussion_r98097515

jayvdb:
@nkprince007 , that should be an error.  See https://git-scm.com/docs/git-commit#git-commit---cleanupltmodegt , which I mentioned in my previous comment.  It cleans up multiple consecutive blank lines.

Makman2:
@jayvdb @nkprince007 filed https://github.com/coala/coala-bears/issues/1351 for that^^

nkprince007:
Could this be merged then @Makman2 ?

bears/vcs/git/GitCommitBear.py
Makman2: @jayvdb I think `strip` is alright, git should guarantee that only a traling \n is there. Also when someone writes commit messages with 10 newlines between body and shortlog, we catch it nicely away. Imo `body.strip()` contains then the real commit body :+1:

Makman2:
ack a8a4418 a25fb0f a2d55e8 bcd0f39

nkprince007:
ack 5302246 3e5c9fa 684befa 41a372c 

**CC** @Makman2

Makman2:
@rultor merge

PR #1339, opened 2017-01-23
deps.sh: Reduce the length of output

Techievena:
Reduce the length of output in ci
log by removing download progress
information

Closes https://github.com/coala/coala-bears/issues/1123

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -1,5 +1,6 @@
 set -e
 set -x
+TERM=dumb
 
 # Choose the python versions to install deps for
 case $CIRCLE_NODE_INDEX in


jayvdb:
Tests are failing on https://travis-ci.org/Techievena/coala-bears/builds . Why submit a pull request?

Techievena:
@jayvdb I'm unable to figure out why the CI are failing

Techievena:
Previous log
[log.txt](https://github.com/coala/coala-bears/files/726058/log.txt)
New log
[log_new.txt](https://github.com/coala/coala-bears/files/726061/log_new.txt)

__ONLY deps.sh part__

Techievena:
9888 lines --> 5220 lines

.ci/deps.sh
```diff
@@ -67,8 +67,8 @@ fi
 
 deps_perl="perl libperl-critic-perl"
 
-sudo apt-get -y update
-sudo apt-get -y --no-install-recommends install $deps $deps_perl $deps_infer
+sudo apt-get -y update|grep -v '\%'
```
Mixih: why don't you just redirect it to /dev/null? imo it's clearer what you're trying to do than this.

.ci/deps.sh
```diff
@@ -67,8 +67,8 @@ fi
...
+sudo apt-get -y --no-install-recommends install $deps $deps_perl $deps_infer|grep -v '\%'
```
Mixih: see above

.ci/deps.sh
```diff
@@ -89,9 +89,9 @@ if [[ -z "$(which hlint)" ]]; then
     # This is the same build as xenial hlint
     hlint_deb_url="https://launchpad.net/ubuntu/+source/hlint/1.9.26-1/+build/8831318/+files/${hlint_deb_filename}"
     hlint_deb=~/.apt-cache/$hlint_deb_filename
-    wget -O $hlint_deb $hlint_deb_url
+    wget -O $hlint_deb $hlint_deb_url 2>&1|grep -v '\%'
```
Mixih: see above

Techievena:
@jayvdb Any comments on @Mixih 's review. Do we have to remove the download progress information or not?

.ci/deps.sh
Techievena: But we need to use `|grep -v '\%'` in some cases. For clarity issues I think leaving a comment can be better i.e. about what grep is doing over here.

Techievena:
@Mixih But I think there is no point in checking out when the build dies in ci logs. I mean it will obviously report if the download fails or succeed. I mean, why do we need to know if download stops at 57% or 67% ?

.ci/deps.sh
Techievena: Moreover redirecting the data to /dev/null will completely silent the data. We don't want that. Output with `grep -v '%'` will be like this. 
image
But with redirecting it to /dev/null the output will be null

Techievena: And @Mixih without grep it will be like this
image

Mixih:
@Techievena oh I see what you actually did. Sorry for misinterpreting. Yes, this is fine

.ci/deps.sh
jayvdb: You could use `curl -sS` instead. That help on the docker repo.

.ci/deps.sh
jayvdb: Apt-get doesnt have a command line argument to omit these % messages?

.ci/deps.sh
```diff
@@ -105,8 +105,8 @@ npm install
 # R commands
 echo '.libPaths( c( "'"$R_LIB_USER"'", .libPaths()) )' >> .Rprofile
 echo 'options(repos=structure(c(CRAN="http://cran.rstudio.com")))' >> .Rprofile
-R -q -e 'install.packages("lintr")'
-R -q -e 'install.packages("formatR")'
+R -q -e 'install.packages("lintr")' 2>&1|grep -v '=\|downloaded'
```
jayvdb: I dont think this is worthwhile. It isnt a percentage progress bar. It is only two additional lines per package, and these R pacakes are very noisy with configure and build phases.

.ci/deps.sh
```diff
@@ -158,19 +158,19 @@ sudo luarocks install luacheck --deps-mode=none
...
   opam pin add --yes --no-action infer .
   opam install --deps-only --yes infer
-  ./build-infer.sh java
+  ./build-infer.sh java 2>&1|grep -v '\%'
 fi
 
 # PMD commands
 if [ ! -e ~/pmd-bin-5.4.1/bin ]; then
-  wget -nc -O ~/pmd.zip https://github.com/pmd/pmd/releases/download/pmd_releases%2F5.4.1/pmd-bin-5.4.1.zip
+  wget -nc -O ~/pmd.zip https://github.com/pmd/pmd/releases/download/pmd_releases%2F5.4.1/pmd-bin-5.4.1.zip 2>&1|grep -v '\%'
```
jayvdb: `%2F` can be just`/`

.ci/deps.sh
Mixih: we can prune the build stage if we must later

.ci/deps.sh
yukiisbored: I think you can set the `TERM` environment variable to `dumb` to remove the percentage.

Techievena: @yukiisbored Setting TERM environment to dumb will have effect on console window but will it have any effect on log file because @Makman2 had pointed out this.
image

Makman2: A log file is a "dumb" term environment, so let's try :+1:

yukiisbored: @Techievena No, that wouldn't do anything bad to the log files, because they're basically "just output text, lol"

Techievena: @yukiisbored I am afraid you misinterpreted I didn't mean it would harm log files, I was asking if setting term environment to dumb will have any effect on log files at all as it changes the environment variable of terminal window not the log file.

Techievena:
Perhaps it is done. Log file http://paste.ubuntu.com/23910613/ Thanks @yukiisbored :smiley:

jayvdb:
ack 72dccb5

jayvdb:
@rultor merge

PR #1356, opened 2017-01-29
PyImportSort: Option to treat imports separately

meetmangukiya:
If this option is set, then the imports separated by a line will be
fed independently to the SortImports function and the results will be
yielded accordingly.

Fixes https://github.com/coala/coala-bears/issues/879

### DIFF
--- a/bears/python/PyImportSortBear.py
+++ b/bears/python/PyImportSortBear.py
@@ -18,6 +18,64 @@ class PyImportSortBear(LocalBear):
     LICENSE = 'AGPL-3.0'
     CAN_FIX = {'Formatting'}
 
+    @staticmethod
+    def _seperate_imports(file):
+        import_stmts = []
+        tmp = []
+        paren = False
+        for lineno, lines in enumerate(file, start=1):
+            if 'import' in lines.split() or paren:
+                # To ensure that
+                # from x import ( y,
+                #                 z) type of imports are not treated as
+                # different sections
+                if '(' in lines and ')' not in lines:
+                    paren = True
+                if ')' in lines:
+                    paren = False
+                tmp.append((lineno, lines))
+            else:
+                if tmp:
+                    import_stmts.append(tmp)
+                tmp = []
+        # To ensure that if the last line of a file is an import statement
+        # it doesn't get ignored
+        if tmp:
+            import_stmts.append(tmp)
+            tmp = []
+        return import_stmts
+
+    def _get_diff(self):
+        if self.treat_seperated_imports_independently:
+            import_stmts = PyImportSortBear._seperate_imports(self.file)
+            sorted_imps = []
+            for units in import_stmts:
+                sort_imports = SortImports(file_contents=''.
+                                           join([x[1] for x in units]),
+                                           **self.isort_settings)
+                sort_imports = sort_imports.output.splitlines(True)
+                sorted_imps.append((units, sort_imports))
+
+            diff = Diff(self.file)
+            for old, new in sorted_imps:
+                start = old[0][0]
+                end = start + len(old) - 1
+                diff.delete_lines(start, end)
+                assert isinstance(new, list)
+                diff.add_lines(start, list(new))
+
+            if diff.modified != diff._file:
+                return diff
+        else:
+            sort_imports = SortImports(file_contents=''.join(self.file),
+                                       **self.isort_settings)
+
+            new_file = tuple(sort_imports.output.splitlines(True))
+            if new_file != tuple(self.file):
+                diff = Diff.from_string_arrays(self.file, new_file)
+                return diff
+        return None
+
     @deprecate_settings(indent_size='tab_width')
     def run(self, filename, file,
             use_parentheses_in_import: bool=True,
@@ -47,7 +105,8 @@ class PyImportSortBear(LocalBear):
             known_third_party_imports: typed_list(str)=(),
             known_standard_library_imports: typed_list(str)=None,
             max_line_length: int=79,
-            imports_forced_to_top: typed_list(str)=()):
+            imports_forced_to_top: typed_list(str)=(),
+            treat_seperated_imports_independently: bool=False):
         """
         Raise issues related to sorting imports, segregating imports into
         various sections, and also adding comments on top of each import
@@ -145,6 +204,9 @@ class PyImportSortBear(LocalBear):
             dependencies that occur in many projects.
         :param max_line_length:
             Maximum number of characters for a line.
+        :param treat_seperated_imports_independently:
+            Treat import statements seperated by one or more blank line or any
+            statement other than an import statement as an independent bunch.
         """
         isort_settings = dict(
             use_parentheses=use_parentheses_in_import,
@@ -177,12 +239,15 @@ class PyImportSortBear(LocalBear):
             isort_settings['known_standard_library'] = (
                 known_standard_library_imports)
 
-        sort_imports = SortImports(file_contents=''.join(file),
-                                   **isort_settings)
-        new_file = tuple(sort_imports.output.splitlines(True))
+        self.isort_settings = isort_settings
+        self.file = file
+        self.filename = filename
+        self.treat_seperated_imports_independently = \
+            treat_seperated_imports_independently
 
-        if new_file != tuple(file):
-            diff = Diff.from_string_arrays(file, new_file)
+        diff = self._get_diff()
+
+        if diff:
             yield Result(self,
                          'Imports can be sorted.',
                          affected_code=diff.affected_code(filename),
--- a/tests/python/PyImportSortBearTest.py
+++ b/tests/python/PyImportSortBearTest.py
@@ -1,5 +1,11 @@
 from bears.python.PyImportSortBear import PyImportSortBear
 from coalib.testing.LocalBearTestHelper import verify_local_bear
+import unittest
+from queue import Queue
+
+from coalib.settings.Section import Section
+
+from coalib.results.Diff import Diff
 
 PyImportSortBearTest = verify_local_bear(PyImportSortBear,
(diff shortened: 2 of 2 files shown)


bears/python/PyImportSortBear.py
```diff
@@ -176,13 +177,36 @@ def run(self, filename, file,
...
+                                                            for x in units]),
+                                           **isort_settings)
+                new_file = tuple(sort_imports.output.splitlines(True))
+
+                if new_file != tuple(file):
+                    diff = Diff.from_string_arrays(file, new_file)
+                    yield Result(self,
+                                 'Imports can be sorted.',
+                                 affected_code=diff.affected_code(filename),
+                                 diffs={filename: diff})
```
meetmangukiya: Would be better if I could not duplicate the code, wrap the reused code in a function maybe :confused: :question:

nemaniarjun:
@meetmangukiya Coverage didnt reach a 100%, Marking WIP :)

bears/python/PyImportSortBear.py
```diff
@@ -176,13 +177,53 @@ def run(self, filename, file,
...
 
-        sort_imports = SortImports(file_contents=''.join(file),
-                                   **isort_settings)
-        new_file = tuple(sort_imports.output.splitlines(True))
+        if treat_seperated_imports_independently:
+            import_stmts = []
+            tmp = []
+            for lineno, lines in enumerate(file, start=1):
+                print(lines)
+                if 'import' in lines:
```
meetmangukiya: ThIs looks like a fair point

sils:
Your commit messsage no verb.

bears/python/PyImportSortBear.py
```diff
@@ -47,7 +49,8 @@ def run(self, filename, file,
             known_third_party_imports: typed_list(str)=(),
             known_standard_library_imports: typed_list(str)=None,
             max_line_length: int=79,
-            imports_forced_to_top: typed_list(str)=()):
+            imports_forced_to_top: typed_list(str)=(),
+            treat_seperated_imports_independently: bool=False):
```
sils: missing documentation

meetmangukiya:
There is a problem currently..
I am splitting the import into sections by checking if given line contains `import` in it which does this:
- 
```
import this
import y

import z
``` 
gets grouped into (`import this`, `import y`) and (`import z`)

But this leads to 
- 
```
import this
import (bearlib,
            coalib)
```
 to be grouped as ('import this', 'import (bearlib, ') which is **so** wrong

meetmangukiya:
# Reviewers 

Please see the `seperate_imports` function, that tries to solve https://github.com/coala/coala-bears/pull/1356#issuecomment-285959401 There is even a test that ensures that it works as expected, if you find any more edge cases, please report :)

tests/python/PyImportSortBearTest.py
```diff
@@ -21,3 +29,81 @@
...
+
+    def test_import_seperation(self):
+        self.file = (
+            """
+from x import (y,
+               z,
+               w)
+import this
+
+import x, y
```
jayvdb: `y` here shadows `y` in the first import statement.
pyflakes would scream about that.
maybe use `a, b` instead here.

tests/python/PyImportSortBearTest.py
```diff
@@ -21,3 +29,81 @@
...
+class PyImportSortBearTest(unittest.TestCase):
+
+    def setUp(self):
+        self.queue = Queue()
+        self.section = Section('PyImportSortBear')
+        self.uut = PyImportSortBear(self.section,
+                                    self.queue)
+
+    def test_treat_seperated_imports_independently(self):
+        self.file = (
```
jayvdb: no need for `self.` here, as the attribute isnt used outside of the method.

Also normally these sample files are declared as module level variables, as they can often be reused in multiple tests.

bears/python/PyImportSortBear.py
```diff
@@ -145,6 +175,9 @@ def run(self, filename, file,
             dependencies that occur in many projects.
         :param max_line_length:
             Maximum number of characters for a line.
+        :param treat_seperated_imports_independently:
+            Treat import statements seperated by more then one line as
```
jayvdb: then -> than

"..seperated by more then one line.." -- this sounds like two blank lines needed between import groups?? that would be a pep8 violation, and isnt what is in your tests.

I think "..separated by a blank line.." is better.

bears/python/PyImportSortBear.py
```diff
@@ -176,13 +209,38 @@ def run(self, filename, file,
             isort_settings['known_standard_library'] = (
                 known_standard_library_imports)
 
-        sort_imports = SortImports(file_contents=''.join(file),
-                                   **isort_settings)
-        new_file = tuple(sort_imports.output.splitlines(True))
+        if treat_seperated_imports_independently:
+            import_stmts = PyImportSortBear.seperate_imports(file)
```
jayvdb: the revised sorting should be a separate function, so we dont have so much logic here in the run() method.

bears/python/PyImportSortBear.py
```diff
@@ -176,13 +209,38 @@ def run(self, filename, file,
...
+            diff = Diff(file)
+            for old, new in sorted_imps:
+                start = old[0][0]
+                end = start + len(old) - 1
+                diff.delete_lines(start, end)
+                assert isinstance(new, list)
+                diff.add_lines(start, list(new))
+
+            if diff.modified != diff._file:
+                yield Result(self,
```
jayvdb: this `Result` being yielded is identical to the existing `Result` being yielded.  The code needs to be adjusted so there is only one `Result` being yielded.

meetmangukiya:
@jayvdb I have created another commit `Rework` since I brought very much changes to previous code, and didn't just want to throw away previous commit. If the new commit looks fine(with little modifications) then I will throw away the previous commit. I hope previous commit doesn't bother for now. :)

meetmangukiya:
I have removed the earlier tests using verify local bear and replaced those with new ones, if that is not ok, comment.

bears/python/PyImportSortBear.py
jayvdb: Not done

tests/python/PyImportSortBearTest.py
```diff
@@ -1,23 +1,109 @@
...
+""".splitlines(True)
+
+        self.assertEqual(list(self.uut.run('', test_file)), [])
+        self.assertEqual(list(self.uut.run('',
+                                           test_file2))[0].diffs[''].modified,
+                         ['\n', 'import os\n', 'import re\n', '\n',
+                          'import requests\n'])
+
+    def test_treat_seperated_imports_independently(self):
+        test_file = (
```
jayvdb: It is too big. Can you create a simpler example of the bug being fixed.

meetmangukiya:
Since there are no objections, I assume it is okay https://github.com/coala/coala-bears/pull/1356#issuecomment-287405083

bears/python/PyImportSortBear.py
jayvdb: Done

tests/python/PyImportSortBearTest.py
jayvdb: Done

bears/python/PyImportSortBear.py
```diff
@@ -145,6 +206,9 @@ def run(self, filename, file,
             dependencies that occur in many projects.
         :param max_line_length:
             Maximum number of characters for a line.
+        :param treat_seperated_imports_independently:
+            Treat import statements seperated by one or more than one blank
```
jayvdb: This is hard to read.
'Treat' is a meaningless word.
'one or more than one' is rarely used English. 'one or more' is simpler.

meetmangukiya: I made it Treat import statements ... as independent bunch, better?

meetmangukiya:
@jayvdb please take a look again

tests/python/PyImportSortBearTest.py
```diff
@@ -1,29 +1,93 @@
...
-                                         ('import os\nimport sys\n',
-                                          'import os\nimport sys\n'),
-                                         ('import sys\nimport os\n',
-                                          'import sys\nimport os\n'))
-
-PyImportSortBearConfigsTest = verify_local_bear(
-    PyImportSortBear,
-    ('from os import read\nfrom sys import *\n',),
-    ('from os import read\nfrom os import *\n',),
-    settings={'combine_star_imports': True})
```
jayvdb: Why are a lot of test classes like `PyImportSortBearConfigsTest` being removed?
It looks like some useful test scenarios are being completely removed..?

meetmangukiya: None of the test cases are useful IMO.
They are just testing of the setting is being passed correctly, which is being tested in my test case and as we are moving away from verify_local_bear and since I already have written tests, I didn't want to make the test file bloated. If you want I can revert this change(I am not with this decision though).
Here, the tests are only testing the settings, and having various test cases altering each setting is not quite useful, unless we are testing the tool, which we are not, because it is not our job to test the tool.

jayvdb: The decision to remove the tests should be a separate issue, and definitely a separate commit.

The reason for a separate issue is I'd prefer to see all existing tests pass with your PR , so I can feel safe there is no regression hiding in here , and then we talk about killing some of the tests after.

The utility of testing the tool is a reasonable debate to have.  IMO our users need to know our settings all work correctly.  And we need to know when a setting is broken by a major release of upstream, where they may delete functionality they do not like to maintain.  Or they change the setting name.  If we are not testing these things, the breakage is seen by our users.

jayvdb:
ack 6fa0800

jayvdb:
@rultor merge

PR #1367, opened 2017-01-30
ShellCheckBear.py: Importing CabalRequirement

naveentata:
CabalRequirement is imported to ShellCheckBear

Fixes https://github.com/coala/coala-bears/issues/1359

### DIFF
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -135,7 +135,7 @@ class GitCommitBear(GlobalBear):
         :param shortlog_length:          The maximum length of the shortlog.
                                          The newline character at end does not
                                          count to the length.
-        :param regex:                    A regex to check the shortlog with.
+        :param shortlog_regex:           A regex to check the shortlog with.
         :param shortlog_trailing_period: Whether a dot shall be enforced at end
                                          end or not (or ``None`` for "don't
                                          care").


bears/shell/ShellCheckBear.py
```diff
@@ -1,4 +1,6 @@
 from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.CabalRequirement import (
```
Techievena: Why the braces?

Techievena:
Importing is not in imperative form. Also use `Closes` instead of `Fixes`

bears/shell/ShellCheckBear.py
```diff
@@ -11,9 +15,14 @@ class ShellCheckBear:
 
     A gallery of bad code that can be detected is available at
     <https://github.com/koalaman/shellcheck/blob/master/README.md>.
+
+    Check <https://hackage.haskell.org/package/ShellCheck> for more
+    information.
     """
 
     LANGUAGES = {'sh', 'bash', 'ksh', 'dash'}
+    REQUIREMENTS = {CabalRequirement(package='shellcheck', version='0.4.4'),
```
jayvdb: see `.ci/deps.shellcheck.sh` for the appropriate version.

jayvdb:
ack a972fcc

jayvdb:
needs rebase

sils:
I object to this PR as per https://github.com/coala/coala-bears/issues/1359#issuecomment-277489451

jayvdb:
blocked by sils

jayvdb:
Since adding both is blocked by sils atm, could you only add the `DistributionRequirement` for now, as it is via apt-get that we are installing it currently (and .deb) and thus it is where the version is coming from.

Also in your commit message should describe where/why the version was chosen.

naveentata:
This has been moved to https://github.com/coala/coala-bears/pull/1496

PR #1386, opened 2017-02-01
Only test non-Python bears on Python 3.6

jayvdb:


jayvdb:
This is mostly a WIP,, as it needs to be split into multiple commits and some of the voodoo bash stuff needs to be refined, and some additional `Requirement` need to be added for the current algorithm to correctly split Python bears from non-Python bears.

However, the concept behind it, of not testing non-Python bears on each version of Python, needs some arch review.
The main performance improvement is that any job which doesnt require `sudo` can run start (and finish) much faster, which gives better feedback for many times of -bear work.  Due to the PPAs, some jobs need to use `sudo`, and these wont run any faster, so the actual elapsed completion time wont be reduced, at least not much.

There are many further improvements that can be made, such as splitting node and ruby and python bears into separate jobs (these are already manageable by Requirement classes), and R and Haskell and Java should follow soon with Requirement classes being added.  That will allow testing using multiple versions of each runtime concurrently.

sils:
@jayvdb gorgeous idea! Put th concept explanation into the commit message please or maybe right into the code as a comment.

.ci/bears.prune.sh
```diff
@@ -0,0 +1,7 @@
+# Delete empty bear directories
+
+for dir in bears/*/; do
+  if [[ -z "$(ls $dir/ 2>/dev/null | egrep -v '(__|.xml)')" ]]; then
```
jayvdb: So this looks for directories without any Bear's in it..

The `__` is for `__init__` and `__pycache__`
`.xml` is because of `bears/scala/scalastyle_config.xml`

I think a much better approach would be to look for one file matching `*Bear.py`

.travis.yml
```diff
@@ -101,6 +107,17 @@ before_script:
     if [[ "$TRAVIS_PYTHON_VERSION" != "3.3" ]]; then
       python -m nltk.downloader punkt maxent_treebank_pos_tagger averaged_perceptron_tagger
     fi
+  # Only keep non-Python bears on Python 3.4
+  - >
+    if [[ "$BEARS" != "all" ]]; then
+      rm $(comm -23 <(ls bears/*/[A-Za-z]*.py | grep -v general | sort) <(grep -l PipRequirement bears/*/*.py | sort))
+      rm bears/general/CPDBear.py
+      rm bears/generate_package.py tests/generate_packageTest.py
```
jayvdb: the special case for `bears/generate_package.py` -> `tests/generate_packageTest.py` could be added to `tests.prune.sh`, instead of here in .travis.yml

.travis.yml
```diff
@@ -101,6 +107,17 @@ before_script:
     if [[ "$TRAVIS_PYTHON_VERSION" != "3.3" ]]; then
       python -m nltk.downloader punkt maxent_treebank_pos_tagger averaged_perceptron_tagger
     fi
+  # Only keep non-Python bears on Python 3.4
+  - >
+    if [[ "$BEARS" != "all" ]]; then
+      rm $(comm -23 <(ls bears/*/[A-Za-z]*.py | grep -v general | sort) <(grep -l PipRequirement bears/*/*.py | sort))
```
jayvdb: `grep -v general` -- all `bears/general/` are Python language bears without any dependencies, except CPDBear (see next line).

This needs to be tidied up somehow.

bears/ruby/RuboCopBear.py
```diff
@@ -19,8 +19,7 @@ class RuboCopBear:
     """
 
     LANGUAGES = {'Ruby'}
-    REQUIREMENTS = {GemRequirement('rubocop'),
-                    PipRequirement('pyyaml', '3.12')}
```
jayvdb: This was done because the shell foo in .travis.yml needs to find non-Python bears.  It is using the presence of `PipRequirement`, and there are two non-Python bears which use PipRequirement.

.travis.yml logic needs to be rewritten, and then this isnt necessary.

.ci/generate_bear_requirements.py
```diff
@@ -81,6 +82,15 @@ def get_inherited_requirements():
     in_inherited = False
     with open(os.path.join(PROJECT_DIR, 'requirements.txt'), 'r') as file:
         for line in file.read().splitlines():
+            line = line.strip()
+            if not line.startswith('#'):
+                m = re.match('^[A-Za-z._-]+', line)
```
jayvdb: Adding more requirements parsing ;-(  Should be using an existing tool for this.
This entire modification to this script wont be necessary when the pyyaml related changes are backed-out.

.ci/bears.prune.sh
jayvdb: Oh, the problem with *Bear is we have directories with only subdirectories, like `bears/vcs`.  Solvable, and probably still clearer to take this approach.

jayvdb:
Note bca64a0 was an intentional breakage to show codecov reports less than 100% coverage when one bear doesnt work. See it on
https://codecov.io/gh/jayvdb/coala-bears/branch/travis-matrix-python-speedup/commits

jayvdb:
The circle breakage has been addressed in https://github.com/coala/coala-bears/pull/1406 , which only solves part of this as it is easier to test that part by itself.

Mixih:
@jayvdb marking WIP because title is WIP. If ready for review please change the title.

.ci/bears.prune.sh
```diff
@@ -0,0 +1,160 @@
...
+  exit 1
+fi
+
+echo Other exe $executable_linter_bears $executable_other_bears
+
+python_bears="$pip_only_requirement_bears $clang_bears $other_bears"
+
+non_python_bears=$(comm -23 <(ls $bears) <(ls $python_bears))
+
+cabal_requirement_bears="$cabal_requirement_bears bears/haskell/HaskellLintBear.py bears/shell/ShellCheckBear.py"
```
anctartica: `ShellCheckBear` has a Cabal Requirement now

Ref: https://github.com/coala/coala-bears/pull/1496

anctartica: Also this can needs to be solved: https://github.com/coala/coala-bears/issues/1358 so, that we can remove `HaskellLintBear`

.ci/bears.dirs.prune.sh
```diff
@@ -0,0 +1,13 @@
...
+
+dirs=$(find bears -depth -type d -and -not -name '__pycache__')
+
+for dir in $dirs; do
+  bears=$(ls $dir/*Bear.py 2>/dev/null || true)
+  subdirs=$(ls -d $dir/*/ 2>/dev/null || true)
+  if [[ -z "$bears""$subdirs" ]]; then
+    echo Removing $dir
+    rm -rf $dir
+  fi
```
anctartica: Why is it important to delete the empty directories? What if we left them as it is?

Godeps/Godeps.json
```diff
@@ -0,0 +1,25 @@
+{
```
jayvdb: https://github.com/coala/coala-bears/issues/2481

DESCRIPTION
```diff
@@ -0,0 +1,3 @@
+Package: coala
```
jayvdb: https://github.com/coala/coala-bears/issues/2482

Makefile.PL
```diff
@@ -0,0 +1,6 @@
+use ExtUtils::MakeMaker;
```
jayvdb: https://github.com/coala/coala-bears/issues/2483

.ci/bears.dirs.prune.sh
jayvdb: I forget.

probably another bear will complain about a missing `__init__.py`, or we have missing coverage of those `__init__.py`

.travis.yml
```diff
@@ -69,6 +69,157 @@ jobs:
...
+      node_js: 6
+
+    - language: ruby
+      rvm: 2.2
+    - language: ruby
+      rvm: 2.1
+    - language: ruby
+      rvm: 2.0
+
+    - language: haskell
```
jayvdb: this job isnt working yet in the new PR https://github.com/coala/coala-bears/pull/2910

the `TRAVIS_GHC_ROOT=/opt/ghc` env setting below might be relevant.

setup.py
```diff
@@ -413,6 +414,18 @@ def read_requirements(filename):
              if not any(req.startswith(ignore)
                         for ignore in ignore_requirements)]
 
+python_bears_deps = True
+
+config = configparser.ConfigParser()
+config.read('setup.cfg')
+if 'setup:custom' in config.sections():
```
jayvdb: This is a separate issue, and hasnt been fixed yet.

PR #1391, opened 2017-02-02
WIP: CPDBear: Use Redundancy aspects

Asnelchristian:


sils:
CC @userzimmermann instead of me :P

bears/general/CPDBear.py
```diff
@@ -91,7 +89,8 @@ def run(self, language: str, minimum_tokens: int=20,
         files = ','.join(self.file_dict.keys())
         executable = which('pmd') or which('run.sh')
         arguments = ('bash', executable, 'cpd', '--skip-lexical-errors',
-                     '--minimum-tokens', str(minimum_tokens),
+                     '--minimum-tokens', str(
+                            Redundancy.Clone.min_clone_tokens.default),
```
sils: uhm coala doesn't populate those yet properly right so we can't use it that way yet

userzimmermann: Those arguments will then also be generated automatically from the aspects definitions...

Asnelchristian: yeah, i was wondering about that, particularly on how the value of the tastes would be accessed ( modified ) by the user

sils: @userzimmermann does that work yet? @Asnelchristian can we have a test for the setting?

Asnelchristian: nope it doesn't work yet, i just forgot to remove this 
So actually, there is no mechanism, to access the tastes of a given aspects and change their value

Asnelchristian: Maybe we should write a decorator for that: something like 
````
@linter(...)
@set_aspects( {'Redundancy.Clone.minimun_tokens':'1}
```
Having something like 
```
LANGUAGES = {....}
ASPECTS = { #a set of aspectbase element} 
```

bears/general/CPDBear.py
```diff
@@ -127,4 +128,5 @@ def run(self, language: str, minimum_tokens: int=20,
                         'that you have more code than you need. Consider'
                         ' refactor your code to remove one of the'
                         ' occurrences. For more information go here:'
-                        'http://tinyurl.com/coala-clone'))
+                        'http://tinyurl.com/coala-clone'),
```
sims1253: Does tinyurl support previewing where you go similar to bit.ly? if not, I'd prefer using bit.ly or just the long url

sims1253: to clarify: http://bit.ly/line-spots+ sends you to a page that shows you where http://bit.ly/line-spots leads to. (The plus at the end does that)

bears/general/CPDBear.py
```diff
@@ -5,6 +5,7 @@
 from coalib.misc.Shell import run_shell_command
 from coalib.results.Result import Result
 from coalib.results.SourceRange import SourceRange
+from coalib.bearlib.aspects.Redundancy import Redundancy
 
 
 class CPDBear(GlobalBear):
```
userzimmermann: @Asnelchristian Hi :) I'm wondering if we could do something like:

```python
class CPDBear(GlobalBear, aspects=[...]):
    ...
```

@sils Would look fancy, right? :D

Let's try to implement it with some metaclass magic... As aspect team task!

sils:
Sounds great as long as it doesn't delay the whole thing by a lot

bears/general/CPDBear.py
Asnelchristian: uhm... i like the idea

userzimmermann:
@Asnelchristian @sils @sims1253 @Mixih Let's get https://github.com/coala/coala/issues/3966 through for making PRs like this one here really work...

userzimmermann:
Hey @Asnelchristian :) You can add those aspects to the bear now...

bears/general/CPDBear.py
jayvdb: We shouldnt be using external shorteners.  We have our own shortener at coala.io :/

PR #1396, opened 2017-02-03
ShellCheckBear: Add option to ignore specific rules

adhikasp:
This extend the ShellCheckBear to use **ignore rules** feature provided by shellcheck, see
https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-errors-in-one-specific-run

The commandline argument for ignoring rules is `-e SCXXX1,SCXXX2,SCXXX3`

Closes https://github.com/coala/coala-bears/issues/1329

### DIFF
--- a/bears/shell/ShellCheckBear.py
+++ b/bears/shell/ShellCheckBear.py
@@ -20,8 +20,14 @@ class ShellCheckBear:
     CAN_DETECT = {'Syntax', 'Security', 'Undefined Element', 'Unused Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file, shell: str='sh'):
+    def create_arguments(filename, file, config_file, shell: str='sh',
+                         shellcheck_ignore: list=None):
         """
         :param shell: Target shell being used.
+        :param shellcheck_ignore: List of linting rules that should be ignored.
         """
-        return '--f', 'gcc', '-s', shell, filename
+        args = ('--f', 'gcc', '-s', shell, filename)
+        if shellcheck_ignore:
+            args += ('-e', ','.join(shellcheck_ignore))
+
+        return args
--- a/tests/shell/ShellCheckBearTest.py
+++ b/tests/shell/ShellCheckBearTest.py
@@ -17,7 +17,55 @@ X+=2    # += operator is supported by bash and not by sh
 echo $X
 """
 
+trigger_sc2164 = """
+#!/usr/bin/env sh
+for dir in */
+do
+  cd "$dir"
+  convert index.png index.jpg
+done
+"""
 
+trigger_sc2060 = """
+#!/usr/bin/env sh
+tr -cd [:digit:]
+"""
+
+good_zero_byte = ''
+good_only_eol = '\n'
+good_only_eol_eol = '\n\n'
+good_only_comment = '# this is a bash comment'
+
+invalid_file_list = (invalid_file, trigger_sc2164, trigger_sc2060,)
 ShellCheckBearTest = verify_local_bear(ShellCheckBear,
                                        valid_files=(valid_file,),
-                                       invalid_files=(invalid_file,))
+                                       invalid_files=invalid_file_list,
+                                       )
+
+small_files = (good_zero_byte, good_only_eol, good_only_eol_eol,
+               good_only_comment,)
+SmallFileTest = verify_local_bear(ShellCheckBear,
+                                  valid_files=small_files,
+                                  invalid_files=(),
+                                  )
+
+IgnoreSC2164Test = verify_local_bear(ShellCheckBear,
+                                     valid_files=(trigger_sc2164,),
+                                     invalid_files=(),
+                                     settings={
+                                         'shellcheck_ignore': ['SC2164']},
+                                     )
+IgnoreSC2060Test = verify_local_bear(ShellCheckBear,
+                                     valid_files=(trigger_sc2060,),
+                                     invalid_files=(),
+                                     settings={
+                                         'shellcheck_ignore': ['SC2060']},
+                                     )
+MultipleIgnoreTest = verify_local_bear(ShellCheckBear,
+                                       valid_files=(trigger_sc2164,
+                                                    trigger_sc2060,),
+                                       invalid_files=(),
+                                       settings={
+                                           'shellcheck_ignore': ['SC2164',
+                                                                 'SC2060']},
+                                       )


bears/shell/ShellCheckBear.py
```diff
@@ -20,8 +20,16 @@ class ShellCheckBear:
...
 
     @staticmethod
-    def create_arguments(filename, file, config_file, shell: str='sh'):
+    def create_arguments(filename, file, config_file, shell: str='sh',
+                         ignore_rules: list=[]):
         """
         :param shell: Target shell being used.
         """
-        return '--f', 'gcc', '-s', shell, filename
+        args = ('--f', 'gcc', '-s', shell)
```
sims1253: a set doesn't have guaranteed or order. Does order matter?

bears/shell/ShellCheckBear.py
```diff
@@ -20,8 +20,16 @@ class ShellCheckBear:
...
+        if ignore_rules:
+            appended_rules = ','.join(ignore_rules)
+            args += ('-e', appended_rules)
```
sims1253: I think you can just inline the join here

adhikasp:
Updated according to review

Asnelchristian:
needs rebase

bears/shell/ShellCheckBear.py
```diff
@@ -20,8 +20,13 @@ class ShellCheckBear:
     CAN_DETECT = {'Syntax', 'Security', 'Undefined Element', 'Unused Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file, shell: str='sh'):
+    def create_arguments(filename, file, config_file, shell: str='sh',
+                         ignore_rules: list=[]):
         """
         :param shell: Target shell being used.
         """
```
Asnelchristian: what about ``:param ignore_rules:``, you have to add the docs for this setting too

Asnelchristian:
This looks good, just add that doc and rebase, and we are good to go :tada:

Asnelchristian:
ooh you commit shortlog should be ``ShellCheckBear: Add ``ignore_rules`` option``

aptrishu:
`This extend` -> `This extends` in the commit message..

bears/shell/ShellCheckBear.py
```diff
@@ -20,8 +20,14 @@ class ShellCheckBear:
     CAN_DETECT = {'Syntax', 'Security', 'Undefined Element', 'Unused Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file, shell: str='sh'):
+    def create_arguments(filename, file, config_file, shell: str='sh',
+                         ignore_rules: list=[]):
```
jayvdb: `ignore_rules` should be be `shellcheck_ignore` or `shellcheck_ignore_codes`.
See `$ git grep _ignore`

supergr35:
Please rebase. Also marking wip because of failed tests

Madhukar-SH:
@adhikasp Are you working on this yet?

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
-ShellCheckBearTest = verify_local_bear(ShellCheckBear,
-                                       valid_files=(valid_file,),
-                                       invalid_files=(invalid_file,))
+# Check that ignore_rules option working
+ShellCheckBear2Test = verify_local_bear(
+    ShellCheckBear,
+    valid_files=(valid_with_ignore_rules_file,),
+    invalid_files=(),
+    settings={
+        'ignore_rules': 'SC2103, SC2164'})
```
Mixih: shouldn't you be passing a list here for consistency?

bears/shell/ShellCheckBear.py
```diff
@@ -20,8 +20,14 @@ class ShellCheckBear:
     CAN_DETECT = {'Syntax', 'Security', 'Undefined Element', 'Unused Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file, shell: str='sh'):
+    def create_arguments(filename, file, config_file, shell: str='sh',
+                         shellcheck_ignore: list=[]):
```
jayvdb: default should be None unless you need a empty list.

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
+cd /home
+for dir in */
+do
+  cd "$dir"
+  convert index.png index.jpg
+  cd ..
+done
+"""
+
+ShellCheckBear1Test = verify_local_bear(
```
jayvdb: Using `1` and `2` is not descriptive.

Keep this test named `ShellCheckBearTest`, as that avoids unnecessary modifications being added to your patch, and means `git blame` continues to show the authors of these tests.

btw, nice adding `valid_with_ignore_rules_file` to the invalid list of this test.

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
+    ShellCheckBear,
+    valid_files=(valid_file,),
+    invalid_files=(invalid_file, valid_with_ignore_rules_file))
 
-ShellCheckBearTest = verify_local_bear(ShellCheckBear,
-                                       valid_files=(valid_file,),
-                                       invalid_files=(invalid_file,))
+# Check that ignore_rules option working
+ShellCheckBear2Test = verify_local_bear(
```
jayvdb: Name this new test descriptively.

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
+    ShellCheckBear,
+    valid_files=(valid_with_ignore_rules_file,),
+    invalid_files=(),
+    settings={
+        'shellcheck_ignore': ['SC2103', 'SC2164']})
```
jayvdb: The dict will fit on one line, and the settings can have a trailing `,`, and the final `)` on the final line by itself. i.e.
```py
    settings={...},
)
```
Then, additional parameters can be added at the end of the function call without your lines being modified.

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
+cd /home
+for dir in */
+do
+  cd "$dir"
+  convert index.png index.jpg
+  cd ..
+done
+"""
+
+ShellCheckBearTest = verify_local_bear(
```
adhikasp: @jayvdb I think i must break this line (and thus make the entire code block modified) 
because the invalid_files is too long. Any idea about this?

jayvdb: create a variable `invalid_files` or `invalid_file_list`

bears/shell/ShellCheckBear.py
```diff
@@ -20,8 +20,14 @@ class ShellCheckBear:
     CAN_DETECT = {'Syntax', 'Security', 'Undefined Element', 'Unused Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file, shell: str='sh'):
+    def create_arguments(filename, file, config_file, shell: str='sh',
+                         shellcheck_ignore: list=None):
         """
         :param shell: Target shell being used.
+        :param shellcheck_ignore: List of linting rules that should be ignored.
```
all3fox: It is safe to assume that in this case the user knows about shellcheck. So lets rename `shellcheck_ignore` to what it is called in shellcheck: `exclude`

https://github.com/koalaman/shellcheck/blob/22c86256ace897e2fcd780da46bc173fb1bbd5f5/shellcheck.hs#L77

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
+#!/usr/bin/env sh
+echo "Path is $PATH"
+cd /home
+for dir in */
+do
+  cd "$dir"
+  convert index.png index.jpg
+  cd ..
+done
+"""
```
all3fox: Let's split this into two smaller test cases. The first one would test `SC2164`, the second one will test `SC2103`. The first test file can then be as simple as:

```
#!/usr/bin/env sh
cd
```
Please figure out the minimal file for the second error then.

Also, you could add cross-test case: exclude one error but feed the other file and check that the other error is still triggered.

all3fox: Also, please rename the variable to something shorter, like `triggers_sc2164`

tests/shell/ShellCheckBearTest.py
```diff
@@ -17,7 +17,27 @@
...
-                                       invalid_files=(invalid_file,))
+                                       invalid_files=invalid_file_list,
+                                       )
+
+ShellCheckBearTestIgnoreRules = verify_local_bear(
+    ShellCheckBear,
+    valid_files=(valid_with_ignore_rules_file,),
+    invalid_files=(),
+    settings={'shellcheck_ignore': ['SC2103', 'SC2164']},
+)
```
all3fox: Your indentation is not in sync with coala style here. See how `ShellCheckBearTest` is indented above. I suggest you rename `ShellCheckBearTestIgnoreRules` to something shorter, like `IgnoreRulesTest`.

bears/shell/ShellCheckBear.py
jayvdb: This is not correct. See my earlier review comment.

tests/shell/ShellCheckBearTest.py
jayvdb: Some good tests to add are https://github.com/coala/coala-bears/issues/1495

bears/shell/ShellCheckBear.py
adhikasp: I think i'll pass this one because coala mainly use term `_ignore`  rather than `exclude`, see https://github.com/coala/coala-bears/pull/1396#pullrequestreview-24900209

tests/shell/ShellCheckBearTest.py
adhikasp: @all3fox 
For note, cross-test case is always failing in unit test because some kind of bugs, so I can't add them 😞

jayvdb:
ack e193af8

jayvdb:
reack 5b5d7de

jayvdb:
@rultor merge

PR #1401, opened 2017-02-03
PycodestyleBear.py: Remove extra "'" from message

sahilarora535:
Fixes https://github.com/coala/coala-bears/issues/1393

### DIFF
--- a/bears/python/PycodestyleBear.py
+++ b/bears/python/PycodestyleBear.py
@@ -35,7 +35,7 @@ class PycodestyleBear:
         :param max_line_length:
             Limit lines to this length.
         """
-        arguments = [r"--format='%(row)d %(col)d %(code)s %(text)s'"]
+        arguments = [r'--format=%(row)d %(col)d %(code)s %(text)s']
 
         if pycodestyle_ignore:
             arguments.append('--ignore=' + pycodestyle_ignore)


dracarys09:
ack 2555651b264d947208f4badb705a3da0a6ea9941

bears/python/PycodestyleBear.py
```diff
@@ -35,7 +35,7 @@ def create_arguments(
         :param max_line_length:
             Limit lines to this length.
         """
-        arguments = [r"--format='%(row)d %(col)d %(code)s %(text)s'"]
+        arguments = [r"--format='%(row)d %(col)d %(code)s %(text)s"]
```
jayvdb: But the original is `r"--format='...'"` with two single quotes.
I suspect the problem is elsewhere...?  but maybe I dont understand correctly.
The issue says `Fix the regex of PycodestyleBear.`

jayvdb:
unack 2555651

bears/python/PycodestyleBear.py
dracarys09: @jayvdb I didn't notice that... :worried: 

@sahilarora535 ```'``` is not even fixing the problem... have you actually tested it locally before submitting PR ?

dracarys09: @jayvdb I didn't notice that... :worried: 

@sahilarora535 look at how other arguments are appended to ```arguments``` and try doing it in the same way... that will work but this is just a hacky way to go about it...  :smile:

sahilarora535: @dracarys09 I actually tried experimenting, and put something like: 
` arguments = [r"--format='%(row)d %(col)d %(code)s %(text)s'Hello, world!"]`
The output was: `E501 line too long (213 > 79 characters)'Hello, world!`
Hence I suspected that this would be the error, though I myself thought that this has to be the closing for that initial single quote. Will see that.

sahilarora535: @jayvdb I removed the double quotes and replaced with single quotes, and removed the single quotes from inside. This is the convention which was followed elsewhere in other bears. Please review

jayvdb:
This is very wrong.  Please discuss on the issue.

jayvdb:
The main problem is that the discussion should occur on https://github.com/coala/coala-bears/issues/1393 if the solution is different to the recommended solution in the issue.

dracarys09:
it was established in https://github.com/coala/coala-bears/issues/1393#issuecomment-277520777 that this PR will solve the issue... this should be reopended...

CC @jayvdb @Makman2

jayvdb:
ack ed6651b

meetmangukiya:
reack 6d8d105

meetmangukiya:
reack 6d8d105

meetmangukiya:
ack 6d8d105

meetmangukiya:
ack 6d8d1057b230f02e882914700db321a01ebab53b

meetmangukiya:
@rultor merge!

PR #1404, opened 2017-02-04
Add GoTOMLBear

Vamshi99:
This bear uses gotoml to know the type of every key value used in a file.

Closes https://github.com/coala/coala-bears/issues/1375

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -117,6 +117,7 @@ go get -u golang.org/x/tools/cmd/goimports
 go get -u sourcegraph.com/sqs/goreturns
 go get -u golang.org/x/tools/cmd/gotype
 go get -u github.com/kisielk/errcheck
+go get -u github.com/BurntSushi/toml/cmd/tomlv
 
 # Ruby commands
 bundle install --path=vendor/bundle --binstubs=vendor/bin --jobs=8 --retry=3
--- a/bears/go/GoTOMLBear.py
+++ b/bears/go/GoTOMLBear.py
@@ -1,0 +1,23 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.GoRequirement import GoRequirement
+
+
+@linter(executable='gotoml',
+        use_stdin=True,
+        output_format='corrected')
+class GoTOMLBear:
+    """
+    It is used to the know the type of every key used in a file.
+    """
+    LANGUAGES = {'Go'}
+    REQUIREMENTS = {
+               GoRequirement(package='github.com/BurntSushi/toml/cmd/tomlv',
+                             version='0.1.0')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Data type'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file):
+        return ()


bears/go/GoTOMLBear.py
```diff
@@ -0,0 +1,22 @@
...
+
+@linter(executable='gotoml',
+        use_stdin=True,
+        output_format='corrected')
+class GoTOMLBear:
+    """
+    It is used to the know the type of every key used in a file.
+    """
+    LANGUAGES = {'Go'}
+    REQUIREMENTS = {GoRequirement(
```
jayvdb: Can you format it like
```
REQUIREMENTS = {GoRequirement('github.com/BurntSushi/toml/cmd/tomlv',
                              flag='-u')}
```

It is helpful if the first parameter is on the same line as the `*Requirement`

However, the `-u` flag is a bit silly.  That just updates it, which isnt sensible when we have versioning.

.ci/deps.sh
```diff
@@ -117,6 +117,7 @@ go get -u golang.org/x/tools/cmd/goimports
 go get -u sourcegraph.com/sqs/goreturns
 go get -u golang.org/x/tools/cmd/gotype
 go get -u github.com/kisielk/errcheck
+go get github.com/BurntSushi/toml/cmd/tomlv
```
jayvdb: no `-u` ?

bears/go/GoTOMLBear.py
```diff
@@ -0,0 +1,22 @@
...
+
+@linter(executable='gotoml',
+        use_stdin=True,
+        output_format='corrected')
+class GoTOMLBear:
+    """
+    It is used to the know the type of every key used in a file.
+    """
+    LANGUAGES = {'Go'}
+    REQUIREMENTS = {GoRequirement(package='github.com/BurntSushi/toml/cmd/tomlv',
```
Vamshi99: @jayvdb This is the problem when I write in this format. Should I write in the previous format?

jayvdb: No.  Go back and read me previous message.

bears/go/GoTOMLBear.py
```diff
@@ -0,0 +1,23 @@
...
+@linter(executable='gotoml',
+        use_stdin=True,
+        output_format='corrected')
+class GoTOMLBear:
+    """
+    It is used to the know the type of every key used in a file.
+    """
+    LANGUAGES = {'Go'}
+    REQUIREMENTS = {
+               GoRequirement(package='github.com/BurntSushi/toml/cmd/tomlv',
```
jayvdb: "package=" is not necessary.
See my comment from 3 hrs ago:
https://github.com/coala/coala-bears/pull/1404#discussion_r99460985

bears/go/GoTOMLBear.py
```diff
@@ -0,0 +1,23 @@
...
+                             version='0.1.0')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Data type'}
```
yukiisbored: I think `Formatting` and `Syntax` is better, because the linter itself is for TOML and doesn't detect the "type"

Vamshi99: @yukiisbored Ok, I will change it to Formatting.

Vamshi99:
This pull request is moved to https://github.com/coala/coala-bears/pull/1409

PR #1406, opened 2017-02-04
coala-bears.cabal: Create and use

jayvdb:
coala has three haskell dependencies:

- ghc-mod 5.6.0.0
- hlint 1.9.26
- ShellCheck 0.4.1

Currently CI pulls these haskell packages from various sources
including Ubuntu apt-get package database, manually installing debs,
and running `cabal install foo` using two different Cabal versions.

This barely works, and those versions are not compatible.
The closest compatible combination requires bumping hlint to 1.9.27.

The dependencies have been placed into a `coala-bears.cabal` so they
can be easily fetched and installed using `cabal install`, which
will look for a combination of versions that meets all criteria.

On Travis, removing 'haskell-platform' from the apt-get package list
reduces the number of packages fetched, most of which were not
the correct version needed by the coala dependencies.
Removing this prevents conflicts, and speeds up the build.

Using cabal properly with a .cabal file also solves the
cabal caching problems.

Fixes https://github.com/coala/coala-bears/issues/1380


.ci/deps.cabal.sh
```diff
@@ -0,0 +1,13 @@
+set -e
+
+# cabal-install 1.22 is the minimum tested version
+cabal --version
+
+cabal update
+
+rm $HOME/.cabal/config
```
jayvdb: This maybe unnecessary.
If the version of cabal-install is too low, and needs re-installing here, the config could be stale and will make the new cabal less functional.

.ci/deps.sh
```diff
@@ -82,28 +79,17 @@ fi
...
-
 # NPM commands
-sudo rm -rf $(which alex)  # Delete ghc-alex as it clashes with npm deps
+if [[ -n "$(which alex)" ]]; then
+  # Delete ghc-alex as it clashes with npm deps
+  echo "Removing $(which alex)"
+  sudo rm -rf $(which alex)
+fi
 npm install
+if [[ -z "$(which alex)" ]]; then
```
jayvdb: This is strangley necessary for travis, where cached npm doesnt include alex. Very odd.

.travis.yml
```diff
@@ -80,9 +80,14 @@ before_install:
   - nvm use 4
   # Remove Ruby directive from Gemfile as this image has 2.2.5
   - sed -i '/^ruby/d' Gemfile
-  - mkdir -p ~/.apt-cache
+  - which alex || true
```
jayvdb: This is debugging for the npm problem.

.travis.yml
```diff
@@ -103,6 +108,7 @@ before_script:
     fi
 
 script:
+  - which alex || true
```
jayvdb: Remove.

coala-bears.cabal
```diff
@@ -0,0 +1,13 @@
+Name: coala-bears
+Version: 0.10
+cabal-version: >= 1.2
+build-type: Simple
+
+-- safe 0.3.9 is needed to coexist with Cabal 1.16
+-- safe 0.3.10 requires Cabal 1.18
+
+-- Note trusty apt package "haskell-platform" includes mtl-2.1.2
```
jayvdb: Check other ubuntu to give a more useful note.

jayvdb:
Would like confirmation creating a .cabal is the best approach.
It means we can easily re-use it in the docker.

Alt. The shell script could create a temp dir + file.

.ci/deps.sh
jayvdb: The reason is that after the first time, if `haskell-platform` isnt installed, the next time the npm alex is deleted.

coala-bears.cabal
jayvdb: Confirmed.  http://packages.ubuntu.com/trusty/haskell-platform actually has two version
```
libghc-mtl-dev (<< 2.1.2+)
    Haskell monad transformer library for GHC 

libghc-mtl-dev (>= 2.1.2)
```

Im not sure how to pick the desired one on travis.

xenial + only have the desired version 2.2.1. http://packages.ubuntu.com/xenial/libghc-mtl-dev

jayvdb:
https://travis-ci.org/jayvdb/coala-bears/builds/201733391 and https://circleci.com/gh/jayvdb/coala-bears/622 are rebuilds without cache.

.travis.yml
```diff
@@ -80,9 +80,8 @@ before_install:
   - nvm use 4
   # Remove Ruby directive from Gemfile as this image has 2.2.5
   - sed -i '/^ruby/d' Gemfile
-  - mkdir -p ~/.apt-cache
   - if [[ "$UNSUPPORTED" != true ]]; then bash .ci/deps.sh; fi
-  - if [[ "$UNSUPPORTED" != true ]]; then bash .ci/deps.shellcheck.sh; fi
+  - if [[ "$UNSUPPORTED" != true ]]; then bash .ci/deps.cabal.sh; fi
```
Mixih: cabal deps should go before maindeps because alex?

jayvdb: it makes no difference on travis, as its image doesnt have ghc's alex pre-installed.

Mixih: ah ok

Mixih:
ack 06329a5

.ci/deps.cabal.sh
```diff
@@ -0,0 +1,11 @@
+set -e
+
+# cabal-install 1.22 is the minimum tested version
+cabal --version
```
Mixih: point of this if we don't act on it in any way? debug info?

jayvdb: ya, tracking it, as it changes on circle.

jayvdb:
reack 06329a5

jayvdb:
@rultor merge

PR #1413, opened 2017-02-06
The YAML markup language is called "YAML", not "YML"

adrienverge:
As [someone said](https://github.com/ceph/s3-tests/commit/e17c56a) on the internet:

    Say ".yaml" not ".yml".
    This is not MS-DOS, and YML is a Yahoo XML dialect.

Similarly, we use 'json', not 'jsn'.
This commit renames the `yml` bear directory to `yaml`.

References:
https://github.com/ceph/s3-tests/commit/e17c56a
https://github.com/adrienverge/yamllint/commit/41733fc
http://yaml.org/spec/1.2/spec.html

### Checklist

- [x] I read the [commit guidelines](http://coala.io/commit) and I've followed them.
      pass on every single commit!


tests/yml/RAMLLintBearTest.py
```diff
@@ -1,4 +1,4 @@
-from bears.yml.RAMLLintBear import RAMLLintBear
+from bears.yaml.RAMLLintBear import RAMLLintBear
```
meetmangukiya: Why not rename the test directory as well, then ?

meetmangukiya: Also as gitmate said, please reduce the length of shortlog to 50 chars

adrienverge: @meetmangukiya Thanks for reviewing.

> Why not rename the test directory as well, then ?

Sure!

> Also as gitmate said, please reduce the length of shortlog to 50 chars

It is already, the comment refers to an older commit.

adrienverge:
Updated.

jayvdb:
There is also `tests/ruby/RuboCopBearTest.py` and `tests/swift/TailorBearTest.py` which use `tests/ruby/test_files/rubocop_config.yml` and `tests/swift/test_files/tailor.yml` respectively.  Those need renaming.

`.ci/appveyor.yml` can also be renamed, but that should be a separate PR.
`.travis.yml` cant be renamed, as far as I know.
`.codecov.yml` is more likely able to be renamed; if not, raise a bug upstream - the author is very likely to be receptive.
I dont know about `circle.yml`

adrienverge:
@jayvdb thanks for having a look here!

I intentionally didn't touch file extensions as they aren't directly related to this change (which is about the language name).

Since you raise this point, I updated this PR for `rubocop_config.yaml` and `tailor.yaml`.

About CI config files, you're right: if done, it should be in a separated PR.

jayvdb:
ack 4a101b7

jayvdb:
ping @meetmangukiya

sils:
@rultor merge

sils:
Hey @adrienverge thanks for your contribution! Do you have any idea on how the contribution process or using coala could have been smoother? (See also coala.io/usability)

sils:
@adrienverge FYI rultor will release it as a prerelease so you'll be able to use it right now if you install coala with `--pre`

adrienverge:
@sils The contribution process was pretty smooth actually. Thanks!

PR #1421, opened 2017-02-08
JSONFormatBear: Parse Exception message

Sanny26:
This change parses the default exception message to print
only 'Expecting property name enclosed in double quotes.'
instead of 'Expecting property name enclosed in double
quotes: line 2 column 7 (char 8)'.It also shows line and
column where the error is raised.

Closes https://github.com/coala/coala-bears/issues/35

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/js/JSONFormatBear.py
+++ b/bears/js/JSONFormatBear.py
@@ -1,5 +1,6 @@
 import json
 from collections import OrderedDict
+from re import match
 
 from coala_utils.param_conversion import negate
 from coalib.bearlib import deprecate_settings
@@ -37,10 +38,14 @@ class JSONFormatBear(LocalBear):
             json_content = json.loads(''.join(file),
                                       object_pairs_hook=OrderedDict)
         except JSONDecodeError as err:
+            err_content = match(r'(.*): line (\d+) column (\d+)', str(err))
             yield Result.from_values(
                 self,
-                'This file does not contain parsable JSON. ' + repr(str(err)),
-                file=filename)
+                'This file does not contain parsable JSON. ' +
+                err_content.group(1) + '.',
+                file=filename,
+                line=int(err_content.group(2)),
+                column=int(err_content.group(3)))
             return
 
         corrected = json.dumps(json_content,
--- a/bears/js/JSONFormatBear.py
+++ b/bears/js/JSONFormatBear.py
@@ -34,6 +34,13 @@ class JSONFormatBear(LocalBear):
         :param escape_unicode: Whether or not to escape unicode values using
                                ASCII.
         """
+        # Output a meaningful message if empty file given as input
+        if len(file) == 0:
+            yield Result.from_values(self,
+                                     'This file is empty.',
+                                     file=filename)
+            return
+
         try:
             json_content = json.loads(''.join(file),
                                       object_pairs_hook=OrderedDict)
--- a/tests/js/JSONFormatBearTest.py
+++ b/tests/js/JSONFormatBearTest.py
@@ -1,5 +1,10 @@
-from bears.js import JSONFormatBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
+from queue import Queue
+
+from bears.js.JSONFormatBear import JSONFormatBear
+from coalib.testing.LocalBearTestHelper import (verify_local_bear,
+                                                LocalBearTestHelper)
+from coalib.results.Result import Result
+from coalib.settings.Section import Section
 
 
 test_file1 = """{
@@ -25,7 +30,32 @@ unicode_file = """{
 }"""
 
 
-JSONFormatBearTest = verify_local_bear(JSONFormatBear.JSONFormatBear,
+test_file4 = """{
+    a: 5
+}"""
+
+
+class JSONTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('')
+        self.uut = JSONFormatBear(self.section, Queue())
+
+    def test_exception_result(self):
+        self.check_results(
+            self.uut,
+            test_file4.split('\n'),
+            [Result.from_values('JSONFormatBear',
+                                'This file does not contain parsable JSON. '
+                                'Expecting property name enclosed in '
+                                'double quotes.',
+                                file='default',
+                                line=2,
+                                column=5)],
+            filename='default')
+
+
+JSONFormatBearTest = verify_local_bear(JSONFormatBear,
                                        valid_files=(test_file1, test_file2),
                                        invalid_files=(test_file3,
                                                       unicode_file,
@@ -34,20 +64,20 @@ JSONFormatBearTest = verify_local_bear(JSONFormatBear.JSONFormatBear,
                                                       '{"a":5,"b":5}'))
 
 
-JSONFormatBearSortTest = verify_local_bear(JSONFormatBear.JSONFormatBear,
+JSONFormatBearSortTest = verify_local_bear(JSONFormatBear,
                                            valid_files=(test_file1,),
                                            invalid_files=(test_file2,),
                                            settings={'json_sort': 'true'})
 
 
-JSONFormatBearTabWidthTest = verify_local_bear(JSONFormatBear.JSONFormatBear,
+JSONFormatBearTabWidthTest = verify_local_bear(JSONFormatBear,
                                                valid_files=(test_file3,),
                                                invalid_files=(test_file2,),
                                                settings={
                                                    'indent_size': '3'})
 
 
-JSONFormatBearUnicodeTest = verify_local_bear(JSONFormatBear.JSONFormatBear,
+JSONFormatBearUnicodeTest = verify_local_bear(JSONFormatBear,
                                               valid_files=(unicode_file,),
                                               invalid_files=(),
                                               settings={'escape_unicode':
--- a/tests/js/JSONFormatBearTest.py
+++ b/tests/js/JSONFormatBearTest.py
@@ -54,6 +54,15 @@ class JSONTest(LocalBearTestHelper):
                                 column=5)],
             filename='default')
 
+    def test_exception_empty_file(self):
+        self.check_results(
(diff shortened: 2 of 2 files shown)


bears/js/JSONFormatBear.py
```diff
@@ -37,10 +38,16 @@ def run(self, filename, file,
             json_content = json.loads(''.join(file),
                                       object_pairs_hook=OrderedDict)
         except JSONDecodeError as err:
+            err_content = str(err).split(':')
+            err_msg = err_content[0]+'.'
+            position = findall(r'\d+', err_content[1])
```
Makman2: I think you should make a single regex and use capture groups accordingly, so your parsing just requires a single match:
```
m = match(r'(.*): line (\d+) column (\d+)'
err_msg = m.group(1) + '.'
...
    line=int(m.group(1))
    column=int(m.group(2))
```

bears/js/JSONFormatBear.py
```diff
@@ -37,10 +38,16 @@ def run(self, filename, file,
...
             yield Result.from_values(
                 self,
-                'This file does not contain parsable JSON. ' + repr(str(err)),
-                file=filename)
+                'This file does not contain parsable JSON. ' +
+                repr(str(err_msg)),
```
Makman2: though you haven't introduced it, if you want you can push another commit which corrects this to just `err_msg` instead of `repr(str(err_msg))` (as `err_msg` is already a string, `str` has no effect + `repr` also renders the string like it has to look in a python file when you write this string, just the message itself is okay).

bears/js/JSONFormatBear.py
```diff
@@ -37,10 +38,16 @@ def run(self, filename, file,
             json_content = json.loads(''.join(file),
                                       object_pairs_hook=OrderedDict)
         except JSONDecodeError as err:
+            err_content = str(err).split(':')
```
aptrishu: use regex and write a test for it.. to verify the output.

aptrishu: Ok, @Makman2 has pointed it out.. :+1:

Sanny26:
@Makman2 If a user runs it over an empty file, the output it prints is 'The line belonging to the following result cannot be printed because it refers to a line that doesn't seem to exist in the given file'. 
As a user, I think just the error message 'Expecting value' is sufficient. What do you suggest? Should that message be displayed? 

And, yeah I'l change it to single expression. I was just tempted to use split because I remember reading that split is fast as compared to re.split and regex is generally slow.

bears/js/JSONFormatBear.py
Sanny26: @aptrishu coverage is 100%, so is a test needed?

Makman2:
> @Makman2 If a user runs it over an empty file, the output it prints is 'The line belonging to the following result cannot be printed because it refers to a line that doesn't seem to exist in the given file'.
As a user, I think just the error message 'Expecting value' is sufficient. What do you suggest? Should that message be displayed?

If you can catch this case it would be good. Though "Expecting value" is bad imo, I mean if I would read the message my brain would be totally confused. Rather say "file is empty", because that's the case ;)

bears/js/JSONFormatBear.py
Makman2: Problem is that those tests don't check for the actual result message. If possible write some tests that check whether your regex is working correctly^^

Sanny26:
@Makman2 @aptrishu I added the tests. Could you review them?

bears/js/JSONFormatBear.py
```diff
@@ -33,14 +39,24 @@ def run(self, filename, file,
         :param escape_unicode: Whether or not to escape unicode values using
                                ASCII.
         """
+        if stat(filename).st_size == 0:
```
Makman2: naaah please use `file`:
`if len(file) == 0:`
\+ I would also add a little comment explaining why you explicitly catch this case
\+ I would move this into a separate commit

bears/js/JSONFormatBear.py
```diff
@@ -33,14 +39,24 @@ def run(self, filename, file,
...
+            yield Result.from_values(self,
+                                     'This file is empty',
```
Makman2: Missing period at end

tests/js/JSONFormatBearTest.py
```diff
@@ -25,6 +26,20 @@
...
+    def test_parse_err_msg(self):
+        valid_msg = ['Expecting value: line 20 column 30 (char 45)',
+                     'Expecting value: line 1 column 1 ']
+        invalid_msg = ['Expecting property change',
+                       'Expecting line 20 column 30',
+                       'Expecting value: line column']
+        for msg in valid_msg:
+            self.assertTrue(JSONFormatBear.parse_err_msg(msg))
+        for msg in invalid_msg:
+            self.assertFalse(JSONFormatBear.parse_err_msg(msg))
```
Makman2: Could you rather try to assert for the values? See https://github.com/coala/coala-bears/blob/master/tests/vcs/git/GitCommitBearTest.py for how this works^^

Makman2: Then you can also inline `parse_err_msg` again^^

Sanny26: For tests, I can assert group(1), group(2), group(3) for the expected output. Would that be sufficient?
And by inline, did you mean something like this?
```
def my_function():
    def other_function():
        pass
```

Makman2: inlining means in this case to inline the content, not just the function ;)

> For tests, I can assert group(1), group(2), group(3) for the expected output. Would that be sufficient?

I think you should still assert for the bear results, as they get delivered to the user, and that's where we want to test ideally.

Asnelchristian:
@Sanny26 would you address @Makman2 issues ? :sweat_smile:

Asnelchristian:
needs a rebase

Sanny26:
@Asnelchristian I was busy with few evaluations....will complete by sunday.

bears/js/JSONFormatBear.py
```diff
@@ -1,5 +1,7 @@
 import json
+from os import stat
```
sahilarora535: As @Makman2 suggested, use `len`, `stat` no required. Also, if the file has say a newline, or any other whitespace character, and is not empty, then your code will give the same error since the file would not be empty in size but actually is a `null` JSON.

Sanny26:
@Makman2  I made the suggested changes. For tests, I used check_results function from LocalBearTestHelper. Or am I supposed to write it in a similar fashion as GitCommitBearTest only?
And, if these tests look fine, then I'll move the empty file test case part to a new commit.

tests/js/JSONFormatBearTest.py
```diff
@@ -1,6 +1,9 @@
-from bears.js import JSONFormatBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
-
+from queue import Queue
+from bears.js.JSONFormatBear import JSONFormatBear
+from coalib.testing.LocalBearTestHelper import (verify_local_bear,
+                                                LocalBearTestHelper)
+from coalib.results.Result import Result
+from coalib.settings.Section import Section
```
Makman2: add a newline :)

tests/js/JSONFormatBearTest.py
```diff
@@ -1,6 +1,9 @@
-from bears.js import JSONFormatBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
-
+from queue import Queue
```
Makman2: could you add a newline here to separate system imports from package imports? :)

tests/js/JSONFormatBearTest.py
```diff
@@ -25,7 +28,40 @@
...
+        self.section = Section('')
+        self.uut = JSONFormatBear(self.section, Queue())
+
+    def test_exception_result(self):
+        self.check_results(
+            self.uut,
+            test_file4.split('\n'),
+            [Result.from_values('JSONFormatBear',
+                                '''This file does not contain parsable JSON. \
+Expecting property name enclosed in double quotes.''',
```
Makman2: that looks bad :)
rather use single quotes and do this:
```
Result.from_values('JSONFormatBear',
                   'This file does not contain parsable JSON. '
                       'Expecting property name enclosed in '
                       'double quotes.'
```
or
```
Result.from_values(
    'JSONFormatBear',
    'This file does not contain parsable JSON. Expecting property '
        'name enclosed in double quotes.',
    file='default',
    ...
```

Makman2:
Exactly what I wanted to see in the tests :)

Looks good, next iteration should be mergable :+1:

Sanny26:
@Makman2  Changes made as suggested! :)

bears/js/JSONFormatBear.py
```diff
@@ -34,6 +34,13 @@ def run(self, filename, file,
         :param escape_unicode: Whether or not to escape unicode values using
```
Makman2: commit message: Add `JSONFormatBear` as tag :)

Makman2:
ack 44dd843

Makman2:
unack ce733cb

Sanny26:
Added JSONFormatBear tag and issue reference as well.

bears/js/JSONFormatBear.py
Makman2: Ah one last: Please remove the "Closes XXXX" reference in the commit message, it does not really apply :)

Makman2:
unack a971c85

Sanny26:
Removed issue reference as suggested.

Makman2:
ack c4ceb3f

Makman2:
rebase and lgtm :)

Sanny26:
Rebase done!

Makman2:
reack 5e7b526 462b953

Makman2:
@rultor merge

Issue #1417, opened 2017-02-07
Most YAML documents use document starts (---)


adrienverge:
Hi,

I am the creator of yamllint, the linter coala uses for YAML.

Since #965 was merged three months ago, coala fails on many projects like Ansible, OpenStack and even yamllint itself, because coala doesn't accept document start markers (`---`) anymore.

Document start markers are commonly used, and required when declaring multiple documents in a single `.yaml` file (see [the spec](http://yaml.org/spec/1.2/spec.html#id2800132)).

The proposed fix in the original issue (#923) was to disable the rule, but the implemented fix (#965) made document starts forbidden.

My opinion is that coala should either require document starts, or disable the rule by default.

Makman2:
Agree.

So task is to use `None` as default value [here](https://github.com/coala/coala-bears/blob/master/bears/yaml/YAMLLintBear.py#L27) and if it's used, leave the config value for the rule, if `True` or `False` are passed, configure the rule with `True`/`False`.

Makman2:
Btw one can remove
```
https://github.com/coala/coala-bears/blob/master/bears/yaml/YAMLLintBear.py#L42
```
and directly feed the value of `document-start` into the configuration dictionary^^

adhikasp:
For reference, this fix would be very similliar with https://github.com/coala/coala-bears/pull/1335.

Issue #923, opened 2016-10-23
YAMLLintBear does not like .travis.yml due to missing start/end


jayvdb:
Many yaml files have no need for start and end markers.  It should be an option to disable `document-start` and IMO it should be disabled by default

```
$ yamllint .travis.yml
.travis.yml
  1:1       warning  missing document start "---"  (document-start)
```

PR #965, opened 2016-10-31
YAMLLintBear: Disable `document-start` rule by default

rafidaslam:
Disable YAMLLint rule `document-start` by default because
many YAML files have no need for start and end markers, but
user still can enable it by using `document_start=true`
setting.

Add unit tests for YAMLLintBear.

Closes https://github.com/coala/coala-bears/issues/923

PR #1422, opened 2017-02-08
yamllint: Disable `document-start` by default

adrienverge:
Some projects use documents starts (`---`) in YAML documents (for example Ansible, OpenStack, yamllint), others don't. Since these markers are required when declaring multiple documents in a single .yaml file (see [the spec](http://yaml.org/spec/1.2/spec.html#id2800132)), it is a bad idea to forbid them.

This commit disables the `document-start` rule by default, instead of forcing / forbidding the use of these markers.

Closes: #1417
Fixes: #923 #965

### DIFF
--- a/bears/yaml/YAMLLintBear.py
+++ b/bears/yaml/YAMLLintBear.py
@@ -24,7 +24,7 @@ class YAMLLintBear:
 
     @staticmethod
     def generate_config(filename, file,
-                        document_start: bool=False):
+                        document_start: bool=None):
         """
         :param document_start:
             Use this rule to require or forbid the use of document start
@@ -33,13 +33,10 @@ class YAMLLintBear:
         yamllint_configs = {
             'extends': 'default',
             'rules': {
-                'document-start': {
-                    'present': False
-                 }
-            }
+                'document-start': 'disable' if document_start is None
+                                  else {'present': document_start},
+            },
         }
-        if document_start:
-            yamllint_configs['rules']['document-start']['present'] = True
 
         return yaml.dump(yamllint_configs)
 
--- a/tests/yaml/YAMLLintBearTest.py
+++ b/tests/yaml/YAMLLintBearTest.py
@@ -70,11 +70,16 @@ with prepare_file(config_file,
                                               'yamllint_config': conf_file})
 
 YAMLLintBear3Test = verify_local_bear(YAMLLintBear,
-                                      valid_files=(no_start_yaml_file,),
-                                      invalid_files=(with_start_yaml_file,))
+                                      valid_files=(no_start_yaml_file,
+                                                   with_start_yaml_file,),
+                                      invalid_files=())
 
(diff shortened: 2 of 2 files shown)


bears/yaml/YAMLLintBear.py
```diff
@@ -33,13 +33,12 @@ def generate_config(filename, file,
...
-                    'present': False
-                 }
+                'document-start': 'disable',
             }
         }
-        if document_start:
-            yamllint_configs['rules']['document-start']['present'] = True
+        if document_start is not None:
+            yamllint_configs['rules']['document-start'] = {
+                'present': document_start}
```
Makman2: For the sake of readability I think this is better:

```
    ...
    'rules': {}
    ...

# Don't forget to wrap the line to meet the 80 chars limit :D
yamllint_configs['rules']['document-start'] = 'disable' if document_start is None else {'present': document_start}
```

Instead of patching the dict depending on a value, we directly assign the correct value :)

Makman2:
Commit message:

When you want to make use of auto-close features of GitHub, don't place a colon:
```
Closes xxxx
instead of
Closes: xxxx
```
also the "Fixed" issues are already closed, and this commit does not really "fix" them. You could write `Related to xxxx` if you want to refer those :)

Makman2:
Also if you want to work on an issue @adrienverge, tell us so we can assign you and nobody's doing double-work :)
(or come by at gitter and assign yourself with `cobot assign <link-to-issue>` :D)

adrienverge:
> For the sake of readability I think this is better:

You're totally right, thanks! Updated.

> When you want to make use of auto-close features of GitHub, don't place a colon:

GitHub works fine with colons, but I removed them if you prefer this way.

> also the "Fixed" issues are already closed

I see a difference between "fixed" and "closed". My commit closes a discussion, but really fixes 1) the problem described in #923, 2) the other problem brought by commit in #965.

Makman2:
> GitHub works fine with colons, but I removed them if you prefer this way.

Oh really? Sorry then :3

> I see a difference between "fixed" and "closed". My commit closes a discussion, but really fixes 1) the problem described in #923, 2) the other problem brought by commit in #965.

I think this is thought 3 corners around. Though those issues have not really identified the overall problem (by making the force-markers setting optional), they addressed a more specific issue which was itself solved by those other commits.

Also we use "Fixes" specifically for fixing bugs (we actually track bug statistics via commits), and "Closes" for every other issue we want to close. Just you know :)

Don't want to be nitpicking, but for those "Fixes xxx" issues I would insist that you change them to "Related to xxx" ^^

adrienverge:
> Don't want to be nitpicking, but for those "Fixes xxx" issues I would insist that you change them to "Related to xxx" ^^

Alright, done.

Makman2:
ack 064abf4

Makman2:
@rultor merge

PR #1442, opened 2017-02-18
bears/general: Add LicenseCheckBear

yash-nisar:
Closes https://github.com/coala/coala-bears/issues/823

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/LicenseCheckBear.py
+++ b/bears/general/LicenseCheckBear.py
@@ -1,0 +1,40 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'.*: .*UNKNOWN$',
+        result_message='No license found.')
+class LicenseCheckBear:
+    """
+    Attempts to check the given file for a license, by searching the start
+    of the file for text belonging to various licenses.
+
+    For Ubuntu/Debian users, the ``licensecheck_lines`` option has to be used
+    in accordance with the ``licensecheck_tail`` option.
+    """
+    LANGUAGES = {'All'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='devscripts',
+                                            dnf='licensecheck')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'License'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=60,
+                         licensecheck_tail: int=5000):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed for
+            license information. Set to 0 to parse the whole file (and ignore
+            ``licensecheck_tail``).
+        :param licensecheck_tail:
+            Specify how many bytes to parse at end of file. Set to 0 to disable
+            parsing from end of file.
+        """
+        return ('--lines', str(licensecheck_lines), '--tail',
+                str(licensecheck_tail), filename)
--- a/tests/general/LicenseCheckBearTest.py
+++ b/tests/general/LicenseCheckBearTest.py
@@ -1,0 +1,67 @@
+import os
+from queue import Queue
+
+from bears.general.LicenseCheckBear import LicenseCheckBear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.results.Result import Result
+from coalib.settings.Section import Section
+
+
+def get_testfile_path(name):
+    return os.path.join(os.path.dirname(__file__),
+                        'licensecheck_test_files',
+                        name)
+
+
+def load_testfile(name):
+    return open(get_testfile_path(name)).readlines()
+
+
+@generate_skip_decorator(LicenseCheckBear)
+class LicenseCheckBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.uut = LicenseCheckBear(Section('name'), Queue())
+
+    def test_license(self):
+        file_contents = load_testfile('mit_license.py')
+        self.check_results(
+            self.uut,
+            file_contents,
+            [],
+            filename=get_testfile_path('mit_license.py'))
+
+    def test_license_without_copyright(self):
+        file_contents = load_testfile('apache_license_without_copyright.py')
+        self.check_results(
+            self.uut,
+            file_contents,
+            [],
+            filename=get_testfile_path('apache_license_without_copyright.py'),
+            settings={'licensecheck_lines': 70,
+                      'licensecheck_tail': 0})
+
+    def copyright_without_license(self):
+        file_contents = load_testfile('copyright_without_license.py')
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('LicenseCheckBear',
+                                'No license found.',
+                                file=get_testfile_path('copyright_without_'
+                                                       'license.py'))],
+            filename=get_testfile_path('copyright_without_license.py'),
+            settings={'licensecheck_lines': 0,
+                      'licensecheck_tail': 0})  # Parse entire file
+
+    def no_license(self):
+        file_contents = load_testfile('no_license.py')
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('LicenseCheckBear',
+                                'No license found.',
+                                file=get_testfile_path('no_license.py'))],
+            filename=get_testfile_path('no_license.py'),
+            settings={'licensecheck_lines': 10})
--- a/tests/general/licensecheck_test_files/mit_license.py
+++ b/tests/general/licensecheck_test_files/mit_license.py
@@ -1,0 +1,27 @@
+# Copyright (c) 2008 <name of author>
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+def add_two_numbers(a, b):
+    return a + b
+
+
+print(add_two_numbers(4, 5))
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -13,7 +13,7 @@ esac
 # apt-get commands
 export DEBIAN_FRONTEND=noninteractive
 
-deps="libclang1-3.4 indent mono-mcs chktex r-base julia golang-go luarocks verilator cppcheck flawfinder"
+deps="libclang1-3.4 indent mono-mcs chktex r-base julia golang-go luarocks verilator cppcheck flawfinder devscripts"
 deps_infer="m4 opam"
 
 case $CIRCLE_BUILD_IMAGE in
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ addons:
       - chktex
       - clang-3.4
       - cppcheck
+      - devscripts
       - flawfinder
       - gfortran
       - ghc
--- a/tests/general/licensecheck_test_files/apache_license_without_copyright.py
+++ b/tests/general/licensecheck_test_files/apache_license_without_copyright.py
@@ -1,0 +1,11 @@
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
--- a/tests/general/licensecheck_test_files/copyright_without_license.py
+++ b/tests/general/licensecheck_test_files/copyright_without_license.py
@@ -1,0 +1,1 @@
+# Copyright (c) 2008 <name of author>
--- a/tests/general/licensecheck_test_files/no_license.py
+++ b/tests/general/licensecheck_test_files/no_license.py
@@ -1,0 +1,2 @@
+# No license in this file
+print('Hello world')


yash-nisar:
@Makman2 Appveyor build failing. However the circle ci and travis builds have passed. I have added the dependencies in .travis.yml and deps.sh. What could be the reason ?

yash-nisar:
@Makman2 @adtac Please have a look at this when you are free. :)

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,49 @@
...
+            Specify that files/directories matching the regular expression
+            should be ignored when checking files.
+        :param licensecheck_check:
+            Specify a pattern indicating which files should be checked.
+        :param licensecheck_copyright:
+            Display the file's copyright
+        """
+        args = ()
+        if licensecheck_lines != 60:
+            args += ('--lines', str(licensecheck_lines))
```
adtac: Just give this directly, there's no need for a conditional statement.

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,49 @@
...
+    LANGUAGES = {'ALL'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='devscripts')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'License', 'Copyright Notice'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int = 60,
```
adtac: No spaces around the `=` (see also the other lines below)

tests/general/LicenseCheckBearTest.py
```diff
@@ -0,0 +1,78 @@
+from queue import Queue
+import os.path
```
adtac: We usually just do `import os` :)

Any particular reasons for `import os.path`?

tests/general/LicenseCheckBearTest.py
```diff
@@ -0,0 +1,78 @@
...
+@generate_skip_decorator(LicenseCheckBear)
+class LicenseCheckBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.uut = LicenseCheckBear(Section('name'), Queue())
+
+    def test_license(self):
+        file_contents = [load_testfile(
+            get_testfile_path('apache_license_2.py'))]
+        test_file_path = get_testfile_path('apache_license_2.py')
```
adtac: move this one line above and reuse `test_file_path`

tests/general/LicenseCheckBearTest.py
```diff
@@ -0,0 +1,78 @@
...
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('LicenseCheckBear',
+                                test_file_path + ': Apache (v2.0)',
```
adtac: Wait, this will return a result all the time? I thought this bear is something like "Make sure all files have this particular license; if not, return a result" and that it'll be quiet if a file is okay?

adtac: Or is this a helper bear for other bears?

tests/general/LicenseCheckBearTest.py
yash-nisar: Idk why I imported os.path :laughing:

tests/general/LicenseCheckBearTest.py
yash-nisar: Unlike other bears that check validity,
> This bear attempts to determine the license that applies to each file passed to
    it, by searching the start of the file for text belonging to various
    licenses.

It will return a result all the time to determine not just the license but also the type of license.

yash-nisar:
Done @adtac

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,48 @@
...
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.
+        :param licensecheck_ignore:
+            Specify that files/directories matching the regular expression
+            should be ignored when checking files.
+        :param licensecheck_check:
+            Specify a pattern indicating which files should be checked.
+        :param licensecheck_copyright:
+            Display the file's copyright
```
sils: I don't see what this is good for or even what it does

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,48 @@
...
+                         licensecheck_copyright: bool=False):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.
+        :param licensecheck_ignore:
+            Specify that files/directories matching the regular expression
+            should be ignored when checking files.
+        :param licensecheck_check:
+            Specify a pattern indicating which files should be checked.
```
sils: can you add examples? Solely from this description I'd not know how to use it

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,48 @@
...
+                         licensecheck_ignore: str='',
+                         licensecheck_check: str='',
+                         licensecheck_copyright: bool=False):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.
+        :param licensecheck_ignore:
+            Specify that files/directories matching the regular expression
+            should be ignored when checking files.
```
sils: this shouldn't be given, coala does ignoring and we shouldn't repeat the functionality as a setting

bears/general/LicenseCheckBear.py
yash-nisar: We can remove this parameter too IMO because the user can explicitly specify it either by the --files arg or in the .coafile.

bears/general/LicenseCheckBear.py
yash-nisar: We were actually planning to write a bear that would check for copyright notices in source files as mentioned [here](https://github.com/savoirfairelinux/flake8-copyright). However we have wrapped this functionality in a tool called as `licensecheck` which determines the license that applies to each file and checks for copyright notices too (displays it if present).

yash-nisar:
Updated @Makman2 @sils

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'(?P<message>\S+.*)')
+class LicenseCheckBear:
+    """
+    Attempts to determine the license that applies to each file passed to
+    it, by searching the start of the file for text belonging to various
+    licenses.
+    """
+    LANGUAGES = {'ALL'}
```
Makman2: just `All`, not `ALL` :)

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=60,
+                         licensecheck_copyright: bool=False):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.
+        :param licensecheck_copyright:
+            Display the file's copyright.
```
Makman2: Display?

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        args = ()
+        args += ('--lines', str(licensecheck_lines))
```
Makman2: initialize `args` directly with:
```
args = ('--lines', str(licensecheck_lines))
```

tests/general/LicenseCheckBearTest.py
Makman2: this is then definitely suited for a dependency bear imo, and we should add a bear that uses that information. Like a `LicenseConsistencyBear` or so^^

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+    LANGUAGES = {'ALL'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='devscripts')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'License', 'Copyright Notice'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=60,
```
satwikkansal: Just curious, why the default value for `licensecheck_lines` is 60? :sweat_smile:

tests/general/LicenseCheckBearTest.py
adtac: I'd never have a bear that always returns results. This should definitely be a dependency bear. And in that case, you should use `HiddenResult`s instead of `Result`s.

yash-nisar: In the case of a native bear, I can silence the results in the run method but how do I silence the results given by the `licensecheck` tool in this case ?

bears/general/LicenseCheckBear.py
yash-nisar: Maybe because most of the licenses fit in that threshold but that was the default length in the wrapped tool too.

bears/general/LicenseCheckBear.py
yash-nisar: Didn't get you ?

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'(?P<message>\S+.*)')
+class LicenseCheckBear:
+    """
+    Attempts to determine the license that applies to each file passed to
```
aptrishu: I thought our objective for this kind-of bear was to determine whether a file contains a licence or not.
So, use HiddenResults maybe so other bears in future can use it.
Also can we have it do something like it checks a directory for a licence file ? or other bear which depends on it would do that..

bears/general/LicenseCheckBear.py
Makman2: Where does it display?

tests/general/LicenseCheckBearTest.py
Makman2: `@linter` currently doesn't support yielding `HiddenResult`s currently, you maybe want to implement support for it^^

bears/general/LicenseCheckBear.py
yash-nisar: On the console. Just like https://github.com/savoirfairelinux/flake8-copyright

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'(?P<message>\S+.*)')
+class LicenseCheckBear:
```
Makman2: I would call the bear `LicenseDetectionBear`, as you detect every used license, you don't perform checks on it :+1:

yash-nisar: Completely agree with it.

tests/general/LicenseCheckBearTest.py
yash-nisar: So, till the time yielding HiddenResult is implemented for linter bears, what do we do with this ?

meetmangukiya:
Change the commit message to incorporate new name

tests/general/LicenseDetectionBearTest.py
```diff
@@ -0,0 +1,66 @@
...
+        self.uut = LicenseDetectionBear(Section('name'), Queue())
+
+    def test_license(self):
+        test_file_path = get_testfile_path('apache_license_2.py')
+        file_contents = [load_testfile(test_file_path)]
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('LicenseDetectionBear',
+                                test_file_path + ': Apache (v2.0)',
```
jayvdb: This is not helpful.  coala is a checker, not a printer.  The user should be able to add  `license=Apache (v2.0)` to `.coafile` and the bear will verify these files use that copyright.

tests/general/LicenseDetectionBearTest.py
```diff
@@ -0,0 +1,66 @@
...
+        test_file_path = get_testfile_path('mit_license.py')
+        file_contents = [load_testfile(test_file_path)]
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('LicenseDetectionBear',
+                                test_file_path + ': MIT/X11 (BSD like)',
+                                file=test_file_path),
+             Result.from_values('LicenseDetectionBear',
+                                '[Copyright: 2008 <name of author>]',
```
jayvdb: again, not helpful.  The user should be able to specify a regex that the copyright must match.

bears/general/LicenseDetectionBear.py
```diff
@@ -0,0 +1,37 @@
...
+    LANGUAGES = {'All'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='devscripts')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'License', 'Copyright Notice'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         lines: int=60,
```
jayvdb: this setting `lines` is very ambiguous.
I guess that the variable requires that the license appears in the top x (60) lines.
If so, describe that as a setting.

yash-nisar:
I wanted to discuss about this before proceeding, @Makman2 @jayvdb you guys around ?

tests/general/LicenseDetectionBearTest.py
yash-nisar: IIUC, there is not way this can be achieved since there is no option to specify a regex that the copyright can match unlike https://pypi.python.org/pypi/flake8-copyright.

yash-nisar: Perhaps we can implement another CopyrightBear ?

tests/general/LicenseDetectionBearTest.py
yash-nisar: We will have to provide a number of options to the user(total available licenses that our bear can detect) and the user may not have that knowledge. I suggest that our bear should check whether the given file has a license or not. It will only complain if the file doesn't have a license. Something similar to: 
```
        stdout, stderr = output
        result = filename + ': *No copyright* UNKNOWN'
        if result.strip() == stdout.strip():
            yield from self.process_output_regex(stdout, filename, file,
                                                 self._output_regex,
                                                 result_message='No license'
                                                                ' found.')
```
where we process that output and check if the file has a license or not.
Here is the output : 
```
s.py
|    | [NORMAL] LicenseDetectionBear:
|    | No license found.
|    | *0: Do nothing
|    |  1: Open file(s)
|    |  2: Add ignore comment
|    | Enter number (Ctrl-D to exit): 0
```
We can thus rename it to LicenseCheckBear since it serves the purpose. :smile:

yash-nisar:
Travis builds failing. Getting some errors related to CheckStyleBear ?! I'm unable to diagnose the problem.

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,45 @@
...
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=60):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.
+        """
+
+        args = ('--lines', str(licensecheck_lines))
+        return args + (filename,)
```
Makman2: immediately return the complete tuple, no need to waste performance for tuple concatenation:
```
return '--lines', str(licensecheck_lines), filename
```

tests/general/LicenseCheckBearTest.py
```diff
@@ -0,0 +1,48 @@
...
+
+@generate_skip_decorator(LicenseCheckBear)
+class LicenseCheckBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.uut = LicenseCheckBear(Section('name'), Queue())
+
+    def test_license(self):
+        test_file_path = get_testfile_path('mit_license.py')
+        file_contents = [load_testfile(test_file_path)]
```
Makman2: `load_testfile` does call `get_testfile_path`, so I think you don't need
```
test_file_path = get_testfile_path('mit_license.py')
```

Makman2: Also `check_results` expects a list of lines, not a list of file-strings with the full content :) I think you should improve your `load_testfile` with

```
return file.splitlines(True)
```
Then you also don't need the square brackets here.

yash-nisar:
Doneeeeee @Makman2 , please have a look when you are free. :smile:

tests/general/LicenseCheckBearTest.py
yash-nisar: Done. :+1:

yash-nisar: Done :+1:

bears/general/LicenseCheckBear.py
yash-nisar: Roger that. :smile:

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,44 @@
...
+                         licensecheck_lines: int=60):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.
+        """
+
+        return '--lines', str(licensecheck_lines), filename
+
+    def process_output(self, output, filename, file):
```
Makman2: you still override `process_output`, don't you? :D

I think this is possible instead:
```python
@linter(
    executable='licensecheck',
    use_stdout=True,
    # use_stderr=True,  <--- you don't need stderr btw !!!
    output_format='regex',
    output_regex=r'.*: \*No copyright\* UNKNOWN',
    result_message='No license found.' # <-- I think 'No license found' sounds somehow better than 'License not found', can't fully explain you why, so it's up to you +1
```
then you don't need `_output_regex` being defined manually in the class + you don't have to override `process_output` :+1:

Makman2: (And also no `re` import :D)

yash-nisar: - [ ] Added `@linter(executable='licensecheck',
        use_stdout=True,
        output_format='regex',
        output_regex=r'.*: \*No copyright\* UNKNOWN',
        result_message='No license found.')`
- [ ]  Removed the unused re import
- [ ]  Tests passing on my local machine
- [ ]  We finally have a LicenseCheckBear thats checks whether the source file has a license or not ;)

yash-nisar: Over to you @Makman2

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        use_stdout=True,
```
Makman2: redundant, `True` by default :)

yash-nisar: Roger that. :smile:

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
...
+    Attempts to check the given file for a license, by searching the start
+    of the file for text belonging to various licenses.
+    """
+    LANGUAGES = {'All'}
+    REQUIREMENTS = {DistributionRequirement(apt_get='devscripts',
+                                            dnf='licensecheck')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'License'}
```
Makman2: what about "Copyright Note" you had before?

yash-nisar: Since we are not checking the copyright thing, I removed it. Do I add it ?

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=None,
+                         licensecheck_tail: int=None):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.(Default: 60)
```
Makman2: Why not providing the default inside the params?
```
licensecheck_lines: int=60
```

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
...
+        :param licensecheck_tail:
+            Specify how many bytes to parse at end of file.(Default: 5000)
```
Makman2: same here

bears/general/LicenseCheckBear.py
yash-nisar: I had a talk with @jayvdb yesterday regarding it. The executable is producing different results for different distros. See https://gitter.im/coala/coala-bears?at=58b9abd521d548df2c87e284

bears/general/LicenseCheckBear.py
Makman2: okay, just wanted to know :+1:

Makman2: (so don't add it)

bears/general/LicenseCheckBear.py
Makman2: Shouldn't we enforce a standard then?

Makman2: (and if the default does change within various distros, you shouldn't document it^^ Or better that it's distro-dependent)

yash-nisar: The default is same, the behaviour is different. The `--lines` wasn't working for me without `--tail` (I use Lubuntu 16.04 LTS) whereas it worked for jayvdb (Fedora) without the `--tail` argument. So we decided to let the users figure out what to use. @Makman2 Is that okay ?

Makman2: then just provide the defaults and always use `--lines` and `--tail`, then it should work always, shouldn't it?

yash-nisar: `--lines` was working for me when I set `--tail` to 0. It will show unexpected behaviour when I set it to defaults. :worried: I think we should let the users figure out this. What say @Makman2 ?

Makman2: That's kind of bad. The whole purpose of coala is to be a unified interface, I mean this breaks it somehow. If you really have no immediate idea on how to solve that, I'm still in for merging this, but file an issue for that (after merge).
\+
Remove those default mentions, if the user needs to figure that out and it's anyway platform dependent, our values are not correct.

yash-nisar: We shouldn't merge anything until it is perfect. I made changes according to what @jayvdb quoted here https://gitter.im/coala/coala-bears?at=58b9abd521d548df2c87e284, but I want to know your opinion so I can make changes that are best suited for this bear @Makman2 :D

yash-nisar: A workaround can be if I directly install the .deb package by fetching the .deb file via wget. This will ensure consistency and since it is an upgraded version, we can expect that strange things wont happen. :P

yash-nisar: Another workaround that I think will help is :  
* If the user mentions `--lines 30`, we can set `--tail` to 0 to make sure that the file is parsed only for the first 30 lines.
* If the user mentions `--tail 200`, we can set `--lines` to 0 to make sure that the file is parsed only for the last 200 bytes of the file.
* If he mentions both, the results will be due to parsing from either sides, which we get in normal circumstances. 
Maybe that is what is happening in the upgraded version.
* So, if I specify `licensecheck --lines 20 xyz.py`, I might expect an error if the license is not defined within the first 20 lines but the default for `--tail` is 5000, due to which the file will also be parsed from the bottom. We can achieve that by setting the other option to 0. Maybe this is what they mean when they say `(and ignore --tail)` [here](paste.ubuntu.com/24102936/). Also, note jayvd's output [here](https://gitter.im/coala/coala-bears?at=58b99a7bf1a33b627576cabf).

yash-nisar: So, I think it basically boils down to what you said, ie. provide the defaults. The user will have to supply an Arg as 0 if he wants to parse the file from the top or bottom. @Makman2 Do you agree ?

Makman2: > We shouldn't merge anything until it is perfect.

I'm also for this attitude, but sometimes you can drive this too much ;) Better get something reasonable than nothing at all. Our PRs do already require a lot of time because of our high standard.

not everyone has support for .deb, I use fedora, so default is .rpm

Maybe you can assert the program version and change behavior depending on that? `check_prerequisites` could do the trick^^

Makman2: > So, I think it basically boils down to what you said, ie. provide the defaults. The user will have to supply an Arg as 0 if he wants to parse the file from the top or bottom. @Makman2 Do you agree ?

Don't know, loosing track a bit. Like said I would merge either behavior that is reasonable and let's improve later^^

yash-nisar: As you said, I shall provide the defaults,
> then just provide the defaults and always use --lines and --tail, then it should work always, shouldn't it?

@Makman2

yash-nisar: Use cases where things are clear:
```
 ~  licensecheck -l 1 --tail 0 mit_license.py 
mit_license.py: UNKNOWN
(Avoiding parsing from the bottom of the file)
 ~  licensecheck -l 0 --tail 0 mit_license.py 
mit_license.py: MIT/X11 (BSD like)
(-l set to 0 to parse the entire file)
```
@Makman2 Solved !!!!

tests/general/LicenseCheckBearTest.py
```diff
@@ -0,0 +1,45 @@
...
+        self.uut = LicenseCheckBear(Section('name'), Queue())
+
+    def test_license(self):
+        file_contents = load_testfile('mit_license.py')
+        self.check_results(
+            self.uut,
+            file_contents,
+            [],
+            filename=get_testfile_path('mit_license.py'),
+            settings={'licensecheck_lines': 'int = 70',
```
jayvdb: Do the tests fail on Ubuntu/Debian if you dont provide `licensecheck_lines` and/or `licensecheck_tail`?

I can report that these two tests pass on Fedora. \o/

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+    CAN_DETECT = {'License'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=60,
+                         licensecheck_tail: int=5000):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information.(Default: 60). Set to 0 to parse the whole
```
jayvdb: you dont need to describe defaults here, as these defaults are the same as the setting defaults in the actual arguments above, so they will automatically be documented.  c.f. https://github.com/coala/bear-docs/blob/master/docs/LineLengthBear.rst#settings

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+            file (and ignore --tail).
```
jayvdb: `(and ignore --tail)` here isnt contextually accurate, as `--tail` should probably be `licensecheck_tail` and 'ignore' isnt a good description.

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'.*: .*UNKNOWN',
+        result_message='No license found.')
+class LicenseCheckBear:
+    """
+    Attempts to check the given file for a license, by searching the start
+    of the file for text belonging to various licenses.
```
jayvdb: The problems you encountered with Ubuntu/Debian should be described here.

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,37 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'.*: .*UNKNOWN',
```
jayvdb: Can you be a bit more precise than `.*` -- it is probably going to be OK, but better to be safer to avoid strange breakages later on.

tests/general/LicenseCheckBearTest.py
yash-nisar: Nope, they pass.

yash-nisar:
Surprisingly, everything is green again. Over to you, @Makman2 @jayvdb.

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,39 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'(?P<filename>.*): (?P<copyright>.*)UNKNOWN',
```
Makman2: unknown capturing groups, please remove them as coala will otherwise raise a warning on execution^^

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,39 @@
...
+        result_message='No license found.')
+class LicenseCheckBear:
+    """
+    Attempts to check the given file for a license, by searching the start
+    of the file for text belonging to various licenses. For Ubuntu/Debian
```
Makman2: Would layout the docs like this:
```
Attempts to check the given file for a license, by searching the start
of the file for text belonging to various licenses.

For Ubuntu/Debian ...
```
As the Ubuntu/Debian part is detailed :) Rule of thumb: The first line/sentence is the short description which shall describe the bear as short as possible, everything following an empty line is detailed description :)

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,39 @@
...
+    users, the licensecheck_lines option has to be used in accordance with the
+    licensecheck_tail option.
```
Makman2: Enclose `licensecheck_lines` and `licensecheck_tail` in double backticks :)

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         licensecheck_lines: int=60,
+                         licensecheck_tail: int=5000):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information. Set to 0 to parse the whole
+            file (and ignore licensecheck_tail).
```
Makman2: Enclose `licensecheck_tail` in double backticks

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,39 @@
...
+        :param licensecheck_tail:
+            Specify how many bytes to parse at end of file. Set to 0 to
+            avoid parsing from end of file.
```
Makman2: Would rephrase to
```
Set to 0 to disable parsing from end of file.
```

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        return '--lines', str(licensecheck_lines), '--tail', \
+               str(licensecheck_tail), filename
```
Makman2: use braces instead of `\`:
```
return ('--lines', ..., ..., 
      'folllowing-args', ...)
```

tests/general/licensecheck_test_files/mit_license.py
```diff
@@ -0,0 +1,20 @@
...
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
Makman2: what about adding some code to these files?

Makman2: (to some of them)

yash-nisar:
Done @Makman2 :)

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
...
+        output_format='regex',
+        output_regex=r'.*: .*UNKNOWN',
+        result_message='No license found.')
+class LicenseCheckBear:
+    """
+    Attempts to check the given file for a license, by searching the start
+    of the file for text belonging to various licenses.
+
+    For Ubuntu/Debian users, the `licensecheck_lines` option has to be used in
+    accordance with the `licensecheck_tail` option.
```
Makman2: we use double backticks, and I'm not sure if RST handles single ones properly. Could you change that everywhere? :)

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
...
+                         licensecheck_lines: int=60,
+                         licensecheck_tail: int=5000):
+        """
+        :param licensecheck_lines:
+            Specify how many lines of the file header should be parsed
+            for license information. Set to 0 to parse the whole
+            file (and ignore `licensecheck_tail`).
+        :param licensecheck_tail:
+            Specify how many bytes to parse at end of file. Set to 0 to disable
+             parsing from end of file.
```
Makman2: indentation off

bears/general/LicenseCheckBear.py
yash-nisar: I'm sorry. :(

bears/general/LicenseCheckBear.py
yash-nisar: How did I forget that. :confused:

tests/general/LicenseCheckBearTest.py
```diff
@@ -0,0 +1,67 @@
...
+    def no_license(self):
+        file_contents = load_testfile('no_license.py')
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('LicenseCheckBear',
+                                'No license found.',
+                                file=get_testfile_path('no_license.py'))],
+            filename=get_testfile_path('no_license.py'),
+            settings={'licensecheck_lines': 'int = 10'})
```
Makman2: `int = 10` ?

Makman2: I think you mean just `'10'`

yash-nisar: will check the first 10 lines and parse 5000 bytes from the bottom of the file. :)

Makman2: but the `int` is invalid, isn't it? it's parsed as a string then^^

yash-nisar: Yes @Makman2 , you are absolutely correct, actually I had copied that statement from the docs and added quotes because of some error in haste. Have filed an issue for the same. :D

bears/general/LicenseCheckBear.py
yash-nisar: Fixed. :)

bears/general/LicenseCheckBear.py
yash-nisar: Fixed. :)

tests/general/LicenseCheckBearTest.py
yash-nisar: :)

yash-nisar: Fixed*

tests/general/licensecheck_test_files/mit_license.py
yash-nisar: Added some code too and update the regex. :D

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'(^(.*/)*)([^/]*): .*UNKNOWN$',
```
Makman2: for paths I would use `.*`, as this may not work on windows ;)

Makman2: (we maybe need some regex composition library, where you could do something like `regexes.path + r': .*UNKNOWN$'`

Makman2: (just an idea, don't seach for one ;D)

yash-nisar: I had the same thought initially but we don't have support for windows so..

yash-nisar: So now our bear is future proof. :D

yash-nisar:
Done @Makman2 :D

Makman2:
ack 01d4f24

Makman2:
@rultor merge

bears/general/LicenseCheckBear.py
```diff
@@ -0,0 +1,40 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.DistributionRequirement import (
+    DistributionRequirement)
+
+
+@linter(executable='licensecheck',
+        output_format='regex',
+        output_regex=r'.*: .*UNKNOWN$',
```
jayvdb: sigh.
The `.*` should have been something like `[^:]+`.

PR #1462, opened 2017-02-27
CheckstyleBear: Use checkstyle 6.19

jayvdb:


jayvdb:
Prefer https://github.com/coala/coala-bears/pull/1466

PR #1480, opened 2017-03-04
bears/elm: Add ElmLintBear

yash-nisar:
Closes https://github.com/coala/coala-bears/issues/1073

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -137,6 +137,13 @@ if [ ! -e ~/bakalint-0.4.0/bakalint.pl ]; then
   tar xf ~/bl.tar.gz -C ~/
 fi
 
+# elm-format Installation
+if [ ! -e ~/elm-format-0.18/elm-format ]; then
+  mkdir -p ~/elm-format-0.18
+  curl -fsSL -o elm-format.tgz https://github.com/avh4/elm-format/releases/download/0.5.2-alpha/elm-format-0.17-0.5.2-alpha-linux-x64.tgz
+  tar -xvzf elm-format.tgz -C ~/elm-format-0.18
+fi
+
 # Julia commands
 julia -e "Pkg.add(\"Lint\")"
 
--- a/bears/elm/ElmLintBear.py
+++ b/bears/elm/ElmLintBear.py
@@ -1,0 +1,37 @@
+from shutil import which
+
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.NpmRequirement import NpmRequirement
+
+
+@linter(executable='elm-format',
+        use_stderr=True,
+        use_stdout=False,
+        output_format='regex',
+        output_regex='(?P<line>\d+)│(?P<message>[\s\S]*)')
+class ElmLintBear:
+    """
+    This bear formats the Elm source code according to a standard set of rules.
+
+    See <https://github.com/avh4/elm-format> for more information.
+    """
+
+    LANGUAGES = {'Elm'}
+    REQUIREMENTS = {NpmRequirement('elm', '0.18')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Syntax', 'Formatting'}
+
+    @classmethod
+    def check_prerequisites(cls):  # pragma: no cover
+        if which('elm-format') is None:
+            return ('elm-format is missing. Download it from '
+                    'https://github.com/avh4/elm-format/blob/master/README.md'
+                    '#for-elm-018 and put it into your PATH.')
+        else:
+            return True
+
+    @staticmethod
+    def create_arguments(filename, file, config_file):
+        return filename, '--yes'
--- a/tests/elm/ElmLintBearTest.py
+++ b/tests/elm/ElmLintBearTest.py
@@ -1,0 +1,145 @@
+from queue import Queue
+
+from bears.elm.ElmLintBear import ElmLintBear
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.Section import Section
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coala_utils.ContextManagers import prepare_file
+
+good_file_no_unexpected_comma = """
+ab =
+  "Hello World!"
+"""
+
+bad_file_unexpected_comma = """
+a, b =
+  "Hello World!"
+"""
+
+good_file_no_unexpected_end_of_input = """
+foo =
+  100
+"""
+
+bad_file_unexpected_end_of_input = """
+foo =
+"""
+
+good_file_not_empty = """
+add x y = x + y
+"""
+
+bad_file_empty = """
+"""
+
+good_file_comment = """
+foo =
+  5 * 3 - 8 / 1
+-- A single line comment
+"""
+
+bad_file_comment = """
+-- A single line comment
+"""
+
+good_file_list = """
+l =
+  ["one", "two", "three"]
+"""
+
+bad_file_list = """
+l =
+  ["one", "two", "three"
+"""
+
+good_file_function = """
+import Html exposing (text)
+four =
+  sqrt 16
+main =
+  text (toString [four])
+"""
+
+bad_file_function = """
+import Html exposing (text)
+four =
+  sqr 16
+main =
+  text (toString [four)
+"""
+
+result_message_bad_function = """\
+text (toString [four)\x1b[31m
+                         ^\x1b[0m
+I am looking for one of the following things:
+
+    ","
+    "]"
+    a field access like .name
+    an expression
+    an infix operator like +
+    an infix operator like `andThen`
+    more letters in this name
+    whitespace"""
+
+UnexpectedCommaTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_no_unexpected_comma,),
+    invalid_files=(bad_file_unexpected_comma,))
+
+UnexpectedEndOfInputTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_no_unexpected_end_of_input,),
+    invalid_files=(bad_file_unexpected_end_of_input,))
+
+EmptyFileTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_not_empty,),
+    invalid_files=(bad_file_empty,))
+
+CommentTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_comment,),
+    invalid_files=(bad_file_comment,))
+
+ListTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_list,),
+    invalid_files=(bad_file_list,))
+
+
+@generate_skip_decorator(ElmLintBear)
+class ElmLintBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.uut = ElmLintBear(Section('name'), Queue())
+
+    def test_bad_function(self):
+        prepared = prepare_file(bad_file_function.splitlines(),
+                                filename=None, create_tempfile=True)
+
+        with prepared as (file, fname):
+            self.check_results(
+                self.uut,
+                bad_file_function.splitlines(),
+                [Result.from_values('ElmLintBear',
+                                    message=result_message_bad_function,
+                                    file=fname,
+                                    line=6,
+                                    severity=RESULT_SEVERITY.NORMAL)],
+                filename=fname,
+                create_tempfile=False)
+
+    def test_good_function(self):
+        prepared = prepare_file(good_file_function.splitlines(),
+                                filename=None, create_tempfile=True)
+        with prepared as (file, fname):
+            self.check_results(
+                self.uut,
+                good_file_function.splitlines(),
+                [],
+                filename=fname,
+                create_tempfile=False)
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,7 @@ cache:
     - ~/infer-linux64-v0.7.0
     - ~/pmd-bin-5.4.1
     - ~/bakalint-0.4.0
+    - ~/elm-format-0.18
     - ~/dart-sdk/bin
     - ~/.local/tailor/
 
@@ -67,7 +68,7 @@ env:
     - USE_PPAS="marutter/rdev staticfloat/juliareleases ondrej/golang"
     - R_LIB_USER=~/R/Library
     - LINTR_COMMENT_BOT=false
-    - PATH="/opt/cabal/1.24/bin:$PATH:$TRAVIS_BUILD_DIR/node_modules/.bin:$TRAVIS_BUILD_DIR/vendor/bin:$HOME/dart-sdk/bin:$HOME/.cabal/bin:$HOME/infer-linux64-v0.7.0/infer/bin:$HOME/pmd-bin-5.4.1/bin:$HOME/bakalint-0.4.0:$HOME/.local/tailor/tailor-latest/bin"
+    - PATH="/opt/cabal/1.24/bin:$PATH:$TRAVIS_BUILD_DIR/node_modules/.bin:$TRAVIS_BUILD_DIR/vendor/bin:$HOME/dart-sdk/bin:$HOME/.cabal/bin:$HOME/infer-linux64-v0.7.0/infer/bin:$HOME/pmd-bin-5.4.1/bin:$HOME/bakalint-0.4.0:$HOME/elm-format-0.18:$HOME/.local/tailor/tailor-latest/bin"
 
 before_install:
   - export TRAVIS_PYTHON_VERSION_MAJOR=${TRAVIS_PYTHON_VERSION%.[0-9]}
--- a/circle.yml
+++ b/circle.yml
@@ -28,6 +28,7 @@ dependencies:
     - echo 'export PATH=$PATH:~/pmd-bin-5.4.1/bin' >> ~/.circlerc
     - echo 'export PATH=$PATH:~/bakalint-0.4.0' >> ~/.circlerc
     - echo 'export PATH=$PATH:~/.local/tailor/tailor-latest/bin' >> ~/.circlerc
+    - echo 'export PATH=$PATH:~/elm-format-0.18' >> ~/.circlerc
     - echo 'export R_LIB_USER=~/.RLibrary' >> ~/.circlerc
     - sed -i '/source \/home\/ubuntu\/virtualenvs\//d' ~/.circlerc
     - mkdir -p ~/.RLibrary
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "complexity-report": "~2.0.0-alpha",
     "csslint": "~1",
     "dockerfile_lint": "~0",
+    "elm": "~0.18",
     "eslint": "~2",
     "eslint-plugin-import": "~1",
     "happiness": "~7.1.2",


yash-nisar:
Why wip ? https://github.com/coala/coala-bears/issues/1472 is due to which the tests are failing. :) The tests passed locally though. :D

aptrishu:
That's fine, but we can not review until the tests pass. They might fail even after fixing that issue. So, It'd be a waste of time for others to review it until then. It's completely fine, you can label them `pending review` once the tests pass. :)

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,63 @@
...
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+good_file_no_unexpected_comma = """
+module UnexpectedComma (..) where
+
+
+var =
+"Hello World!"
+"""
+
```
jayvdb: only one space between variables

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,63 @@
+from bears.elm.ElmLintBear import ElmLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+good_file_no_unexpected_comma = """
+module UnexpectedComma (..) where
+
```
jayvdb: can you avoid (all|excessive|consecutive) blank lines in these sample files; it makes it hard to read, *especially* if syntax highlighting isnt present.

bears/elm/ElmLintBear.py
```diff
@@ -0,0 +1,37 @@
+from shutil import which
+
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.NpmRequirement import NpmRequirement
+
+
+@linter(executable='elm-format',
+        use_stdin=True,
```
jayvdb: stdin should be avoided if possible.  The filename cant be specified on the command line?

bears/elm/ElmLintBear.py
```diff
@@ -0,0 +1,37 @@
...
+        if which('elm-format') is None:
+            return ('elm-format is missing. Download it from '
+                    '<https://github.com/avh4/elm-format/blob/master/README.md'
+                    '#for-elm-018> and put it in your PATH.')
+        else:
+            return True
+
+    @staticmethod
+    def create_arguments(filename, file, config_file):
+        return '--stdin',
```
jayvdb: wtf!?  the coverage results say this line isnt being covered, ... ??

yash-nisar: Exactly. I'm trying to figure out why !

bears/elm/ElmLintBear.py
yash-nisar: The filename can be specified but it doesn't yield results on the console. Instead, it prompts to overwrite the existing file with the changes whereas `--stdin` writes output to `--stdout`.
Quoting the usage:
> elm-format --stdin  # Format input from stdin and write to stdout
Also see https://github.com/avh4/elm-format#usage for more details.

jayvdb: See other options for `output_format`

tests/elm/ElmLintBearTest.py
yash-nisar: Sure. :)

tests/elm/ElmLintBearTest.py
yash-nisar: Didn't get you ?

bears/elm/ElmLintBear.py
yash-nisar: @Makman2 @jayvdb , any idea why isn't this line being covered ?

jayvdb: Optimisation?
Bug in coverage? (same version local vs CI?)

tests/elm/ElmLintBearTest.py
jayvdb: Pep8. Only one blank line is desirable between variables.

bears/elm/ElmLintBear.py
yash-nisar: Haven't optimised. It's the same bug. @jayvdb

Makman2:
Tests are skipped, seems like the executable can't be found^^

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,56 @@
+from bears.elm.ElmLintBear import ElmLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+good_file_no_unexpected_comma = """
+module UnexpectedComma (..) where
+var =
+"Hello World!"
+"""
+
+
```
jayvdb: Can you remove this blank line, and the ones on lines 17, 24, 30, 36, etc

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,56 @@
...
+good_file_not_empty = """
+var =
+100
+"""
+
+
+bad_file_empty = """
```
jayvdb: This makes me happy.
https://github.com/coala/coala-bears/issues/1495

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+
+bad_file_unexpected_end_of_input = """
+module UnexpectedEndOfInput (..) where
+var =
+"""
+
+
+good_file_not_empty = """
```
jayvdb: This isnt a particularly good example of a non-empty file, as it has a leading blank line, and a line feed in the middle of an assignment (implicit continuation).  Simplify it, so that it only tests one thing.  Ideally also add a few from https://github.com/coala/coala-bears/issues/1495

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,56 @@
+from bears.elm.ElmLintBear import ElmLintBear
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+
+good_file_no_unexpected_comma = """
+module UnexpectedComma (..) where
+var =
```
jayvdb: using `var` as a variable name is unnecessary.  If the `=` doesnt explain that the lhs is a variable name, the reader should give up programming.

use either foo / bar / baz , or a / b / c, or some other similar pattern.

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,56 @@
...
+"Hello World!"
+"""
+
+
+bad_file_unexpected_comma = """
+module UnexpectedComma (..) where
+va,r =
```
jayvdb: part of the reason to avoid `var` above is that here you seem to be implying that `var` is a thing.

Some languages (notably Python) multiple variable (tuple) assignments are ok, but Elm isnt one of them.
There are much simpler 'unexpected comma' sample files.

But if you want to keep the structure of this sample file, you should use `a, b = \n...`

tests/elm/ElmLintBearTest.py
yash-nisar: Sure. :)

tests/elm/ElmLintBearTest.py
yash-nisar: I picked them up from https://github.com/avh4/elm-format/tree/master/tests/test-files/bad repository.

tests/elm/ElmLintBearTest.py
yash-nisar: IIUC, we are testing a non-empty file with some valid content. The line feed in the middle of an assignment is a part of the syntax of the elm language.

yash-nisar: However, I have changed it to add a comment to satisfy https://github.com/coala/coala-bears/issues/1495

jayvdb: According to http://elm-lang.org/try , this doesnt compile.
It would need to be
```
foo =
     100
```
But that is testing too many things.  Need minimal tests; the smallest 'not_empty' using as few language features as possible.

Then add more good sample files for more elaborate language features.

yash-nisar: Roger that. Also, adding indentation and more tests. :)

.ci/deps.sh
```diff
@@ -175,3 +175,12 @@ fi
 echo '' > dummy
 coala-ci --bears CheckstyleBear --files dummy --no-config --bear-dirs bears || true
 coala-ci --bears ScalaLintBear --files dummy --no-config --bear-dirs bears || true
+
+if [ ! -e ~/elm_format/elm-format ]; then
+  mkdir -p ~/elm_format
+  export PATH=$PATH:~/elm_format
```
jayvdb: Is this really needed? circle and travis yml each add it to the path.

.ci/deps.sh
```diff
@@ -175,3 +175,12 @@ fi
...
+  wget -nc -O elm-format.tgz https://github.com/avh4/elm-format/releases/download/0.5.2-alpha/elm-format-0.17-0.5.2-alpha-linux-x64.tgz
+  tar -xvzf elm-format.tgz
+  mv elm-format ~/elm_format
```
jayvdb: this can and should be easily avoided.

.ci/deps.sh
```diff
@@ -175,3 +175,12 @@ fi
 echo '' > dummy
 coala-ci --bears CheckstyleBear --files dummy --no-config --bear-dirs bears || true
 coala-ci --bears ScalaLintBear --files dummy --no-config --bear-dirs bears || true
+
+if [ ! -e ~/elm_format/elm-format ]; then
+  mkdir -p ~/elm_format
+  export PATH=$PATH:~/elm_format
+  wget -nc -O elm-format.tgz https://github.com/avh4/elm-format/releases/download/0.5.2-alpha/elm-format-0.17-0.5.2-alpha-linux-x64.tgz
```
jayvdb: I'd prefer we start using curl only, as `wget` isnt installed on the docker (whereas curl is mandatory in the docker), and it would be good for the docker to be directly using the coala-bears install scripts, rather than having a slightly different set of commands.  see https://github.com/coala/docker-coala-base/blob/master/Dockerfile

.ci/deps.sh
yash-nisar: I have to work on them, main focus was to get all green ticks. Marking WIP. :D Thanks a lot for your meaningful reviews.

yash-nisar:
@Makman2 @jayvdb Are the changes good ?

bears/elm/ElmLintBear.py
```diff
@@ -0,0 +1,37 @@
+from shutil import which
+
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.NpmRequirement import NpmRequirement
+
+
+@linter(executable='elm-format',
+        use_stdin=True,
+        use_stderr=True,
```
Makman2: why do you enable `use_stderr`, I thought results are put into stdout?

bears/elm/ElmLintBear.py
Makman2: right `regex` doesn't look right then

bears/elm/ElmLintBear.py
```diff
@@ -0,0 +1,37 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Syntax', 'Formatting'}
+
+    @classmethod
+    def check_prerequisites(cls):  # pragma: no cover
+        if which('elm-format') is None:
+            return ('elm-format is missing. Download it from '
+                    '<https://github.com/avh4/elm-format/blob/master/README.md'
+                    '#for-elm-018> and put it in your PATH.')
```
Makman2: as this is displayed on console, I would remove `<` and `>` from the link
\+
put it in -> put it into :)

bears/elm/ElmLintBear.py
yash-nisar: Nope, since there are errors, they are put into stderr.

yash-nisar:
Use cases : 
1. Unexpected comma : 
```
s.elm
|   1| a,·b·=
|    | [NORMAL] ElmLintBear:
|    | a, b =
|    |     ^
|    | I am looking for one of the following things:
|    | 
|    |     =
|    |     a pattern
|    |     more letters in this name
|    |     whitespace
|    | *0: Do nothing
|    |  1: Open file(s)
|    |  2: Add ignore comment
|    | Enter number (Ctrl-D to exit): 0
```
2. Empty file :
```
s.elm
|   1| The line belonging to the following result cannot be printed because it refers to a line that doesn't seem to exist in the given file.
|    | [NORMAL] ElmLintBear:
|    | 
|    |    ^
|    | I am looking for one of the following things:
|    | 
|    |     "{-|"
|    |     a module declaration
|    |     a port declaration
|    |     a type declaration
|    |     a value definition
|    |     an import
|    |     an infix declaration
|    |     whitespace
|    | *0: Do nothing
|    |  1: Open file(s)
|    |  2: Add ignore comment
|    | Enter number (Ctrl-D to exit): 0
```
3. Bad list : 
```
s.elm
|   2| ··["one",·"two",·"three"
|    | [NORMAL] ElmLintBear:
|    | ["one", "two", "three"
|    |                            ^
|    | I am looking for one of the following things:
|    | 
|    |     ","
|    |     "]"
|    |     an expression
|    |     an infix operator like +
|    |     an infix operator like `andThen`
|    |     whitespace
|    | *0: Do nothing
|    |  1: Open file(s)
|    |  2: Add ignore comment
|    | Enter number (Ctrl-D to exit): 0
```
All the examples have been taken from my tests. :) 
@Makman2 @jayvdb

bears/elm/ElmLintBear.py
yash-nisar: Tried removing `use_stderr=True` but the console doesn't show any output.

bears/elm/ElmLintBear.py
yash-nisar: Done. :)

bears/elm/ElmLintBear.py
Makman2: alright :)
if you don't need `stdout` then, you can disable it by `use_stdout = False`  :+1:

yash-nisar: Done @Makman2 , is this also ready for merge ? :smile:

bears/elm/ElmLintBear.py
yash-nisar: We have `output_format = regex` and `output_format = corrected`. Since it only detects the error, I think `corrected` cannot be used and since we don't want to use regex, should we go for overriding `process_output` ? @Makman2

Makman2: So what's actually the output of `elm-format`? Do you have a sample?

yash-nisar: Yes @Makman2 , here they are : 
Unexpected comma :
```
 yash  ~  elm-format --stdin
a, b =
  "Hello World!"
ERRORS
-- SYNTAX PROBLEM ------------------------------------------------------ <STDIN>

I ran into something unexpected when parsing your code!

1│ a, b =
    ^
I am looking for one of the following things:

    =
    a pattern
    more letters in this name
    whitespace


 yash  ~  

```
Empty file :
```
 yash  ~  elm-format --stdin
ERRORS
-- SYNTAX PROBLEM ------------------------------------------------------ <STDIN>

I ran into something unexpected when parsing your code!

1│ 
   ^
I am looking for one of the following things:

    "{-|"
    a module declaration
    a port declaration
    a type declaration
    a value definition
    an import
    an infix declaration
    whitespace


 yash  ~  

```
Bad list : 
```
 yash  ~  elm-format --stdin
l =
  ["one", "two", "three"
ERRORS
-- SYNTAX PROBLEM ------------------------------------------------------ <STDIN>

I ran into something unexpected when parsing your code!

3│ 
   ^
I am looking for one of the following things:

    ","
    "]"
    whitespace


 yash  ~  

```

bears/elm/ElmLintBear.py
```diff
@@ -0,0 +1,38 @@
...
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.NpmRequirement import NpmRequirement
+
+
+@linter(executable='elm-format',
+        use_stdin=True,
+        use_stderr=True,
+        use_stdout=False,
+        output_format='regex',
+        output_regex='(?P<line>\d+)│(?P<message>[\s\S]*)')
```
Makman2: `[\s\S]` will match everything. And it is be greedy, so you will swallow all following messages too

bears/elm/ElmLintBear.py
Makman2: Though complex output, `regex` is the right choice. Can you control the output format of `elm-format` somehow maybe?

bears/elm/ElmLintBear.py
yash-nisar: `elm-format` reports only one error at a time, so that will not be a problem IIUC. Here is a use case: 
```
 yash  ~  elm-format --stdin
l =
  ["one", "two", "three"
a, b =
  "Hello World!"
ERRORS
-- SYNTAX PROBLEM ------------------------------------------------------ <STDIN>

I ran into something unexpected when parsing your code!

3│ a, b =
   ^
I am looking for one of the following things:

    ","
    "]"
    whitespace


 yash  ~  

```
CC @Makman2

bears/elm/ElmLintBear.py
sils: @jayvdb why should we avoid stdin? If we use stdin things will be kept in memory and there's less IO right? It also means we can use it for things that don't exist on file systems making the bear easier to reuse in our efforts to get languages in languages analyzed or testing.

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,94 @@
...
+
+ListTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_list,),
+    invalid_files=(bad_file_list,))
+
+FunctionTest = verify_local_bear(
+    ElmLintBear,
+    valid_files=(good_file_function,),
+    invalid_files=(bad_file_function,))
```
sils: can we have on eexploratory test here? Like to see a result

bears/elm/ElmLintBear.py
sils: https://github.com/avh4/elm-format#usage given that it seems like it can actually also auto reformat the file

tests/elm/ElmLintBearTest.py
yash-nisar: Sure, why not :)

bears/elm/ElmLintBear.py
yash-nisar: @sils , we wanted to use `--stdin` right from https://github.com/coala/coala-bears/pull/1093#discussion_r91087892 ? or should I add it as an option ?

sils: doesn't mean we can't use other options as well! We gotta use what's best

yash-nisar: I agree, adding it as an option. :)

tests/elm/ElmLintBearTest.py
yash-nisar: I have addressed your comments @sils . Please have a look when you are free. ;)

bears/elm/ElmLintBear.py
jayvdb: Firstly, IO related to a normal source file will be statistically insignificant compared to the IO of starting `elm-format`.  And "language in language" can use temporary named files (and will need to, as many linters dont support stdin), so there is no problems there.

It isnt easy for the external command to handle stdin.  It is typically done as an afterthought, and not done properly.
If there is a problem reading a file, fallback mechanism can be attempted in normal exception handling of the linter, and it is fairly normal to do this.
If there is a problem reading stdin, ... the program must have kept a literal bytestream in order to re-attempt to read the file.  Most programs dont bother, letting their stdin support lag behind their filename handling.
Then there are all the problems with stdin encoding detection, which is not the same as file handling because the language runtimes believe they should 'help' the three default handles, especially wrt utf8.

For a very practical example of the types of problems we'll be encountering, I offer https://gitlab.com/pycqa/flake8/issues?scope=all&search=stdin&state=all
And that is just Python.)
Every other runtime has its own lovely oddities.
(I expect I could find at least five difficult bugs in coala's subprocess creation, in 24 hrs)

Also filenames help with logic around paths, which obviously cant occur with stdin. (some of the above flake8 bugs are because of this; flake8 allows passing the filename to be used in conjunction with stdin to work around some of these problems).

There is also the aspect of `overwrite_file` below needing the filename anyway, so why not always do that instead of having two code paths that achieve the same thing.

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,138 @@
...
+        self.uut = ElmLintBear(Section('name'), Queue())
+
+    def test_bad_function(self):
+        prepared = prepare_file(bad_file_function.splitlines(),
+                                filename=None, create_tempfile=False)
+        with prepared as (file, fname):
+            self.check_results(
+                self.uut,
+                bad_file_function.splitlines(),
+                [Result.from_values('ElmLintBear',
```
jayvdb: can you store this string in a module variable with multi-line `"""..."""` so it is easier to read.

yash-nisar: Done. :)

.ci/deps.sh
```diff
@@ -175,3 +175,10 @@ fi
 echo '' > dummy
 coala-ci --bears CheckstyleBear --files dummy --no-config --bear-dirs bears || true
 coala-ci --bears ScalaLintBear --files dummy --no-config --bear-dirs bears || true
+
+if [ ! -e ~/elm-format-0.18/elm-format ]; then
+  mkdir -p ~/elm-format-0.18
+  curl -fsSL -o elm-format.tgz https://github.com/avh4/elm-format/releases/download/0.5.2-alpha/elm-format-0.17-0.5.2-alpha-linux-x64.tgz
+  tar -xvzf elm-format.tgz -C ~/elm-format-0.18
+fi
+
```
yash-nisar: SpaceConsistencyBear had complained about it. :(, nvm removed the blank line. :)

.ci/deps.sh
```diff
@@ -175,3 +175,10 @@ fi
 echo '' > dummy
 coala-ci --bears CheckstyleBear --files dummy --no-config --bear-dirs bears || true
 coala-ci --bears ScalaLintBear --files dummy --no-config --bear-dirs bears || true
+
```
yash-nisar: Resolved merge conflicts and moved it higher in the file. :)

.ci/deps.sh
yash-nisar: Added it again due to gitmate issues.

tests/elm/ElmLintBearTest.py
```diff
@@ -0,0 +1,145 @@
...
+bad_file_function = """
+import Html exposing (text)
+four =
+  sqr 16
+main =
+  text (toString [four)
+"""
+
+result_message_bad_function = """\
+text (toString [four)\x1b[31m
```
jayvdb: can color be disabled?
It isnt helping anyone.

jayvdb: I am guessing the only way to do this is to set `TERM=dumb` before invoking haskell.
But that may not help as [hSupportsANSI](https://github.com/avh4/elm-format/search?utf8=%E2%9C%93&q=hSupportsANSI) isnt checked, and the [ANSI](https://github.com/avh4/elm-format/search?utf8=%E2%9C%93&q=ANSI) doesnt appear to otherwise care much about color, so we'd need to dig through the libraries to know -- or do a build.
Maybe later.

jayvdb:
ack f04f8bc

jayvdb:
@rultor merge

PR #1491, opened 2017-03-06
LineCountBear.py: Add exclude_blank_lines option

yashtrivedi96:
Add option to exclude blank lines. The setting is disabled by default.

Closes https://github.com/coala/coala-bears/issues/1478

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/LineCountBear.py
+++ b/bears/general/LineCountBear.py
@@ -1,3 +1,5 @@
+import re
+
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.Result import Result
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
@@ -10,14 +12,24 @@ class LineCountBear(LocalBear):
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting'}
 
-    def run(self, filename, file, max_lines_per_file: int):
+    def _get_blank_line_count(self, file):
+        num_blank_lines = len(
+            list(filter(lambda x: re.match(r'^\s*$', x), file)))
+        return num_blank_lines
+
+    def run(self, filename, file, max_lines_per_file: int,
+            exclude_blank_lines: bool=False):
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: "True" if blank lines are to be excluded.
         """
         file_length = len(file)
+        if exclude_blank_lines:
+            num_blank_lines = self._get_blank_line_count(file)
+            file_length = file_length - num_blank_lines
         if file_length > max_lines_per_file:
             yield Result.from_values(
                 origin=self,
--- a/tests/general/LineCountBearTest.py
+++ b/tests/general/LineCountBearTest.py
@@ -25,3 +25,18 @@ class LineCountBearTest(LocalBearTestHelper):
             filename='default')
         self.check_validity(self.uut, ['1 line'])
         self.check_validity(self.uut, [])  # Empty file
+
+    def test_exclude_blank_lines(self):
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.section.append(Setting('max_lines_per_file', 2))
+        self.check_results(
+            self.uut, ['line 1', ' ', 'line 2',
+                       'line 3', '\n', '\t', ' line 4',
+                       'line 5 ', ' line 6 ', '\t\tline 7',
+                       '', '\t \n ', ' \t\n '],
+            [Result.from_values('LineCountBear',
+                                'This file had 7 lines, which is 5 lines more '
+                                'than the maximum limit specified.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+            filename='default')


tests/general/LineCountBearTest.py
```diff
@@ -15,13 +15,28 @@ def setUp(self):
 
     def test_run(self):
         self.section.append(Setting('max_lines_per_file', 1))
+        self.section.append(Setting('exclude_blank_lines', False))
```
meetmangukiya: You can remove this

tests/general/LineCountBearTest.py
```diff
@@ -15,13 +15,28 @@ def setUp(self):
...
+        self.section.append(Setting('max_lines_per_file', 2))
+        self.check_results(
+            self.uut, ['line 1', ' ', 'line 2',
+                       'line 3', '\n', '\t'],
+            [Result.from_values('LineCountBear',
+                                'This file had 3 lines, which is 1 lines more '
+                                'than the maximum limit specified.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+
```
meetmangukiya: Remove this blank line

bears/general/LineCountBear.py
```diff
@@ -10,14 +10,24 @@ class LineCountBear(LocalBear):
...
+        :param exclude_blank_lines: Enable to exclude blank lines.
         """
         file_length = len(file)
+
+        blank_line_counter = 0
+        file_length = len(file)
+        for line in file:
+            if line.split() == []:
+                blank_line_counter += 1
+        if exclude_blank_lines is True:
```
meetmangukiya: Just `if exclude_blank_lines` should do fine

bears/general/LineCountBear.py
```diff
@@ -10,14 +10,24 @@ class LineCountBear(LocalBear):
...
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: Enable to exclude blank lines.
         """
         file_length = len(file)
+
+        blank_line_counter = 0
+        file_length = len(file)
+        for line in file:
+            if line.split() == []:
+                blank_line_counter += 1
```
meetmangukiya: This can be reduced to`len(list(filter(lambda x: re.match(r'\s+')), file)` . Actually am not sure if we'll have to convert to list explicitly to get the length, try if we can get length of an iterator ;)

tests/general/LineCountBearTest.py
yashtrivedi96: OK line number 41

bears/general/LineCountBear.py
yashtrivedi96: Yeah I'll try this one :)

bears/general/LineCountBear.py
yashtrivedi96: OK

bears/general/LineCountBear.py
yashtrivedi96: I think re.match() takes at least 2 arguments.
len(list(filter(lambda x: re.match(r'\s+',x),file)))

yashtrivedi96: this should work len(list(filter(lambda x: x.strip(),file)))

meetmangukiya: Oh, yes, that's what I meant :P

meetmangukiya: Seriously, I don't like using x.strip here, use regex :)

yashtrivedi96: for this 
import re
a = ['line 1', 'line 2', 'line 3','\n','\t line 4',' line 5','     ', '\n\n']

lines=list(filter(lambda x: re.match(r'\s+',x),a))
print(lines)

the output is
['\n', '\t line 4', ' line 5', '     ', '\n\n']

and for x.strip()
a = ['line 1', 'line 2', 'line 3','\n','\t line 4',' line 5','     ', '\n\n']

lines=list(filter(lambda x: x.strip(),a))
print(lines)
the output is
['line 1', 'line 2', 'line 3', '\t line 4', ' line 5']

meetmangukiya: Well, this is unexpected, will get back to you in 18 hours, not on computer now, and am going to take a nap

yashtrivedi96: Ok :)

meetmangukiya: @yashtrivedi96 go with `re.fullmatch`. Confirmed works :+1:

tests/general/LineCountBearTest.py
```diff
@@ -16,12 +16,26 @@ def setUp(self):
     def test_run(self):
         self.section.append(Setting('max_lines_per_file', 1))
         self.check_results(
-            self.uut, ['line 1', 'line 2', 'line 3'],
+            self.uut, ['line 1', 'line 2', 'line 3', '\n'],
             [Result.from_values('LineCountBear',
-                                'This file had 3 lines, which is 2 lines more '
+                                'This file had 4 lines, which is 3 lines more '
```
yash-nisar: I would suggest not to modify the existing tests. Just adding an additional test should be fine. :)

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,20 @@ class LineCountBear(LocalBear):
...
 
-    def run(self, filename, file, max_lines_per_file: int):
+    def run(self, filename, file, max_lines_per_file: int,
+            exclude_blank_lines: bool=False):
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: Enable to exclude blank lines.
```
yash-nisar: Can be better framed as `True if blank lines are to be excluded.`

tests/general/LineCountBearTest.py
```diff
@@ -16,12 +16,26 @@ def setUp(self):
...
         self.check_validity(self.uut, [])  # Empty file
+
+    def test_exclude_blank_lines(self):
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.section.append(Setting('max_lines_per_file', 2))
+        self.check_results(
+            self.uut, ['line 1', ' ', 'line 2',
+                       'line 3', '\n', '\t'],
+            [Result.from_values('LineCountBear',
+                                'This file had 3 lines, which is 1 lines more '
```
yash-nisar: 1 lines -> 1 line

tests/general/LineCountBearTest.py
```diff
@@ -16,12 +16,26 @@ def setUp(self):
...
                                 file='default')],
             filename='default')
         self.check_validity(self.uut, ['1 line'])
         self.check_validity(self.uut, [])  # Empty file
+
+    def test_exclude_blank_lines(self):
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.section.append(Setting('max_lines_per_file', 2))
+        self.check_results(
+            self.uut, ['line 1', ' ', 'line 2',
```
yash-nisar: I would suggest you to add all the possible cases, ie. 
* leading whitespaces/tabs before the `line x` -> '   line x'
* trailing whitespaces/tabs after the `line x` -> 'line x   '
* both of the above cases `line x` -> '   line x   '
* some lines with multiple tabs

tests/general/LineCountBearTest.py
meetmangukiya: No, it is correct, see the result generated in the bear

tests/general/LineCountBearTest.py
yashtrivedi96: Ok

tests/general/LineCountBearTest.py
```diff
@@ -25,3 +25,18 @@ def test_run(self):
...
+                       'line 3', '\n', '\t', ' line 4',
+                       'line 5 ', ' line 6 ', '\t\tline 7'],
+            [Result.from_values('LineCountBear',
+                                'This file had 7 lines, which is 5 lines more '
+                                'than the maximum limit specified.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+
```
yash-nisar: This blank line needs to be removed. :)

bears/general/LineCountBear.py
```diff
@@ -2,6 +2,8 @@
 from coalib.results.Result import Result
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 
+import re
```
yash-nisar: The order of imports can be preserved, it can be something like : 
```
import re

from coalib.results.Result import 
from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY 


class LineCountBear...
```

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,20 @@ class LineCountBear(LocalBear):
...
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: True if blank lines are to be excluded.
         """
         file_length = len(file)
+        blank_lines = len(
+            list(filter(lambda x: re.match(r'^\s*$', x), file)))
```
yash-nisar: Why breaking this to 2 lines, try joining it to a single line as: 
`blank_lines = len(list(filter(lambda x: re.match(r'^\s*$', x), file)))`

yash-nisar: It fits in the 79 char range. :)

yashtrivedi96: Ok

tests/general/LineCountBearTest.py
jayvdb: But it is still a bug...

sils:
ack https://github.com/coala/coala-bears/pull/1491/commits/82ef6716711b81d6b55cbf43d34c78b9bfda446e

jayvdb:
needs a rebase.

yashtrivedi96:
Ok

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,19 @@ class LineCountBear(LocalBear):
...
+            exclude_blank_lines: bool=False):
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: True if blank lines are to be excluded.
         """
         file_length = len(file)
+        blank_lines = len(list(filter(lambda x: re.match(r'^\s*$', x), file)))
```
nemaniarjun: Seems good! 
Some suggestions:
 - Calculate the blank lines only if exclude_blank_lines is true
 - Consider compiling the regex only once, since the length of the file can be a lot, this might be a considerable performance issue 
(Correct me if I am wrong.)
Also for the second part this might be useful http://stackoverflow.com/questions/938429/scope-of-python-lambda-functions-and-their-parameters

tests/general/LineCountBearTest.py
```diff
@@ -25,3 +25,17 @@ def test_run(self):
             filename='default')
         self.check_validity(self.uut, ['1 line'])
         self.check_validity(self.uut, [])  # Empty file
+
+    def test_exclude_blank_lines(self):
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.section.append(Setting('max_lines_per_file', 2))
+        self.check_results(
+            self.uut, ['line 1', ' ', 'line 2',
+                       'line 3', '\n', '\t', ' line 4',
```
jayvdb: probably also a good idea to have one empty line with multiple mixed whitespace characters in it; currently they are all single character lines.

bears/general/LineCountBear.py
jayvdb: The compiled RE will definitely be cached in this pipeline (see note in [the docs](https://docs.python.org/2/library/re.html#re.compile), so a method level compiled variable would be useless.
This RE will probably be cached for the entire section being processed. The same bear is run on each file as part of the same stage of processing, and the RE cache probably wont be overflowed if only a few bears are running in a single stage of processing.

Moving this RE to a compiled variable at the module or class level means extra work must be done at startup, which isnt a good idea as this bear may not even be needed by `.coafile`, so the extra work is wasted.

jayvdb: Please move this algorithm into a `_get_blank_line_count` function, just to encapsulate the logic.

The assignment of `blank_lines` should be inside the `if ... :` block below.  It is unused computation if `exclude_blank_lines` is disabled.

tests/general/LineCountBearTest.py
yashtrivedi96: Yes I will add some more cases in the list.

bears/general/LineCountBear.py
yashtrivedi96: Ok

nemaniarjun: Thanks for the new piece of knowledge :D
Will keep this in mind!! :)

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,24 @@ class LineCountBear(LocalBear):
...
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: "True" if blank lines are to be excluded.
         """
         file_length = len(file)
+        if exclude_blank_lines:
+            num_blank_lines = LineCountBear.get_blank_line_count(self, file)
```
pratyushprakash: `get_blank_line_count` should be a static method as it is not using the bear instance

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,25 @@ class LineCountBear(LocalBear):
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting'}
 
-    def run(self, filename, file, max_lines_per_file: int):
+    @staticmethod
+    def get_blank_line_count(file):
```
yash-nisar: Renaming the function with a leading underscore would be a good practise. See https://www.python.org/dev/peps/pep-0008/ for more details. Also, you can make it an instance method like `_get_blank_line_count(self, file):`

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,25 @@ class LineCountBear(LocalBear):
...
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: "True" if blank lines are to be excluded.
         """
         file_length = len(file)
+        if exclude_blank_lines:
+            num_blank_lines = LineCountBear.get_blank_line_count(file)
```
yash-nisar: Call the method with an object like `self._get_blank_line_count..`

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,25 @@ class LineCountBear(LocalBear):
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting'}
 
-    def run(self, filename, file, max_lines_per_file: int):
+    @staticmethod
```
yash-nisar: `@staticmethod` can be removed since you are calling it on an object.

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,25 @@ class LineCountBear(LocalBear):
...
+    def _get_blank_line_count(seilf, file):
```
yash-nisar: `seilf` -> `self`

bears/general/LineCountBear.py
```diff
@@ -10,14 +12,25 @@ class LineCountBear(LocalBear):
...
         """
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
         :param max_lines_per_file: Number of lines allowed per file.
+        :param exclude_blank_lines: "True" if blank lines are to be excluded.
         """
         file_length = len(file)
+        if exclude_blank_lines:
+            num_blank_lines = self._get_blank_line_count(self, file)
```
yash-nisar: `self` in the parenthesis needs to be removed. :)

yash-nisar:
ack 34f1be48490397b9d4e6781406e4b766c09ddc23

Asnelchristian:
@rultor merge

yashtrivedi96:
Thanks :)

PR #1514, opened 2017-03-14
DuplicateFileBear: Add DuplicateFileBear

Siddhartha1234:
This bear checks for files with same content .

Closes https://github.com/coala/coala-bears/issues/1479

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+from coalib.bears.GlobalBear import GlobalBear
+from coalib.results.Result import Result
+
+
+class DuplicateFileBear(GlobalBear):
+    LANGUAGES = {'All'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Duplicates'}
```
Techievena: Duplication instead. Please refer to the standard aspect names in use.

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def run(self):
+        """
+        Checks for Duplicate Files using filecmp module
+        of Standard Library
+        """
+
+        unique_tuples = [fileTuple for fileTuple in
```
jayvdb: fileTuple doesnt comply with PEP8 naming

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+                         itertools.combinations(self.file_dict, 2)]
+
+        if not unique_tuples:
+            yield Result(self, 'You included no files '
+                         'or only a single file.')
+        else:
+
```
jayvdb: blank line why?

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        Checks for Duplicate Files using filecmp module
+        of Standard Library
+        """
+
+        unique_tuples = [fileTuple for fileTuple in
+                         itertools.combinations(self.file_dict, 2)]
+
+        if not unique_tuples:
+            yield Result(self, 'You included no files '
```
jayvdb: "no files" error should probably part of GlobalBear.

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+                         'or only a single file.')
```
jayvdb: "single file" can be determined before building `unique_tuples`

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+        else:
+
+            for fileTuple in unique_tuples:
+                if filecmp.cmp(fileTuple[0], fileTuple[1]):
+                    firstFileName = os.path.basename(fileTuple[1])
+                    secondFileName = os.path.basename(fileTuple[1])
+                    message = 'File ' + firstFileName + ' identical' \
```
jayvdb: please never use `\` (line continuation) -- use round brackets, which will then require you align the parts in brackets correctly.

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,36 @@
...
+                         itertools.combinations(self.file_dict, 2)]
+
+        if not unique_tuples:
+            yield Result(self, 'You included no files '
+                         'or only a single file.')
+        else:
+
+            for fileTuple in unique_tuples:
+                if filecmp.cmp(fileTuple[0], fileTuple[1]):
+                    firstFileName = os.path.basename(fileTuple[1])
```
jayvdb: why not show the full filename?
The same filename could appear multiple times in the repository.

tests/general/DuplicateFileBearTest.py
```diff
@@ -0,0 +1,34 @@
...
+        for file in self.files:
+            self.file_dict[file] = ''
+
+        self.uut = DuplicateFileBear(self.file_dict, self.section, self.queue)
+        return list(self.uut.run())
+
+    def check_results(self):
+        messages = self.obtain_results()
+        self.assertEqual(messages, [
+            'File bigFirst.txt identical to bigSecond.txt',
```
jayvdb: what is the benefit of having such large files in the repository?

Siddhartha1234:
I tried to address the issues you both mentioned ( @jayvdb  and @Techievena ) . I'll rebase once it is approved and tests pass(I tested Locally) . I also removed the big test files.

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def run(self):
+        """
+        Checks for Duplicate Files using filecmp module
+        of Standard Library
+        """
+        if not self.file_dict:
+            yield Result(self, 'You did not add any file to compare')
+        else:
+            unique_tuples = [fileTuple for fileTuple in
```
meetmangukiya: use snake case for `fileTuple`

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,32 @@
...
+                             itertools.combinations(self.file_dict, 2)]
+
+            for file_pair in unique_tuples:
+                if filecmp.cmp(file_pair[0], file_pair[1]):
+                    firstFileName = file_pair[0]
```
meetmangukiya: use snake_case again

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,32 @@
...
+                    secondFileName = file_pair[1]
```
meetmangukiya: same

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,32 @@
...
+                    message = ('File ' + firstFileName + ' identical'
```
meetmangukiya: Change the message to `File x is identical to file y` maybe :/

nano.save
```diff
@@ -0,0 +1,4 @@
+DuplicateFileBear: Add DuplicateFileBear
```
meetmangukiya: :/ ? remove this file, please.

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,32 @@
...
+        of Standard Library
+        """
+        if not self.file_dict:
+            yield Result(self, 'You did not add any file to compare')
+        else:
+            unique_tuples = [fileTuple for fileTuple in
+                             itertools.combinations(self.file_dict, 2)]
+
+            for file_pair in unique_tuples:
+                if filecmp.cmp(file_pair[0], file_pair[1]):
```
meetmangukiya: I think we can improve in performance here. The files are already loaded by `coala`. use something like `file_dict[file_pair[0]] == file_dict[file_pair[1]]`. ? We are providing file names here, I guess python loads the files again and then compares(not sure though).

bears/general/DuplicateFileBear.py
```diff
@@ -0,0 +1,32 @@
...
+            unique_tuples = [file_tuple for file_tuple in
+                             itertools.combinations(self.file_dict, 2)]
+
+            for file_pair in unique_tuples:
+                if (self.file_dict[file_pair[0]] ==
+                        self.file_dict[file_pair[1]]):
+                    first_file_name = file_pair[0]
+                    second_file_name = file_pair[1]
+                    message = ('File ' + first_file_name + ' is identical'
+                               ' to File ' + second_file_name)
```
meetmangukiya: I think jayvdb requested to have absolute path here

Siddhartha1234: Yes, It shows absolute path when i run it locally .

Siddhartha1234: @meetmangukiya Checking for duplicates can not be done in less than O(N2) as we have to check each pair if they are same.

meetmangukiya: @Siddhartha1234 am not familiar with the `O`s(I know I should :sweat: ) Is the change I requested, an improvement or not ?

Siddhartha1234: @meetmangukiya  Let's say you take the 0th indexed file. You can't guarantee that there are no duplicates for it unless you traverse all the 1, 2, .... n -1 files and we have to repeat the process for other n - 1 files. Total N^2 comparisions .But if you consider only unique tuples : 
```
for i in range(n-1):
    for j in range(i+1,n) : 
        Compare i'th and j'th file 
```
You reduce the number of comparisions but it is still O(N^2) (In simple terms, there exists k such that it is bounded by k* N^2 no of comparisions) . 
So , my method is optimal for any deterministic method to find duplicates :smile:  .

meetmangukiya: the lines shown in the CI log next to the DuplicateFileBear in the test report are never executed by your test cases. Build a test case where all the lines are executed, and that'll get you fully covered.

Siddhartha1234: But its just skipping the Test entirely it seems. Running py.test locally doesn't run test for only that particular Test file and also my test cases cover all the cases ( there can only be three , since its a simple if else block) . Any reason why that happens ?

bears/general/DuplicateFileBear.py
jayvdb: This check is missing from 1f3b556

Siddhartha1234: @jayvdb : Checking for single file isn't necessary.Also, no output means "No duplicates found".I just added a message for "no files" as it might have been due to mentioning wrong pattern in config file or param.As you said, I checked for that before building `unique_tuples` .

jayvdb: Well, DuplicateFileBear with only a single file is IMO a config error, as it is doing nothing so the "No duplicates found" is misleading as nothing was actually attempted.

tests/general/DuplicatefileBearTest.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_results_no_duplicates(self):
+        results = self.get_results([self.test_files[2],
+                                    self.test_files[3]])
+        messages = [result.message for result in results]
+        self.assertEquals(messages, [])
+
+    def test_results_empty(self):
+        results = self.get_results([])
+        messages = [result.message for result in results]
```
jayvdb: This should check the level of the message is MAJOR (and it should be set as severity MAJOR)

jayvdb:
ack 63d31e1

jayvdb:
@rultor merge

PR #1518, opened 2017-03-16
bears/__init__.py: Check version with coala core

adhikasp:
Print a warning if the installed version of coala is not 
same with coala-bears.

Closes https://github.com/coala/coala/issues/3042

### DIFF
--- a/bears/__init__.py
+++ b/bears/__init__.py
@@ -4,6 +4,8 @@ the bears that are officially supported by coala.
 """
 
 
+from distutils.version import StrictVersion
+import logging
 import os
 import sys
 
@@ -20,3 +22,27 @@ def assert_supported_version():  # pragma: no cover
     if sys.version_info < (3, 4):
         print('coala supports only python 3.4 or later.')
         exit(4)
+
+
+def check_coala_version():
+    """
+    Check the installed coala and coala-bears version.
+    """
+    try:
+        import coalib
+        # check only MAJOR.MINOR version
+        coalib_version = '.'.join(coalib.__version__.split('.')[:2])
+        bears_version = '.'.join(__version__.split('.')[:2])
+        if StrictVersion(coalib_version) > StrictVersion(bears_version):
+            logging.warning('Version mismatch between coala ({}) and '
+                            'coala-bears ({}). This may or may not cause '
+                            'errors. If you encounter a problem, try to '
+                            'update coala and coala-bears to latest version '
+                            "with 'pip3 install -U coala coala-bears'."
+                            .format(coalib.__version__, __version__))
+    except ImportError:  # pragma: no cover
+        logging.error('Module coalib is not found. Cannot do version '
+                      'checking.')
+
+
+check_coala_version()
--- a/tests/coalaBearTest.py
+++ b/tests/coalaBearTest.py
@@ -1,0 +1,25 @@
+import logging
+import unittest
+
+import bears
+import coalib
+
+
+class coalaBearTest(unittest.TestCase):
+
+    def test_check_coala_version(self):
+        bears.__version__ = coalib.__version__ = '1.0.0.dev999999999'
+        logger = logging.getLogger()
+        with self.assertLogs(logger, 'WARNING') as cm:
+            bears.check_coala_version()
+            self.assertEqual(0, len(cm.output))
+
+            bears.__version__ = '1.1.0.dev999999999'
+            bears.check_coala_version()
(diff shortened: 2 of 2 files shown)


bears/__init__.py
```diff
@@ -20,3 +20,25 @@ def assert_supported_version():  # pragma: no cover
...
+    """
+    Check the installed coala and coala-bears version when the coalib loads
+    coala-bears module.
+    """
+    try:
+        import coalib
+        if coalib.__version__ != VERSION:
+            logging.warning('Version mismatch between coala ({}) and '
+                            'coala-bears ({}). This may or may not cause '
+                            'errors. If you encounter problem, try to update '
```
NiklasMM: "encounter problem" -> "encounter a problem"

bears/__init__.py
```diff
@@ -20,3 +20,25 @@ def assert_supported_version():  # pragma: no cover
...
+                            'coala and coala bears to latest version with '
```
NiklasMM: "coala bears" -> "coala-bears"

NiklasMM:
The issue you're referencing does not exists. Please fix the link!

bears/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 import os
 import sys
-
+import logging
```
jayvdb: logging should be first, so that standard library imports are in alpha order.  PEP8

tests/coalaBearTest.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+import logging
+import bears
+import coalib
```
jayvdb: Please follow PEP8

bears/__init__.py
```diff
@@ -20,3 +20,25 @@ def assert_supported_version():  # pragma: no cover
...
+
+
+def check_coala_version():
+    """
+    Check the installed coala and coala-bears version when the coalib loads
+    coala-bears module.
+    """
+    try:
+        import coalib
+        if coalib.__version__ != __version__:
```
jayvdb: this is too simple.

coala-bears version might be greater than coala version.  Isnt that OK?
You need to do a proper version comparison.  There are classes that make this very easy.

adhikasp: @jayvdb 

> coala-bears version might be greater than coala version.

Could you explain further about the version comparison case? AFAIU, coala use `MAJOR.MINOR.PATCH.BUILD`. Do we need to just compare that `MAJOR.MINOR` of coalib is **greater** than bears?

```python
coalib_version = '.'.join(coalib.__version__.split('.')[:-2])
bears_version = '.'.join(bears.__version__.split('.')[:-2])
if StrictVersion(coalib_version) > StrictVersion(bears_version):
    print('warning')
```

jayvdb: Yes, that is much better logic.

The patch level of each package runs independently, as those patches should be bugs only within that package, and released very quickly, and thus lots of bugs in a minor release of coalib could mean its patch level increments more quickly than the bears patch level.

tests/CoalaBearTest.py
```diff
@@ -0,0 +1,21 @@
...
+import unittest
+
+import bears
+import coalib
+
+
+class CoalaBearTest(unittest.TestCase):
+
+    def test_check_coala_version(self):
+        bears.__version__ = '1.0.0.999999999'
```
jayvdb: should be `1.0.0.dev999999999` here and below.  See contents of `coalib/VERSION`.

tests/CoalaBearTest.py
```diff
@@ -0,0 +1,21 @@
...
+        coalib.__version__ = bears.__version__
```
jayvdb: And you can do `coalib.__version__ = bears.__version__ = '1.0.0.dev999999999'` to make it more clear

bears/__init__.py
```diff
@@ -20,3 +22,28 @@ def assert_supported_version():  # pragma: no cover
     if sys.version_info < (3, 4):
         print('coala supports only python 3.4 or later.')
         exit(4)
+
+
+def check_coala_version():
+    """
+    Check the installed coala and coala-bears version when the coalib loads
```
jayvdb: Just "Check the installed coala and coala-bears version." so it fits onto one line (pep257), and ... it doesnt matter how the coala-bears package is loaded ... this will run.

jayvdb:
ack 774f509

jayvdb:
reack a5c01a0

jayvdb:
@rultor merge

PR #1521, opened 2017-03-17
setup.py: Fix vim-vint installation bug

Madhukar-SH:
Fixed vim-vint setup command which was causing issues
in systems having python version older than 3.4

Fixes https://github.com/coala/coala/issues/3878

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,17 @@ with open('requirements.txt') as requirements:
 with open('bear-requirements.txt') as requirements:
     bear_required = requirements.read().splitlines()
 
+version = sys.version_info  # Checks if python version is <3.4
+# If yes changes the vim-vint~=x.x to vim-vint=x.x
+if(version[0] == 3 and version[1] <= 4):
+    vim_pkg = 'vim-vint'
+    vim_package_line = [line for line in bear_required if vim_pkg in line]
+    if (vim_package_line):
+        for package in vim_package_line:
+            bear_required.remove(package)
+            vim_pkg = package.replace('~', '=')
+            bear_required.append(vim_pkg)
+
 with open('test-requirements.txt') as requirements:
     test_required = requirements.read().splitlines()
 


Madhukar-SH:
Sorry coala is broken on my system for some reason, hence not able to run it and check for code errors.

setup.py
```diff
@@ -55,7 +66,6 @@ def run(self):
 required += [req for req in bear_required
              if not req.startswith('language-check')]
 
-
```
meetmangukiya: unrelated change

meetmangukiya: You are not supposed to make this change

Madhukar-SH: Removing empty line?

Madhukar-SH: Undo-ing it

setup.py
```diff
@@ -43,6 +43,17 @@ def run(self):
 with open('bear-requirements.txt') as requirements:
     bear_required = requirements.read().splitlines()
 
+version = sys.version_info  # Checks if python version is <3.4
+# If yes changes the vim-vint~=x.x to vim-vint=x.x
+if(version[0] == 3 and version[1] <= 4):
+    vim_pkg = 'vim-vint'
+    vim_package_line = [line for line in bear_required if vim_pkg in line]
```
Techievena: Why the braces?

Techievena: `line for line in bear_required if vim_pkg in line` Please split this

Madhukar-SH: Braces because its a list comprehension https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions

Madhukar-SH: Is it really necessary to split it and write it as `for loop with if inside` ?

Madhukar-SH: List comprehension is a bit faster actually

Techievena: Well didn't knew about such a thing. Then I think its fine. :+1:

setup.py
```diff
@@ -43,6 +43,17 @@ def run(self):
 with open('bear-requirements.txt') as requirements:
     bear_required = requirements.read().splitlines()
 
+version = sys.version_info  # Checks if python version is <3.4
+# If yes changes the vim-vint~=x.x to vim-vint=x.x
+if(version[0] == 3 and version[1] <= 4):
```
jayvdb: See https://github.com/coala/coala/issues/3878#issuecomment-288269752 .
This doesnt include when the version is 2.x, which is where the problem is.

jayvdb:
Closing as the bug hasnt been reproduced yet.
Cant fix what you cant see.

setup.py
Madhukar-SH: I made this since we never use python 2.x for installing coala-bears

Madhukar-SH:
As the issue post says, this was problem with specific versions of Linux mint 17.x where newer versions of python wasn't available

Madhukar-SH:
I felt this was pretty much it https://www.bountysource.com/issues/38134858-177-breaks-pip-install-for-python2
Linux mint 17.3 64 bit

Check the comment of dezza (27days ago from today)

jayvdb:
Please continue the analysis discussion on https://github.com/coala/coala/issues/3878 , not here.

PR #1522, opened 2017-03-20
generate_package.py: Upload specific bears

madhukar01:
This script is now able to upload single bear with `-b Bearname`
as argument.
Tests have been added for the same

Fixes https://github.com/coala/coala-bears/issues/845

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/generate_package.py
+++ b/bears/generate_package.py
@@ -98,6 +98,8 @@ def create_upload_parser():
                         action='store_true')
     parser.add_argument('-u', '--upload', help='Upload the packages on PyPi',
                         action='store_true')
+    parser.add_argument('-b', '--bear', help='Name of the bear to upload',
+                        nargs='+')
     return parser
 
 
@@ -114,7 +116,28 @@ def main():
     else:
         bear_version = repr(bear_version) + '.' + str(int(time.time()))
 
-    for bear_file_name in sorted(set(glob('bears/**/*Bear.py'))):
+    bear_files = []
+    local_files = sorted(set(glob('bears/**/*Bear.py')))
+
+    """
+    File names in local_files will be of pattern 'bears/somefolder/somebear.py'
+    Example: 'bears/css/CSSLintBear.py'
+    To check if the input bear exist, we need to check if it is
+    installed (i.e present in local_files)
+    Input will be just bearname, we need to extract the bearname from the
+    filename and then compare to check if input bear exists
+    """
+    if args.bear:
+        for bearname in args.bear:
+            for bear_file in local_files:
+                installed_bear = bear_file.split('/')[-1].strip('.py')
+                if bearname == installed_bear:
+                    bear_files.append(bear_file)
+                    break
+    else:
+        bear_files = local_files
+
+    for bear_file_name in bear_files:
         bear_object = next(iimport_objects(
             bear_file_name, attributes='kind', local=True),
             None)
--- a/tests/generate_packageTest.py
+++ b/tests/generate_packageTest.py
@@ -155,5 +155,77 @@ class MainTest(unittest.TestCase):
 
     def tearDown(self):
         shutil.rmtree(os.path.join(self.BEARS_UPLOAD_PATH))
+        os.chdir(self.orig_dir)
 
+
+class MainSpecificBearsTest(unittest.TestCase):
+
+    BEARS_PATH = os.path.abspath(os.path.join(
+        os.path.dirname(__file__), '..', 'bears'))
+    BEARS_UPLOAD_PATH = os.path.join(BEARS_PATH, 'upload')
+    CSS_BEAR_SETUP_PATH = os.path.join(
+        BEARS_UPLOAD_PATH, 'CSSLintBear', 'setup.py')
+    coala_BEAR_SETUP_PATH = os.path.join(
+        BEARS_UPLOAD_PATH, 'coalaBear', 'setup.py')
+
+    def setUp(self):
+        self.orig_dir = os.getcwd()
+
+        self.argv = ['generate_package.py', '-b', 'CSSLintBear', 'coalaBear']
+        argv_patcher = patch.object(sys, 'argv', self.argv)
+        self.addCleanup(argv_patcher.stop)
+        self.argv_mock = argv_patcher.start()
+
+    def test_sp_main(self):
+        main()
+        self.assertTrue(os.path.exists(self.BEARS_UPLOAD_PATH))
+        with open(self.CSS_BEAR_SETUP_PATH) as fl:
+            setup_py = fl.read()
+        self.assertIn('Check code for syntactical or semantical', setup_py)
+        with open(self.coala_BEAR_SETUP_PATH) as fl:
+            setup_py = fl.read()
+        self.assertIn('None', setup_py)
+
+    def test_sp_main_bear_version_prod(self):
+        fake_prod_version = '99.99.99'
+        with patch('bears.generate_package.VERSION', fake_prod_version):
+            main()
+        with open(self.CSS_BEAR_SETUP_PATH) as fl:
+            setup_py = fl.read()
+        self.assertIn(fake_prod_version, setup_py)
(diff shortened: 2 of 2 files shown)


Madhukar-SH:
It would be great If you guys would help me with the travis CI.
https://travis-ci.org/coala/coala-bears/jobs/213292747#L3393

I have added a testcase where it covers the single bear input too. But travis seems to tell me there is no full coverage
@sils @Adrianzatreanu

tests/generate_packageTest.py
```diff
@@ -90,7 +90,7 @@ class MainTest(unittest.TestCase):
     NO_BEAR_PATH = os.path.join('bears', 'BadBear', 'InvalidBear.py')
 
     def setUp(self):
-        self.argv = ['generate_package.py']
+        self.argv = ['generate_package.py -b CSSLintBear']
```
NiklasMM: This should probably be changed to `['generate_package.py', '-b',  'CSSLintBear']`

bears/generate_package.py
```diff
@@ -159,12 +161,23 @@ def main():
...
             bear_dist_name = bear_name + '-' + bear_version
-            if args.register:
+            if args.register and not args.bear:
                 perform_register(os.path.join('bears', 'upload', bear_name),
                                  bear_dist_name)
-            if args.upload:
+            if args.upload and not args.bear:
                 perform_upload(os.path.join('bears', 'upload', bear_name))
 
+    if args.bear:
```
Adrianzatreanu: i think we should be able to upload multiple bears such as running:
"-b CSSLintBear CSharpBear" or so, it would be nice.

Adrianzatreanu: also, shouldnt we have this at the start and just do a "return" "if args.bear" so we wouldnt have to check "if args.register and not args.bear" each time? as in it wouldnt even reach this "if" if args.bear was set.

Madhukar-SH: Added this :+1:

Madhukar-SH: I had a better code as you said in previous commits, but wasnt covering full with travis.
So was trying to figure out how the tests work. Hence ended up with this.

Anyways, have fixed it in latest commit

vijeth-aradhya:
LGTM! ack df3d1f2b3535c40fa967c28ad1fa710ed9921b9a

supergr35:
@vijeth-aradhya in order for an ack to go through, it must be the only thing in a comment.

supergr35:
Commit message can be changed to describe the change made in the imperative mood. Something more like `Add the ablility to upload` and then the rest of the commit message. Also change the line about tests 👍

bears/generate_package.py
```diff
@@ -98,12 +98,16 @@ def create_upload_parser():
                         action='store_true')
     parser.add_argument('-u', '--upload', help='Upload the packages on PyPi',
                         action='store_true')
+    parser.add_argument('-b', '--bear', help='Name of the bear to upload',
+                        nargs='+', default=None)
```
meetmangukiya: No need to use `default=None`

bears/generate_package.py
```diff
@@ -113,7 +117,13 @@ def main():
     else:
         bear_version = repr(bear_version) + '.' + str(int(time.time()))
 
-    for bear_file_name in sorted(set(glob('bears/**/*Bear.py'))):
+    if args.bear:
+        bear_files = [f for f in sorted(set(glob('bears/**/*Bear.py')))
+                      for x in args.bear if x in f]
```
meetmangukiya: I don't get this expression :/ Please break it down if possible

bears/generate_package.py
```diff
@@ -113,7 +117,13 @@ def main():
...
+    else:
+        bear_files = sorted(set(glob('bears/**/*Bear.py')))
```
meetmangukiya: You are using `sorted(set(glob('bears/**/*Beary.py')))` in both the `if` and `else` branches, why not store it in another variable and then use it.

bears/generate_package.py
NiklasMM: This is quite complicated to read, could you please add a comment explaining what's happening here.

NiklasMM: Also this introduces the hard requirement, that Bear classes must be named *exactly* like their file. I'm not sure if we can do that?

madhukar01: Sure will add comments explaining them. It is actually a `list comprehension` in python. It is actually reducing use of a `for loop` and `two if statements`

madhukar01: Sure

bears/generate_package.py
madhukar01: Alright

supergr35:
commit body: `upload bears` -> `upload a single bear`

NiklasMM:
Also, [this](https://github.com/coala/coala-bears/pull/1522#discussion_r108598690) comment is still not addressed. With this commit we are creating the hard constraint, that bear classes must be named exactly like the file. Is that something we want/can do? cc @sils @jayvdb

bears/generate_package.py
madhukar01: At the current stage, all bear files have the bear name as a part of their file names. So I dont think this should be an issue

madhukar01:
We just have to make sure that bear name is a part of file name too. I'm not sure If we have to do this.
I thought of this since the previous edition was just taking all the files and creating packages. 
So I came up with adding the constraint there

madhukar01:
@supergr35 The bear can also upload multiple bears on providing multiple names, so is it necessary to specify `single bear`?

bears/generate_package.py
```diff
@@ -113,7 +117,18 @@ def main():
     else:
         bear_version = repr(bear_version) + '.' + str(int(time.time()))
 
-    for bear_file_name in sorted(set(glob('bears/**/*Bear.py'))):
+    local_files = sorted(set(glob('bears/**/*Bear.py')))
+
+    if args.bear:  # Checks if bear name entered is available on the machine
```
Adrianzatreanu: why "x" and "f"? cant you find better variable names :( ?

bears/generate_package.py
```diff
@@ -113,7 +117,18 @@ def main():
...
+        for f in local_files:
```
Adrianzatreanu: so why exactly is this done?
local_files is a list of strings, right? why would you iterate through it? cant you just check whether each of the bears given as arg is in this list?
the only way i would see it the way youre doing it is so that you could give partial names like "coala" instead of "coalaBear", is this it? but wouldnt this give some false positives in some cases? like writing RBear and getting something like "Cpprbear" or so (cpprbear doesnt exist, im just making a scenario in which it would be bad)?

tests/generate_packageTest.py
```diff
@@ -83,6 +83,41 @@ def test_command(self, call_mock):
         call_mock.assert_called_with(['twine', 'upload', './dist/*'])
 
 
+class MainSingleTest(unittest.TestCase):
```
Adrianzatreanu: good tests but a bit incomplete. we should have tests for all situations. for example, 2 or 3 bears given as "-b A B C D".
also a test for no matching bears (incorrect names).

bears/generate_package.py
madhukar01: Lets say there are 2 bears coalaBear and coalaliintBear, giving input as 'coala' would lead to both bears to be uploaded.

madhukar01: `cant you just check whether each of the bears given as arg is in this list?` 
This is what I am doing.
Main thing is, `local_files` will have all the bear files named `somebear.py` . 
So the user will enter `somebear` and not `somebear.py`. 
Hence we can't do it the other way round (i.e check if the arg is in the list. Because name in arg will not exactly match name in list)

Adrianzatreanu:
this seems to fail a lot of tests and is outdated, can you fix it please @madhukar01 ?

supergr35:
Looks like it's not quite at 100% coverage

PR #1527, opened 2017-03-21
CheckstyleBear: Detect errors

jayvdb:
Since 9227382, errors from checkstyle were ignored.

Also set java files to be native EOL in .gitattributes,
to match default EOL expected by Checkstyle rule
NewlineAtEndOfFile.

Fixes https://github.com/coala/coala-bears/issues/1464

### DIFF
--- a/.ci/appveyor.yml
+++ b/.ci/appveyor.yml
@@ -51,6 +51,10 @@ install:
 build: false  # Not a C# project, build stuff at the test step instead.
 
 test_script:
+  # Force DOS format, as Checkstyle configs enable NewlineAtEndOfFile,
+  # which defaults to CRLF on Windows, and Appveyor CI ignores .gitattributes
+  # http://help.appveyor.com/discussions/problems/5687-gitattributes-changes-dont-have-any-effect
+  - unix2dos tests/java/test_files/CheckstyleGood.java
   # Clang DLLs x64 were nowadays installed, but the x64 version hangs, so we
   # exclude according tests. See https://github.com/appveyor/ci/issues/495 and
   # https://github.com/appveyor/ci/issues/688
--- a/bears/java/CheckstyleBear.py
+++ b/bears/java/CheckstyleBear.py
@@ -33,7 +33,7 @@ def known_checkstyle_or_path(setting):
 
 @linter(executable='java',
         output_format='regex',
-        output_regex=r'\[(?P<severity>WARN|INFO)\].*?'
+        output_regex=r'\[(?P<severity>ERROR|WARN|INFO)\].*?'
                      r'(?P<line>\d+):?(?P<column>\d+)?. '
                      r'(?P<message>.*?) *\[(?P<origin>[a-zA-Z]+?)\]')
 class CheckstyleBear:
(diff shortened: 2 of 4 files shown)


.ci/appveyor.yml
```diff
@@ -51,6 +51,10 @@ install:
 build: false  # Not a C# project, build stuff at the test step instead.
 
 test_script:
+  # Force DOS format, as Checkstyle configs enable NewlineAtEndOfFile,
+  # which defaults to CRLF on Windows, and Appveyor CI ignores .gitattributes
+  # http://help.appveyor.com/discussions/problems/5687-gitattributes-changes-dont-have-any-effect
+  - unix2dos tests/java/test_files/CheckstyleGood.java
```
sils: let's do this with the bad file as well so we see the differences between the configs.

jayvdb: I dont object to doing that; certainly makes sense.

However pushing back a little bit, and only a little bit, this line fixes one problem occurring in `test_style_android`, and it occurs only on Windows.  The bad files already fail, for other reasons, and fail on unix, so the only thing which more `unix2dos` would find is more Windows specific issues wrt checkstyle.

When I upgraded to Checkstyle 7.6 (https://github.com/coala/coala-bears/pull/1466), there are three failed method - `test_style_sun` and `test_style_google` also fail [on Appveyor](https://ci.appveyor.com/project/coala/coala-bears/build/1.0.5492/job/jx37mkvuaa0faidm) and not on Travis/Circle , so there are more Windows specific problems somewhere in there, which will be found in the next set of patches for CheckstyleBear.

jayvdb: @sils [replied on gitter](https://gitter.im/coala/coala?at=58d23e1358ad4eb0283d73f4)

bears/java/CheckstyleBear.py
```diff
@@ -33,7 +33,7 @@ def known_checkstyle_or_path(setting):
 
 @linter(executable='java',
         output_format='regex',
-        output_regex=r'\[(?P<severity>WARN|INFO)\].*?'
+        output_regex=r'\[(?P<severity>WARN|INFO|ERROR)\].*?'
```
sims1253: Small thing:
Why not order it info warn error?

jayvdb: Good catch.  It seems (`git grep '<severity>'`) ERROR|WARN|INFO would be the more common ordering.

tests/java/CheckstyleBearTest.py
```diff
@@ -24,24 +24,24 @@ def setUp(self):
...
-
     def test_style_google(self):
         self.section['checkstyle_configs'] = 'google'
         self.check_validity(self.uut, [], self.good_file)
+        self.check_validity(self.uut, [], self.bad_file, valid=False)
 
     def test_style_sun(self):
         self.section['checkstyle_configs'] = 'sun'
-        self.check_validity(self.uut, [], self.good_file)
+        self.check_validity(self.uut, [], self.good_file, valid=False)
```
adtac: `valid=False`? Is that non-Sun-compliant?

jayvdb: yup.  I'm surprised it is compliant with any ruleset to be honest.

Nosferatul:
ack 0d4d895

Nosferatul:
needs rebase

jayvdb:
reack d9234ef

jayvdb:
@rultor merge

PR #1529, opened 2017-03-22
.travis.yml: Test version check on Python 2.7

jayvdb:
Avoid installation of mypy-lang and install 3to2 early
as it is needed by guess-language and language-check
as they do not install cleanly on Python 2 without it.

Avoid compilation of Python modules to avoid Python 3
syntax errors in build logs.

Places jobs for unsupported Python versions after the
supported versions, so newcomers dont read those first.

Related to https://github.com/coala/coala/issues/3310
Related to https://github.com/coala/coala/pull/3614

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,16 @@ sudo: required
 
 language: python
 python:
-  # 2.7 is rejected by dependency mypy-lang
-  - 3.3
   - 3.4
   - 3.5
   - 3.6
+
+matrix:
+  include:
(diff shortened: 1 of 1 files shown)


sils:
ack https://github.com/coala/coala-bears/pull/1529/commits/cd51fd89289ad6403511fc312fbf4015dd4a9375?diff=split though I wonder why the test takes so long

sils:
@rultor merge

PR #1530, opened 2017-03-22
FilenameBear.py: Enable default naming convention

yash-nisar:
Add default naming conventions that are widely accepted for
languages like java, javascript and python.

Closes https://github.com/coala/coala-bears/issues/1083

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/FilenameBear.py
+++ b/bears/general/FilenameBear.py
@@ -19,14 +19,22 @@ class FilenameBear(LocalBear):
                           'snake': to_snakecase,
                           'space': to_spacecase}
 
+    _language_naming_convention = {
+                                   '.java': 'pascal',
+                                   '.js': 'kebab',
+                                   '.py': 'snake',
+                                  }
+
     def run(self, filename, file,
-            file_naming_convention: str='snake',
+            file_naming_convention: str=None,
             ignore_uppercase_filenames: bool=True):
         """
         Checks whether the filename follows a certain naming-convention.
 
         :param file_naming_convention:
             The naming-convention. Supported values are:
+            - ``auto`` to guess the correct convention. Defaults to ``snake``
+            if the correct convention cannot be guessed.
             - ``camel`` (``thisIsCamelCase``)
             - ``kebab`` (``this-is-kebab-case``)
             - ``pascal`` (``ThisIsPascalCase``)
@@ -38,6 +46,23 @@ class FilenameBear(LocalBear):
         """
         head, tail = os.path.split(filename)
         filename_without_extension, extension = os.path.splitext(tail)
+
+        if file_naming_convention is None:
+            self.warn('Please specify a file naming convention explicitly'
+                      ' or use "auto".')
+            file_naming_convention = 'auto'
+        else:
+            file_naming_convention = file_naming_convention.lower()
+
+        if file_naming_convention == 'auto':
+            if extension in self._language_naming_convention:
+                file_naming_convention = self._language_naming_convention[
+                    extension]
+            else:
+                self.warn('The file naming convention could not be guessed. '
+                          'Using the default "snake" naming convention.')
+                file_naming_convention = 'snake'
+
         try:
             new_name = self._naming_convention[file_naming_convention](
                 filename_without_extension)
--- a/tests/general/FilenameBearTest.py
+++ b/tests/general/FilenameBearTest.py
@@ -21,14 +21,13 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
 
         log = self.uut.message_queue.get(timeout=1)
         self.assertEqual(log.message,
-                         'Invalid file-naming-convention provided: INVALID')
+                         'Invalid file-naming-convention provided: invalid')
         self.assertEqual(log.log_level, LOG_LEVEL.ERROR)
 
     def test_snake_case(self):
+        self.section['file_naming_convention'] = 'snake'
         self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py')
         self.check_validity(self.uut, [''], filename='XYZ/__init__.py')
-
-        self.section['file_naming_convention'] = 'snake'
         self.check_validity(self.uut, [''], filename='/a/camCase', valid=False)
 
     def test_camel_case(self):
@@ -46,11 +45,12 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
         self.check_validity(self.uut, [''], filename='/a/kebab-case')
 
     def test_pascal_case(self):
-        self.section['file_naming_convention'] = 'pascal'
-        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py',
-                            valid=False)
-        self.check_validity(self.uut, [''], filename='XYZ/__Init__.py')
-        self.check_validity(self.uut, [''], filename='/a/PascalCase')
+        for convention in ('Pascal', 'pascal', 'PASCAL'):
+            self.section['file_naming_convention'] = convention
+            self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py',
+                                valid=False)
+            self.check_validity(self.uut, [''], filename='XYZ/__Init__.py')
+            self.check_validity(self.uut, [''], filename='/a/PascalCase')
 
     def test_space_case(self):
         self.section['file_naming_convention'] = 'space'
@@ -63,8 +63,58 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
         self.check_validity(self.uut, [''], filename='/a/Space Case')
 
     def test_ignore_upper(self):
+        self.section['file_naming_convention'] = 'auto'
         self.check_validity(self.uut, [''], filename='/LICENSE')
 
         self.section['ignore_uppercase_filenames'] = 'nope'
 
         self.check_validity(self.uut, [''], filename='/LICENSE', valid=False)
+
+    def test_default_naming_java(self):
+        self.section['file_naming_convention'] = 'auto'
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.java',
+                            valid=False)
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x-y.java',
+                            valid=False)
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x y.java',
+                            valid=False)
+        self.check_validity(self.uut, [''], filename='/Home/xyz/XY.java')
+
+    def test_default_naming_javascript(self):
+        self.section['file_naming_convention'] = 'auto'
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.js',
+                            valid=False)
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x y.js',
+                            valid=False)
+        self.check_validity(self.uut, [''], filename='/Home/xyz/XY.js')
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x-y.js')
+
+    def test_default_file_naming_python(self):
+        self.section['file_naming_convention'] = 'auto'
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py')
+        self.check_validity(self.uut, [''], filename='XYZ/__init__.py')
+        self.check_validity(self.uut, [''], filename='/a/x y.py', valid=False)
+
+    def test_none_file_naming_convention_warning(self):
+        self.uut.execute('filename', [])
+
+        # Automatic bear messages we want to ignore.
+        self.uut.message_queue.get()
+
+        log = self.uut.message_queue.get(timeout=1)
+        self.assertEqual(log.message,
+                         'Please specify a file naming convention explicitly'
+                         ' or use "auto".')
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py')
+
+    def test_auto_file_naming_convention_warning(self):
+        self.section['file_naming_convention'] = 'auto'
+        self.uut.execute('filename.xyz', [])
+
+        # Automatic bear messages we want to ignore.
+        self.uut.message_queue.get()
+
+        log = self.uut.message_queue.get(timeout=1)
+        self.assertEqual(log.message,
+                         'The file naming convention could not be guessed. '
+                         'Using the default "snake" naming convention.')
--- a/tests/general/FilenameBearTest.py
+++ b/tests/general/FilenameBearTest.py
@@ -6,7 +6,7 @@ from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.settings.Section import Section
 
 
-class SpaceConsistencyBearTest(LocalBearTestHelper):
+class FilenameBearTest(LocalBearTestHelper):
 
     def setUp(self):
         self.section = Section('test section')


yash-nisar:
@jayvdb Please have a look when you're free. :)

meetmangukiya:
This can be better implemented and maintained this way IMO:
- Create a dict that contains extensions as keys and lambdas that accept filename as argument and return the standard naming convention compliant name
- Then try 
```
try: 
  dict[extension](filename) 
except KeyError:
  # some nice message ?
```

This way there will be much less `if`s if we want to add other language supports in future :)

bears/general/FilenameBear.py
```diff
@@ -37,8 +37,14 @@ def run(self, filename, file,
...
         filename_without_extension, extension = os.path.splitext(tail)
         try:
+            if extension == '.java':
+                file_naming_convention = 'pascal'
+
             new_name = self._naming_convention[file_naming_convention](
                 filename_without_extension)
+
+            if extension == '.js':
+                new_name = new_name.replace('_', '-')
```
jayvdb: the JS style should have its own proper name defined in the dict.

Create a separate issue & PR, as required, as this would be a feature.

bears/general/FilenameBear.py
```diff
@@ -37,8 +37,14 @@ def run(self, filename, file,
         head, tail = os.path.split(filename)
         filename_without_extension, extension = os.path.splitext(tail)
         try:
+            if extension == '.java':
```
jayvdb: youch.  this is in a try block designed to catch the KeyError ... nothing else should be in here.

yash-nisar:
@Makman2 @javvdb Please have a look when you're free. :)

bears/general/FilenameBear.py
```diff
@@ -19,14 +20,20 @@ class FilenameBear(LocalBear):
                           'snake': to_snakecase,
                           'space': to_spacecase}
 
+    _language_naming_convention = {'.java': 'pascal',
+                                   '.js': 'kebab',
+                                   '.py': 'snake'}
+
     def run(self, filename, file,
-            file_naming_convention: str='snake',
+            file_naming_convention: str=None,
```
Makman2: maybe make `auto` the default behaviour?

bears/general/FilenameBear.py
```diff
@@ -19,14 +20,20 @@ class FilenameBear(LocalBear):
                           'snake': to_snakecase,
                           'space': to_spacecase}
 
+    _language_naming_convention = {'.java': 'pascal',
+                                   '.js': 'kebab',
+                                   '.py': 'snake'}
```
Makman2: I would allow it for this merge, but we should definitely introduce a simple system somewhere else mapping file endings to language identifiers, so we have a single interface for that kind of functionality :+1:

bears/general/FilenameBear.py
yash-nisar: Okay :+1:

yash-nisar: According to @jayvdb ,
> Ideally, we change the default file_naming_convention to None, but then we have two paths forward if the user doesnt set a value, 
> 1. use `snake`, but warn the user that they should be explicit, and allow auto for the user to explicitly request that we guess the right convention which will make it backwards compatible.

bears/general/FilenameBear.py
jayvdb: https://github.com/AlejandroFrias/case-conversion might have one. If not, they might be a good maintainer of a simple mapping.
coala would eventually have a Language class approach

jayvdb: This should be a module level private variable, and it will be extended, so the layout should be friendly to additions.
```py
x = {
  a,
  b,
}
```

bears/general/FilenameBear.py
```diff
@@ -19,14 +19,20 @@ class FilenameBear(LocalBear):
...
     def run(self, filename, file,
-            file_naming_convention: str='snake',
+            file_naming_convention: str=None,
             ignore_uppercase_filenames: bool=True):
         """
         Checks whether the filename follows a certain naming-convention.
 
         :param file_naming_convention:
             The naming-convention. Supported values are:
+            - ``auto`` to guess the right convention. Defaults to
```
jayvdb: s/right/correct/

bears/general/FilenameBear.py
jayvdb: Instead of `str`, I believe we have a LOV type specifier, which means you can remove the `.lower()` below

yash-nisar: The tests were failing when I provided `Auto` instead of `auto` so I think we need to keep the `.lower()`. @jayvdb

yash-nisar:
@jayvdb @Makman2 Changes that I've performed since the last push:
* Added friendly layout for `_language_naming_convention = {
                                   '.java': 'pascal',
                                   '.js': 'kebab',
                                   '.py': 'snake',
                                  }`
* Changed `right` -> `correct` in docstring ` ``auto`` to guess the right convention...`
* Since the tests were failing when I removed the `.lower()` and supplied `Auto`, I chose to keep the `.lower()` or maybe we could open an issue for this in the core.

jayvdb:
> Since the tests were failing when I removed the .lower()

Did you find out why they failed?
Have you investigated replacing type annotation str with a list of values?

yash-nisar:
We haven't incorporated a conversion function yet IIUC because when I run 
`coala -V -b FilenameBear -S file_naming_convention=Snake -f xyz.py --flush-cache`, the value of the setting is `Snake` and when I run
`coala -V -b FilenameBear -S file_naming_convention=snake -f xyz.py --flush-cache`, the value of the setting is `snake` which I confirmed by using the debug mode.

yash-nisar:
Should we have this feature or should we expect the users to type in the exact setting ? @jayvdb

yash-nisar:
I've removed `.lower()` but this wont work if the user enters a setting as `AUTO` or `Auto` instead of `auto. @jayvdb

bears/general/FilenameBear.py
```diff
@@ -19,14 +19,22 @@ class FilenameBear(LocalBear):
...
-            file_naming_convention: str='snake',
+            file_naming_convention: str=None,
             ignore_uppercase_filenames: bool=True):
         """
         Checks whether the filename follows a certain naming-convention.
 
         :param file_naming_convention:
             The naming-convention. Supported values are:
+            - ``auto`` to guess the correct convention. Defaults to
+            ``snake`` if the correct convention cannot be guessed.
```
meetmangukiya: Align so that it starts just below the backticks of `auto` ?

bears/general/FilenameBear.py
```diff
@@ -38,6 +46,21 @@ def run(self, filename, file,
         """
         head, tail = os.path.split(filename)
         filename_without_extension, extension = os.path.splitext(tail)
+
+        if file_naming_convention is None:
+            self.err('Please specify a file naming convention explicitly. '
```
meetmangukiya: IMO, this is more of a warning, since we are falling back to something default. It should be an error if the bear can't run.

bears/general/FilenameBear.py
```diff
@@ -38,6 +46,21 @@ def run(self, filename, file,
...
+                     'Using the default "snake" naming convention...')
+            file_naming_convention = 'snake'
+
+        if file_naming_convention == 'auto':
+            if extension not in self._language_naming_convention.keys():
+                self.err('The file naming convention could not be guessed. '
```
meetmangukiya: Again, this should be warning, refer previous comments

bears/general/FilenameBear.py
yash-nisar: I think it is alright since it is the continuation of the first point.

yash-nisar:
Addressed your comments @meetmangukiya

yash-nisar:
@Makman2 Is it okay now ?

bears/general/FilenameBear.py
Makman2: hm right I think it improves readability in source when you align this to `auto` :+1:

bears/general/FilenameBear.py
```diff
@@ -19,14 +19,22 @@ class FilenameBear(LocalBear):
...
 
+    _language_naming_convention = {
+                                   '.java': 'pascal',
+                                   '.js': 'kebab',
+                                   '.py': 'snake',
+                                  }
+
     def run(self, filename, file,
-            file_naming_convention: str='snake',
+            file_naming_convention: str=None,
```
Makman2: so what about making this `auto` by default? :)
(I think I commented that already...)

yash-nisar: @jayvdb wanted a `None` option by default so we can warn the user to explicitly set the `file_naming_convention`

Makman2: alright :+1:

bears/general/FilenameBear.py
```diff
@@ -38,6 +46,21 @@ def run(self, filename, file,
         """
         head, tail = os.path.split(filename)
         filename_without_extension, extension = os.path.splitext(tail)
+
+        if file_naming_convention is None:
+            self.warn('Please specify a file naming convention explicitly. '
+                      'Using the default "snake" naming convention...')
+            file_naming_convention = 'snake'
```
Makman2: not `auto`?

bears/general/FilenameBear.py
```diff
@@ -38,6 +46,21 @@ def run(self, filename, file,
...
+
+        if file_naming_convention.lower() == 'auto':
+            if extension not in self._language_naming_convention.keys():
```
Makman2: omit `.keys()`, as this is the same:
`extension not in self._language_naming_convention`

bears/general/FilenameBear.py
```diff
@@ -38,6 +46,21 @@ def run(self, filename, file,
...
+                self.warn('The file naming convention could not be guessed. '
+                          'Using the default "snake" naming convention...')
+                file_naming_convention = 'snake'
+            else:
+                file_naming_convention = self._language_naming_convention[
+                    extension]
```
Makman2: I would reverse order of the if, then you don't need to invert condition:
```
if extension in ...
    file_naming_convention = self._language_....
else:
    self.warn...
```

bears/general/FilenameBear.py
yash-nisar: The first point [here](https://github.com/coala/coala-bears/issues/1083#issuecomment-288518740) states that we use `snake` if the user doesn't mention explicitly. Is that okay ? @Makman2

bears/general/FilenameBear.py
Makman2: you could also use `try-catch` with `KeyError`. Up to you :+1: :)

yash-nisar: Roger that. :)

yash-nisar: @Makman2 Should these naming conventions be case insensitive ?

tests/general/FilenameBearTest.py
```diff
@@ -63,8 +62,58 @@ def test_space_case(self):
...
+
+        log = self.uut.message_queue.get(timeout=1)
+        self.assertEqual(log.message,
+                         'Please specify a file naming convention explicitly. '
+                         'Using the default "snake" naming convention...')
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py')
+
+    def test_auto_file_naming_convention_warning(self):
+        self.section['file_naming_convention'] = 'auto'
+        self.uut.execute('filename.cpp', [])
```
Makman2: use some weirdo extension, cpp might be implemented :)

yash-nisar: Alright @Makman2 , I'm on it now. :)

tests/general/FilenameBearTest.py
```diff
@@ -6,7 +6,7 @@
 from coalib.settings.Section import Section
 
 
-class SpaceConsistencyBearTest(LocalBearTestHelper):
+class FilenameBearTest(LocalBearTestHelper):
```
Makman2: ohh damn copy paste virus :3
mind to put that into another commit? :)

yash-nisar: Sure, why not ?As you say :D

bears/general/FilenameBear.py
yash-nisar: Done. :)

bears/general/FilenameBear.py
yash-nisar: Done. :)

bears/general/FilenameBear.py
Makman2: yeah though we should maybe use `auto` still, so the user directly gets the language-default. When you supply Java files and it defaults to snake you think "huh what the... why snake?".

Makman2: Oh and other point: I would add to the warning message:
```
Please specify a file naming convention explicitly or use "auto".
```
"explicitly specifying" would mean for me that you shouldn't use "auto", but instead `snake`, `camel` or others

yash-nisar:
To make things easier for reviewers, things that I've included since the last push: 
* Aligned `auto` with `snake`. See https://github.com/coala/coala-bears/pull/1530#discussion_r108040695
* Used weirdo extension of `filename.xyz`. See https://github.com/coala/coala-bears/pull/1530#discussion_r108041075
* Reversed the order of the `if` condition. See https://github.com/coala/coala-bears/pull/1530#discussion_r108040853
* Removed the `.keys()`. See https://github.com/coala/coala-bears/pull/1530#discussion_r108040835
* Added a separate commit for changing the classname.
* Added `.lower()` so `file_naming_convention` is case insensitive and added a test for the same. :smile: 
@Makman2 over to you. :)

yash-nisar:
@Makman2 cast that magic spell :P

bears/general/FilenameBear.py
yash-nisar: Did not align because it caused indentation problems https://circleci.com/gh/coala/coala-bears/5890#tests/containers/1

bears/general/FilenameBear.py
```diff
@@ -19,14 +19,22 @@ class FilenameBear(LocalBear):
...
-            file_naming_convention: str='snake',
+            file_naming_convention: str=None,
             ignore_uppercase_filenames: bool=True):
         """
         Checks whether the filename follows a certain naming-convention.
 
         :param file_naming_convention:
             The naming-convention. Supported values are:
+            - ``auto`` to guess the correct convention. Defaults to ``snake``
+            if the correct convention cannot be guessed.
```
Makman2: alignment :3

bears/general/FilenameBear.py
```diff
@@ -38,8 +46,23 @@ def run(self, filename, file,
...
+                file_naming_convention = self._language_naming_convention[
+                    extension]
+            else:
+                self.warn('The file naming convention could not be guessed. '
+                          'Using the default "snake" naming convention...')
+                file_naming_convention = 'snake'
+
         try:
-            new_name = self._naming_convention[file_naming_convention](
+            new_name = self._naming_convention[file_naming_convention.lower()](
```
Makman2: maybe you should lower `file_naming_convention` in the beginning in L54, then you don't need to lower it for comparison and here too. (You could do for example sth like this:
```
if file_naming_convention is None:
    self.warn...
    file_naming ... = auto
else:
    file_naming_convention = file_naming_convention.lower()
```
)

tests/general/FilenameBearTest.py
```diff
@@ -46,7 +45,7 @@ def test_kebab_case(self):
         self.check_validity(self.uut, [''], filename='/a/kebab-case')
 
     def test_pascal_case(self):
-        self.section['file_naming_convention'] = 'pascal'
+        self.section['file_naming_convention'] = 'Pascal'
```
Makman2: up to you: you could test both, by wrapping into a loop:
```python
for convention in ('pascal', 'Pascal', 'PASCAL')
    self.section = ...
    self.check_validity...
```

tests/general/FilenameBearTest.py
```diff
@@ -63,8 +62,58 @@ def test_space_case(self):
...
+                         'Please specify a file naming convention explicitly'
+                         ' or use "auto".')
+        self.check_validity(self.uut, [''], filename='/Home/xyz/x_y.py')
+
+    def test_auto_file_naming_convention_warning(self):
+        self.section['file_naming_convention'] = 'auto'
+        self.uut.execute('filename.xyz', [])
+
+        # Automatic bear messages we want to ignore.
+        self.uut.message_queue.get(timeout=1)
```
Makman2: do you have to specify a timeout? I would invoke that without timeout, our tests do run max 40secs before pytest aborts the test ;)

bears/general/FilenameBear.py
yash-nisar: Builds fail stating as `indentation problem` so I did not align it. Is it okay ? See https://github.com/coala/coala-bears/pull/1530#discussion_r108043976

Makman2: Ah okay nvm then :+1:

bears/general/FilenameBear.py
```diff
@@ -38,6 +46,23 @@ def run(self, filename, file,
...
+        else:
+            file_naming_convention = file_naming_convention.lower()
+
+        if file_naming_convention == 'auto':
+            if extension in self._language_naming_convention:
+                file_naming_convention = self._language_naming_convention[
+                    extension]
+            else:
+                self.warn('The file naming convention could not be guessed. '
+                          'Using the default "snake" naming convention...')
```
Makman2: maybe turn the `...` into a single period `.`? Up to you :+1:

bears/general/FilenameBear.py
yash-nisar: @Makman2 build failures due to indentation problems in yapf bear (completely unrelated to my PR). How should we handle this ?

bears/general/FilenameBear.py
yash-nisar: Alright :)

yash-nisar:
@Makman2 Can you ack so we can get this merged as soon as the issues that are blocking this get fixed ?

yash-nisar:
@Makman2 The tests pass now, one last review ?

Makman2:
ack 1b1bdea 3c9c352

Makman2:
@rultor merge

PR #1595, opened 2017-04-09
pyroma_test_files: Remove unnecessary print

jayvdb:
The print call in the pyroma_test_files complate package ``__init__``
is unnecessary for the test, and causes unexpected output when a test
runner such as ``green`` collects ``test_*.py`` by default.

Fixes https://github.com/coala/coala-bears/issues/1594


Nosferatul:
ack 1096b29

Nosferatul:
@rultor merge

Issue #1573, opened 2017-04-01
aspectsYEAHBear


userzimmermann:
Based on @Techievena's great idea from https://github.com/coala/coala/issues/4039#issuecomment-290886365 , we should grow the `aspectsYEAHBear` that checks if the terms *aspect* and *aspects* are always written lowercase according to https://github.com/coala/cEPs/issues/69 , just like the `coalaBear` does it for the term *coala*: https://github.com/coala/coala-bears/blob/master/bears/general/coalaBear.py

The `aspectsYEAHBear` of course needs an aspect for the above main task and is always open for any ideas about additional tasks to perform.

@coala/aspects-developers

userzimmermann:
The `aspectsYEAHBear` could also check if **aspectsYEAH** is always written correctly :)

@coala/aspects-developers Opinions?

Asnelchristian:
yep it should, and we should probably implement Grammar( or Spelling)
aspect for the this

2017-04-01 16:40 GMT+01:00 Stefan Zimmermann <notifications@github.com>:

Asnelchristian:
annotating the its(aspectsYEAHBear)  results right away

2017-04-01 16:43 GMT+01:00 Christian Asnel Ngoulla <ngoullasob@gmail.com>:

userzimmermann:
@Asnelchristian Something like `Spelling.aspectsYEAH` :)

userzimmermann:
@coala/aspects-developers So! Thanks to @Techievena we now have the `aspectsYEAH` aspect:

https://github.com/coala/coala/pull/4049
https://github.com/coala/coala/blob/master/coalib/bearlib/aspects/Spelling.py#L12

Time for the `aspectsYEAHBear`! @pratyushprakash

jayvdb:
Creating a bear to do this, without being disabled by default in coala-quickstart, and removed from the docker, will cause errors in CI for other users repositories.

This should be merged into coalaBear (which is already special-cased to avoid this problem) , and made into a more generic tool, where the default is not going to destroy other peoples repositories who dont adhere to this rule.

jayvdb:
"importance/high" - wtf.  Please have some perspective.  We have far more important problems.

userzimmermann:
@jayvdb It should not be merged with the `coalaBear` but handled exactly the same way. I totally agree with you in that latter aspect :)

jayvdb:
So now we will have a separate bear for each component of coala that wants to have its name written a certain way?

[...]

PR #1604, opened 2017-04-11
aspectsYEAHBear.py: Add aspectsYEAHBear

pratyushprakash:
Add aspectsYEAHBear to make sure `aspect` is always
written with a lower case `a` and `aspectsYEAH` is
always written with upper case `YEAH`.

### DIFF
--- a/bears/general/aspectsYEAHBear.py
+++ b/bears/general/aspectsYEAHBear.py
@@ -1,0 +1,67 @@
+import re
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.bearlib.aspects.Spelling import aspectsYEAH
+from coalib.bearlib.languages import Unknown
+
+
+class aspectsYEAHBear(LocalBear, aspects={
+        'detect': [aspectsYEAH],
+        'fix': [aspectsYEAH], }):
+    LANGUAGES = {'All'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Spelling'}
+
+    def run(self, filename, file, aspects=[
+            aspectsYEAH(Unknown),
+    ]):
+        """
+        Check for the correct spelling of ``aspect`` and ``aspectsYEAH``
+        in the file.
+        """
+
+        for aspect in aspects:
+            if type(aspect) is aspectsYEAH:
+                corrected_aspects = []
+                corrected_aspectsYEAH = []
+                for line in file:
+                    wrong_spelling1 = r'(?!aspect)[aA][sS][pP][eE][cC][tT]'
+                    wrong_spelling2 = r'(?!aspectsYEAH)[aA][sS]'
+                    '[pP][eE][cC][tT][sS][yY][eE][aA][hH]'
+                    if not re.match(wrong_spelling2, line):
+                        corrected_aspects += [re.sub(wrong_spelling1,
+                                                     'aspects',
+                                                     line)]
+                    corrected_aspectsYEAH += [re.sub(wrong_spelling2,
+                                                     'aspectsYEAH',
+                                                     line)]
+
+                aspects_diffs = Diff.from_string_arrays(file,
+                                                        corrected_aspects
+                                                        ).split_diff()
+                aspectsYEAH_diffs = \
+                    Diff.from_string_arrays(file,
+                                            corrected_aspectsYEAH
+                                            ).split_diff()
+                for diff in aspects_diffs:
+                    yield Result(self,
+                                 '``aspect`` is always written with'
+                                 ' lower case characters',
+                                 aspect=aspect,
+                                 affected_code=(diff.range(filename),),
+                                 diffs={filename: diff},
+                                 severity=RESULT_SEVERITY.MAJOR)
+                for diff in aspectsYEAH_diffs:
+                    yield Result(self,
+                                 '``aspectsYEAH`` is always written with'
+                                 ' lower case ``aspects`` and upper case '
+                                 '``YEAH``',
+                                 aspect=aspect,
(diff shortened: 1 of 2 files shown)


pratyushprakash:
I get 100% coverage when tested locally, don't know what is the problem here.

userzimmermann:
@pratyushprakash Great work! I will also investagate on your CI issue :)

bears/general/aspectsYEAHBear.py
```diff
@@ -0,0 +1,67 @@
...
+        Check for the correct spelling of ``aspect`` and ``aspectsYEAH``
+        in the file.
+        """
+
+        for aspect in aspects:
+            if type(aspect) is aspectsYEAH:
+                corrected_aspects = []
+                corrected_aspectsYEAH = []
+                for line in file:
+                    wrong_spelling1 = r'(?!aspect)[aA][sS][pP][eE][cC][tT]'
```
userzimmermann: You only check for *'aspect'* but not for *'aspects'* here

tests/general/aspectsYEAHBearTest.py
```diff
@@ -0,0 +1,42 @@
...
+        self.assertEqual(results[0].severity,
+                         RESULT_SEVERITY.MAJOR)
+
+    def test_message(self):
+        bad_file = """
+        Aspects
+        Aspectyeah"""
+        results = list(self.uut.run('bad_file', bad_file.split('\n')))
+        self.assertEqual(results[0].message,
+                         '``aspect`` is always written'
```
userzimmermann:
```
``aspect`` and ``aspects`` are always...
```

bears/general/aspectsYEAHBear.py
```diff
@@ -0,0 +1,67 @@
...
+                    wrong_spelling2 = r'(?!aspectsYEAH)[aA][sS]'
+                    '[pP][eE][cC][tT][sS][yY][eE][aA][hH]'
+                    if not re.match(wrong_spelling2, line):
+                        corrected_aspects += [re.sub(wrong_spelling1,
+                                                     'aspects',
```
userzimmermann: and here you replace all misspelled *'aspect'* with *'aspects'*

userzimmermann:
Just writing this to get some reference to #1573 here for quick access :)

userzimmermann:
Just writing this to get a reference to this PR in #1573 for quick access :)

userzimmermann:
@pratyushprakash: @jayvdb made a goint point in https://github.com/coala/coala-bears/issues/1573#issuecomment-293278280 based on my https://github.com/coala/coala-bears/issues/1573#issuecomment-293277055

We should try to create some shared logic between `coalaBear` and `aspectsYEAHBear` as part of this PR :)

@coala/aspects-developers

bears/general/aspectsYEAHBear.py
```diff
@@ -0,0 +1,67 @@
...
+class aspectsYEAHBear(LocalBear, aspects={
+        'detect': [aspectsYEAH],
+        'fix': [aspectsYEAH], }):
+    LANGUAGES = {'All'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Spelling'}
+
+    def run(self, filename, file, aspects=[
```
Techievena: Why do we need this `aspects` in the run function?

Techievena: Do you wish to run this bear via passing the `aspects` or something like that?

Techievena: Whatever it is please explain `param aspects` in the docstring.

bears/general/aspectsYEAHBear.py
```diff
@@ -0,0 +1,67 @@
...
+            aspectsYEAH(Unknown),
+    ]):
+        """
+        Check for the correct spelling of ``aspect`` and ``aspectsYEAH``
+        in the file.
+        """
+
+        for aspect in aspects:
+            if type(aspect) is aspectsYEAH:
```
Techievena: `type(aspect) is <aspectclass 'Root.Spelling.aspectsYEAH'>` not `aspectsYEAH`

Techievena: Better replace with `type(aspect).__name__ is 'aspectsYEAH':`. :smiley:

bears/general/aspectsYEAHBear.py
Techievena: If you are willing to call it via aspects, the default value has to be `Root` perhaps instead of `aspectsYEAH`. Well we need to make it universal for all the bears therefore modify it in the run method of Bear class. Remember the discussion we had for the default value. I meant to place it here actually and by mistake I triggered a discussion in the instantiation of Bear class.

bears/general/aspectsYEAHBear.py
userzimmermann: @Techievena If `aspect` is an instance of `aspectsYEAH` then `type(aspect)` is the same class object as referenced by `aspectsYEAH`. `"<aspectclass 'Root.Spelling.aspectsYEAH'>"` is the `repr(aspectsYEAH)`. So nothing is wrong with that condition.

userzimmermann: We only need to implement a more generalized way of running bear logic for supported aspects than in such a `for` loop with several `if type(aspect) is ...` checks

Techievena: My bad must have checked properly before reviewing.

userzimmermann:
@coala/aspects-developers To avoid unnecessary discussions about the value of this bear for now, and to speed up development, we should move it into the **coala** repo under `coalib.bearlib.aspects`

Then we could also use it as a simple testbed for the whole aspects machinery in coala

@pratyushprakash What do you think? Wanna create a new PR with this code in **coala**?

sims1253:
This is fun and all buuuut maybe we don't want too many fun bears?
How about we get a "not so serious bear repository" so we keep the useful ones and the fun project ones separate?

jayvdb:
Removing from the queue.  Re-open if there is consensus on the issue that this is the right way forward.

PR #1619, opened 2017-04-18
GitCommitBear: Download nltk data

jayvdb:
The nltk data is a dependency when setting shortlog_imperative_check
is enabled, which is default.  The bear should do this check
instead of requiring the user does it prior to running coala.

Fixes https://github.com/coala/coala-bears/issues/1617


jayvdb:
I need this for the `package_manager` .yml cleanups, as the manual download command is very long, and breaking the line to 80cpl needs very ugly hacks.
Also means we can create newcomer issues to remove the download from every CI script in all repos ;-)

bears/vcs/git/GitCommitBear.py
```diff
@@ -31,6 +31,18 @@ class GitCommitBear(GlobalBear):
...
+
+    @classmethod
+    def setup_dependencies(cls):
+        if not cls._nltk_data_downloaded:
+            nltk.download([
+                'punkt',
+                'maxent_treebank_pos_tagger',
+                'averaged_perceptron_tagger',
+            ])
+            cls._nltk_data_downloaded = True
```
SanketDG: why not use this variable later in the code for a check?

jayvdb: I use it above, to prevent this code being executed twice.

NiklasMM: I assume `nltk.download` does nothing if the data is already there?

bears/vcs/git/GitCommitBear.py
```diff
@@ -31,6 +31,18 @@ class GitCommitBear(GlobalBear):
     }
     CONCATENATION_KEYWORDS = [r',', r'\sand\s']
 
+    _nltk_data_downloaded = False
+
+    @classmethod
+    def setup_dependencies(cls):
```
NiklasMM: This downloads a fairly large dataset (~200MB unzipped). If I'm understanding this correctly `setup_dependencies` is done for all runs of the bear. Ideally we would only want it to be run when `check_imperative` is selected by the user.

bears/vcs/git/GitCommitBear.py
jayvdb: It emits a bunch of junk in the log, but otherwise does nothing ;-)

jayvdb: Note that as a `GlobalBear`, this bear is typically only run once per `coala` invocation, so this variable is going to be especially useful normally, but ... with tests ... it is rather helpful to reduce spamming the logs.

bears/vcs/git/GitCommitBear.py
jayvdb: I believe the `Bear.__init__` logic is:
```
        self.section = section
        ...
        self.setup_dependencies()
        cp = type(self).check_prerequisites()
```
(I may not be correct, especially for `GlobalBear`s)

That means the `setup_dependencies` is called for each instance.
As a `GlobalBear`, this bear is typically only run once per `coala` invocation, so `setup_dependencies` will typically only be called once.

It seems odd that a class method is used to do the `check_prerequisites`, as that means there can not be instance- (configuration-) specific prerequisites.  And the class method `check_prerequisites` is being invoked for every instance.  It should be in `__new__` , or in the meta-class, as it can't be different each instantiation..?

`self.section` is being set before `setup_dependencies` is invoked , so I might be able to get `check_imperative` from there.  (I did think of this, but was being lazy so I didnt investigate beyond wondering why we have `setup_dependencies` and `check_prerequisites` (and `prerequisite_check_command`: https://github.com/coala/coala/issues/2073), and especially why have them in that order, and wtf for the class method.)

NiklasMM: Seems about right. It would be great if you could investigate this. Otherwise we could check and download the nltk data in `GitCommitBear.check_imperative`

jayvdb: Will do.

jayvdb:
@RohanVB `~/nltk_data` still needs to be in the cached directories, otherwise every CI build will refetch the same 200Mb.
It is the explicit fetch which can be removed as newcomer issues, and this PR does remove them from the CI scripts of this repo.

bears/vcs/git/GitCommitBear.py
```diff
@@ -31,6 +31,22 @@ class GitCommitBear(GlobalBear):
     }
     CONCATENATION_KEYWORDS = [r',', r'\sand\s']
 
+    _nltk_data_downloaded = False
+
+    def _check_imperative_enabled(self):
+        imperative_check = self.section.contents.get(
+            'shortlog_check_imperative', True)
+        return bool(imperative_check)
```
sils: shouldn't `return bool(self.section.get('shortlog_check_imperative', True))` do the job in one line? (Note: .contents shouldn't be needed and we should optimize for readability rather than an unnoticable performance benefit)

Second, the default `True` might change, maybe put this in a class level constant?

jayvdb: The tests would fail if the default changes.

bears/vcs/git/GitCommitBear.py
```diff
@@ -193,25 +209,19 @@ def check_imperative(self, paragraph):
...
-            return
+        words = nltk.word_tokenize(nltk.sent_tokenize(paragraph)[0])
+        # VBZ : Verb, 3rd person singular present, like 'adds', 'writes'
+        #       etc
+        # VBD : Verb, Past tense , like 'added', 'wrote' etc
+        # VBG : Verb, Present participle, like 'adding', 'writing'
+        word, tag = nltk.pos_tag(['I'] + words)[1:2][0]
+        if(tag.startswith('VBZ') or
+           tag.startswith('VBD') or
+           tag.startswith('VBG') or
```
RohanVB: Can we have a tuple with the 3 strings and just have the key as argument?

jayvdb: Not in this commit.  git has a special mode for indentation only changes, so that `git blame` can skip them to find real changes.

bears/vcs/git/GitCommitBear.py
jayvdb: I've removed the `.contents` , which makes this shorter and can be moved back into the `if` condition where it was being used.

jayvdb:
This is blocking https://gitlab.com/coala/package_manager/merge_requests/42

bears/vcs/git/GitCommitBear.py
```diff
@@ -31,6 +31,18 @@ class GitCommitBear(GlobalBear):
     }
     CONCATENATION_KEYWORDS = [r',', r'\sand\s']
 
+    _nltk_data_downloaded = False
```
NiklasMM: This variable is only useful for the tests, right? If that's true, I'd recommend removing it entirely and instead use `mock.patch` to check if `nltk.download` was called in the test.

jayvdb: No, not "only".

The variable is used to prevent the `download()` occurring twice.  `download()` isnt cheap; it fetches the index from the nltk website to determine whether the local cache needs updating.

While it is not likely that this bear will be invoked twice in typical coala usage, gitmate does run the bear on each commit in a PR.  `gitmate` probably invokes the coala command for each commit, rather than re-using the same runtime, but that is an example of how multiple instances of this global bear could be used in the same runtime.

This is essentially a very simple form of memoizing; so simple that even [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) is overkill (and doing a custom implementation is simpler to understand and simpler to test).

But I can still mock the tests, to avoid web fetches altogether, if that is requested.

NiklasMM: No, thats fine. Thanks for explaining. As far as I'm concerned you can leave it.

NiklasMM:
ack 81b8cd5

jayvdb:
reack 882d697

nkprince007:
LGTM 👍 @jayvdb

jayvdb:
@rultor merge

PR #1621, opened 2017-04-19
Add argument for include paths

tupaschoal:
CPPCheck supports include paths on its argument processor and this is fundamental for using it in big projects.

Closes https://github.com/coala/coala-bears/issues/1701

### DIFF
--- a/bears/c_languages/CPPCheckBear.py
+++ b/bears/c_languages/CPPCheckBear.py
@@ -29,17 +29,24 @@ class CPPCheckBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
-                         enable: typed_list(str)=[]):
+                         enable: typed_list(str)=[],
+                         include_paths: str=None):
         """
         :param enable:
             Choose specific issues to report. Issues that can be
             reported are: all, warning, style, performance,
             portability, information, unusedFunction,
             missingInclude
+        :param include_paths:
+            Specify include paths to make sure cppcheck
+            can find all the relevant code.
         """
         args = ('--template={line}:{severity}:{id}:{message}',)
 
         if enable:
             args += ('--enable=' + ','.join(enable),)
+
+        if include_paths:
+            args += include_paths,
 
         return args + (filename,)
--- a/tests/c_languages/CPPCheckBearTest.py
+++ b/tests/c_languages/CPPCheckBearTest.py
@@ -1,3 +1,4 @@
+import os
 from bears.c_languages.CPPCheckBear import CPPCheckBear
 from coalib.testing.LocalBearTestHelper import verify_local_bear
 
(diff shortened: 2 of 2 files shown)


satwikkansal:
@tupaschoal 
- Commit message needs a bit of work. Please see https://coala.io/commit
- Need to add test for your change.

Also it is preferable to create an issue before submitting a PR as it allows for other's views and feedbacks.

bears/c_languages/CPPCheckBear.py
```diff
@@ -42,4 +43,7 @@ def create_arguments(filename, file, config_file,
         if enable:
             args += ('--enable=' + ','.join(enable),)
 
+        if include_paths:
+            args += include_paths,
```
satwikkansal: Add tests for this :)

tupaschoal:
I'll ammend the commit message and am working on the test, thanks for the feedback @satwikkansal

bears/c_languages/CPPCheckBear.py
```diff
@@ -29,7 +29,8 @@ class CPPCheckBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
-                         enable: typed_list(str)=[]):
+                         enable: typed_list(str)=[],
+                         include_paths: str=None):
```
Techievena: Include details about this parameter in the docstring. :)

Techievena:
Same here https://github.com/coala/coala-bears/pull/1622#issuecomment-298809263. Also include the issue reference in both the PRs

bears/c_languages/CPPCheckBear.py
tupaschoal: Done.

bears/c_languages/CPPCheckBear.py
tupaschoal: Done!

jayvdb: This is a little confusing.
I assume that `include_paths` would be `-I` arguments, but here you add the paths without `-I`.

If you are adding directories at the end of the list of arguments, you are running `cppcheck` on all files in that directory, which mean it is a `GlobalBear`, not a `LocalBear` which works on one file at a time.

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -26,10 +27,12 @@
     return 0;
 }"""
 
+test_dir = os.path.join(os.path.dirname(__file__), 'test_files')
 
 CPPCheckBearTest1 = verify_local_bear(CPPCheckBear,
                                       valid_files=(good_file, warn_file),
-                                      invalid_files=(bad_file,))
+                                      invalid_files=(bad_file,),
+                                      settings={'include_paths': test_dir})
```
jayvdb: this test worked before ... and works after you add this argument.  Adding this argument here does not test that the argument is used ; it only tests that it doesnt break something.

We need a real test case where `include_paths` is *necessary*, and if omitted the test case would fail.

tupaschoal:
Not intending to work on this, so I'll close it.

PR #1630, opened 2017-04-20
MarkdownBear: Set max_line_length default to None

jayvdb:
The introduction of max_line_length with a default of 80
was a breaking change as previously line length was not
constrained.

This constraint required the use of remark-lint, which
caused MarkdownBear to remove all content from files
if remark-lint could not be found in NODE_PATH.

The default is now again to not check line length,
and not use remark-lint unless the .coafile explicitly
requests a max_line_length.

Fixes https://github.com/coala/coala-bears/issues/1581

### DIFF
--- a/bears/markdown/MarkdownBear.py
+++ b/bears/markdown/MarkdownBear.py
@@ -65,7 +65,7 @@ class MarkdownBear:
                          horizontal_rule: str='*',
                          horizontal_rule_spaces: bool=False,
                          horizontal_rule_repeat: int=3,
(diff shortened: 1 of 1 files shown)


yash-nisar:
ack 0c334d990d5fb8732d938aaff16c7634e3decd97

yash-nisar:
The commit body is very nicely written :+1:

jayvdb:
reack 20c4782

jayvdb:
@rultor merge

PR #1718, opened 2017-05-08
restLintBear: Ignore unknown errors

SanketDG:
Fixes https://github.com/coala/coala-bears/issues/1716

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/rest/reSTLintBear.py
+++ b/bears/rest/reSTLintBear.py
@@ -1,9 +1,23 @@
+import re
+
 from restructuredtext_lint import lint
 
 from coalib.bears.LocalBear import LocalBear
 from dependency_management.requirements.PipRequirement import PipRequirement
 from coalib.results.Result import Result
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+
+
+def _ignore_unknown(errors):
+    for error in errors:
+        msg = error.full_message
+        res = re.search(
+            r'No directive entry for "[\w|\-]+"|'
+            r'Unknown directive type "[\w|\-]+"|'
+            r'No role entry for "[\w|\-]+"|'
+            r'Unknown interpreted text role "[\w|\-]+"', msg)
+        if not res:
+            yield error
 
 
 class reSTLintBear(LocalBear):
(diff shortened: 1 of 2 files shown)


SanketDG:
needs better test names :laughing:

bears/rest/reSTLintBear.py
```diff
@@ -14,13 +28,17 @@ class reSTLintBear(LocalBear):
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting', 'Syntax'}
 
-    def run(self, filename, file):
+    def run(self, filename, file, ignore_unknown_roles: bool=False):
```
jayvdb: directives are not roles, so either this is `ignore_unknown: bool`, or it is split into `ignore_unknown_roles` and `ignore_unknown_directives`.

Even if we did plan on deprecating this bear, this PR still makes sense, as it will help us implement the bear, and thus understand the bear ;-)

SanketDG: gah I am not sure if we deprecate the bear, it will be worth doing this.

But now that I have seen `rstcheck's` way of doing this, I would also like to have two params `ignore_directives`, `ignore_roles`, which takes in a list of directives/roles to ignore. And then obviously do the stuff with regex and the helper function.

jayvdb: I dont see the benefit of replicating `rstcheck` , unless we know of a feature in `restlint` that is missing from `rstcheck`...?

SanketDG: I am not really sure, there is no feature "unimplemented" in restlint because both uses docutils

However, I feel restlint is the "purer" form of a rst linter, because rstcheck has "batteries" included like sphinx support etc. This is the only argument I have,

That said, I am up for deprecating `restlint` for the sake of bear uniformity.

jayvdb: c.f. https://github.com/coala/coala-bears/issues/1723

Also see https://github.com/coala/coala/issues/3044#issuecomment-261856917 , where I think I saw extra error messages from restlint than rstcheck emitted .

I think you might be right about the need for a purer rst linter.  https://pypi.python.org/pypi/restructuredtext_lint reminds me ... we need a rst-checker which validates README.rst so that it can be used for PyPI, which has stricter rules than Sphinx, and stricter rules than GitHub's RST parser.

*Maybe* we can request rstcheck has a strict rst mode.  Oddly enough, [this coala project](http://projects.coala.io/#/projects?project=improve_sphinx_support_in_rstcheck) seemed like it might have created a more clear separation between docutils mode and sphinx mode.

And if we need a pure rst linter, then we dont want to allow it to validate sphinx rst.  It should fail! ;-)
And we should implement restLintBear on any README.rst or other file which is included in the packaging for PyPI, and not other RST files.

SanketDG: So close this PR then seems the right way?

SanketDG:
@jayvdb ping.

PR #1738, opened 2017-05-16
Pycodestyle

shreyans800755:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/python/PycodestyleBear.py
+++ b/bears/python/PycodestyleBear.py
@@ -21,7 +21,8 @@ class PycodestyleBear:
     @staticmethod
     def create_arguments(
             filename, file, config_file,
-            pycodestyle_ignore: str='',
+            pycodestyle_ignore: str='E121, E123, E126, E133, E226, E241, '
+                                    'E242, E704, W503',
             pycodestyle_select: str='',
             max_line_length: int=79):
         """
--- a/tests/python/PycodestyleBearTest.py
+++ b/tests/python/PycodestyleBearTest.py
@@ -22,6 +22,12 @@ PycodestyleBearTest = verify_local_bear(
     valid_files=(good_file,),
     invalid_files=(bad_file,))
 
+PycodestyleBearNoIgnoreTest = verify_local_bear(
+    PycodestyleBear,
+    valid_files=(good_file,),
+    invalid_files=(bad_file,),
+    settings={'pycodestyle_ignore': ''})
+
 PycodestyleBearConfigIgnoreTest = verify_local_bear(
     PycodestyleBear,
     valid_files=(good_file, bad_file),


bears/python/PycodestyleBear.py
```diff
@@ -37,8 +37,10 @@ def create_arguments(
         """
         arguments = [r'--format=%(row)d %(col)d %(code)s %(text)s']
 
+        ignore_values = 'E121, E123, E126, E133, E226, E241, E242, E704, W503'
```
Makman2: makes those rather the default above:
```
pycodestyle_ignore: str='E121, E123, ...'
```
then you don't have to change any code and the user can enable those implicitly disabled defaults again. Otherwise we would restrict people :+1:

shreyans800755: You mean to put it in parameters of create_arguments ?
If I understood it right, putting default values in param itself will not resolve the issue. Because we need to append the default values to param values, and using param only will replace it.

Makman2: Yes :)
The user can see now which values are default and can copy them to his coafile if he wants to keep them.

Makman2: You would do the same for your other PR modifying our coafile, you copy those defaults and append E303 (I think it was E303... anyway) for the `pycodestyle_ignore` setting.

SanketDG: I feel like this calls for a `+=` operator for `.coafile`. ;)

tests/python/PycodestyleBearTest.py
```diff
@@ -22,6 +22,12 @@ def hello():
     valid_files=(good_file,),
     invalid_files=(bad_file,))
 
+PycodestyleBearTest = verify_local_bear(
```
Makman2: You have to name the test differently, otherwise it overrides the test from before and that's not executed then^^

shreyans800755: Uh okay. I'll fix it.

shreyans800755: Or I can remove the first one. New serves both the purpose.

Makman2: Nah keep it. Actually those tests do different things, one relies on the tool-defaults, one relies on the coafile defaults.

SanketDG:
Given that the user has to copy the whole thing, we should definitely do the setting as a list, and then implement a `+=` operator in the coafile

shreyans800755:
@SanketDG We can add new param for that. Changing this param doesn't make sense. It'll enforce user to ignore all the tags present in default list. Should I create an issue for that, or fix it over here itself ?

Makman2:
@SanketDG already filed ;) --> https://github.com/coala/coala-bears/issues/1739

Makman2:
ack 98ca7cf

shreyans800755:
> @SanketDG already filed ;) --> #1739

@Makman2 That's good.

Rebased.

Makman2:
ack dd85632

Makman2:
@rultor merge

PR #1758, opened 2017-05-23
Add FutureBear for running futurize on code

lphk92:
Closes https://github.com/coala/coala-bears/issues/1606

This bear will run futurize on code and present the user with
options to replace code with the equivalent Python2/3 compatible
statements.

NOTE: This does not working properly right now due to a bug
in coalib! The unified-diff output type is not preserving
new line characters in the corrected output.

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/python/FutureBear.py
+++ b/bears/python/FutureBear.py
@@ -1,0 +1,29 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.PipRequirement import PipRequirement
+
+
+@linter(executable='futurize',
+        output_format='unified-diff',
+        result_message='The code could be futurized.',
+        diff_distance=0,
+        use_stdout=True,
+        use_stderr=False)
+class FutureBear:
+    LANGUAGES = {'Python', 'Python 2'}
+    REQUIREMENTS = {PipRequirement('future', '0.16.0')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         futurize_stage: int = 0):
+        """
+        :param futurize_stage: The desired stage of futurization (0, 1, or 2).
+                               Please refer to the futurize documentation for
+                               more information on the meaning of the
+                               different stages.
+        """
+        stage = '-0'
+        if futurize_stage in (0, 1, 2):
+            stage = '-{}'.format(futurize_stage)
+        return (stage, filename)
--- a/tests/python/FutureBearTest.py
+++ b/tests/python/FutureBearTest.py
@@ -1,0 +1,27 @@
+from queue import Queue
+from shutil import which
+from unittest.case import skipIf
+
+from bears.python.FutureBear import FutureBear
+from coalib.settings.Section import Section
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+
+
+@skipIf(which('futurize') is None, 'futurize is not installed')
+class FutureBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('test section')
+        self.uut = FutureBear(self.section, Queue())
+
+        self.test_lines = [
+            "print 'Hello World!'\n\n",
+            "d = {1: 'a', 2: 'b', 3: 'c'}\n\n",
+            'for k, v in d.iteritems():\n',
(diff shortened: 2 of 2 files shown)


lphk92:
I could use some help from a more experienced member of the team as to why the tests are failing for this one. Any pointers or ideas what is causing the issue?

underyx:
First, I wonder if closing https://github.com/coala/coala/issues/3834 fixed your diff application problem somehow.

Second, @lphk92 I think the test failures are still the same as we had at the sprint: `pytest` sees the `.py` file so it tries to run it, but it has a Python 3 syntax error when it sees your Python 2 print statement, which breaks the testing process. I'd suggest testing only changes that are Python 3 syntax compatible (like `dict.iteritems() -> dict.items()`.)

jayvdb:
> I think the test failures are still the same as we had at the sprint: pytest sees the .py file so it tries to run it,

pytest can be told what files to load, and what filenames to ignore.  See our setup.cfg.

Try renaming "future_test.py" to "python_2_print.py"

tests/python/FutureBearTest.py
```diff
@@ -0,0 +1,30 @@
...
+class FutureBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('test section')
+        self.uut = FutureBear(self.section, Queue())
+
+        self.test_lines = [
+            "print 'Hello World!'\n\n",
+            "d = {1: 'a', 2: 'b', 3: 'c'}\n\n",
+            "for k, v in d.iteritems():\n",
```
jayvdb: This review request by gitmate is not perfect, but our tool (QuotesBear) isnt smart enough to know why double quotes should be used here.
Would you kindly switch to single quotes here, so the build passes, and we can do a human code review.

jayvdb:
A few minor fixups according to gitmate, and could you rebase please.

bears/python/FutureBear.py
```diff
@@ -0,0 +1,29 @@
...
+    REQUIREMENTS = {PipRequirement('future', '0.16.0')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+
+    @staticmethod
+    def create_arguments(filename, file, config_file,
+                         futurize_stage: int = 0):
+        """
+        :param futurize_stage: The desired stage of futurization (0, 1, or 2).
+                               Please refer to the futurize documentation for
```
sils: link please

bears/python/FutureBear.py
```diff
@@ -0,0 +1,29 @@
...
+from dependency_management.requirements.PipRequirement import PipRequirement
+
+
+@linter(executable='futurize',
+        output_format='unified-diff',
+        result_message='The code could be futurized.',
+        diff_distance=0,
+        use_stdout=True,
+        use_stderr=False)
+class FutureBear:
```
sils: there's no documentationon what this actually does, how are our users supposed to know that without googling?

sils:
> NOTE: This does not working properly right now due to a bug in coalib! The unified-diff output type is not preserving new line characters in the corrected output.

Uhm, is there an issue filed about that somewhere?

bears/python/FutureBear.py
```diff
@@ -0,0 +1,29 @@
...
+    def create_arguments(filename, file, config_file,
+                         futurize_stage: int = 0):
+        """
+        :param futurize_stage: The desired stage of futurization (0, 1, or 2).
+                               Please refer to the futurize documentation for
+                               more information on the meaning of the
+                               different stages.
+        """
+        stage = '-0'
+        if futurize_stage in (0, 1, 2):
```
sils: I think instead of lines 26 and 27 we should rather do an assert on the stages to be in the range

tests/python/FutureBearTest.py
```diff
@@ -0,0 +1,27 @@
...
+            "print 'Hello World!'\n\n",
+            "d = {1: 'a', 2: 'b', 3: 'c'}\n\n",
+            'for k, v in d.iteritems():\n',
+            "    print '{}: {}'.format(k, v)\n"
+        ]
+
+    def test_base(self):
+        self.check_validity(self.uut,
+                            self.test_lines,
+                            valid=False)
```
sils: it'd be good to have a testcase that actually tests the patch or result message I think.

Also we should have one test for a file that passes

PR #1777, opened 2017-05-26
.ci/deps.sh: Fix dartsdk caching

jayvdb:
`-v` always fails as that is not a valid command line argument.
Use `--version` instead.

Fixes https://github.com/coala/coala-bears/issues/1776

### DIFF
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -24,7 +24,7 @@ go get -u github.com/BurntSushi/toml/cmd/tomlv
 bundle install --path=vendor/bundle --binstubs=vendor/bin --jobs=8 --retry=3
 
 # Dart Lint commands
-if ! dartanalyzer -v &> /dev/null ; then
+if [ ! -e ~/dart-sdk/lib ]; then
   wget -nc -O ~/dart-sdk.zip https://storage.googleapis.com/dart-archive/channels/stable/release/1.14.2/sdk/dartsdk-linux-x64-release.zip
   unzip -n ~/dart-sdk.zip -d ~/
 fi
--- a/.ci/deps.sh
+++ b/.ci/deps.sh
@@ -29,6 +29,7 @@ if [ ! -e ~/dart-sdk/lib ]; then
   unzip -n ~/dart-sdk.zip -d ~/
 fi
 
+
 # VHDL Bakalint Installation
 if [ ! -e ~/bakalint-0.4.0/bakalint.pl ]; then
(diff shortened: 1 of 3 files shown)


underyx:
`darksdk` in commit message should be `dartsdk`

underyx:
Any idea why tests fail now?

jayvdb:
My guess is the caching of only `dartsdk/bin` is the problem.

jayvdb:
So now the question is, does having a larger cache perform better than always doing a fetch from a reliable host and an unzip.

https://github.com/coala/coala-bears/pull/1769 will make it easier to answer that question, but I guess @li3n3 might know the answer.

The usual advice from CI providers seems to be to not cache things like this.

jayvdb:
Always hard to prove anything with CI performance, without doing a lot of silly identical test runs, but ..

- [no dart caching](https://circleci.com/gh/jayvdb/coala-bears/742) - 35secs to restore cache -  24secs to do `deps.sh`
- [fixed dart caching](https://circleci.com/gh/jayvdb/coala-bears/739) - 50secs to restore cache - 29secs to do `deps.sh`

So based on one run, we should be removing the dart caching.
But of course, it is illogical that `deps.sh` would take longer when it has less to do.

jayvdb:
Rebased https://circleci.com/gh/jayvdb/coala-bears/772 and restoring the cache with dartsdk is 50s again,
and restoring cache without dartsdk https://circleci.com/gh/jayvdb/coala-bears/775 is 39s .

This time, the cached sdk `deps.sh` is 32s and the uncached sdk `deps.sh` is 34s.

[Adding timing](https://circleci.com/gh/jayvdb/coala-bears/777) to uncached, the wget is 0.63s (real), and the unzip is 0.34s.  By [using `wget -q`](https://circleci.com/gh/jayvdb/coala-bears/778) its time goes down to 0.469s.  i.e. installing the uncached dart sdk costs is easily less than 1s operation, assuming no network problems.

But https://circleci.com/gh/jayvdb/coala-bears/777 also has a 50s cache restore step, so it is still possible to have higher cache restore times without dart sdk in the cache.  And https://circleci.com/gh/coala/coala-bears/6625 had a 36s cache restore, so low cache restores are possible with the dart sdk bin in the cache.

sils:
how does https://github.com/coala/coala-bears/pull/1777/commits/c16111bc5fb8b8e2b23e34e60d61aa13004ec283 force an uncached build? :confused:

sils:
Given the timings it seems it's pretty irrelevant if we have caching for this or not, right? I'm ok with either of those, caching is kind of a complexity but also helps consistency just caching everything.

jayvdb:
Irrelevant due to https://github.com/coala/coala-bears/pull/2910

PR #1787, opened 2017-05-30
VintBear: Deactivate bear

jayvdb:
The bear is broken, and it breaks all coala, due to
its requirement chardet~=2.3.0 which is incompatible
with requests adding requirement chardet~=3.0.2.

Related to https://github.com/coala/coala-bears/issues/1785

### DIFF
--- a/bears/vimscript/VintBear.py
+++ b/bears/vimscript/VintBear.py
@@ -1,5 +1,6 @@
 from coalib.bearlib.abstractions.Linter import linter
-from dependency_management.requirements.PipRequirement import PipRequirement
+from dependency_management.requirements.PackageRequirement import (
+    PackageRequirement)
 
 
 @linter(executable='vint', output_format='regex',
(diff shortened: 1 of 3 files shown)


bears/vimscript/VintBear.py
```diff
@@ -12,7 +13,7 @@ class VintBear:
     """
 
     LANGUAGES = {'VimScript'}
-    REQUIREMENTS = {PipRequirement('vim-vint', '0.3.10')}
+    REQUIREMENTS = {PackageRequirement('broken', 'vim-vint', '0.3.10')}
```
meetmangukiya: what does broken indicate as a package requirement :/

meetmangukiya:
ack 29a9de8

bears/vimscript/VintBear.py
jayvdb: Nothing.
It is just a value for the mandatory `type`, and this is a bad hack which can be easily backported.

sils: can't we just comment the requirements with a comment eplaining the situatoin?

sils: though I assume you want the is_installed to fail so the hack is better for now until we make a dependency class that is never met

sils:
@rultor merge

PR #1796, opened 2017-06-01
StylintBear: Implement settings

yash-nisar:
Use the implemented settings to generate a `config_file` and use
that file if the user does not explicitly mention `stylint_config`.

Closes https://github.com/coala/coala-bears/issues/1795

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/stylus/StylintBear.py
+++ b/bears/stylus/StylintBear.py
@@ -1,5 +1,11 @@
+import json
+
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.NpmRequirement import NpmRequirement
+
+_setting_map = {True: 'always',
+                False: 'never',
+                None: 'false'}
 
 
 @linter(executable='stylint',
@@ -36,12 +42,250 @@ class StylintBear:
     SEE_MORE = 'https://github.com/SimenB/stylint'
 
     @staticmethod
+    def generate_config(filename, file,
+                        block_keyword: bool=None,
+                        brackets: bool=False,
+                        colons_for_property_declaration: bool=True,
+                        color_variables_for_hex_values: bool=True,
+                        spaces_after_commas: bool=True,
+                        spaces_after_comments: bool=True,
+                        allow_trailing_whitespace: bool=False,
+                        no_css_literals: bool=False,
+                        max_selector_depth: bool=False,
+                        check_duplicates: bool=True,
+                        efficient_properties: bool=True,
+                        extend_preference: str=None,
+                        indent_size: int=0,
+                        leading_zero: bool=None,
+                        max_errors: int=0,
+                        max_warnings: int=0,
+                        mixed_spaces_and_tabs: bool=False,
+                        variable_naming_convention: str=None,
+                        strict_naming_convention: bool=False,
+                        none_keyword: bool=False,
+                        check_no_important_keyword: bool=True,
+                        spaces_inside_parentheses: bool=None,
+                        placeholder: bool=True,
+                        prefix_vars_with_dollar: bool=True,
+                        semicolons: bool=False,
+                        sort_order: str='alphabetical',
+                        stacked_properties: bool=False,
+                        check_property_validity: bool=True,
+                        preferred_quotation: str=None,
+                        zero_units: bool=False,
+                        z_index_normalize_base: int=0,
+                        stylint_config: str=''):
+        """
+        :param block_keyword:
+            When ``True`` expect the ``@block`` keyword when defining block
+            variables. When ``False``, expect no ``@block`` keyword when
+            defining block variables.
+        :param brackets:
+            When ``True``, expect ``{}`` when declaring a selector. When
+            ``False``, expect no brackets when declaring a selector.
+        :param colons_for_property_declaration:
+            When ``True``, expect ``:`` when declaring a property. When
+            ``False``, expect no ``:`` when declaring a property.
+        :param color_variables_for_hex_values:
+            When ``True``, enforce variables when defining hex values.
+        :param spaces_after_commas:
+            Enforce or disallow spaces after commas.
+        :param spaces_after_comments:
+            Enforce or disallow spaces after line comments.
+            For example: If set to ``True``, prefer
+            ``// comment`` over ``//comment``.
+        :param allow_trailing_whitespace:
+            If ``True``, ignores trailing whitespace. If ``False``, trailing
+            whitespace will throw a warning.
+        :param no_css_literals:
+            By default Stylint ignores ``@css`` blocks. If set to ``True``
+            however, warnings are thrown if ``@css`` is used.
+        :param max_selector_depth:
+            Set the max selector depth. If set to 4, max selector depth will
+            be 4 indents. Pseudo selectors like ``&:first-child`` or
+            ``&:hover`` won't count towards the limit.
+        :param check_duplicates:
+            Checks if selectors or properties are duplicated unnecessarily.
+        :param efficient_properties:
+            Check for places where properties can be written more efficiently.
+        :param extend_preference:
+            Pass in either ``@extend`` or ``@extends`` and then enforce that.
+            Both are valid in Stylus. It doesn't really matter which one
+            you use.
+            For example: If set to ``@extends``, prefer
+            ``@extends $some-var`` instead of ``@extend $some-var``
+            or if set to ``@extend``, prefer
+            ``@extend $some-var`` over ``@extends $some-var``.
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``0``.
+            For example: If set to ``2``, prefer
+            ``/s/smargin: 0`` over ``/s/s/smargin: 0``.
+        :param leading_zero:
+            When ``True``, prefer leading zeroes on decimal points. When
+            ``False``, disallow leading zeroes on decimal points.
+        :param max_errors:
+            Set maximum number of errors. If ``0``, all errors will
+            be shown without a limit.
+        :param max_warnings:
+            Set maximum number of warnings. If ``0``, all errors will
+            be shown without a limit.
+        :param mixed_spaces_and_tabs:
+            If a non-negative number is passed to ``indent_size``,
+            soft tabs (i.e. spaces) are assumed, and if
+            ``0`` is passed to ``indent_size``, hard tabs are assumed.
+            For example: If ``indent_size = 4`` and
+            ``mixed_spaces_and_tabs = True``, prefer
+            ``/s/s/s/smargin 0`` over ``/tmargin 0``
+            or if ``indent_size = 0`` and
+            ``mixed_spaces_and_tabs = True``, prefer
+            ``/tmargin 0`` over ``/s/s/s/smargin 0``.
+        :param variable_naming_convention:
+            Enforce a particular naming convention when declaring variables.
+            Throws a warning if you don't follow the convention.
+            Supported values are ``lowercase-dash``, ``lowercase_underscore``,
+            ``camelCase`` and ``BEM``.
+        :param strict_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``strict_naming_convention`` is set to ``True``,
+            ``variable_naming_convention`` will also look at class and
+            ID names.
+        :param none_keyword:
+            If ``True`` check for places where ``none`` used instead of ``0``.
+            If ``False`` check for places where ``0`` could be used
+            instead of ``none``.
+        :param check_no_important_keyword:
+            If ``True``, show warning when ``!important`` is found.
+            For example: If set to ``True``, this will throw a warning::
+
+                div
+                    color red !important
+
+        :param spaces_inside_parentheses:
+            Enforce or disallow use of extra spaces inside parentheses.
+            For example: If set to ``True``, prefer
+            ``my-mixin( $myParam )`` over ``my-mixin($myParam)``
+            or if set to ``False``, prefer
+            ``my-mixin($myParam)`` over ``my-mixin( $myParam )``.
+        :param placeholder:
+            Enforce extending placeholder vars when using ``@extend(s)``.
+        :param prefix_vars_with_dollar:
+            Enforce use of ``$`` when defining a variable. In Stylus, using a
+            ``$`` when defining a variable is optional, but is a good idea
+            if you want to prevent ambiguity. Not including the ``$`` sets up
+            situations where you wonder: "Is this a variable or a value?"
+            For instance: ``padding $default`` is easier to understand than
+            ``padding default``.
+            For example: If set to ``True``, prefer
+            ``$my-var = 0`` over ``my-var = 0``
+            or if set to ``False``, prefer
+            ``my-var = 0`` over ``$my-var = 0``.
+        :param semicolons:
+            When ``True``, enforce semicolons. When ``False``, disallow
+            semicolons.
+        :param sort_order:
+            Enforce a particular sort order when declaring properties. Throws
+            a warning if you don't follow the order.
+            For example: If set to ``alphabetical``, prefer this::
+
+                .some-class
+                    display block
+                    float left
+                    position absolute
+                    right 10px
+                    top 0
+
+            over this::
+
+                .some-class
+                    position absolute
+                    top 0
+                    right 10px
+                    display block
+                    float left
+
+            Supported values are ``alphabetical`` and ``grouped``.
+        :param stacked_properties:
+            No one-liners. Enforce putting properties on new lines.
+            For example: If set to ``False``, prefer::
+
+                .className
+                    padding 0
+
+            over
+            ``.className { padding 0 }``
+        :param check_property_validity:
+            Check that a property is valid CSS or HTML.
+        :param preferred_quotation:
+            Your preferred quotation character, e.g. ``double`` or ``single``.
+        :param zero_units:
+            Looks for instances of ``0px``. You don't need the ``px``.
+            Checks all units, not just ``px``.
+        :param z_index_normalize_base:
+            Enforce some basic z-index sanity. Any number passed in
+            will be used as the base for your z-index values.
+            Use ``0`` to disable this feature.
+            For example: If set to ``5``, prefer
+            ``z-index 10`` over ``z-index 9``
+            or if set to ``10``, prefer
+            ``z-index 20`` over ``z-index 15``.
+        """
+        if stylint_config:
+            return None
+        else:
+            options = {
+                'blocks': _setting_map[block_keyword],
+                'brackets': _setting_map[brackets],
+                'colons': _setting_map[colons_for_property_declaration],
+                'colors': _setting_map[color_variables_for_hex_values],
+                'commaSpace': _setting_map[spaces_after_commas],
+                'commentSpace': _setting_map[spaces_after_comments],
+                'cssLiteral': _setting_map[no_css_literals],
+                'depthLimit': max_selector_depth,
+                'duplicates': check_duplicates,
+                'efficient': _setting_map[efficient_properties],
+                'extendPref': extend_preference,
+                'indentPref': indent_size,
+                'leadingZero': _setting_map[leading_zero],
+                'maxErrors': max_errors,
+                'maxWarnings': max_warnings,
+                'mixed': mixed_spaces_and_tabs,
+                'namingConvention': variable_naming_convention,
+                'namingConventionStrict': strict_naming_convention,
+                'none': _setting_map[none_keyword],
+                'noImportant': check_no_important_keyword,
+                'parenSpace': _setting_map[spaces_inside_parentheses],
+                'placeholders': _setting_map[placeholder],
+                'prefixVarsWithDollar': _setting_map[prefix_vars_with_dollar],
+                'quotePref': preferred_quotation,
+                'semicolons': _setting_map[semicolons],
+                'sortOrder': sort_order,
+                'stackedProperties': _setting_map[stacked_properties],
+                'trailingWhitespace':
+                    False if allow_trailing_whitespace is True else 'never',
+                'valid': check_property_validity,
+                'zeroUnits': _setting_map[zero_units],
+                'zIndexNormalize':
+                    False if z_index_normalize_base == 0
+                    else z_index_normalize_base,
+                'groupOutputByFile': True,
+                'reporterOptions': {
+                    'columns': ['lineData', 'severity', 'description', 'rule'],
+                    'columnSplitter': '  ',
+                    'showHeaders': False,
+                    'truncate': True
+                }
+            }
+
+            return json.dumps(options)
+
+    @staticmethod
     def create_arguments(filename, file, config_file, stylint_config: str=''):
         """
         :param stylint_config:
             The location of the ``.stylintrc`` config file.
         """
-        if stylint_config:
-            return '--config', stylint_config, filename
-        else:
-            return filename,
+        return ('--config',
+                stylint_config if stylint_config else config_file,
+                filename)
--- a/tests/stylus/stylint_test_files/test_bad_placeholder_space_color.styl
+++ b/tests/stylus/stylint_test_files/test_bad_placeholder_space_color.styl
@@ -1,0 +1,9 @@
+.message
+    padding: 10px
+body
+  @extend .message
+  color: $var_name
+  color #f00
+  font: 12px Helvetica, Arial, sans-serif
+  //comment
+  rgba(0,0,0,.18)
--- a/tests/stylus/StylintBearTest.py
+++ b/tests/stylus/StylintBearTest.py
@@ -145,6 +145,42 @@ class StylintBearTest(LocalBearTestHelper):
             filename=get_testfile_path(filename),
             settings={'stylint_config': get_testfile_path('.stylintrc')})
 
+    def test_bad_placeholder_space_color(self):
+        filename = 'test_bad_placeholder_space_color.styl'
+        file_contents = load_testfile(filename)
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('StylintBear',
+                                message='always use a placeholder variable '
+                                        'when extending',
+                                file=get_testfile_path(filename),
+                                line=4,
+                                severity=RESULT_SEVERITY.NORMAL),
+             Result.from_values('StylintBear',
+                                message='hexidecimal color should '
+                                        'be a variable',
+                                file=get_testfile_path(filename),
+                                line=6,
+                                column=8,
+                                severity=RESULT_SEVERITY.NORMAL),
+             Result.from_values('StylintBear',
+                                message='line comments require a space '
+                                        'after //',
+                                file=get_testfile_path(filename),
+                                line=8,
+                                column=2,
+                                severity=RESULT_SEVERITY.NORMAL),
+             Result.from_values('StylintBear',
+                                message='commas must be followed '
+                                        'by a space for readability',
+                                file=get_testfile_path(filename),
+                                line=9,
+                                column=6,
+                                severity=RESULT_SEVERITY.NORMAL),
+             ],
+            filename=get_testfile_path(filename))
+
     def test_valid_file(self):
         filename = 'test_valid_file.styl'
         file_contents = load_testfile(filename)


yash-nisar:
CC @Makman2

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,210 @@
...
+            'groupOutputByFile': True}
+
+        options['reporterOptions'] = {
+            'columns': ['lineData', 'severity', 'description', 'rule'],
+            'columnSplitter': '  ',
+            'showHeaders': False,
+            'truncate': True
+        }
+
+        return json.dumps(options)
```
Makman2: please check here if `stylint_config` is set, so you can avoid creating a temporary file in that case.

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,210 @@
...
+    @staticmethod
+    def create_arguments(filename, file, config_file, stylint_config: str=''):
+        """
+        :param stylint_config:
+            The location of the ``.stylintrc`` config file.
+        """
+        if stylint_config:
+            return '--config', stylint_config, filename
+        else:
+            return '--config', config_file, filename
```
Makman2: can be made shorter:
```
return '--config', stylint_config or config_file , filename
```

Makman2: or if you wanna make it more verbose ;)
```
return '--config', stylint_config if stylint_config else config_file, filename
```

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,210 @@
...
+            'zeroUnits': allow_zero_units,
+            'zIndexNormalize': z_index_normalize,
+            'groupOutputByFile': True}
+
+        options['reporterOptions'] = {
+            'columns': ['lineData', 'severity', 'description', 'rule'],
+            'columnSplitter': '  ',
+            'showHeaders': False,
+            'truncate': True
+        }
```
Makman2: wanna inline that directly into the dict above?
```
options = {
    '...': ...,
    'reporterOptions': {
        'columns': [...]
        ...
```
I think that makes it easier to understand, as there's more data to read instead of code ;)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

yash-nisar:
CC @Makman2 , blocked until we get the StylintBear merged.

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,219 @@
...
+                        allow_spaces_inside_parens: bool = False,
+                        allow_placeholder: str = 'always',
+                        prefix_vars_with_dollar: str = 'always',
+                        allow_semicolons: str = 'never',
+                        sort_order: str = 'alphabetical',
+                        allow_stacked_properties: str = 'never',
+                        check_valid_property: bool = True,
+                        allow_zero_units: str = 'never',
+                        z_index_normalize: bool = False,
+                        stylint_config: str = ''):
```
Makman2: `styling_config: str=''` (no space around `=`)
This is PEP8, so we should follow.
Same applies for all other params above (sorry :3)

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,219 @@
...
+            Looks for instances of ``0px``. You don't need the px. Checks all
+            units, not just px.
+        :param z_index_normalize:
+            Enforce some (very) basic z-index sanity. Any number passed in
+            will be used as the base for your z-index values. Throws an error
+            if your value is not normalized.
+        :param stylint_config:
+            The location of the ``.stylintrc`` config file.
+        """
+        if not stylint_config:
```
Makman2: rather invert the if ;)
```
if styling_config:
    return None
else:
    # make json dict
```

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,219 @@
...
+            Check that a property is valid CSS or HTML.
+        :param allow_zero_units:
+            Looks for instances of ``0px``. You don't need the px. Checks all
+            units, not just px.
+        :param z_index_normalize:
+            Enforce some (very) basic z-index sanity. Any number passed in
+            will be used as the base for your z-index values. Throws an error
+            if your value is not normalized.
+        :param stylint_config:
+            The location of the ``.stylintrc`` config file.
```
Makman2: you don't have to document `stylint_config` twice. Hm though I think you have to remove the documentation from `create_arguments` so it appears properly. There's something written in the linter docs on how to document params^^

bears/stylus/StylintBear.py
```diff
@@ -0,0 +1,219 @@
...
+            return None
+
+    @staticmethod
+    def create_arguments(filename, file, config_file, stylint_config: str=''):
+        """
+        :param stylint_config:
+            The location of the ``.stylintrc`` config file.
+        """
+        return ('--config', stylint_config if stylint_config
+                else config_file, filename)
```
Makman2: maybe like this? easier to read imo, but up to you:
```
return ('--config',
        stylint_config if stylint_config else config_file,
        filename)
```

bears/stylus/StylintBear.py
yash-nisar: In the JSHintBear, it is documented in `create_arguments`..

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Removed it from `create_arguments` though.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

yash-nisar:
Updated @Makman2

bears/stylus/StylintBear.py
Makman2: > Documentation:
    Bear description shall be provided at class level.
    If you document your additional parameters inside ``create_arguments``,
    ``generate_config`` and ``process_output``, beware that conflicting
    documentation between them may be overridden. Document duplicated
    parameters inside ``create_arguments`` first, then in ``generate_config``
    and after that inside ``process_output``.

So I was actually wrong, it should be only in `create_arguments`^^

See http://api.coala.io/en/latest/_modules/coalib/bearlib/abstractions/Linter.html

yash-nisar:
Updated @Makman2

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,181 @@ class StylintBear:
...
         """
         :param stylint_config:
             The location of the ``.stylintrc`` config file.
         """
-        if stylint_config:
-            return '--config', stylint_config, filename
-        else:
-            return filename,
+        return ('--config',
+                stylint_config if stylint_config else config_file, filename)
```
Makman2: up to you, but I think having each item on a single row is easier to read :)
```
('--config',
 stylint_config if stylint_config else config_file,
 filename)
```

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+                        set_naming_convention: bool=False,
+                        set_strict_naming_convention: bool=False,
+                        allow_none_keyword: str='never',
+                        check_no_important_keyword: bool=True,
+                        allow_spaces_inside_parens: bool=False,
+                        allow_placeholder: str='always',
+                        prefix_vars_with_dollar: str='always',
+                        allow_semicolons: str='never',
+                        sort_order: str='alphabetical',
+                        allow_stacked_properties: str='never',
```
Makman2: could you make all those `always` and `never` settings to bools? It seems that there are no more values^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param allow_efficient_properties:
+            Check for places where properties can be written more efficiently.
+        :param set_extend_preference:
+            Pass in either ``@extend`` or ``@extends`` and then enforce that.
+            Both are valid in Stylus. It doesn't really matter which one
+            you use.
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``false``.
```
Makman2: but it's bool, how can we provide a number here?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param allow_leading_zero:
+            Enforce or disallow unnecessary leading zeroes on decimal points.
+        :param set_max_errors:
+            Set 'max' number of Errors.
+        :param set_max_warnings:
+            Set 'max' number of Warnings.
+        :param mixed_spaces_and_tabs:
+            Returns ``true`` if mixed spaces and tabs are found. If a number is
+            passed to indent_size, it assumes soft tabs (ie, spaces)
+            , and if ``false`` is passed to indentPref it assumes hard tabs.
```
Makman2: no sole commas, put them into the previous line.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param set_naming_convention:
+            Enforce a particular naming convention when declaring classes, ids,
+            and variables. Throws a warning if you don't follow the convention.
+        :param set_strict_naming_convention:
```
Makman2: don't prefix settings with `set`. Those are no getters and setters, they behave like properties. That makes the `set_` redundant and also irritating^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            By default, ``set_naming_convention`` only looks at variable names.
+            If ``set_strict_naming_convention`` is set to true,
+            ``set_naming_convention`` will  also look at class and id names.
```
Makman2: rename the setting to resemble this better. I wouldn't have thought that the meaning of "strict" means to also check classes.
Can we have something like `variable_naming_convention` and `class_naming_convention`.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            Set 'max' number of Errors.
+        :param set_max_warnings:
+            Set 'max' number of Warnings.
+        :param mixed_spaces_and_tabs:
+            Returns ``true`` if mixed spaces and tabs are found. If a number is
+            passed to indent_size, it assumes soft tabs (ie, spaces)
+            , and if ``false`` is passed to indentPref it assumes hard tabs.
+        :param set_naming_convention:
+            Enforce a particular naming convention when declaring classes, ids,
+            and variables. Throws a warning if you don't follow the convention.
```
Makman2: hm that doesn't really match what's written for `set_strict_naming_convention`. It say that class-naming conventions are only checked when `set_strict_naming_convention` is enabled.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            By default, ``set_naming_convention`` only looks at variable names.
+            If ``set_strict_naming_convention`` is set to true,
+            ``set_naming_convention`` will  also look at class and id names.
+        :param allow_none_keyword:
+            If ``always`` check for places where ``none`` used instead of 0. If
+            ``never`` check for places where 0 could be used instead of none.
+        :param check_no_important_keyword:
+            If ``true``, show warning when ``!important`` is found.
+        :param allow_spaces_inside_parens:
+            Enforce or disallow use of extra spaces inside parens.
```
Makman2: `parens`: no typo? or is `parents` meant?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param allow_placeholder:
+            Enforce extending placeholder vars when using ``@extend(s)``.
+        :param prefix_vars_with_dollar:
+            Enforce use of ``$`` when defining a variable. In Stylus using a
+            ``$`` when defining a variable is optional, but is a good idea
+            if you want to prevent ambiguity. Not including the ``$`` sets up
+            situations where you wonder: "Is this a variable or a value?"
+            For instance: padding ``$default`` is easier to understand than
+            padding ``default``.
```
Makman2: what's `padding`? :)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param allow_semicolons:
+            Enforce or disallow semicolons.
```
Makman2: And if we don't care? Should we enforce semicolons or disallow them completely?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param sort_order:
+            Enforce a particular sort order when declaring properties. Throws
+            a warning if you don't follow the order. If set to ``false``, allow
+            any order.
```
Makman2: please document all possible values.

Makman2: and maybe instead of `false` let's take `none` or something like that. A "false sort order" is something different for me than "no false order"

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param allow_stacked_properties:
+            No one-liners. Enforce putting properties on new lines.
+        :param check_valid_property:
+            Check that a property is valid CSS or HTML.
+        :param allow_zero_units:
+            Looks for instances of ``0px``. You don't need the px. Checks all
```
Makman2: px in double backticks

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            units, not just px.
```
Makman2: here too: px in double backticks

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param z_index_normalize:
+            Enforce some (very) basic z-index sanity. Any number passed in
```
Makman2: (very) is redundant.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            will be used as the base for your z-index values. Throws an error
+            if your value is not normalized.
```
Makman2: imo there's only one normalized number: 1

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            For instance: padding ``$default`` is easier to understand than
+            padding ``default``.
+        :param allow_semicolons:
+            Enforce or disallow semicolons.
+        :param sort_order:
+            Enforce a particular sort order when declaring properties. Throws
+            a warning if you don't follow the order. If set to ``false``, allow
+            any order.
+        :param allow_stacked_properties:
+            No one-liners. Enforce putting properties on new lines.
```
Makman2: "allow" or "enforce"? And what are stacked properties in this context?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``false``.
+        :param allow_leading_zero:
+            Enforce or disallow unnecessary leading zeroes on decimal points.
+        :param set_max_errors:
+            Set 'max' number of Errors.
+        :param set_max_warnings:
+            Set 'max' number of Warnings.
```
Makman2: why `'max'` and not just `max`? Or write it out: `maximum`

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+        :param set_extend_preference:
+            Pass in either ``@extend`` or ``@extends`` and then enforce that.
+            Both are valid in Stylus. It doesn't really matter which one
+            you use.
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``false``.
+        :param allow_leading_zero:
+            Enforce or disallow unnecessary leading zeroes on decimal points.
```
Makman2: name your settings accordingly: either allow or enforce

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+            By default Stylint ignores ``@css`` blocks. If set to true however,
+            it will throw a warning if ``@css`` is used.
+        :param max_selector_depth:
+            Set the max selector depth. If set to 4, max selector depth will
+            be 4 indents. Pseudo selectors like ``&:first-child`` or
+            ``&:hover`` won't count towards the limit.
+        :param check_duplicates:
+            Checks if selector or property duplicated unnecessarily. By
+            default, only checks on a file-by-file basis, but if
+            ``globalDupes: true`` is set, then it will also check for
```
Makman2: where can we set `globalDupes` here? Seems to be not appropriate for coafiles^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+                        allow_stacked_properties: str='never',
+                        check_valid_property: bool=True,
+                        allow_zero_units: str='never',
+                        z_index_normalize: bool=False,
+                        stylint_config: str=''):
+        """
+        :param allow_block_keyword:
+            When ``always`` expect the ``@block`` keyword when defining block
+            variables. When ``never``, expect no ``@block`` keyword when
+            defining block variables. When ``false``, do not check either way.
```
Makman2: tri-state setting: Use `bool` as type that's by default `None`.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +38,182 @@ class StylintBear:
...
+                        prefix_vars_with_dollar: str='always',
+                        allow_semicolons: str='never',
+                        sort_order: str='alphabetical',
+                        allow_stacked_properties: str='never',
+                        check_valid_property: bool=True,
+                        allow_zero_units: str='never',
+                        z_index_normalize: bool=False,
+                        stylint_config: str=''):
+        """
+        :param allow_block_keyword:
```
Makman2: when you use a tri-state bool: Call this just `block_keyword`. Makes more sense when reading:
```
block_keyword = False # Disallow block keywords
block_keywords = True # Yes use block kewords
... # Don't care about block_keywords
```

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: `parens` -> `parenthesis`. Will change it to `parenthesis`.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
Makman2: Ah yeah that's better :) Btw, the plural of paranthesis is parantheses :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Some stylus propertly (just like we have one in CSS)

bears/stylus/StylintBear.py
yash-nisar: We can choose `None` if we don't care. Fixed.

bears/stylus/StylintBear.py
yash-nisar: Documented values where I felt was necessary.

yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed. :) Added some examples for clarity.

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

yash-nisar:
Updated @Makman2

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            ``.className { padding 0 }``
+        :param check_valid_property:
+            Check that a property is valid CSS or HTML.
+        :param zero_units:
+            Looks for instances of ``0px``. You don't need the ``px``.
+            Checks all units, not just ``px``.
+        :param z_index_normalize:
+            Enforce some basic z-index sanity. Any number passed in
+            will be used as the base for your z-index values. Throws an error
+            if your value is not normalized. Use ``0`` to disable this feature.
```
Makman2: I still don't get what a normalized value is in this context ;)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            Example if set to ``5``: prefer ``z-index 10`` over ``z-index 9``
+            Example if set to ``10``: prefer ``z-index 20`` over ``z-index 15``
```
Makman2: write that into a normal sentence, not just
Example if set blub ...
Example if set blub ...

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            ```
+            .some-class
+                position absolute
+                top 0
+                right 10px
+                display block
+                float left
+            ```
+        :param stacked_properties:
+            No one-liners. Allow putting properties on new lines.
```
Makman2: Allow or enforce? "No one-liners" Means for me "enforce".

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            Example if ``False``: prefer
```
Makman2: --> For example: If ``False``, prefer

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param semicolons:
+            When ``True``, allow semicolons. When ``False``, disallow
+            semicolons. When ``None``, do not check either way.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param sort_order:
+            Enforce a particular sort order when declaring properties. Throws
+            a warning if you don't follow the order. If set to ``None``, allow
+            any order.
+            Example if ``alphabetical``:
+            prefer this:
```
Makman2: --> For example: If ``alphabetical``, prefer this:

Single line + no colon in the middle

Makman2: Are there others than ``alphabetical``?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param prefix_vars_with_dollar:
+            Enforce use of ``$`` when defining a variable. In Stylus using a
+            ``$`` when defining a variable is optional, but is a good idea
+            if you want to prevent ambiguity. Not including the ``$`` sets up
+            situations where you wonder: "Is this a variable or a value?"
+            For instance: padding ``$default`` is easier to understand than
+            padding ``default``.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param semicolons:
+            When ``True``, allow semicolons. When ``False``, disallow
```
Makman2: allow or enforce?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            semicolons. When ``None``, do not check either way.
+            Supported values are ``True``, ``False`` and ``None``.
```
Makman2: `None` is not supported from outside, please don't mention it. Say that by default either way is allowed without mentioning `None`.
We can just use `None` in the signature as default values bypass the type-check by coala.

bears/stylus/StylintBear.py
Makman2: If it's code, then it needs to be enclosed in double backticks ;)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param placeholder:
+            Enforce extending placeholder vars when using ``@extend(s)``.
+        :param prefix_vars_with_dollar:
+            Enforce use of ``$`` when defining a variable. In Stylus using a
+            ``$`` when defining a variable is optional, but is a good idea
+            if you want to prevent ambiguity. Not including the ``$`` sets up
+            situations where you wonder: "Is this a variable or a value?"
+            For instance: padding ``$default`` is easier to understand than
+            padding ``default``.
+            Supported values are ``True``, ``False`` and ``None``.
```
Makman2: `None` is not supported by the user, only from our side in the function signature. See below^^

Makman2: By default we just don't care, and the user just shall not use this setting in his coafile^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            If ``True`` check for places where ``none`` used instead of 0. If
+            ``False`` check for places where 0 could be used instead of none.
+        :param check_no_important_keyword:
+            If ``true``, show warning when ``!important`` is found.
+        :param spaces_inside_parentheses:
+            Enforce or disallow use of extra spaces inside parentheses.
+        :param placeholder:
+            Enforce extending placeholder vars when using ``@extend(s)``.
+        :param prefix_vars_with_dollar:
+            Enforce use of ``$`` when defining a variable. In Stylus using a
```
Makman2: In Stylus, (<-- comma)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param class_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``class_naming_convention`` is set to ``true``,
+            ``variable_naming_convention`` will  also look at class and
+            id names.
+        :param none_keyword:
+            If ``True`` check for places where ``none`` used instead of 0. If
+            ``False`` check for places where 0 could be used instead of none.
+        :param check_no_important_keyword:
+            If ``true``, show warning when ``!important`` is found.
```
Makman2: true -> True

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            ``camelCase``, ``BEM`` and ``False``.
+        :param class_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``class_naming_convention`` is set to ``true``,
+            ``variable_naming_convention`` will  also look at class and
+            id names.
+        :param none_keyword:
+            If ``True`` check for places where ``none`` used instead of 0. If
+            ``False`` check for places where 0 could be used instead of none.
+        :param check_no_important_keyword:
```
Makman2: actually does the keyword `!important` mean that there's something important or that something is unimportant?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param variable_naming_convention:
+            Enforce a particular naming convention when declaring classes, ids,
+            and variables. Throws a warning if you don't follow the convention.
+            Supported values are ``lowercase-dash``, ``lowercase_underscore``,
+            ``camelCase``, ``BEM`` and ``False``.
+        :param class_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``class_naming_convention`` is set to ``true``,
+            ``variable_naming_convention`` will  also look at class and
+            id names.
```
Makman2: hm can't we allow the same values for `class_naming_convention` like for `variable_naming_convention`? This is currently kind of confusing^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param max_errors:
+            Set 'max' number of Errors. Use ``0`` to disable this feature.
+        :param max_warnings:
+            Set 'max' number of Warnings. Use ``0`` to disable this feature.
+        :param mixed_spaces_and_tabs:
+            Returns ``true`` if mixed spaces and tabs are found. If a
+            non-negative number is passed to ``indent_size``, it assumes soft
+            tabs (ie, spaces), and if ``0`` is passed to ``indent_size``
```
Makman2: ie, -> i.e.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param leading_zero:
+            When ``True``, prefer leading zeroes on decimal points. When
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param max_errors:
+            Set 'max' number of Errors. Use ``0`` to disable this feature.
+        :param max_warnings:
+            Set 'max' number of Warnings. Use ``0`` to disable this feature.
+        :param mixed_spaces_and_tabs:
+            Returns ``true`` if mixed spaces and tabs are found. If a
```
Makman2: returns `true`?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            Example if ``2``: prefer
+            ``/s/smargin: 0`` over ``/s/s/smargin: 0``
+        :param leading_zero:
+            When ``True``, prefer leading zeroes on decimal points. When
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param max_errors:
+            Set 'max' number of Errors. Use ``0`` to disable this feature.
+        :param max_warnings:
+            Set 'max' number of Warnings. Use ``0`` to disable this feature.
```
Makman2: why max in single quotes?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``0``.
+            Example if ``2``: prefer
+            ``/s/smargin: 0`` over ``/s/s/smargin: 0``
+        :param leading_zero:
+            When ``True``, prefer leading zeroes on decimal points. When
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param max_errors:
+            Set 'max' number of Errors. Use ``0`` to disable this feature.
```
Makman2: why max in single quotes?

Makman2: I think you should rephrase
```
Use 0 to disable this feature.
```
to
```
If 0, all errors will be shown without a limit.
```
or similar^^

bears/stylus/StylintBear.py
Makman2: I think you should rephrase

Use 0 to disable this feature.

to

If 0, all errors will be shown without a limit.

or similar^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            Enforce or disallow spaces after commas.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param spaces_after_comments:
+            Enforce or disallow spaces after line comments.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param trailing_whitespace:
+            If ``None``, ignores trailing white space. If ``False``, trailing
+            white space will throw a warning.
+        :param css_literal:
+            By default Stylint ignores ``@css`` blocks. If set to true however,
```
Makman2: true -> `True`

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+            Supported values are ``True`` and ``None``.
+        :param spaces_after_commas:
+            Enforce or disallow spaces after commas.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param spaces_after_comments:
+            Enforce or disallow spaces after line comments.
+            Supported values are ``True``, ``False`` and ``None``.
+        :param trailing_whitespace:
+            If ``None``, ignores trailing white space. If ``False``, trailing
+            white space will throw a warning.
```
Makman2: Change None, False to True, False and rename the setting to fit those values better^^

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Normalized means a multiple of the number that you pass in. So, if you pass in 5, it will prefer multiples of 5 like 10, 15, 20, etc.

bears/stylus/StylintBear.py
yash-nisar: Enforce :+1:

yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Shame on me. My bad. Fixed.

bears/stylus/StylintBear.py
yash-nisar: Okay, so I'm removing `None` from the sentence -> Supported values are ``True``, ``False`` and ``None``. So, we assume that the user will use `None` when he wants to disable the setting ?

yash-nisar: Fixed.

yash-nisar: > None is not supported from outside, please don't mention it. Say that by default either way is allowed without mentioning None.
We can just use None in the signature as default values bypass the type-check by coala.

I've used the default values on https://github.com/SimenB/stylint (best possible settings) and making things simpler for the user because all he has to do is just run coala without being worried. If we deviate from the defaults and use `None`, he might have to configure the `.coafile` manually and go through the docs. Your opinion ?

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Added an example for clarity. Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Oops I had changed it to `maximum`. Happened because I changed my branch in git. Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Done

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,230 @@ class StylintBear:
...
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``0``.
+            Example if ``2``: prefer
+            ``/s/smargin: 0`` over ``/s/s/smargin: 0``
+        :param leading_zero:
+            When ``True``, prefer leading zeroes on decimal points. When
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True``, ``False`` and ``None``.
```
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Fixed.

bears/stylus/StylintBear.py
yash-nisar: Changed the setting name to `allow_trailing_whitespace` as that is a common setting that we use in some place like the `SpaceConsistencyBear` too.

yash-nisar: Done

bears/stylus/StylintBear.py
yash-nisar: Removed it and added an example for clarity.

bears/stylus/StylintBear.py
yash-nisar: Fixed. :)

bears/stylus/StylintBear.py
yash-nisar: Changed it to: 
```
:param variable_naming_convention:
            Enforce a particular naming convention when declaring variables.
            Throws a warning if you don't follow the convention.
```
and
```
:param strict_naming_convention:
            By default, ``variable_naming_convention`` only looks at variable
            names. If ``strict_naming_convention`` is set to ``True``,
            ``variable_naming_convention`` will  also look at class and
            id names.
```
Reasonable ?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+                        colons: bool=True,
+                        colors: bool=True,
+                        spaces_after_commas: bool=True,
+                        spaces_after_comments: bool=True,
+                        allow_trailing_whitespace: bool=False,
+                        css_literal: bool=False,
+                        max_selector_depth: bool=False,
+                        check_duplicates: bool=True,
+                        efficient_properties: bool=True,
+                        extend_preference: str='False',
```
Makman2: use `None` instrad of `'False'`

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+                        indent_size: int=0,
+                        leading_zero: bool=False,
+                        max_errors: int=0,
+                        max_warnings: int=0,
+                        mixed_spaces_and_tabs: bool=False,
+                        variable_naming_convention: str='False',
```
Makman2: Use `None`

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+                        check_valid_property: bool=True,
+                        zero_units: bool=False,
+                        z_index_normalize: bool=False,
+                        stylint_config: str=''):
+        """
+        :param block_keyword:
+            When ``True`` expect the ``@block`` keyword when defining block
+            variables. When ``False``, expect no ``@block`` keyword when
+            defining block variables.
+            Supported values are ``True`` and ``False``.
```
Makman2: you don't have to write it, people know what to feed in when having a `bool` ;)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param brackets:
+            When ``True``, expect ``{}`` when declaring a selector. When
+            ``False``, expect no brackets when declaring a selector.
+            Supported values are ``True`` and ``False``.
```
Makman2: see above

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param colons:
+            When ``True``, expect ``:`` when declaring a property. When
+            ``False``, expect no ``:`` when declaring a property.
+            Supported values are ``True`` and ``False``.
```
Makman2: see above

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param colors:
+            When ``True``, enforce variables when defining hex values.
```
Makman2: hm maybe we should call the setting differently? `colors` does not describe this so well imo

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            Check for places where properties can be written more efficiently.
+            Supported values are ``True`` and ``False``.
+        :param extend_preference:
+            Pass in either ``@extend`` or ``@extends`` and then enforce that.
+            Both are valid in Stylus. It doesn't really matter which one
+            you use.
+            For example: If set to ``@extends``, prefer
+            ``@extends $some-var`` over ``@extend $some-var``
+            For example: If set to ``@extend:``, prefer
+            ``@extend $some-var`` over ``@extends $some-var``
```
Makman2: connect those example sentences together so they don't sound so stiff:
```
For example: If set to ``@extends`` prefer ... instead of ... or if set to ``@extend`` prefer ...
```

Makman2: And don't forget the period at the end, it's still a sentence ;D

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``0``.
+            For example: If set to ``2``, prefer
+            ``/s/smargin: 0`` over ``/s/s/smargin: 0``
```
Makman2: Missing period at end-of-sentence.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param leading_zero:
+            When ``True``, prefer leading zeroes on decimal points. When
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True`` and ``False``.
+        :param max_errors:
+            Set maximum number of Errors. If ``0``, all errors will
```
Makman2: why is errors written upper-case?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            be shown without a limit.
+        :param max_warnings:
+            Set maximum number of Warnings. If ``0``, all errors will
```
Makman2: why is "Warnings" written upper-case?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            be shown without a limit.
+        :param mixed_spaces_and_tabs:
+            If a non-negative number is passed to ``indent_size``, it assumes
+            soft tabs (i.e., spaces), and if ``0`` is passed to ``indent_size``
```
Makman2: no comma after `i.e.` :)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            ``False``, disallow leading zeroes on decimal points.
+            Supported values are ``True`` and ``False``.
+        :param max_errors:
+            Set maximum number of Errors. If ``0``, all errors will
+            be shown without a limit.
+        :param max_warnings:
+            Set maximum number of Warnings. If ``0``, all errors will
+            be shown without a limit.
+        :param mixed_spaces_and_tabs:
+            If a non-negative number is passed to ``indent_size``, it assumes
```
Makman2: who/what is "it"?
Use passive form: "xxx is assumed"

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            soft tabs (i.e., spaces), and if ``0`` is passed to ``indent_size``
+            it assumes hard tabs.
+            For example: If ``indent_size = 4`` and
+            ``mixed_spaces_and_tabs = True``, prefer
+            ``s\s\s\smargin\s0`` over ``\tmargin\s0``
+            For example: If ``indent_size = 0`` and
+            ``mixed_spaces_and_tabs = True``, prefer
+            ``\tmargin\s0`` over ``\s\s\s\smargin\s0``
```
Makman2: see above

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param variable_naming_convention:
+            Enforce a particular naming convention when declaring variables.
+            Throws a warning if you don't follow the convention.
+            Supported values are ``lowercase-dash``, ``lowercase_underscore``,
+            ``camelCase``, ``BEM`` and ``False``.
```
Makman2: use `None` as default.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param strict_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``strict_naming_convention`` is set to ``True``,
+            ``variable_naming_convention`` will  also look at class and
+            id names.
```
Makman2: hm maybe write id upper-case? -> ID

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param none_keyword:
+            If ``True`` check for places where ``none`` used instead of 0. If
+            ``False`` check for places where 0 could be used instead of none.
```
Makman2: none not inside double backticks :)

Maybe you can also place 0 into double backticks, as it refers to a value in stylint code :+1:

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+        :param check_no_important_keyword:
+            If ``True``, show warning when ``!important`` is found.
+            For example: If set to ``True``, this will throw a warning:
+            ```
+            div
+                color red !important
+            ```
```
Makman2: Use double colon notation, rst has no support for markdown comment blocks :)

```
For example: If set to ... throw a warning:: # <-- note these double colons

    div
        color red !important

# and note the indentation ;)
```

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            over this:
+            ```
+            .some-class
+                position absolute
+                top 0
+                right 10px
+                display block
+                float left
+            ```
```
Makman2: see above, use double colon notation because of rst ;)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            Supported values are ``alphabetical`` and ``grouped``.
+        :param stacked_properties:
+            No one-liners. Enforce putting properties on new lines.
+            For example: If set to ``False``, prefer
+            ```
+            .className
+                padding 0
+            ```
```
Makman2: see above

bears/stylus/StylintBear.py
Makman2: ahh so "throwing an error" means it yields a result. This sounds like "if you don't put into this setting a normalized number, coala will fail". Could you rephrase the sentence a bit? :)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,242 @@ class StylintBear:
...
+            Looks for instances of ``0px``. You don't need the ``px``.
+            Checks all units, not just ``px``.
+        :param z_index_normalize:
+            Enforce some basic z-index sanity. Any number passed in
+            will be used as the base for your z-index values. Throws an error
+            if your value is not normalized. Use ``0`` to disable this feature.
+            For example: If set to ``5``, prefer
+            ``z-index 10`` over ``z-index 9``
+            For example: If set to ``10``, prefer
+            ``z-index 20`` over ``z-index 15``
```
Makman2: see above

bears/stylus/StylintBear.py
yash-nisar: Removing the sentence due to ambiguity.

bears/stylus/StylintBear.py
yash-nisar: `variable_over_hex_value` ?

yash-nisar:
Updated @Makman2

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+                        z_index_normalize: bool=False,
+                        stylint_config: str=''):
+        """
+        :param block_keyword:
+            When ``True`` expect the ``@block`` keyword when defining block
+            variables. When ``False``, expect no ``@block`` keyword when
+            defining block variables.
+        :param brackets:
+            When ``True``, expect ``{}`` when declaring a selector. When
+            ``False``, expect no brackets when declaring a selector.
```
Makman2: Is the default of `False` for this setting reasonable? Just want to ask whether you think having this by default `False` is a good standard^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param colons:
+            When ``True``, expect ``:`` when declaring a property. When
+            ``False``, expect no ``:`` when declaring a property.
```
Makman2: does this only affect properties? If so, maybe we should call it `colons_for_property_declaration` or similar. Not sure... What do you think?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param color_variable_over_hex_value:
+            When ``True``, enforce variables when defining hex values.
```
Makman2: over -> for? "Over" sounds like it should look like this:
```
red
0xFFFF0000
```
instead of
```
red = 0xFFFF0000
```
(then also append a pluralizing "s" again, sounds a bit more general)

bears/stylus/StylintBear.py
yash-nisar: Yup, see `brackets` in https://github.com/SimenB/stylint#options

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param spaces_after_commas:
+            Enforce or disallow spaces after commas.
+        :param spaces_after_comments:
+            Enforce or disallow spaces after line comments.
```
Makman2: What do "spaces after line comments" mean?
If I have a line comment, any spaces after the comment would be redundant spaces at an end-of-line.
Or is it meant like
```
# This is a comment
                                           # <-- here's the space, aka a newline.
code = code.code.code
```

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param allow_trailing_whitespace:
+            If ``True``, ignores trailing white space. If ``False``, trailing
```
Makman2: white space -> whitespace

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+            white space will throw a warning.
```
Makman2: white space -> whitespace

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param css_literal:
```
Makman2: maybe rename to `no_css_literals`? As when you set it to `True`, a warning is thrown^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+            By default Stylint ignores ``@css`` blocks. If set to ``True``
+            however, it will throw a warning if ``@css`` is used.
```
Makman2: use passive:
```
If set to ``True`` however, warnings are thrown if ``@css`` is used.
```
Maybe we should rephrase to "checks if ``@css`` is used" or similar, as I'm not sure if people misunderstand what's meant with "throwing a warning".

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param max_selector_depth:
+            Set the max selector depth. If set to 4, max selector depth will
+            be 4 indents. Pseudo selectors like ``&:first-child`` or
+            ``&:hover`` won't count towards the limit.
+        :param check_duplicates:
+            Checks if selector or property duplicated unnecessarily.
```
Makman2: property **are**

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+        :param efficient_properties:
+            Check for places where properties can be written more efficiently.
```
Makman2: optional:
maybe a bit more precise on what the optimizations are?
But we can do this after a merge^^

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+            For example: If set to ``@extends``, prefer
+            ``@extends $some-var`` instead of ``@extend $some-var``
+            or if set to ``@extend``, prefer
+            ``@extend $some-var`` over ``@extends $some-var``.
+        :param indent_size:
+            This works in conjunction with ``max_selector_depth``. If you
+            indent with spaces this is the number of spaces you indent with.
+            If you use hard tabs, set this value to ``0``.
+            For example: If set to ``2``, prefer
+            ``/s/smargin: 0`` over ``/s/s/smargin: 0``.
```
Makman2: are those `/s` rendered as whitespaces?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+                        spaces_after_commas: bool=True,
+                        spaces_after_comments: bool=True,
+                        allow_trailing_whitespace: bool=False,
+                        css_literal: bool=False,
+                        max_selector_depth: bool=False,
+                        check_duplicates: bool=True,
+                        efficient_properties: bool=True,
+                        extend_preference: str=None,
+                        indent_size: int=0,
+                        leading_zero: bool=False,
```
Makman2: hm not sure we should set this to `None` by default^^

bears/stylus/StylintBear.py
yash-nisar: See https://github.com/SimenB/stylint#commentspace--default-always-always--never--false-

yash-nisar: However, I'm adding an example for clarity. :+1:

bears/stylus/StylintBear.py
yash-nisar: Nope, this is just so the user can differenciate between tabs and spaces and see how the setting works.

bears/stylus/StylintBear.py
yash-nisar: changing it to plural as are goes well with plurals -> `Checks if selectors or properties are duplicated unnecessarily`.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+            Supported values are ``lowercase-dash``, ``lowercase_underscore``,
+            ``camelCase``, ``BEM`` and ``False``.
+        :param strict_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``strict_naming_convention`` is set to ``True``,
+            ``variable_naming_convention`` will  also look at class and
+            ID names.
+        :param none_keyword:
+            If ``True`` check for places where none used instead of ``0``. If
+            ``False`` check for places where 0 could be used instead of none.
```
Makman2: both `none` in double backticks :)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,233 @@ class StylintBear:
...
+                        max_warnings: int=0,
+                        mixed_spaces_and_tabs: bool=False,
+                        variable_naming_convention: str=None,
+                        strict_naming_convention: bool=False,
+                        none_keyword: bool=False,
+                        check_no_important_keyword: bool=True,
+                        spaces_inside_parentheses: bool=None,
+                        placeholder: bool=True,
+                        prefix_vars_with_dollar: bool=True,
+                        semicolons: bool=False,
```
Makman2: What's a common default? Maybe we should make it `None`? Don't know what's best practice for stylint^^

bears/stylus/StylintBear.py
Makman2: Ahhh yeah an example would be really good :)

bears/stylus/StylintBear.py
yash-nisar: `False` is the common default. See https://github.com/SimenB/stylint#semicolons--default-never-always--never--false-

bears/stylus/StylintBear.py
yash-nisar: Added and repushed. :)

bears/stylus/StylintBear.py
Makman2: Alright, if someone doesn't want to check semicolons at all, he/she shall open an issue :P

yash-nisar: Also, see http://stylus-lang.com/docs/css-style.html#example where they don't use semicolons (and have mentioned it as optional).

yash-nisar:
Updated @Makman2

tests/stylus/stylint_test_files/test_bad_placeholder_space_color.styl
```diff
@@ -0,0 +1,10 @@
+.message
+    padding: 10px
+body
+  @extend .message
+  color: $var_name
+  color #f00
+  font: 12px Helvetica, Arial, sans-serif
+  //comment
+  rgba(0,0,0,.18)
+
```
yash-nisar: Ugh.

bears/stylus/StylintBear.py
Makman2: Hm alright, fine with me :+1: I think this should be readable :*1:

bears/stylus/StylintBear.py
Makman2: You should remove the `False` option then :+1:

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,236 @@ class StylintBear:
...
+            ``\tmargin\s0`` over ``\s\s\s\smargin\s0``.
+        :param variable_naming_convention:
+            Enforce a particular naming convention when declaring variables.
+            Throws a warning if you don't follow the convention.
+            Supported values are ``lowercase-dash``, ``lowercase_underscore``,
+            ``camelCase``, ``BEM`` and ``False``.
+        :param strict_naming_convention:
+            By default, ``variable_naming_convention`` only looks at variable
+            names. If ``strict_naming_convention`` is set to ``True``,
+            ``variable_naming_convention`` will  also look at class and
```
Makman2: two spaces behind "will" and before "also"

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,236 @@ class StylintBear:
...
+            instead of ``none``.
+        :param check_no_important_keyword:
+            If ``True``, show warning when ``!important`` is found.
+            For example: If set to ``True``, this will throw a warning::
+
+                div
+                    color red !important
+
+        :param spaces_inside_parentheses:
+            Enforce or disallow use of extra spaces inside parentheses.
```
Makman2: hm maybe an example when extra space is enforced and one when it's disallowed?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,236 @@ class StylintBear:
...
+        :param placeholder:
+            Enforce extending placeholder vars when using ``@extend(s)``.
+        :param prefix_vars_with_dollar:
```
Makman2: what happens if set to `False`? Is then `$` forbidden? Or does it mean "don't care"?

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,236 @@ class StylintBear:
...
+                        check_no_important_keyword: bool=True,
+                        spaces_inside_parentheses: bool=None,
+                        placeholder: bool=True,
+                        prefix_vars_with_dollar: bool=True,
+                        semicolons: bool=False,
+                        sort_order: str='alphabetical',
+                        stacked_properties: bool=False,
+                        check_valid_property: bool=True,
+                        zero_units: bool=False,
+                        z_index_normalize: bool=False,
```
Makman2: Shouldn't this be an int?
If that's the case, we should rename the setting to `z_index_normalization_base` (or similar, at least the `base` should occur describing that the passed number acts as some kind of base-value :+1:)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,236 @@ class StylintBear:
...
+                        efficient_properties: bool=True,
+                        extend_preference: str=None,
+                        indent_size: int=0,
+                        leading_zero: bool=None,
+                        max_errors: int=0,
+                        max_warnings: int=0,
+                        mixed_spaces_and_tabs: bool=False,
+                        variable_naming_convention: str=None,
+                        strict_naming_convention: bool=False,
+                        none_keyword: bool=False,
```
Makman2: not sure, is this value a developer preference? If so we should maybe make it to `None`

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,236 @@ class StylintBear:
...
+                        check_no_important_keyword: bool=True,
+                        spaces_inside_parentheses: bool=None,
+                        placeholder: bool=True,
+                        prefix_vars_with_dollar: bool=True,
+                        semicolons: bool=False,
+                        sort_order: str='alphabetical',
+                        stacked_properties: bool=False,
```
Makman2: Is this one a developer preference? I can also assume that many people do mix them if the properties are simple^^

bears/stylus/StylintBear.py
```diff
@@ -1,6 +1,12 @@
+import json
+
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.NpmRequirement import NpmRequirement
 
+setting_map = {True: 'always',
```
Makman2: ah maybe precede it with an underscore so we "mark it as private". Not that people think they can import this :P

bears/stylus/StylintBear.py
yash-nisar: Yup, this is the default, see https://github.com/SimenB/stylint#none--default-never-options-always--never--false--

bears/stylus/StylintBear.py
yash-nisar: Yup, this is the default too, see https://github.com/SimenB/stylint#stackedproperties--default-never-never--false-

bears/stylus/StylintBear.py
yash-nisar: Example added.

bears/stylus/StylintBear.py
yash-nisar: Example added.

yash-nisar:
@Makman2 Updated.

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,246 @@ class StylintBear:
...
+        :param max_warnings:
+            Set maximum number of warnings. If ``0``, all errors will
+            be shown without a limit.
+        :param mixed_spaces_and_tabs:
+            If a non-negative number is passed to ``indent_size``,
+            soft tabs (i.e. spaces) are assumed, and if
+            ``0`` is passed to ``indent_size``, hard tabs are assumed.
+            For example: If ``indent_size = 4`` and
+            ``mixed_spaces_and_tabs = True``, prefer
+            ``s\s\s\smargin\s0`` over ``\tmargin\s0``
```
Makman2: missing `\` before first `s`

btw this will be escaped by python and will be made a normal whitespace. I'm not sure if `rst` will render it properly^^
Double escape `\` so those backslashes will be visible

Makman2: Btw you use normal slashes above, maybe you should here too^^

(You only have to show the first spaces, not the ones in the middle imo: `/s/s/s/smargin 0` instead of `/s/s/s/smargin/s0`)

bears/stylus/StylintBear.py
```diff
@@ -36,12 +42,250 @@ class StylintBear:
...
+            No one-liners. Enforce putting properties on new lines.
+            For example: If set to ``False``, prefer::
+
+                .className
+                    padding 0
+
+            over
+            ``.className { padding 0 }``
+        :param check_valid_property:
+            Check that a property is valid CSS or HTML.
```
Makman2: ah sorry for missing this: we should call this setting `check_property_validity`, as we don't check valid-properties for something, but we check that they are actually valid :)

Makman2:
ack e4a9d34

Makman2:
@rultor merge

PR #1819, opened 2017-06-07
Use sphinx-1.6.2 for building docs and avoid warnings by cleaning up output_regex of StyleLintBear

userzimmermann:
According to https://github.com/coala/coala/pull/4330 and https://github.com/coala/coala/pull/4333

Partial continuation of https://github.com/coala/coala-bears/pull/1569

Closes https://github.com/coala/coala-bears/issues/1554

Fixes https://github.com/coala/coala-bears/issues/1821 as side-effect

And changes `circle.yml` to use Python 3.4.3 for building docs instead of default 3.5.1, since the latter's `typing` package causes recent `sphinx` versions to fail

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -70,6 +70,12 @@ test:
     # and node 1 has python 3.4.3
     - ? |
           if [ $CIRCLE_NODE_INDEX == 1 ] ; then
+            pip install -r docs-requirements.txt
+          fi
+      :
+        parallel: true
+    - ? |
+          if [ $CIRCLE_NODE_INDEX == 1 ] ; then
             python setup.py docs
           fi
       :
--- a/circle.yml
+++ b/circle.yml
@@ -66,5 +66,11 @@ test:
         parallel: true
     - codecov:
         parallel: true
-    - python setup.py docs:
+    # sphinx fails with python 3.5.1 due to an issue in typing package,
+    # and node 1 has python 3.4.3
+    - ? |
+          if [ $CIRCLE_NODE_INDEX == 1 ] ; then
+            python setup.py docs
+          fi
+      :
         parallel: true
--- a/bears/css/StyleLintBear.py
+++ b/bears/css/StyleLintBear.py
@@ -6,7 +6,7 @@ from dependency_management.requirements.NpmRequirement import NpmRequirement
 @linter(executable='stylelint',
         output_format='regex',
         output_regex=r'\s*(?P<filename>.+)\s*(?P<line>\d+):(?P<column>\d+)\s*'
-                     r'(\D)\s*(?P<message>.+)',
+                     r'\D\s*(?P<message>.+)',
         config_suffix='.json')
 class StyleLintBear:
     """
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,1 +1,2 @@
-sphinx==1.3.5
+sphinx~=1.6.2
+sphinx_rtd_theme~=0.2.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # NOTE: This file is parsed by .ci/generate_bear_requirements.py
 # Use >= for development versions so that source builds always work
-coala>=0.12.0.dev20170703154304
+coala>=0.12.0.dev20170715183139
 # Dependencies inherited from coala
 # libclang-py3
 # coala_utils


userzimmermann:
Hmm...

```
Warning, treated as error:
StyleLintBear: Using unnecessary capturing groups affects the performance of coala. You should use '(?:<pattern>)' instead of '(<pattern>)' for your regex.
make: *** [html] Error 1
make: Leaving directory `/home/travis/build/coala/coala-bears/docs'
```

userzimmermann:
Hmm... again :)

```
Warning, treated as error:
/home/ubuntu/coala-bears/docs/API/bears.c_languages.rst:57:error while formatting arguments for bears.c_languages.ClangBear.ClangBear.run: Optional[t] requires a single type. Got <function typed_list.<locals>.<lambda> at 0x2acae1fda2f0>.
make: *** [html] Error 1
make: Leaving directory `/home/ubuntu/coala-bears/docs'
```

jayvdb:
We need an issue for StyleLintBear problem (725a6170) as that is a user facing warning.

And it looks like you've now hit the exact same "Circle - only" bug that @yash-nisar has been struggling with on .

userzimmermann:
Just found out that same CircleCI fail is already addressed in https://github.com/coala/coala-bears/issues/1575

userzimmermann:
@jayvdb @yash-nisar That `error while formatting arguments ...` problem as pasted in https://github.com/coala/coala-bears/pull/1819#issuecomment-306759739 and also happening in https://github.com/coala/coala-bears/issues/1575 originates from Python's internal `typing._Optional.__getitem__` calling `typing._type_check`

Only strange that this check is only failing on CircleCI o.O Maybe there was a bug that was fixed in some Python micro version release, which is not available in CircleCI... Or can we chose the exact Python versions there?

cc @sils @Makman2 @AbdealiJK

circle.yml
```diff
@@ -61,6 +64,8 @@ test:
         parallel: true
     - codecov:
         parallel: true
+    - pip install -r docs-requirements.txt
+    - pip install git+https://github.com/coala/coala.git@zimmermann/typed_list -I --no-deps
```
userzimmermann: This was just for experimenting. It will go away of course :)

jayvdb:
Note that switching to Circle Trusty is https://github.com/coala/coala-bears/issues/869 .  It should not be changed unilaterally without understanding the issue.  I have restored it to Circle Precise for the time being.

jayvdb:
Also note I already verified that the problem didnt occur on Circle Trusty https://github.com/coala/coala-bears/issues/1575#issuecomment-292710519

userzimmermann:
@jayvdb With Precise this PR will fail again :/ And all tests passed with Trusty, cached and uncached...

jayvdb:
You can still test on Circle Trusty by setting up enabling Circle on your fork and changing the configuration on your fork : https://circleci.com/gh/userzimmermann/coala-bears = 404
Please also time the uncached build and update https://github.com/coala/coala-bears/issues/869 with your findings.
And fwiw, it is easy to install different versions of Python onto a Circle CI VM, but of course that also causes the builds to take longer, so we need to evaluate the options to find the best approach.

Finally, we don't really need Circle CI on this repo if we enable GitCommitBear on either Travis or on AppVeyor, so that is another strategy we can follow.

circle.yml
```diff
@@ -61,6 +61,11 @@ test:
...
         parallel: true
-    - rm -rf docs/API && make -C docs clean
+    - pyenv local 3.4.3:
+        parallel: true
+    - pip install -r docs-requirements.txt:
+        parallel: true
+    - rm -rf docs/API && make -C docs clean:
+        parallel: true
     - python setup.py docs:
         parallel: true
```
jayvdb: Another option is to simply not run this on the effected Python version.  The docs are being build by Travis on three versions, and AppVeyor on two versions.  Circle can do one build instead of two, and the world will not end ;-)

jayvdb:
@userzimmermann , once https://github.com/coala/coala/pull/4342 is merged, does that solve the problem with Circle Precise 3.5.1 ?

circle.yml
userzimmermann: @jayvdb I meanwhile chose a similar approach by just setting `pyenv local 3.4.3` before building docs:

```
    - pyenv local 3.4.3:  # sphinx fails with 3.5.1
        parallel: true
    - pip install $(ls ./dist/*.whl)"[alldeps]":
        parallel: true
    - pip install -r docs-requirements.txt:
        parallel: true
    - rm -rf docs/API && make -C docs clean:
        parallel: true
    - python setup.py docs:
        parallel: true
```

The additional `pip install $(ls ./dist/*.whl)"[alldeps]"` is almost no overhead because all requirements are already installed at that point

userzimmermann:
@jayvdb Unfortunately, https://github.com/coala/coala/pull/4342 doesn't solve the issue

circle.yml
jayvdb: I dont understand why you prefer to install the same python version on both nodes (it is already installed on one of them) and run Sphinx on the same Python version on two nodes.  That gives us nothing.  Just dont run it on one node.

userzimmermann: @jayvdb I didn't change anything in the installation process

userzimmermann: And I'm not familiar with CircleCI :D So if you can tell me how to restrict steps under `test:` to certain nodes, I would be very happy...

circle.yml
```diff
@@ -61,6 +61,13 @@ test:
...
     - codecov:
         parallel: true
-    - rm -rf docs/API && make -C docs clean
+    - pyenv local 3.4.3:  # sphinx fails with 3.5.1
```
jayvdb: This (re)installs Python 3.4.3 on both nodes (it is already installed on one of them), which means Sphinx is run on the same Python version on two nodes. That gives us nothing. Just dont run it on one node.

circle.yml
jayvdb: > I didn't change anything in the installation process

I've made a note on the specific line.

> So if you can tell me how to restrict steps under test: to certain nodes, I would be very happy...

I cant explain it better than the Circle CI documentation [on the topic](https://circleci.com/docs/1.0/parallelism/).

userzimmermann:
@jayvdb Thanks! Now I know how that Circle config works... A bit strange :D But anyway... The docs are now only built on the Python 3.4.3 node. And this PR is hopefully done now ;)

circle.yml
```diff
@@ -61,6 +61,11 @@ test:
         parallel: true
     - codecov:
         parallel: true
-    - rm -rf docs/API && make -C docs clean
-    - python setup.py docs:
+    # sphinx fails with python 3.5.1 due to an issue in typing package,
+    # and node 1 has python 3.4.3
+    - if [ $CIRCLE_NODE_INDEX == 1 ] ; then pip install -r docs-requirements.txt ; fi:
+        parallel: true
+    - if [ $CIRCLE_NODE_INDEX == 1 ] ; then rm -rf docs/API && make -C docs clean ; fi:
```
jayvdb: Is this line even needed?
If not, approve https://github.com/coala/coala-bears/pull/1826 , or rebase on top of `pull/1826head` and they can all go in together.

userzimmermann: @jayvdb Removing `docs/API` is important to make autodoc always regenerate it from the python modules, what can lead to errors as we had them here... That should always be checked!

jayvdb: Replied at https://github.com/coala/coala-bears/pull/1826#issuecomment-307840747

jayvdb: #1826 was merged, so this part needs to be revised, but otherwise all is good.

jayvdb:
ack 2164aec 2046f4c 8aae605

yash-nisar:
@userzimmermann Mind sharing how did you find out that `error while formatting arguments...` was due to Python's internal issues ? :open_mouth:

yash-nisar:
@userzimmermann Maybe you can reference https://github.com/coala/coala-bears/issues/1554 (I can't change the version in our title to `1.6.2` since the issue has been filed by cobot) in your commit ?

circle.yml
userzimmermann: Thanks @jayvdb for your investigation and their results in https://github.com/coala/coala-bears/pull/1826#issuecomment-307840747 :) So I will rebase now, keeping the `rm ...` stuff deleted, and then finally merge this

userzimmermann:
@yash-nisar Regarding https://github.com/coala/coala-bears/pull/1819#issuecomment-308379182 ... Very simply ;)

```
> for /f "usebackq" %P in (`python -c "print(__import__('sys').prefix)"`) do (
      grep -Rn "Optional\[t\] requires a single type." %P)

C:\Users\Zimmermann\Miniconda3\envs\coala/Lib/typing.py:786:        arg = _type_check(arg, "Optional[t] requires a single type.")
```

userzimmermann:
reack https://github.com/coala/coala-bears/pull/1819/commits/28c2eb719f80d991084d97bb2acccf7427931d0d https://github.com/coala/coala-bears/pull/1819/commits/d308f54234513a078a9fec3f3aa9e6c26c0305e7 https://github.com/coala/coala-bears/pull/1819/commits/0c938e1d808efc5ab9ad2929f8cf91929127485d

userzimmermann:
WTF is going on here? This was done! :D 

https://circleci.com/gh/coala/coala-bears/6570#tests/containers/1 : 

```
Warning, treated as error:
param_start setting has already been defined in section PYTHON. The previous setting will be overridden.
make: *** [html] Error 1
make: Leaving directory `/home/ubuntu/coala-bears/docs'
```

https://travis-ci.org/coala/coala-bears/jobs/243706208#L4479 : 

```
Warning, treated as error:
doc-marker1 setting has already been defined in section JAVA. The previous setting will be overridden.
make: *** [html] Error 1
make: Leaving directory `/home/travis/build/coala/coala-bears/docs'
```

userzimmermann:
reack https://github.com/coala/coala-bears/pull/1819/commits/b9ef43c265a7d593e68d8830014144ae8bc117cd https://github.com/coala/coala-bears/pull/1819/commits/a10ec0c8ebffd0ef8f34a20dad9fc1d331cd2749 https://github.com/coala/coala-bears/pull/1819/commits/c6e7a95ac57ccb25dd719c8e8b81d16c0e6ae3cb

sils:
ack https://github.com/coala/coala-bears/pull/1819/commits/7e4bbd8608f6a5158075e40f6756fb18c280dc49

userzimmermann:
reack https://github.com/coala/coala-bears/pull/1819/commits/7d6301f052917c9a36d8492e249ce19b72921554 https://github.com/coala/coala-bears/pull/1819/commits/790a3a90eb05e035f533b747e7a35a366a176d1e https://github.com/coala/coala-bears/pull/1819/commits/c9be1bfc84542b92c640b57050f98a2270199ed8

sils:
@rultor merge

PR #1826, opened 2017-06-08
CI: Cache Sphinx build directory

jayvdb:
Speeds up the builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ addons:
 cache:
   pip: true
   directories:
+    - docs/_build
     # Installed language package caches
     - ~/.cabal
     - ~/.ghc
@@ -128,7 +129,6 @@ script:
     fi
   - coala --non-interactive
   - codecov
-  - rm -rf docs/API && make -C docs clean
   - python setup.py docs
 
 notifications:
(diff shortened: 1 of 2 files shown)


jayvdb:
Same as https://github.com/coala/coala/pull/4338

jayvdb:
@mixih, do you know why we dont use Sphinx caching.
I remember we had some problem with the `make` command, but I think that is unrelated.

jayvdb:
Over on https://github.com/coala/coala-bears/pull/1819#discussion_r120952016 , @userzimmermann has said:

> Removing `docs/API` is important to make autodoc always regenerate it from the python modules, what can lead to errors as we had them here... That should always be checked!

jayvdb:
I've checked the custom `setup.py` commands `docs`, and it uses `sphinx-apidoc -f` which forces overwriting.  But that is quite unnecessary anyway, as the `docs/API` area is not cached between CI builds and it isnt generated in any earlier steps of the CI build, so `docs/API` is empty each CI build when (the next line) `python setup.py docs` is executed. (but even if it isnt, the `docs` command will clean it, so the `rm -rf docs/API` is absolutely useless, unless there is a bug in sphinx-autodoc's `-f`, which we should fix upstream.

yash-nisar:
ack 210d78a6976585ce6e2f9105cc883614a73acbbb

jayvdb:
reack 0650689

jayvdb:
@rultor merge

PR #1836, opened 2017-06-13
FilenameBear.py: Add prefix/suffix matching feature

AMR-KELEG:
FilenameBearTest.py: Add tests

Add a new feature to the FilenameBear.
The bear can check if the filename matches a certain prefix/suffix.

Closes https://github.com/coala/coala-bears/issues/1817

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/FilenameBear.py
+++ b/bears/general/FilenameBear.py
@@ -27,7 +27,9 @@ class FilenameBear(LocalBear):
 
     def run(self, filename, file,
             file_naming_convention: str=None,
-            ignore_uppercase_filenames: bool=True):
+            ignore_uppercase_filenames: bool=True,
+            filename_prefix: str='',
+            filename_suffix: str=''):
         """
         Checks whether the filename follows a certain naming-convention.
 
@@ -43,6 +45,12 @@ class FilenameBear(LocalBear):
         :param ignore_uppercase_filenames:
             Whether or not to ignore fully uppercase filenames completely,
             e.g. COPYING, LICENSE etc.
+        :param filename_prefix:
+            Check whether the filename uses a certain prefix.
+            The file's extension is ignored.
+        :param filename_suffix:
+            Check whether the filename uses a certain suffix.
+            The file's extension is ignored.
         """
         head, tail = os.path.split(filename)
         filename_without_extension, extension = os.path.splitext(tail)
@@ -63,6 +71,8 @@ class FilenameBear(LocalBear):
                           'Using the default "snake" naming convention.')
                 file_naming_convention = 'snake'
 
+        messages = []
+
         try:
             new_name = self._naming_convention[file_naming_convention](
                 filename_without_extension)
@@ -71,15 +81,33 @@ class FilenameBear(LocalBear):
                      file_naming_convention)
             return
 
+        if new_name != filename_without_extension:
+            messages.append(
+                'Filename does not follow {} naming-convention.'.format(
+                    file_naming_convention))
+
+        if not filename_without_extension.startswith(filename_prefix):
+            new_name = filename_prefix + new_name
+            messages.append(
+                'Filename does not use the prefix {!r}.'.format(
+                    filename_prefix))
+
+        if not filename_without_extension.endswith(filename_suffix):
+            new_name = new_name + filename_suffix
+            messages.append(
+                'Filename does not use the suffix {!r}.'.format(
+                    filename_suffix))
+
         if ignore_uppercase_filenames and filename_without_extension.isupper():
             return
 
-        if new_name != filename_without_extension:
+        if messages:
             diff = Diff(file, rename=os.path.join(head, new_name + extension))
+            message = ('\n'.join('- ' + mes for mes in messages)
+                       if len(messages) > 1 else messages[0])
 
             yield Result(
                 self,
-                'Filename does not follow {} naming-convention.'.format(
-                    file_naming_convention),
+                message,
                 diff.affected_code(filename),
                 diffs={filename: diff})
--- a/tests/general/FilenameBearTest.py
+++ b/tests/general/FilenameBearTest.py
@@ -2,6 +2,7 @@ from queue import Queue
 
 from bears.general.FilenameBear import FilenameBear
 from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.results.Result import RESULT_SEVERITY, Result
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.settings.Section import Section
 
@@ -107,3 +108,46 @@ class FilenameBearTest(LocalBearTestHelper):
         self.assertEqual(log.message,
                          'The file naming convention could not be guessed. '
                          'Using the default "snake" naming convention.')
+
+    def test_file_prefix(self):
+        self.section['filename_prefix'] = 'pre'
+        self.check_invalidity(
+            self.uut, [''], filename='filename.xyz')
+        self.check_validity(
+            self.uut, [''], filename='prefilename.xyz')
+        self.check_results(
+            self.uut,
+            [''],
+            [Result.from_values('FilenameBear',
+                                "Filename does not use the prefix 'pre'.",
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='filename.xyz')],
+            filename='filename.xyz')
+
+    def test_file_suffix(self):
+        self.section['filename_suffix'] = 'fix'
+        self.check_invalidity(
+            self.uut, [''], filename='filename.xyz')
+        self.check_validity(
+            self.uut, [''], filename='filenamesuffix.xyz')
+        self.check_results(
+            self.uut,
+            [''],
+            [Result.from_values('FilenameBear',
+                                "Filename does not use the suffix 'fix'.",
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='filename.xyz')],
+            filename='filename.xyz')
+
+    def test_file_prefix_suffix(self):
+        self.section['filename_prefix'] = 'pre'
+        self.section['filename_suffix'] = 'fix'
+        self.check_results(
+            self.uut,
+            [''],
+            [Result.from_values('FilenameBear',
+                                "- Filename does not use the prefix 'pre'.\n"
+                                "- Filename does not use the suffix 'fix'.",
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='filename.xyz')],
+            filename='filename.xyz')


bears/general/FilenameBear.py
```diff
@@ -71,6 +77,18 @@ def run(self, filename, file,
...
+            yield Result.from_values(
+                origin=self,
+                message='Filename does not use the {} prefix.'.format(prefix),
+                file=filename)
+
+        if not filename_without_extension.endswith(suffix):
+            yield Result.from_values(
+                origin=self,
+                message='Filename does not use the {} suffix.'.format(suffix),
+                file=filename)
```
Makman2: Is it possible to suggest a patch?

bears/general/FilenameBear.py
```diff
@@ -27,7 +27,9 @@ class FilenameBear(LocalBear):
 
     def run(self, filename, file,
             file_naming_convention: str=None,
-            ignore_uppercase_filenames: bool=True):
+            ignore_uppercase_filenames: bool=True,
+            prefix: str='',
+            suffix: str=''):
```
Makman2: I would call those settings `filename_prefix` and `filename_suffix` (or `file_name_prefix` / `file_name_suffix`, what you like better). There are certainly other prefixes and suffixes that mean something different in another context^^

bears/general/FilenameBear.py
```diff
@@ -43,6 +45,10 @@ def run(self, filename, file,
         :param ignore_uppercase_filenames:
             Whether or not to ignore fully uppercase filenames completely,
             e.g. COPYING, LICENSE etc.
+        :param prefix:
+            Check whether the filename uses a certain prefix.
+        :param suffix:
+            Check whether the filename uses a certain suffix.
```
Makman2: You should definitely mention that the file ending is not checked. Otherwise I would have though that I need to include the file-ending^^ (as a filename consists of `name.ending`)

bears/general/FilenameBear.py
Makman2: Like just appending the suffix^^
(and for the prefix: just appending the prefix ;D)

AMR-KELEG: Do you mean that the bear should be able to rename the file such that the file name matches the prefix/suffix?

bears/general/FilenameBear.py
AMR-KELEG: I thought that this is obvious especially for the suffix part.
Anyway, i will update the description.

bears/general/FilenameBear.py
Makman2: If you suggest patches, we maybe need to take care if a file has not both: correct prefix and suffix. Then we would get a merge conflict for the patches^^ We could merge the results into a single result and merge the messages together (depending what combination we get).

Makman2: oh we even need to merge it below with https://github.com/coala/coala-bears/pull/1836/commits/23360b43da416d3afae8573e646d24ecb83ae6fe#diff-8da09498fa5dc8583a01d5e2f76de875R98 :/

AMR-KELEG: Ok, i will dig dipper into the Result class and try to figure out how to suggest patches.
Thanks for your comments.

Makman2: See https://github.com/coala/coala/blob/master/coalib/results/Diff.py :) There's a `rename` field you can use for this purpose ;)

bears/general/FilenameBear.py
```diff
@@ -74,12 +83,44 @@ def run(self, filename, file,
...
+                                         if
+                                         filename_without_extension.endswith(
+                                                               filename_suffix)
+                                         else
+                                         correct_prefix_name + filename_suffix)
+
+            diff = Diff(file, rename=os.path.join(
+                head, correct_prefixsuffix_name + extension))
+
+            FilenameBear_messages = [
```
Makman2: why did you prefix with `FilenameBear`? I think just `messages` is totally fine :)

bears/general/FilenameBear.py
```diff
@@ -74,12 +83,44 @@ def run(self, filename, file,
...
+                'Filename does not follow {} naming-convention.'.format(
+                        file_naming_convention),
```
Makman2: we usually use +4 spaces for hanging indents instead of 8^^

bears/general/FilenameBear.py
```diff
@@ -74,12 +83,44 @@ def run(self, filename, file,
...
+                        filename_prefix),
+                'Filename does not use the {} suffix.'.format(
+                        filename_suffix)]
+
+            selectors = [
+                        new_name != filename_without_extension,
+                        not filename_without_extension.startswith(
+                            filename_prefix),
+                        not filename_without_extension.endswith(
+                            filename_suffix)]
```
Makman2: indentation: 4 spaces relative to `selectors`:
```
selectors = [
    new_name != filename_without_extensionm
    not filename_without_extension.startswith(
        filename_prefix),
    not ...
```

bears/general/FilenameBear.py
```diff
@@ -74,12 +83,44 @@ def run(self, filename, file,
...
+                                                               filename_prefix)
+                                   else
+                                   filename_prefix + new_name)
+
+            correct_prefixsuffix_name = (correct_prefix_name
+                                         if
+                                         filename_without_extension.endswith(
+                                                               filename_suffix)
+                                         else
+                                         correct_prefix_name + filename_suffix)
```
Makman2: why you don't move the new filename generation before the `if` clause? So you modify `new_name` before the if, and in the end you don't have to check every condition, but just like before `new_name != filename_without_extension`

Makman2: So into this line: https://github.com/coala/coala-bears/pull/1836/commits/9c4d2ed9295e8d067f7d60fe7210ae6344f4b027#diff-8da09498fa5dc8583a01d5e2f76de875R85

bears/general/FilenameBear.py
```diff
@@ -74,12 +83,44 @@ def run(self, filename, file,
...
-            diff = Diff(file, rename=os.path.join(head, new_name + extension))
+        if (new_name != filename_without_extension or
+            not filename_without_extension.startswith(filename_prefix) or
+                not filename_without_extension.endswith(filename_suffix)):
+            correct_prefix_name = (new_name
+                                   if
+                                   filename_without_extension.startswith(
+                                                               filename_prefix)
+                                   else
+                                   filename_prefix + new_name)
```
Makman2: Should be a new issue, but I think I got a smart idea :P

Instead of just prepending the prefix if we don't match it in the beginning, we try to match it as long as possible. I.e. consider this:
A file is called `hello_world.py`. We want it to have the prefix `python_` (whyever...). Our patch would suggest to correct it to `python_hello_world`. This is fine. But now consider a filename with a typo:
`pythonhello_world2.py`. Our bear would propose now `python_pythonhello_world2.py`. A human would see that we probably meant to type `python_hello_world2.py`. Same with this file: `ython_hello_world3.py` is probably meant to be `python_hello_world3.py`.

Like said this is imo a new issue, but we could try to perform some best-suiting match of the prefix (same idea would also apply for suffixes) and suggest a smarter patch :)

bears/general/FilenameBear.py
Makman2: so you would insert something like this:
```
if not filename_without_extension.startswith(prefix):
    # modify new_name accordingly how it should be.

# same with suffix
```
Then you don't have to change any code in the result, except the message handling. Actually you can append messages inside a message list:

```
if not filename_without_extension.startswith(prefix):
    # modify new_name accordingly how it should be.
    messages.append('my message')
```
Then you don't need `itertools.compress` btw :)

bears/general/FilenameBear.py
```diff
@@ -74,12 +83,44 @@ def run(self, filename, file,
...
+                            filename_prefix),
+                        not filename_without_extension.endswith(
+                            filename_suffix)]
 
             yield Result(
                 self,
-                'Filename does not follow {} naming-convention.'.format(
-                    file_naming_convention),
+                '\n'.join(
+                    itertools.compress(FilenameBear_messages, selectors)),
```
Makman2: I'm maybe thinking of prepending bullets to each message. Instead of
```
message1
message2
...
```
-->
```
- message1
- message2
- ...
```
The `SpaceConsistencyBear` does it like this :) (Okay I have to admit that I've written that code :3)

Makman2: What do you think? :)

Makman2: Oh maybe we can be super smart and have an exception if there's only a single message:
```
if len(messages) == 1:
    # no bullet, just the message
else:
    # use bullets together with messages
```

bears/general/FilenameBear.py
AMR-KELEG: I thought that `messages` isn't that clear.

bears/general/FilenameBear.py
AMR-KELEG: I tried to follow the formatting that is already used in the file (Line 83).
https://github.com/coala/coala-bears/commit/9c4d2ed9295e8d067f7d60fe7210ae6344f4b027#diff-8da09498fa5dc8583a01d5e2f76de875L79

bears/general/FilenameBear.py
AMR-KELEG: Actually, this is much better :+1:  :smiley:

bears/general/FilenameBear.py
AMR-KELEG: I think that ,after adding the bullets, the output looks better to me.
Special thanks goes to the `SpaceConsistencyBear`'s author :wink:.

bears/general/FilenameBear.py
AMR-KELEG: Well, i would like to work on it.
However, fulfilling the achievement that my PR is merged pushes me to say that i prefer working on it as a part of a new issue :thinking:

bears/general/FilenameBear.py
```diff
@@ -74,12 +82,38 @@ def run(self, filename, file,
         if ignore_uppercase_filenames and filename_without_extension.isupper():
             return
 
-        if new_name != filename_without_extension:
-            diff = Diff(file, rename=os.path.join(head, new_name + extension))
+        messages = []
```
Makman2: Up to you:
I have the feeling that this belongs before the part with
```
try:
    new_name = self._naming_conventions[...]
```
as there we start to analyze the filename.

But like said, up to you, as semantically the current code is totally correct.

bears/general/FilenameBear.py
```diff
@@ -74,12 +82,38 @@ def run(self, filename, file,
...
+                    self,
+                    '\n'.join('- ' + message for message in messages),
+                    diff.affected_code(filename),
+                    diffs={filename: diff})
+            else:
+                yield Result(
+                    self,
+                    messages[0],
+                    diff.affected_code(filename),
+                    diffs={filename: diff})
```
Makman2: don't generate 2 results, but two different messages :D
```
if len(messages) > 1:
    result_message = '\n'.join('- ' + message for message in messages)
else:
    result_message = messages[0]

yield Result(
    self, result_message, diff.affected_code(filename), diffs={filename:diff})
```
or you can compact the `if`-statement to an `if`-expression if you like:
```
result_message = ('\n'.join('- ' + message for message in messages)
                  if len(messages) > 1 else
                  messages[0])
```

bears/general/FilenameBear.py
```diff
@@ -74,12 +82,38 @@ def run(self, filename, file,
...
+        if not filename_without_extension.startswith(filename_prefix):
+            new_name = filename_prefix + new_name
+            messages.append(
+                'Filename does not use the {} prefix.'.format(
+                    filename_prefix))
+
+        if not filename_without_extension.endswith(filename_suffix):
+            new_name = new_name + filename_suffix
+            messages.append(
+                'Filename does not use the {} suffix.'.format(
```
Makman2: maybe displaying the `repr`? -->
`Filename does not use the {!r} suffix.`

What do you think?

Makman2: Same above^^

AMR-KELEG: Yes, i agree.
The output messages look much better.

bears/general/FilenameBear.py
```diff
@@ -66,6 +74,24 @@ def run(self, filename, file,
...
+                new_name = filename_prefix + new_name
+                messages.append(
+                    'Filename does not use the {!r} prefix.'.format(
+                        filename_prefix))
+
+            if not filename_without_extension.endswith(filename_suffix):
+                new_name = new_name + filename_suffix
+                messages.append(
+                    'Filename does not use the {!r} suffix.'.format(
+                        filename_suffix))
```
Makman2: ah no I didn't mean to put the whole stuff into the try-block, just moving `messages = []` *before* the try-block :D

The prefix/suffix stuff should come after the `try`-block :+1:

So
```python
        messages = []

        try:
            new_name = self._naming_convention[file_naming_convention](
                filename_without_extension)
        except KeyError:
            self.err('Invalid file-naming-convention provided: ' +
                     file_naming_convention)
            return

        if new_name != filename_without_extension:
            messages.append(
                'Filename does not follow {} naming-convention.'.format(
                    file_naming_convention))

        if not filename_without_extension.startswith(filename_prefix):
            new_name = filename_prefix + new_name
            messages.append(
                'Filename does not use the {!r} prefix.'.format(
                    filename_prefix))

        if not filename_without_extension.endswith(filename_suffix):
            new_name = new_name + filename_suffix
            messages.append(
                'Filename does not use the {!r} suffix.'.format(
                    filename_suffix))
```

Makman2: `try`-blocks should be left as early as possible, as we want to fail fast on other errors and don't want them to slip through :)

bears/general/FilenameBear.py
```diff
@@ -74,12 +100,13 @@ def run(self, filename, file,
         if ignore_uppercase_filenames and filename_without_extension.isupper():
             return
 
-        if new_name != filename_without_extension:
-            diff = Diff(file, rename=os.path.join(head, new_name + extension))
+        if messages:
+            diff = Diff(file, rename=os.path.join(
+                head, new_name + extension))
```
Makman2: is the linebreak necessary here? Would keep the old line without a linebreak:
```
diff = Diff(file, rename=os.path.join(head, new_name + extension))
```
One change less to review ;)

bears/general/FilenameBear.py
AMR-KELEG: Ok

bears/general/FilenameBear.py
```diff
@@ -71,15 +81,32 @@ def run(self, filename, file,
...
+            new_name = filename_prefix + new_name
+            messages.append(
+                'Filename does not use the {!r} prefix.'.format(
+                    filename_prefix))
+
+        if not filename_without_extension.endswith(filename_suffix):
+            new_name = new_name + filename_suffix
+            messages.append(
+                'Filename does not use the {!r} suffix.'.format(
+                        filename_suffix))
```
Makman2: indentation ;)

bears/general/FilenameBear.py
```diff
@@ -71,15 +81,32 @@ def run(self, filename, file,
...
 
+        if new_name != filename_without_extension:
+            messages.append(
+                'Filename does not follow {} naming-convention.'.format(
+                    file_naming_convention))
+
+        if not filename_without_extension.startswith(filename_prefix):
+            new_name = filename_prefix + new_name
+            messages.append(
+                'Filename does not use the {!r} prefix.'.format(
```
Makman2: maybe rephrase slightly to
```
Filename does not use the prefix {!r}
```
Currently it sounds a bit "this is one well-known prefix".

Makman2: Same below for suffix :)

Makman2: But up to you :+1:

bears/general/FilenameBear.py
```diff
@@ -71,15 +81,32 @@ def run(self, filename, file,
...
-        if new_name != filename_without_extension:
+        if messages:
             diff = Diff(file, rename=os.path.join(head, new_name + extension))
 
             yield Result(
                 self,
-                'Filename does not follow {} naming-convention.'.format(
-                    file_naming_convention),
+                '\n'.join('- ' + message for message in messages)
+                if len(messages) > 1 else messages[0],
```
Makman2: optional:
maybe put that into a `message` variable before, as this expression is a bit more complex?
up to you :+1:

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +118,17 @@ def test_auto_file_naming_convention_warning(self):
...
+            self.uut, [''], filename='filename.xyz', valid=False)
+        self.check_validity(
+            self.uut, [''], filename='prefilename.xyz', valid=True)
+
+    def test_file_suffix(self):
+        self.section['filename_suffix'] = 'fix'
+        self.check_validity(
+            self.uut, [''], filename='filename.xyz', valid=False)
+        self.check_validity(
+            self.uut, [''], filename='filenamesuffix.xyz', valid=True)
```
Makman2: Hm I would like to see here tests with `check_results` :)

Makman2: Shouldn't be too hard ;D

bears/general/FilenameBear.py
AMR-KELEG: I was embarrassed that i still have indentation problems but i think this message is the reason why i indented this line.
```
bears/general/FilenameBear.py                                                                                                         
|  99| ················filename_suffix))
|    | [NORMAL] PycodestyleBear (E122):
|    | E122 continuation line missing indentation or outdented
|    | *0: Do nothing
|    |  1: Open file(s)
|    |  2: Add ignore comment

```

AMR-KELEG: @Makman2 Could you please check the message?
I won't push my code until we can figure out the correct way to do it.

Makman2: can't reproduce, a run on coala locally on your PR yields no errors :)

AMR-KELEG: You should first unindent the lines as requested to reproduce the error message.

AMR-KELEG: Well, the error is reproducible on my local env.
Anyway, i will unindent lines 93 and 99 as requested and push the changes without running coala.
Let's see what will gitmate say :smile:

bears/general/FilenameBear.py
```diff
@@ -71,15 +81,33 @@ def run(self, filename, file,
...
+            new_name = filename_prefix + new_name
+            messages.append(
+                'Filename does not use the prefix {!r}.'.format(
+                filename_prefix))
+
+        if not filename_without_extension.endswith(filename_suffix):
+            new_name = new_name + filename_suffix
+            messages.append(
+                'Filename does not use the suffix {!r}.'.format(
+                filename_suffix))
```
AMR-KELEG: @Makman2 Kindly check the error.

Makman2: naaaaah you shall indent like this!
```
            messages.append(
                'Filename does not use the suffix {!r}.'.format(
                    filename_suffix))
```
:D

AMR-KELEG: @Makman2 I know how to fix it.
I just wanted to show you that the error reproducible and the lines shouldn't be unindented as requested.
Tell me if I am misunderstanding the situation :)

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +119,50 @@ def test_auto_file_naming_convention_warning(self):
         self.assertEqual(log.message,
                          'The file naming convention could not be guessed. '
                          'Using the default "snake" naming convention.')
+
+    def test_file_prefix(self):
+        self.section['filename_prefix'] = 'pre'
+        self.check_validity(
+            self.uut, [''], filename='filename.xyz', valid=False)
```
Makman2: use `check_invalidity` instead :)

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +119,50 @@ def test_auto_file_naming_convention_warning(self):
...
+        self.check_validity(
+            self.uut, [''], filename='prefilename.xyz', valid=True)
```
Makman2: you can omit the parameter `valid` :+1:

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +119,50 @@ def test_auto_file_naming_convention_warning(self):
...
+                                'Filename does not use '
+                                'the prefix {!r}.'.format('pre'),
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='filename.xyz')],
+            filename='filename.xyz')
+
+    def test_file_suffix(self):
+        self.section['filename_suffix'] = 'fix'
+        self.check_validity(
+            self.uut, [''], filename='filename.xyz', valid=False)
```
Makman2: same here -> `check_invalidity`

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +119,50 @@ def test_auto_file_naming_convention_warning(self):
...
+        self.check_validity(
+            self.uut, [''], filename='filenamesuffix.xyz', valid=True)
```
Makman2: `valid=True` can be omitted

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +119,50 @@ def test_auto_file_naming_convention_warning(self):
...
+        self.section['filename_prefix'] = 'pre'
+        self.section['filename_suffix'] = 'fix'
+        self.check_results(
+            self.uut,
+            [''],
+            [Result.from_values('FilenameBear',
+                                '\n'.join(['- Filename does not use '
+                                           'the prefix {!r}.'.format('pre'),
+                                           '- Filename does not use '
+                                           'the suffix {!r}.'.format('fix')]),
```
Makman2: instead of copying the functionality with join in the algorithm, try to write the string out:
```
"- Filename does not use the prefix 'pre'\n"
"- Filename does not use the suffix 'fix'"
```
So we directly see what we shall get :)
You could do the same above for the `.format` function uses above in the test.

tests/general/FilenameBearTest.py
```diff
@@ -118,3 +119,50 @@ def test_auto_file_naming_convention_warning(self):
...
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='filename.xyz')],
+            filename='filename.xyz')
```
Makman2: Do you assert for the diffs? That would be cool to test too :)

bears/general/FilenameBear.py
```diff
@@ -63,6 +71,8 @@ def run(self, filename, file,
                           'Using the default "snake" naming convention.')
                 file_naming_convention = 'snake'
 
+        messages = []
```
meetmangukiya: I don't like the idea of collecting all the errors into one diff, what if I want to apply the patch that renames it to follow naming convention, but not the patch that renames it with prefix text :confused: ?

meetmangukiya: Please consider yielding 3 results instead

tests/general/FilenameBearTest.py
AMR-KELEG: Could you please provide another test file that uses diffs?

bears/general/FilenameBear.py
Makman2: Nope this is on purpose, so we can generate a single patch without conflict errors :+1:

Makman2: If we support merging diffs on a file-name basis, we can yield again different results for those :+1:

tests/general/FilenameBearTest.py
Makman2: Actually I was searching for one, and discovered no tests use `diffs`. I was wondering why your tests did already pass, because you definitely generate results with diffs. I found out that `Result` doesn't take diffs into account when comparing with each other. This is super bad and needs definitely a fix. Filed an issue here: https://github.com/coala/coala/issues/4400

Because of this, you don't have to provide a diff as it's meaningless now (via the `diffs` parameter of `Result.__init__`), but we file an issue once we are able to do so :+1:

AMR-KELEG:
@Makman2 I know the organisation is reviewing the GSoC projects now.
In the meanwhile, could you tell me how can i fix the ci/circleci failing problem.
Should i re-push my commit to the branch?
Thanks :smile:

Makman2:
ack f05e746

Makman2:
@AMR-KELEG that is a known issue that is caused by `language-check`. The error is occurring randomly, I restarted the build :+1:

AMR-KELEG:
@Makman2 thanks 😄 
I really enjoyed the whole review process and I am looking forward to collaborating with you in further contributions to coala.

AMR-KELEG:
@Makman2 I have rebased my branch yet ci failed again :cry:

Makman2:
> `bears/documentation/DocumentationStyleBear.py                             42      5     18      1    87%   103-116, 49->exit

looks like it's not your fault. I hope that's not happening also on master...

Makman2:
@AMR-KELEG I hope a rebase does it now, as other PRs do pass :+1:

jayvdb:
ack 404db94

jayvdb:
@rultor merge

Issue #1867, opened 2017-06-25
Implement confidence values in VultureBear


RJ722:
`VultureBear` should report confidence values along with it's results.

CC @jayvdb @jendrikseipp

PR #1868, opened 2017-06-25
VultureBear: Report confidence values

RJ722:
Report confidence values along with results in VultureBear

Closes: #1867 

### Checklist

- [X] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [X] I ran coala over my code locally. (*All commits have to pass
      pass on every single commit!

### DIFF
--- a/bears/python/VultureBear.py
+++ b/bears/python/VultureBear.py
@@ -3,6 +3,15 @@ from coalib.results.Result import Result
 from dependency_management.requirements.PipRequirement import PipRequirement
 from vulture import Vulture
 
+CONFIDENCE_MAP = {
+    'attribute': 70,
+    'class': 70,
+    'function': 70,
+    'import': 95,
+    'property': 70,
+    'variable': 70,
+}
+
 
 def _find_unused_code(filenames):
     """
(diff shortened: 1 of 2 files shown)


RJ722:
CC @jayvdb @jendrikseipp

bears/python/VultureBear.py
```diff
@@ -3,6 +3,19 @@
...
+    'attribute': 50,
+    'class': 50,
+    'function': 50,
+    'import': 100,
+    'property': 50,
+    'variable': 50,
+}
+
+
+def calc_confidence(typ):
```
jendrikseipp: No need for this function. I think we can directly use `CONFIDENCE_MAP[item.typ]` below. We don't want to use get() since it swallows missing confidence values silently.

bears/python/VultureBear.py
```diff
@@ -3,6 +3,19 @@
 from dependency_management.requirements.PipRequirement import PipRequirement
 from vulture import Vulture
 
+CONFIDENCE_MAP = {
+    'attribute': 50,
+    'class': 50,
+    'function': 50,
+    'import': 100,
+    'property': 50,
+    'variable': 50,
```
jendrikseipp: The exact numbers are debatable, but I'd change them to 70 and 95.

jayvdb:
ack 7ae4795

jayvdb:
@rultor merge

PR #1886, opened 2017-07-04
KeywordBearTest: Add language setting in test

refeed:
Fixes https://github.com/coala/coala-bears/issues/1885

### DIFF
--- a/bears/general/KeywordBear.py
+++ b/bears/general/KeywordBear.py
@@ -11,9 +11,6 @@ from bears.general.AnnotationBear import AnnotationBear
 
 
 def _get_comments(dependency_results):
-    if not dependency_results:
-        return
-
     annotation_bear_results = dependency_results.get('AnnotationBear')
     if (not annotation_bear_results or
             not isinstance(annotation_bear_results, list)):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # NOTE: This file is parsed by .ci/generate_bear_requirements.py
 # Use >= for development versions so that source builds always work
-coala>=0.12.0.dev20170627184353
+coala>=0.12.0.dev20170703154304
 # Dependencies inherited from coala
 # libclang-py3
 # coala_utils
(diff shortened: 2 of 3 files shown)


jayvdb:
There is a coverage problem :-(

refeed:
@jayvdb 
err, I think we could delete this code, since there's should be always a `dependency_results`:
from `KeywordBear.py:14`
```py
    if not dependency_results:
        return
```

jayvdb:
> err, I think we could delete this code, since there's should be always a dependency_results:
from KeywordBear.py:14

Well, `if not` is currently also testing for an "empty" `dependency_results`.  Can it ever be empty?  Or, will it be filled with one dict member key "AnnotationBear" with a value of "no annotations"?

If `dependency_results` can be "false-ish" , there is likely a bug here anyway, as b8d38207261 added `list(..)`, which will be `list(None)` which raises `TypeError: 'NoneType' object is not iterable` .  Lines 17-20 also have a similar problem.

That commit comes from issue https://github.com/coala/coala-bears/issues/1274 , which didnt have any tests, but https://github.com/coala/coala-bears/issues/1306 was created for tests, and then https://github.com/coala/coala-bears/issues/1306 had tests but was closed without being merged.

So, I think it is very reasonable to remove `KeywordBear.py:14-15`.  That is the simple fix.

We could potentially skip all of the `KeywordBear` logic if there are no `dependency_results` containing comments, but it needs to be rewritten as `_get_comments` should return an empty list so that `list(..)` doesnt fail.  Then we would need a new test method to confirm (using mock) that the bear does not execute `check_keywords` if there was no comments  `dependency_results`, which will give you the extra coverage needed.

refeed:
@jayvdb 
> Well, if not is currently also testing for an "empty" dependency_results. Can it ever be empty? Or, will it be filled with one dict member key "AnnotationBear" with a value of "no annotations"?

If `AnnotationBear` doesn't yield anything, `dependency_results` will be filled with one dict member `AnnotationBear` with value `[]`

jayvdb:
ack 581827c 0340406 62daca6

jayvdb:
@rultor merge

jayvdb:
@rultor merge

PR #1890, opened 2017-07-05
RuboCopBear: Fix CI breakage

yash-nisar:
https://github.com/bbatsov/rubocop/pull/4226 indicates that `--stdin`
takes filename as an argument which was incorporated. Also, the tests
were enhanced that involved the bear to be run on files with the `.rb`
(ruby) extension because initially the bear was run on the generated
temporary files that was the cause of error.

Fixes https://github.com/coala/coala-bears/issues/1548

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/ruby/RuboCopBear.py
+++ b/bears/ruby/RuboCopBear.py
@@ -19,7 +19,7 @@ class RuboCopBear:
     """
 
     LANGUAGES = {'Ruby'}
-    REQUIREMENTS = {GemRequirement('rubocop', '0.47.1'),
+    REQUIREMENTS = {GemRequirement('rubocop', '0.49.1'),
                     PipRequirement('pyyaml', '3.12')}
     AUTHORS = {'The coala developers'}
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
@@ -36,7 +36,7 @@ class RuboCopBear:
     def create_arguments(filename, file, config_file, rubocop_config: str=''):
         # Need both stdin and filename. Explained in this comment:
         # https://github.com/bbatsov/rubocop/pull/2146#issuecomment-131403694
-        args = (filename, '--stdin', '--format=json')
+        args = ('--stdin', filename, '--format=json')
         if rubocop_config:
             args += ('--config', rubocop_config)
         else:
--- a/tests/ruby/RuboCopBearTest.py
+++ b/tests/ruby/RuboCopBearTest.py
@@ -1,57 +1,111 @@
 import os
+from queue import Queue
+
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 
 from bears.ruby.RuboCopBear import RuboCopBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
-
-good_file = """def good_name
-  test if something
-end
-"""
-
-bad_file = """def badName
-  test if something
-end
-"""
 
 
-RuboCopBearTest = verify_local_bear(RuboCopBear,
-                                    invalid_files=(bad_file,),
-                                    valid_files=(good_file,))
+def get_testfile_path(name):
+    return os.path.join(os.path.dirname(__file__),
+                        'test_files',
+                        name)
 
-# Testing Config
-rubocop_config = os.path.join(os.path.dirname(__file__),
-                              'test_files',
-                              'rubocop_config.yaml')
+
+def load_testfile(name):
+    with open(get_testfile_path(name)) as f:
+        return f.readlines()
+
+
+@generate_skip_decorator(RuboCopBear)
+class RuboCopBearTest(LocalBearTestHelper):
+
+    def setUp(self):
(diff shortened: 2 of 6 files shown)


bears/ruby/RuboCopBear.py
```diff
@@ -36,7 +36,7 @@ class RuboCopBear:
     def create_arguments(filename, file, config_file, rubocop_config: str=''):
         # Need both stdin and filename. Explained in this comment:
         # https://github.com/bbatsov/rubocop/pull/2146#issuecomment-131403694
-        args = (filename, '--stdin', '--format=json')
+        args = ('--stdin', filename, '--format=json')
```
jayvdb: Is this change compatible with earlier versions of rubocop?

yash-nisar: Yes :+1: , tested will versions `0.47.1`, `0.48.1`, `0.49.1`

tests/ruby/RuboCopBearTest.py
```diff
@@ -1,57 +1,91 @@
...
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 
 from bears.ruby.RuboCopBear import RuboCopBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
 
-good_file = """def good_name
```
jayvdb: why do we need to move the text out into real filenames?

Does rubocop require that there is a valid file with this filename?

Cant `LocalBearTestHelper` create temporary files with an appropriate filename/extension?

If it cant, then writing tests is much harder, which is sort of the anti-thesis of your project.

yash-nisar: Yes, rubocop requires that we have a valid filename extension (`.rb`). Temporary files were created in the previous case when we had to use `verify_local_bear(...)` but without a valid extension. See the commit message.

jayvdb: So, you need to improve `LocalBearTestHelper` :P

yash-nisar: Actually, this is a linter specific problem. This is the first linter that I've come across that is careful about the filename extension. Moreover, it would make less sense IMO to improve `LocalBearTestHelper` to take care of this case only. Also we had to somehow improve the tests and use `check_results(...)` instead of `verify_local_bear(...)`.

yash-nisar:
@Makman2 Tagging you since this PR involved improvisation of tests.

bears/ruby/RuboCopBear.py
```diff
@@ -19,7 +19,7 @@ class RuboCopBear:
     """
 
     LANGUAGES = {'Ruby'}
-    REQUIREMENTS = {GemRequirement('rubocop', '0.47.1'),
+    REQUIREMENTS = {GemRequirement('rubocop'),
```
Makman2: why did you remove the version specification?

tests/ruby/RuboCopBearTest.py
```diff
@@ -1,57 +1,91 @@
...
+            [Result.from_values('RuboCopBear (Layout/CommentIndentation)',
+                                message='Incorrect indentation detected '
+                                        '(column 2 instead of 1).',
+                                file=get_testfile_path(filename),
+                                line=2,
+                                column=3,
+                                end_line=2,
+                                end_column=20,
+                                severity=RESULT_SEVERITY.INFO)],
+            filename=get_testfile_path(filename),
```
Makman2: you should also check good files :+1:

Makman2:
You could write into the commit message body that part of the fixing was to improve the tests.

bears/ruby/RuboCopBear.py
yash-nisar: Our motive is to use the latest version of rubocop. We were using `0.47.1` in the earlier case and due to an upstream release i.e. `0.48`, our CI broke. As a temporary fix, I had to pin the version to `0.47.1` to make sure that the tree was green again. Now that we have to solution, I think it would be a good option to get rid of the pinning and install the latest version.

tests/ruby/RuboCopBearTest.py
yash-nisar: Done :+1:

yash-nisar:
Updated @Makman2

bears/ruby/RuboCopBear.py
Makman2: though I would recommend to fix it currently for 0.48, and upgrade to newer releases explicitly with an issue. But up to you :+1: Not sure actually what our requirements strategy is for bears^^

tests/ruby/RuboCopBearTest.py
```diff
@@ -1,57 +1,111 @@
...
+            settings={'indent_size': 1})
+
+    def test_good_indent_size(self):
+        filename = 'bad_indent_file.rb'
```
Makman2: maybe just call the file `indent.rb`, as in this context it isn't bad?

Makman2:
ack 51507c5

Makman2:
reack d35b120

Makman2:
@rultor merge

PR #1898, opened 2017-07-07
PEP8Bear.py: Show PEP8 error description

### LINKED ISSUE #1897: PEP8Bear: Show PEP8 error description
Originally created by @TheMeaningfulEngineer https://github.com/coala/coala/issues/2511

### Overview Description

Coala doesn't give a description of the error what PEP8 does.
# Steps to Reproduce

Files:

example.py

```
# After this comment there should be two lines but I'll only put one.

class Faker:

    def __init__():
        pass
```

.coafile

```
[Default]
bears = PEP8Bear
files = *.py
```
# Actual Results

```
$ coala
Executing section Default...

example.py
|   1| #•After•this•comment•there•should•be•two•lines•but•I'll•only•put•one.
|    | [NORMAL] PEP8Bear:
|    | The code does not comply to PEP8.
|----|    | /home/alan/europython/by_day/friday/coala_contribution/example.py
|    |++++| /home/alan/europython/by_day/friday/coala_contribution/example.py
|   1|   1| # After this comment there should be two lines but I'll only put one.
|    |   2|+
|   2|   3| 
|   3|   4| class Faker:
|   4|   5| 
|    | The following actions are applicable to this result:
|    |  0: Apply no further actions.
|    |  1: Open the affected file(s) in an editor.
|    |  2: Apply the patch automatically.
|    | Please enter the number of the action you want to execute (Ctrl-D to exit). 

```
# Expected Results

It would be nice to get the error description which PEP8 gives somewhere.

```
$ pep8 example.py 
example.py:3:1: E302 expected 2 blank lines, found 1
```

JohnMoutafis:
Add PEP8 error message description for every error
found by autopep8 during PEP8Bear analysis.

Closes #1897

### DIFF
--- a/bears/python/PEP8Bear.py
+++ b/bears/python/PEP8Bear.py
@@ -1,4 +1,5 @@
 import autopep8
+import pycodestyle
 
 from coalib.bearlib import deprecate_settings
 from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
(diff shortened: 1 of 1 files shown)


JohnMoutafis:
Oops! That's what you get when you push all the buttons!!

sidharthskumar (approved):

adhikasp:
Missing tests :(

bears/python/PEP8Bear.py
```diff
@@ -41,14 +42,69 @@ def run(self, filename, file,
                    'max_line_length': max_line_length,
                    'indent_size': indent_size}
 
+        errors = self.list_pep8_errors(source=file, options=options)
```
adhikasp: If I understand this correctly, you run `pycodestyle` independently and then match it with result from `autopep8`?

I think it will be unreliable because the coala-generated `diffs` below is not guaranteed to match 1-on-1 with list from `pycodestyle`.

NiklasMM: I can confirm that. The [documentation](https://pypi.python.org/pypi/autopep8) states that autopep8 fixes some things not reported by `pycodestyle`

JohnMoutafis: `autopep8.fix_code` method fixes the errors reported by `pycodestyle`.
If you check `autopep8`'s code you will see that it runs a `pycodestyle` check and then it applies the fixes on the reported errors.

NiklasMM: Yes, but as I said it does more than that, for example `lib2to3.refactor`

NiklasMM:
It seems that you have put in quite an effort into this. Thank you for doing that!

However, I don't think this is the right approach. It makes many assumptions and does quite a bit of work to fix an issue that could easily be fixed by the upstream tool. Have we ever tried to go upstream and patch autopep8 to actually report the error being fixed?

I'm also curious to hear other peoples oppinion!

bears/python/PEP8Bear.py
```diff
@@ -41,14 +42,69 @@ def run(self, filename, file,
...
-        diffs = Diff.from_string_arrays(file, corrected).split_diff()
+        diffs = enumerate(
+            Diff.from_string_arrays(file, corrected).split_diff())
 
-        for diff in diffs:
+        for i, diff in diffs:
             yield Result(self,
-                         'The code does not comply to PEP8.',
+                         'The code does not comply to PEP8.\nReason: {}'
+                         .format(errors[i]['info'] if i < len(errors) else ''),
```
sils: `''` -> `'Unknown'` would make more sense to me FWIW

sils:
Generally I agree with @NiklasMM - this is awesome stuff but it's nondeterministic. It is already known that our diff splitting isn't always correct so we *know* there will be cases where this'll fail.

We should be really sure if an upstream contribution wouldn't be possible, cause upstream is generally better for the community anyway.

@JohnMoutafis any thoughts on that?

sils:
(marking WIP in the meantime)

JohnMoutafis:
I have added some explanation in the commit body.
I do understand that this is not a perma-fix since it makes assumptions but I do still believe that this can act as a temporary fix and/or a path to a more permanent solution.

sils:
interesting: apparently there's a `--verbose`? https://github.com/hhatto/autopep8/issues/289

sils:
this verbose flag is interesting, it already gives you the lines affected per issue class - so with that we'd be able to get a much more reliable mapping right?

sils:
So if I understand your patch right it assumes that the errors are in the same order, meaning if our patch splitting is wrong all the results get shifted by one and the messages will all be wrong, right? That won't be the case when relying on the lines.

JohnMoutafis:
@sils The `--verbose` option is a good start but I believe that it does not give an error by error analysis. 
I will read on it though to be sure.

PR #1931, opened 2017-07-15
setup: Ignore lxml on avg installation

sils:
@NiklasMM if this is good to you you'll wanna add it to your cherry PR :)

### DIFF
--- a/ignore.txt
+++ b/ignore.txt
@@ -1,0 +1,2 @@
+language-check
+lxml
--- a/setup.py
+++ b/setup.py
@@ -46,14 +46,21 @@ with open('bear-requirements.txt') as requirements:
 with open('test-requirements.txt') as requirements:
     test_required = requirements.read().splitlines()
 
+with open('ignore.txt') as ignore:
+    ignore_requirements = ignore.read().splitlines()
+
 with open('README.rst') as readme:
     long_description = readme.read()
 
 extras_require = {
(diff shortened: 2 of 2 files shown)


jayvdb:
This doesnt help the average user, as they use pip, and pip uses the whl's , and wheel's dont run `setup.py`.

https://github.com/coala/coala-bears/issues/1476 would provide one way to do this, effectively marking some bears as 'use at your own risk'.

And of course https://github.com/coala/coala-bears/issues/1000 will solve this properly. (But that is waiting on improvements to package_manager which nobody seems to wont to work on unless I crack the whip).

sils:
> This doesnt help the average user

It does, right now coala can't be installed without installing lxml via distribution and it gets the user a pretty mad looking error which isn't easy to understand.

With this we just don't install it by default, similar situation as language-check back then.

-------

Another story:

I'm actually looking at dependency_manager/package_manager right now because the build is broken due to an issue there that we never found yet but once I enabled the super() check_prerequisites check for linter we got lots of more testcases.

On linux this worked totally fine but on windows we get exceptions from dependency manager.

Asnelchristian:
ack https://github.com/coala/coala-bears/pull/1931/commits/125acf4bac3ba782b45187e87cabfa9700e389e7

Asnelchristian:
@rultor merge

PR #1937, opened 2017-07-20
bears/python: Add ShebangBear

AMR-KELEG:
Closes https://github.com/coala/coala-bears/issues/1932

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


bears/python/ShebangBear.py
```diff
@@ -0,0 +1,34 @@
...
+class ShebangBear(LocalBear):
+    LANGUAGES = {'Python', 'Python 2', 'Python 3'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Syntax'}
+
+    def run(self, filename, file):
+        """
+            Ensure that the file begins with the Shebang operator.
```
Asnelchristian: remove indentation here, should be 
```
"""
Ensure that the ...
"""
```

bears/python/ShebangBear.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
+            return
+
+        diff = Diff(file)
+        mes = ''
+
+        if '#!' in file[0]:
```
Asnelchristian: just check for ``#!/usr/bin/env``

Asnelchristian: having #! in the first line don't necessarily mean that it was intended to be the shebang operator

Asnelchristian: so you just add the shebangs if ``#!/usr/bin/env`` it is not found

bears/python/ShebangBear.py
```diff
@@ -0,0 +1,34 @@
...
+            mes = 'Use the env variable'
+        else:
+            diff.add_line(0, '#!/usr/bin/env')
+            mes = 'Add the she bangs line'
+
+        yield Result(
+            self,
+            mes,
+            diff.affected_code(filename),
+            diffs={filename: diff})
```
Asnelchristian: and tests for this bear

bears/python/ShebangBear.py
Asnelchristian: or what do you think?

bears/python/ShebangBear.py
AMR-KELEG: Ok :+1:

bears/python/ShebangBear.py
AMR-KELEG: The referenced issue requires that coala enforces using the #!usr/bin/env (https://github.com/coala/coala-bears/issues/1932).
Additionally, I am not familiar with the unix flags so i don't know whether #! can be used for purposes other than the shebang operator.

yukiisbored: A shebang basically specifies the shell for the command to call the interpreter (like `python`, `ruby`, `php`, or another UNIX shell). We want to enforce the usage of using `/usr/bin/env` to get the proper path of the interpreter.

AMR-KELEG: @yukiisbored so first of all, the bear should be moved from the python bears directory.
Now, let's discuss the case where the user uses something like `#!/usr/bin/python` , how can we check that the user doesn't  use the general form of the shebang operator and then deduce that script is written in python(or any other language) ?
Do you think that the current check can be improved somehow? 
Thanks 😃

bears/python/ShebangBear.py
```diff
@@ -0,0 +1,34 @@
...
+            Ensure that the file begins with the Shebang operator.
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
+            return
+
+        diff = Diff(file)
+        mes = ''
+
+        if '#!' in file[0]:
+            diff.modify_line(1, '#!/usr/bin/env')
```
sils: you don't want `#!/usr/bin/env` as your shebang without any binary right? That's just a directory

yukiisbored: s/directory/program that lists the current exported environment/

bears/python/ShebangBear.py
yukiisbored: @AMR-KELEG You can just grab the binary name (without the directory) and add that at the end of `/usr/bin/env `.

ex.
1. `#!/usr/bin/python` -> `#!/usr/bin/env python`
2. `#!/usr/bin/python3` -> `#!/usr/bin/env python3`
3. `#!/usr/bin/asuka --intepreter-mode --no-warnings` -> `#!/usr/bin/env asuka --intepreter-mode --no-warnings`

yukiisbored: Also, I think the fitting category is between `shell` and `general` (depends on how you look at it ;))

bears/python/ShebangBear.py
AMR-KELEG: @sils Yes, you're right.
The modified line should include the binary.
`#!/usr/bin/env` isn't enough.
The modified line should look like: `#!/usr/bin/env python`.
@yukiisbored What do you think?

yukiisbored: @AMR-KELEG It's ok. Please don't wait for our response just do the changes needed ASAP. We're pretty busy with a lot of stuff and we don't want to have this PR delayed just because of us~

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def run(self, filename, file):
+        """
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
+            return
+
+        diff = Diff(file)
+        mes = None
```
meetmangukiya: I assume this is message, use `message` or `msg` or something that is more clear as a message

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,31 @@
...
+
+        if '#!' in file[0]:
+            binary = file[0].split('bin/', 1)[1]
+            diff.modify_line(1, '#!/usr/bin/env {}'.format(binary))
+            mes = 'Use the /usr/bin/env.'
```
meetmangukiya: use `'Use /usr/bin/env {}'.format(binary)` instead ?

bears/general/ShebangBear.py
AMR-KELEG: Ok

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,32 @@
...
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
+            return
+
+        diff = Diff(file)
+        message = None
+
+        if '#!' in file[0]:
+            binary = file[0].split('bin/', 1)[1].strip('\n')
```
yukiisbored: Can you please use [`os.path.basename`](https://docs.python.org/3/library/os.path.html#os.path.basename) instead of this? Since, it causes problems if it's `/usr/sbin/`, `/sbin/`, `/usr/local/sbin`, etc.

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Syntax'}
+
+    def run(self, filename, file):
+        """
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
```
yukiisbored: I think this won't work if a program or an argument is defined into `env`.
Here are a couple of examples on how `env(1)` is used in a script:
```
#!/usr/bin/env bash
```
```
#!/usr/bin/env python3
```
```
#!/usr/bin/env ruby2.4 -w
```
```
#!/usr/bin/env guile --some-arguments ---and-some-more-arguments -a -fc
```

yukiisbored: I also noticed an empty list is `False` and I think it's better to do it that way.
https://stackoverflow.com/questions/53513/best-way-to-check-if-a-list-is-empty#53522

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+            return
+
+        diff = Diff(file)
+        message = None
+
+        if '#!' in file[0]:
+            binary = os.path.basename(file[0]).strip('\n')
+            new_line = '#!/usr/bin/env {}'.format(binary)
+            diff.modify_line(1, new_line)
+            message = 'Use the {}.'.format(repr(new_line))
```
yukiisbored: I find this message to be unclear especially for uses who don't know what is `/usr/bin/env`. Any ideas? I'm not sure how to phrase it as well.

yukiisbored: "Use `/usr/bin/env`" and then a more info "`/usr/bin/env` is preferable to be used because blah blah blah ..."

bears/general/ShebangBear.py
yukiisbored: I think `.startswith('#!/usr/bin/env')` is better...

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
+            return
+
+        diff = Diff(file)
+        message = None
+
+        if '#!' in file[0]:
```
yukiisbored: Same as https://github.com/coala/coala-bears/pull/1937/files#r129461801 but just check for `#!`.

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def run(self, filename, file):
+        """
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if len(file) == 0 or '#!/usr/bin/env' in file[0]:
+            return
+
+        diff = Diff(file)
+        message = None
```
yukiisbored: I find both of the above to be tricky when reading it. I think it's better to put it inside the if statement below and just yield it inside it as well.

yukiisbored: Basically,
```python
if '#!' in file[0]:
    diff = Diff(file)
    ...
    message = "message goes here"
    ...
    yield ...
```

yukiisbored:
Also, Don't forget to squash your commits. Since this is only one "change".

yash-nisar:
Please address @yukiisbored 's comments after which we can have another iteration.

RaiVaibhav:
Rebase please.

sils:
@gitmate-bot rebase

AMR-KELEG:
@Makman2 Here we go 😄

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Syntax'}
+
+    def run(self, filename, file):
+        """
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if not file or file[0].startswith('#!/usr/bin/env'):
+            return
```
Makman2: Maybe we should also return a result if the file is empty, saying there's no shebang?

Makman2: Maybe without a patch, but at least saying it's missing ;)
(oh would be so cool if we would have custom result actions, then we could write one where the user can input immediately the name of the binary... but we don't have it yet :cry:)

Makman2: Could you also add a newline here? I think it improves readability :+1:

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+        if file[0].startswith('#!'):
+            diff = Diff(file)
+            message = None
+            binary = os.path.basename(file[0]).strip('\n')
+            new_line = '#!/usr/bin/env {}'.format(binary)
```
Makman2: I think you are using `#!/usr/bin/env` quite often, I think it's worth to put it inside a variable as a constant :+1:

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+    def run(self, filename, file):
+        """
+        Ensure that the file uses the generic Shebang operator.
+        """
+        if not file or file[0].startswith('#!/usr/bin/env'):
+            return
+        if file[0].startswith('#!'):
+            diff = Diff(file)
+            message = None
+            binary = os.path.basename(file[0]).strip('\n')
```
Makman2: don't you have to strip off the `#!` part in `file[0]`?
So in the end you have `file[0][2:]`?

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+            new_line = '#!/usr/bin/env {}'.format(binary)
+            diff.modify_line(1, new_line)
+            message_desc = ('This eliminates the limitations caused by systems'
+                            ' that have non-standard file system layout')
+            message = 'Use the {}.\n{}.'.format(repr(new_line), message_desc)
```
Makman2: Hmm I'm not that convinced of the message, but can't think of a better one right now... we'll discuss it again ;)

Makman2: But please inline `message_desc`, I think it's just an unnecessary extra variable ;) Then you don't have to use two placeholders ;)

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+
+class ShebangBear(LocalBear):
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Syntax'}
+
+    def run(self, filename, file):
+        """
+        Ensure that the file uses the generic Shebang operator.
```
Makman2: I think you can be precise and describe the idea of using `/usr/bin/env` ;)

tests/general/ShebangBearTest.py
```diff
@@ -0,0 +1,61 @@
...
+        self.check_results(
+            self.uut,
+            file_contents,
+            [Result.from_values('ShebangBear',
+                                message='Use the {}.\n{}.'.format(
+                                    repr(correct_operator),
+                                    message_desc),
+                                line=1,
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file=get_testfile_path(file_name))],
```
Makman2: Can you also assert for the diff? :)

tests/general/ShebangBearTest.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def setUp(self):
+        self.section = Section('')
+        self.uut = ShebangBear(self.section, Queue())
+
+    def test_basic_usage(self):
+        self.check_validity(self.uut,
+                            ['#!/usr/bin/env python \n print hello world'])
+        self.check_validity(self.uut,
+                            [''])
```
Makman2: I think an empty file should be written as an empty list, that's how python handles it :+1:

Makman2: Maybe you should put those into test files too? Up to you :)

Makman2:
The `__init__.py` inside the test files directory is not necessary, as you don't place python modules there for coala use, they are just test files ;)

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
...
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+
+
+class ShebangBear(LocalBear):
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_FIX = {'Syntax'}
```
Makman2: Hmm not sure if it's syntax... do we have a better fitting keyword?

bears/general/ShebangBear.py
```diff
@@ -0,0 +1,33 @@
+import os
```
Makman2: please add a newline here to separate system imports from others ;)

Makman2:
unack 90a7bd6

bears/general/ShebangBear.py
AMR-KELEG: It is only written twice here..
Putting it inside a variable would be only beneficial if we will need to change it later..

bears/general/ShebangBear.py
AMR-KELEG: I don't know if i understood your comment but the basename extracts the binary file out of the path.
https://docs.python.org/2/library/os.path.html#os.path.basename
Why would i strip the #! if i don't care what the first part of the path is?

bears/general/ShebangBear.py
Makman2: it's also for consistency :+1:
Everything that's used more than once shouldn't be inlined ;)

bears/general/ShebangBear.py
Makman2: but is `os.path.basename` able to handle it? As the path starts after `#!`

bears/general/ShebangBear.py
Makman2: You aren't correcting syntax errors... hmm... Could be that you have to create a new keyword, maybe `Shebang` itself works?
You could maybe get some ideas from the community ;)

bears/general/ShebangBear.py
AMR-KELEG: https://github.com/python/cpython/blob/3.6/Lib/posixpath.py#L142
The method just extracts all the characters after the last separator.
IMO, we shouldn't strip off the `#!`.
What do you think?

tests/general/ShebangBearTest.py
AMR-KELEG: I couldn't figure out how to assert it.
https://github.com/coala/coala/blob/master/coalib/results/Result.py#L161

AMR-KELEG: @Makman2 could you please give me a hint?
Thanks :smile:

bears/general/ShebangBear.py
Makman2: Those characters don't belong to the path, that's why I'm asking for it, as conceptually it's not "completely right". I would do `file[0][2:]` for the sake of comprehensibility ;)

tests/general/ShebangBearTest.py
Makman2: just add `diff` to the result here :+1:

PR #1943, opened 2017-07-23
DocumentationStyleBear: Add blankline check

damngamerz:
DocumentationStyleBear now checks for blankline
before and after the docstring and fixes them.

Closes https://github.com/coala/coala/issues/4200

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/documentation/DocumentationStyleBear.py
+++ b/bears/documentation/DocumentationStyleBear.py
@@ -21,15 +21,15 @@ class DocumentationStyleBear(DocBaseClass, LocalBear):
     CAN_FIX = {'Documentation'}
 
     def process_documentation(self,
-                              parsed,
+                              doc_comment: DocumentationComment,
                               allow_missing_func_desc: str=False,
                               indent_size: int=4,
                               expand_one_liners: str=False):
         """
         This fixes the parsed documentation comment.
 
-        :param parsed:
-            Contains parsed documentation comment.
+        :param doc_comment:
+            Contains instance of DocumentationComment.
         :param allow_missing_func_desc:
             When set ``True`` this will allow functions with missing
             descriptions, allowing functions to start with params.
@@ -40,6 +40,7 @@ class DocumentationStyleBear(DocBaseClass, LocalBear):
         :return:
             A tuple of fixed parsed documentation comment and warning_desc.
         """
+        parsed = doc_comment.parse()
         # Assuming that the first element is always the only main
         # description.
         metadata = iter(parsed)
@@ -92,7 +93,36 @@ class DocumentationStyleBear(DocBaseClass, LocalBear):
             new_desc = new_desc.rstrip() + '\n'
 
             new_metadata.append(m._replace(desc=new_desc.lstrip(' ')))
-        return (new_metadata, warning_desc)
+
+        new_comment = DocumentationComment.from_metadata(
+            new_metadata, doc_comment.docstyle_definition,
+            doc_comment.marker, doc_comment.indent, doc_comment.position)
+
+        # Instantiate default padding.
+        class_padding = doc_comment.docstyle_definition.class_padding
+        function_padding = doc_comment.docstyle_definition.function_padding
+        # Check if default padding exist in the coalang file.
+        if (class_padding != DocstyleDefinition.ClassPadding('', '') and
+                function_padding != DocstyleDefinition.FunctionPadding(
+                    '', '')):
+            # Check docstring_type
+            if doc_comment.docstring_type == 'class':
+                new_comment.top_padding = class_padding.top_padding
+                new_comment.bottom_padding = class_padding.bottom_padding
+            elif doc_comment.docstring_type == 'function':
+                new_comment.top_padding = function_padding.top_padding
+                new_comment.bottom_padding = function_padding.bottom_padding
+            else:
+                # Keep paddings as they are originally.
+                new_comment.top_padding = doc_comment.top_padding
+                new_comment.bottom_padding = doc_comment.bottom_padding
+        else:
+            # If there's no default paddings defined. Keep padding as
+            # they are originally.
+            new_comment.top_padding = doc_comment.top_padding
+            new_comment.bottom_padding = doc_comment.bottom_padding
+
+        return (new_comment, warning_desc)
 
     def run(self, filename, file, language: str,
             docstyle: str='default', allow_missing_func_desc: str=False,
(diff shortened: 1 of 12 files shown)


NiklasMM:
This feels wrong. Can you find another way that doesn't move fixing functionality to the diff generation method?

damngamerz:
@NiklasMM I tried but the thing is markers are added to a docstring only after `.assemble()` and thats the part where only I could do the amending of blank spaces. There's no way I could have done this in `process_documentation`

NiklasMM:
Seems like you could fix it in `DocumentationComment` then.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -134,6 +134,11 @@ def run(self, filename, file, language: str,
                 new_metadata, doc_comment.docstyle_definition,
                 doc_comment.marker, doc_comment.indent, doc_comment.position)
 
+            if file[doc_comment.range.start.line - 2] != '\n':
```
NiklasMM: This will now add paddings for all languages. Is this what we want? Don't you want to define this in the style definitions?

damngamerz: Ahh I didnt get you. Yup we only want this. To follow PEP-257 standard, we just check if there's a blank line before and after the docstring. If its there then good if not we apply a fix. Also I would like to have your idea about padding applied in DocumentationAPI. Which will give diversity to other developers :tada:

NiklasMM: But this is not exclusive to Python code is it?

damngamerz: No its not I suppose!! That's why I want your idea to get implemented :smile:

Nosferatul:
looks good to me.

bears/documentation/DocumentationStyleBear.py
jayvdb: this bear is defined as
```
LANGUAGES = {language for docstyle, language in
DocstyleDefinition.get_available_definitions()}
```

you put that there.  It is part of your project to support multiple languages (Java and Python according to https://gitlab.com/coala/GSoC-2017/issues/30).  The bear should be language agnostic.

jayvdb: I note that we already had this 'chat' about not using literal values at the end of the last phase, regarding indent size.  Not happy that you need to be told again to design for multiple languages.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -92,7 +96,16 @@ def process_documentation(self,
             new_desc = new_desc.rstrip() + '\n'
 
             new_metadata.append(m._replace(desc=new_desc.lstrip(' ')))
-        return (new_metadata, warning_desc)
+        new_comment = DocumentationComment.from_metadata(
+            new_metadata, doc_comment.docstyle_definition,
+            doc_comment.marker, doc_comment.indent, doc_comment.position)
+
+        if file[doc_comment.range.start.line - 2] != '\n':
```
adtac: Quick qustion: I haven't been following your work, like at all, but a comment could potentially start in the 0th or 1st line, right? Then you'd be checking the last/last-but-one line (0 - 2 = -2), which is probably not what you intended over here (same applies to the if-clause below).

jayvdb: C.f. earlier discussion https://github.com/coala/coala-bears/pull/1943#discussion_r129030418

tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test.correct
```diff
@@ -1,4 +1,5 @@
 def improper_indents(indents):
+
```
jayvdb: this is incorrect.

tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test.correct
```diff
@@ -9,4 +10,5 @@ def improper_indents(indents):
         3 space indent
         4 space indent
     """
+
```
jayvdb: this is also incorrect

tests/documentation/test_files/DocumentationStyleBear/bad_file.py.test.correct
damngamerz: How? I think this issue is still valid... Based on this https://github.com/coala/coala-bears/pull/699#discussion_r98310815
I and Sanket agreed to have a blank line before and after docstring.

jayvdb: That comment that you link points to PEP 257, and quotes from it, showing why this is wrong.

I accepted that the bear didnt enforce that aspect of PEP257 because it was the first merge of the bear, which would have to improve later.
And now your project is here, and you *are* enforcing a PEP257 violation.

damngamerz: @jayvdb Oh it's the other way around :O . What I and Sanket discussed while before the GSoC started that we should have a "blank line before and after" which I guess you wanted to portray as voilation. Tell me if im wrong? Though I will just take a quick look at PEP - 257.

damngamerz: Okay so we actually do the violation of PEP-257 but not fully.
For one-liners:
"There's no blank line either before or after the docstring. "
For multi-line docstrings:
"Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line. "
And Im confused :confused:

NiklasMM: I think what this merge should (and does) bring to the table is the ability of the bear to actually check and fix paddings both on the top and the bottom, depending on what the style demands. The actual issue that this now violates PEP257 lies in the style definition [here](https://github.com/coala/coala/pull/4557/files#diff-eb7df11f93546c67d1807d6649d54216) <- this is wrong and hence the files in this PR are wrong. 

You should probably write tests that don't rely on the styles defined in coala, but rather use ad-hoc styles. It will make it clearer what happened anyway.

jayvdb: @damngamerz , `def` have different padding rules to `class`.  This is distinction is common in styles of other languages also.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -40,6 +40,7 @@ def process_documentation(self,
         :return:
```
SanketDG: https://github.com/coala/coala-bears/issues/2009

tests/documentation/test_files/DocumentationStyleBear/bad_file3.py.test
```diff
@@ -30,9 +30,11 @@ class docstring_if_indented():
             """
             This is `if` indented block function.
             """
+            return None
```
SanketDG: any reason for adding these, when they were not there previously?

damngamerz: Suitable `return` statements are added just to show our new padding system neatly. Earlier this wasn't a problem but now the below happens when `return` is not there... 
eg:
official function padding->0,0
```
         def hello_venus(self):
             """
             This is `if` indented block function.
             """
 
 
 def docstring_inner_function(dummy):
```
So this will be amended to
```
         def hello_venus(self):
             """
             This is `if` indented block function.
             """
def docstring_inner_function(dummy):
```
This is the intended behaviour :wink: . But according to PEP-257 module function should have 2 lines in between. which is violated in amended result. Hence it's added just to show paddings neatly, also to avoid the violation of PEP-257. :wink:

SanketDG:
ack 89b5913

SanketDG:
@rultor merge

PR #1944, opened 2017-07-23
GitCommitBear: Block invalid email addresses

rhemon:
Uses validate_email to check invalid email
addresses in git commit message.

Fixes https://github.com/coala/coala-bears/issues/1791

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


rhemon:
clueless with the problem in circle
and also not understanding why output is different in travis, while locally test case passes and also properly identifies if email is valid or not
```
(i1791-venv) MBPs-MacBook-Pro:coala-bears rhemon$ pytest -k GitCommitBearTest 
============================= test session starts ==============================
platform darwin -- Python 3.5.2, pytest-3.1.3, py-1.4.34, pluggy-0.4.0
rootdir: /Users/mbp13inch/Documents/Workspace/coala-bears, inifile: setup.cfg
plugins: cov-2.5.1, env-0.6.2, timeout-1.2.0, xdist-1.18.1
timeout: 35.0s method: signal
collected 697 items / 1 skipped 

tests/vcs/git/GitCommitBearTest.py ................

============================= 681 tests deselected =============================
============ 16 passed, 1 skipped, 681 deselected in 64.96 seconds =============
```

rhemon:
`validate_email` seems to be acting differently in travis. I set `debug=True` for validate email and it gave `Server Error ... [Error 99] Cannot assign requested address` (check [here](https://travis-ci.org/rhemon/coala-bears/jobs/256737877#L3880)) which is most probably due to [this line](https://github.com/syrusakbary/validate_email/blob/master/validate_email.py#L102) with `DNS.mxlookup`. My guess is that the address at which its sending request, does not get permission from travis? or could be related to https://github.com/syrusakbary/validate_email/issues/61

so not quite sure about the problem and trying out some other libraries now (`pyisemail`, `python_emailahoy`, etc.) 

if anyone knows how I could deal with this issue it would be great as `validate_email` checks if email exists or not, while `pyisemail`/`python_emailahoy` only checks if the domain exists or not, not necessary for the email to actually exist
of if you know of any other module that checks existence of email and will also support travis?

NiklasMM:
Hey, thanks for working on this!

If I understand correctly `validate_email` checks if the email format matches and also if a valid MX record can be obtained for the domain.

While this is theoretically cool it a bit of an overkill for our usecase. Especially the DNS lookup part makes the coala run dependent on the network topology. Also we are adding two more dependencies. 

I think in our case it's enough to check the string for validity. This can be done with a regex. Read [this](http://emailregex.com/) for some background info and Python regexes ready to be used.

rhemon:
oh ok. sorry i might have misunderstood the issue. so we only need to check of emails are in valid format, right? will do the correction soon. thanks for explaining

NiklasMM:
Well I was just adding my opinion. The issue is not detailed enough to prefer one way of solving it over the other. I think your solution is a valid one! However due to the reasons I listed above I would prefer a solution with a smaller footprint.

rhemon:
then I guess i should first clarify how it should be

rhemon:
so now it first uses regexp to confirm if email is in format, then if optinal setting for dns check is set to `True` then it uses `is_email` to look up if domain exist or not (it will still need the two dependencies)

bears/vcs/git/GitCommitBear.py
```diff
@@ -260,6 +268,25 @@ def check_body(self, body,
                                'Commit body lines should not exceed {} '
                                'characters.'.format(body_line_length))
 
+        if valid_email_check:
+            result_message = 'Body contains these invalid emails:\n'
+            invalid_emails = False
+            for line in body:
+                for email in re.findall(r'\S*@\S*', line):
```
NiklasMM: This regex will also find things that could not be intended to be E-Mail adresses, like for example Github user names, e.g. `@NiklasMM`

I would require at least one charater on each side and maybe even a dot: `r'\S+@\S+\.\S+`

bears/vcs/git/GitCommitBear.py
```diff
@@ -260,6 +268,25 @@ def check_body(self, body,
...
+                    email_check = re.match(r'(^[a-zA-Z0-9_.+-]'
+                                           + r'+@[a-zA-Z0-9-]+\.'
+                                           + r'[a-zA-Z0-9-.]+$)', email)
+                    # if eamil do not match regex there is no need to
```
NiklasMM: `eamil` -> `email`

bears/vcs/git/GitCommitBear.py
```diff
@@ -260,6 +268,25 @@ def check_body(self, body,
...
+                    # look up dns
+                    if email_dns_check and email_check:
+                        email_check = not is_email(email, diagnose=True,
+                                                   check_dns=True).ERROR_CODES
+                    if not email_check:
+                        invalid_emails = True
+                        result_message += ' ' + email + '\n'
```
NiklasMM: An approach a bit more elegant would be to collect the invalid addresses in a list instead of using the `invalid_emails` flag and building the string. You could check if the list is empty in the end and then use `"\n".join()` to build the string more effectively.

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -288,6 +288,24 @@ def test_body_checks(self):
...
+                        'validdomain@gmail.com\n'
+                        'Fix 2017')
+        message = 'Body contains these invalid emails:\n'
+        message += ' invalid#!format!*!@email.com\n'
+        message += ' another.*invalid@format\n'
+        # Test with email_dns_check set to False
+        self.assertEqual(self.run_uut(), [message])
+        # Test with email_dns_check set to True
+        message += ' validformat@email.check\n'
+        self.assertEqual(self.run_uut(email_dns_check=True), [message])
```
NiklasMM: Is the test here actually doing the DNS lookup? If so, please mock it away, we don't want out test suite to be network dependent.

bears/vcs/git/GitCommitBear.py
```diff
@@ -260,6 +268,25 @@ def check_body(self, body,
                                'Commit body lines should not exceed {} '
                                'characters.'.format(body_line_length))
 
+        if valid_email_check:
+            result_message = 'Body contains these invalid emails:\n'
+            invalid_emails = False
+            for line in body:
+                for email in re.findall(r'\S*@\S*', line):
+                    email_check = re.match(r'(^[a-zA-Z0-9_.+-]'
```
NiklasMM: Is this regex taken from the website I suggested? If so, please document this in a comment.

tests/vcs/git/GitCommitBearTest.py
rhemon: yes this does the test with DNS lookup. sorry but I am not clear on what you meant by mock it away? if I dont run it, how else can I get the result ? also without running that it wont reach 100% coverage right?

bears/vcs/git/GitCommitBear.py
rhemon: so cases like `abc@foo` should be ignored ?

tests/vcs/git/GitCommitBearTest.py
NiklasMM: With the [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) module you can temporarily replace certain functions with mocks, so you don't actually have to run them. You can control the output and thus test your code under certain situations. 

In you case you could for example do something lilke this:

``` python
# test successful DNS lookup
with mock.patch("bears/vcs/git/GitCommitBear.is_email", return_value=[]):
   message = "... whatever your expected output is" 
   self.assertEqual(self.run_uut(email_dns_check=True), [message])

# test failing DNS lookup (returning a non-empty list is enough, but you could replace it with an actual error
with mock.patch("bears/vcs/git/GitCommitBear.is_email", return_value=["error"]):
   message = "... whatever your expected output is" 
   self.assertEqual(self.run_uut(email_dns_check=True), [message])

This way you still test all the functionality of the code that you wrote, while not actually going out to the net.

**Comment by rhemon** [comment]
is it alright if I do the author's email check within `check_body` ? or do I need to have separate function for `check_user_email` ?

**Comment by NiklasMM** [comment]
It should be a separate function as both need to provide different error messages. It would also be nice, if users could switch both functionalities on and off independently via setting.

**Comment by rhemon** [comment]
so to get the author's email I am using `run_shell_command('git config --global user.email')`. and am able to make the bear check author's email. but then the problem comes in tests. In some of the test cases `run_shell_command` is mocked and so output changes, and causes those test to fail. so is it alright in those test cases I set the `auth_email_check` to set false? that way in those cases it wont yield the error result for invalid auth email.

**Comment by rhemon** [comment]
https://travis-ci.org/coala/coala-bears/jobs/259042497#L4022
am I not supposed to add them in `bear-requirements.txt` ?

File: bears/vcs/git/GitCommitBear.py [hunk 6]
```diff
@@ -134,6 +145,34 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
...
+        :param auth_email_check:      If True, it will check validity of
+                                      author's email.
+        :param email_dns_check:       If True, it will confirm the domain
+                                      in author's email is valid.
+        """
+        if not auth_email_check:
+            return
+
+        # Used regex provided in http://emailregex.com/
+        email_check = re.match(r'(^[a-zA-Z0-9_.+-]'
```
jayvdb: why do we have our own regex?  validity checks are provided by upstream libraries, and are maintained.
The validity of an email address is actually very complicated to do without network activity, and that algorithm should be maintained upstream, not in a coala bear.

rhemon: alright then I will be just using `is_email` to do that. and about using `validate_email`, it was my first try but it was giving Server Error in travis. See [comment](https://github.com/coala/coala-bears/pull/1944#issuecomment-317323142). thats why i shifted to pyisemail

bears/vcs/git/GitCommitBear.py
```diff
@@ -124,6 +132,9 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
                 yield Result(self, 'HEAD commit has no message.')
             return
 
+        yield from self.check_auth_email(
+            auth_email[0][:-1],  # Taking the email and removing \n
```
NiklasMM: Input sanitation should be done as early as possible, in this case, when you first get the email adress.

bears/vcs/git/GitCommitBear.py
```diff
@@ -134,6 +145,30 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
...
+                         email_dns_check: bool=False):
+        """
+        Checks validity of author's email
+
+        :param auth_email_check:      If True, it will check validity of
+                                      author's email.
+        :param email_dns_check:       If True, it will confirm the domain
+                                      in author's email is valid.
+        """
+        if not auth_email_check:
```
NiklasMM: I would prefer doing this check outside of the function and not calling it if the check is not requested.

bears/vcs/git/GitCommitBear.py
```diff
@@ -134,6 +145,30 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
...
+            return
+
+        if not email_dns_check:
+            email_check = is_email(auth_email)
+        else:
+            email_check = not is_email(auth_email, diagnose=True,
+                                       check_dns=True).ERROR_CODES
+
+        if not email_check:
+            yield Result(self,
```
NiklasMM: Please yield different results for the two possible errors. It's important for the user to know if the email is syntactically incorrect or if we can't resolve the DNS (possible network issue)

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,10 +125,60 @@ def test_empty_message(self):
...
+        self.run_git_command('config', 'user.email coala@coala.io')
+
+        # Valid email with no dns check
+        self.git_commit('Shortlog\n\n'
+                        'some text\n'
+                        'Fix 2017')
+        self.assertEqual(self.run_uut(), [])
+
+        # Return Object for successful mock
+        class ReturnObject:
```
NiklasMM: you don't need to create a class here, just create a Mock object like so:

```Python
return_object = mock.Mock(ERROR_CODES={})
```

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,10 +125,60 @@ def test_empty_message(self):
...
+        with unittest.mock.patch('bears.vcs.git.GitCommitBear.is_email',
+                                 return_value=ReturnObject()):
+            self.git_commit('Shortlog\n\n'
+                            'some text\n'
+                            'Fix 2017')
+            self.assertEqual(self.run_uut(email_dns_check=True), [])
+            self.assert_no_msgs()
+
+        # Return Object for failing mock
+        class ReturnObject:
```
NiklasMM: Same here, use a `Mock` object ;-)

bears/vcs/git/GitCommitBear.py
jayvdb: Regarding `validate_email`, that problem would go away if you disable its smtp validator, which is the appropriate setting anyway... but `is_email` does look to be more mature, with lots of tests, so good choice! :+1:

bear-requirements.txt
```diff
@@ -8,6 +8,7 @@ cmakelint~=1.3
 cppclean~=0.12.0
 cpplint~=1.3
 dennis~=0.8
+dnspython~=1.15.0
```
rhemon: oh ok thanks

rhemon:
@jayvdb cricle still failing 
```
File "caml_startup", line 1:
Error: Error during linking
Command exited with code 2.
make[2]: *** [build_java] Error 10
make[2]: Leaving directory `/home/ubuntu/infer-linux64-v0.7.0/infer/src'
make[1]: *** [java] Error 2
make[1]: Leaving directory `/home/ubuntu/infer-linux64-v0.7.0/infer'
make: *** [java] Error 2

  compilation failure; you can try running

    make clean
    ./build-infer.sh java


bash .ci/deps.opam.sh returned exit code 1

Action failed: bash .ci/deps.opam.sh
```
and travis is still pending?

rhemon:
circle is still failing at `.ci/deps.opam.sh`, and now appveyor is giving some test failures for bears that are not even modified in this PR.
😕

PR #1967, opened 2017-08-01
WIP: Add powershell/PSScriptAnalyzerBear

userzimmermann:
While attending @PPOSHGROUP's https://www.meetup.com/Polish-PowerShell-Group-PPoSh/events/241810275/ I realized that we haven't grown any *powerBear* for analyzing PowerShell code yet 😨 How do we ever want to get really accepted by the Windows community w/o the `PSScriptAnalyzerBear`? Here it is! And needs some mapping to all the https://github.com/PowerShell/PSScriptAnalyzer rules :D And tests of course! So there is still some work left to do...

@mgr32 Hi :) I was talking about coala and PS integration with @mczerniawski and he said you might be interested in such stuff ;) Are there maybe even more PS code analysis tools worth integrating?

cc @sils


mgr32:
Hi @userzimmermann, thanks for letting me know about coala, I've been a long-time user of SonarQube (but it doesn't have plugin for Powershell) 😃 As for other PS analysis tools I had been using ScriptCop, but I believe it's not maintained any more, so I would stick with PSScriptAnalyzer.

userzimmermann:
@Makman2 I added this `@Rulified` for automagically turning all PSScriptAnalyzer rules into Bear settings. I know this purely dynamical approach is not a good solution for a release installation. But what do you anyway think about generalizing this? And then creating some static bundled setting definitions on building `bears` releases from it... For this case maybe looking like:

```python
@linter(executable='powershell', settings={name: (bool, True) for name in rules()})
class PSScriptAnalyzerBear:
    ...
    # @linter can create the wrapper here or we extend the settings discovery routines
    @staticmethod
    def create_arguments(filename, file, config_file, **rules):
        ...
```

cc @sils @jayvdb

userzimmermann:
@mgr32 Thanks for your response :) Already tried coala in the meantime? ;) The nice advantages over SonarQube are the simple extensibility for new languages/tools, as you can see it here for PSScriptAnalyzer, and the CI integration with GitHub/Lab via @gitmate-bot which makes those nice PR comments like above in https://github.com/coala/coala-bears/pull/1967#pullrequestreview-53791214 and blocks the PR from getting merged until cleaned up... If you have further ideas or remarks regarding this PR or coala in general, please let us know :)

Also @mczerniawski @DomBros @bielawb ... If anything comes to your mind regarding this ... Can't wait to read it ;)

Makman2:
`@Rulify` looks promising. However where's the advantage over inlining `rules()` now into the function signature of `create_arguments`? Often parameters are processed further.
Also when having `generate_config` and `create_arguments` together, isn't this slightly overhead to pass the full list of rules/arguments to both functions?

userzimmermann:
@Makman2 I hope I got you right ;) There just doesn't seem to be another way to define Bear settings than adding them as parameters to `create_arguments`, or to `run` in the non-`@linter`-case. Correct me please! :P So if you want to create settings programmatically, you have do such kinds of hacks again and again. That's why I would like to create some generic API for programmatically defining them. The Bear-developer-defined `create_arguments` and `generate_config` methods then only have `**settings` and don't need additional `@decorator`s

Makman2:
Actually you can override `get_metadata` which is exactly doing this. coala takes the information to see what settings are required and optional and passes them to `run`. linter defines some additional "private" functions that further delegate those parameters to the `create_arguments`, `process_output` and `generate_config` functions: https://github.com/coala/coala/blob/master/coalib/bearlib/abstractions/Linter.py#L228-L253

We could make those public and allow to modify those, so you can define parameters like you want. But I want to be sure that this is a good thing to do ;)

However, I'm not sure that what you try to achieve for this bear specifically is a good approach. If you have different versions of PSScriptAnalyzer (not sure how powershell includes this commandlet), the bear signature changes. This is imo not desired, and we should have a fixed and static list of arguments coala supports. Also we are able to adapt style/rule names to our naming scheme (instead of taking them directly from powershell).

userzimmermann:
@Makman2 Thanks! The `get_metadata` override is a good idea :)

> If you have different versions of PSScriptAnalyzer (not sure how powershell includes this commandlet), the bear signature changes. 

This is exactly what I was referring to with *"I know this purely dynamical approach is not a good solution for a release installation."*. And with the solution approach *"And then creating some static bundled setting definitions on building bears releases from it..."* we could always roll out the same pre-defined settings based on the linter version in our build system

> This is imo not desired, and we should have a fixed and static list of arguments coala supports.  Also we are able to adapt style/rule names to our naming scheme (instead of taking them directly from powershell).

Why reinventing the wheel? And we don't have a real naming scheme for bear settings. Every linter bear developer creates them on their own, mostly manually resembling the tool's native options. Of course! Why reinventing the wheel? ;) And this is usually done once. Haven't seen some pull request like `Update ...Bear settings for new linter version`. There is usually no maintenance to keep up with linter development going on. And the number of PSScriptAnalyzer rules is huge! So a (semi-)automatic approach actually seems beneficial to me. I can of course decamelize the PSScriptAnalyzer rule names and remove the `PS` prefixes, so that we have for example `align_assignment_statement` instead of `PSAlignAssignmentStatement`

All that changes of course with aspects... Then we really have to think about it and need to keep the aspects<-->tool-options relations up to date

Makman2:
> Why reinventing the wheel? And we don't have a real naming scheme for bear settings.

Yes we actually have ongoing efforts to unify setting names to include for example prefixes like `prohibit`, `enforce`, `allow` etc. Also between different bears having same setting names for same things is quite important imo.

Another thing are the type annotations: There are settings which are boolean, some which are string, and some are integers. If it's possible to extract this information out of the documentation texts, cool. If not we can only handcraft those ;) It's technically not necessary to annotate the types as we directly pass the settings to the linter, however it's valuable documentation for the user.

It's not that I want to turn down the idea, I'm just really not sure if it's the right direction for us to go.
CC @sils @jayvdb, would like some opinions here^^

jayvdb:
IMO this is not helpful.  We need precise versions of upstream tools , managed with [package managers](http://gitlab.com/coala/package_manager/).
And we need to test each new release of a tool before pushing it downstream to users, and update the settings.

userzimmermann:
@jayvdb Exactly my ideas... I don't see any contradiction :) I only want to additionally generate the settings from that same precise version of `PSScriptAnalyzer` to be used in our build system and store them in the `bears` package. Or I just generate the Python code ;) Instead of manually mapping those 51 rules. And for managing the dependency, I will add a `PSRequirement` or `PowerShellRequirement` class to `dependency_management` ... The `WIP:` here was not intended as short-term status ;)

@Makman2 

> There are settings which are boolean, some which are string, and some are integers. If it's possible to extract this information out of the documentation texts, cool. If not we can only handcraft those ;)

Of course! Automation where possible. Handcraft where needed. That's obvious ;)

jayvdb:
@userzimmermann , how are you going to automatically map those to aspects?

And we dont want `PSScriptAnalyzer` based naming; we have consistent naming for common concepts.

You'll end up with a custom parser that is quite complex, and most importantly that parser doesnt belong in the bear.  Put it in a `.ci` script which checks the linter output to identify new settings not supported by the bear yet, and then maybe recommends the change needed to the bear.

userzimmermann:
@jayvdb I didn't write anything about automatic mapping to aspects... only settings

But it seems like manual adjustments are needed... So it looks like we could need some interactive tool-options<-->settings mapping assistant... For the development process... Something the `bearship` could offer... Should offer! Thanks for making me get this idea! :D That's going to be a great `WIP` ;)

PR #1984, opened 2017-08-07
StandardBear: Add fixing capabilities

Alexander-N:
The corrected code is retrieved by piping to JSStandard. If there
several issues in a single line, the messages are displayed together to
make it easy to understand the proposed fix.

Closes https://github.com/coala/coala-bears/issues/1952

### DIFF
--- a/bears/js/JSStandardBear.py
+++ b/bears/js/JSStandardBear.py
@@ -1,11 +1,17 @@
+from collections import defaultdict
+import re
+from subprocess import Popen, PIPE
+from typing import List, Iterable, Tuple
+
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.NpmRequirement import NpmRequirement
+from coalib.results.Result import Result
+from coalib.results.Diff import Diff
 
 
 @linter(executable='standard',
-        output_format='regex',
-        output_regex=r'\s*[^:]+:(?P<line>\d+):(?P<column>\d+):'
-                     r'\s*(?P<message>.+)')
+        use_stdin=True,
+        use_stderr=True)
 class JSStandardBear:
     """
     One JavaScript Style to Rule Them All.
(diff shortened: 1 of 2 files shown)


Alexander-N:
This PR is ready for review. Anybody?

sils:
hey @Alexander-N can you add a testcase that actually tests that the patch is correct? Like if there's multiple results from `standard` does it properly figure out which part of the diff belongs to the right result? That's a thing that I'd definitely want to see in the tests

Alexander-N:
Ok, done.

Makman2:
Sorry it took so long and we haven't done a review :/ We are a bit out of manpower in general for such things, so I hope you can understand :/

If you still want to give it a try, do you mind to resolve the conflicts and push again? I'm gonna review this then :+1:

Alexander-N:
No Problem, lets do this :-)

Conflicts resolved.

bears/js/JSStandardBear.py
```diff
@@ -24,5 +30,65 @@ class JSStandardBear:
...
         return (filename, '--verbose')
+
+    @staticmethod
+    def _get_corrected_code(old_code: List[str]) -> List[str]:
+        """
+        Pipes the code to JSStandard and returns the corrected code.
+        """
+        p = Popen(
+            ('standard', '--stdin', '--fix'),
+            stdin=PIPE, stdout=PIPE, stderr=PIPE)
```
Makman2: do we have to execute `standard` twice to get the corrected code?

bears/js/JSStandardBear.py
```diff
@@ -24,5 +30,65 @@ class JSStandardBear:
...
+        p.stdin.write(bytes(''.join(old_code), 'UTF-8'))
+        out, err = p.communicate()
```
Makman2: we have to be careful, this can deadlock if the buffers get full (and because we might operate with larger files, this can happen). Are there alternatives that take care of this?

tests/js/JSStandardBearTest.py
```diff
@@ -1,8 +1,12 @@
...
+from queue import Queue
 
+from coalib.results.Diff import Diff
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.testing.LocalBearTestHelper import verify_local_bear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 from bears.js.JSStandardBear import JSStandardBear
 
-
```
Makman2: unnecessary change ;)

bears/js/JSStandardBear.py
```diff
@@ -24,5 +30,65 @@ class JSStandardBear:
...
+        if '--fix' in stderr:
+            corrected_code = self._get_corrected_code(file)
+            if len(file) != len(corrected_code):
+                corrected_code = None
+
+        for line_number, message in self._get_issues(stdout):
+            diff = None
+            if corrected_code:
+                diff = Diff(file)
+                diff.modify_line(line_number, corrected_code[line_number-1])
```
Makman2: can a patch span multiple lines?

bears/js/JSStandardBear.py
Alexander-N: If we want the messages and the corrected code I see no other way.

bears/js/JSStandardBear.py
Alexander-N: Maybe temporary files instead? I checked some bigger files, standard itself is failing if ~500kB files are piped. I should also check the return code at this point.

bears/js/JSStandardBear.py
Alexander-N: Not that I'm aware of.

But looking at this again, it's not robust at all. If standards "--fix" would remove a line and add one at some other place, the patches would be wrong.

Alexander-N:
I'm doubtful if it is worth moving on with this. What do you think? Are there any alternatives I'm not seeing?

bears/js/JSStandardBear.py
Makman2: There should be special functions somewhere to call a program with stdin and retrieve stdout and stderr without deadlocking.

>  I should also check the return code at this point.

Oh yeah good point :+1:

Makman2:
> I'm doubtful if it is worth moving on with this. What do you think? Are there any alternatives I'm not seeing?

Hm this worries me a bit:

> But looking at this again, it's not robust at all. If standards "--fix" would remove a line and add one at some other place, the patches could be wrong.

I don't know the tools behaviour, so I can't say what edge cases might occur.

What's actually best is if the tool itself would print out a format that assigns patches to issue messages, something like
```
L3C3 Invalid identifier
-- @unified-diff patch attached --
@@ ...
<unified diff goes here>
----------
L4C4 Invalid xyz
...
```
Usually tools don't have something like this sadly (could be actually worth to define one - or maybe a few - such formats in a coala cEP). This would be the optimal solution, having an upstream contribution doing this (then we also don't have to invoke `standard` twice ;D). But this is probably very time-consuming^^

Alexander-N:
Yes I agree. Close, since the approach seems problematic.

PR #1997, opened 2017-08-14
ElmLintBear: Remove `pragma: no cover`

yash-nisar:
Remove `pragma: no cover` in `check_prerequisites` by adding
a suitable test.

Closes https://github.com/coala/coala-bears/issues/1996
Related to https://github.com/coala/coala-bears/issues/1618

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/tests/elm/ElmLintBearTest.py
+++ b/tests/elm/ElmLintBearTest.py
@@ -1,4 +1,5 @@
 from queue import Queue
+from unittest.mock import patch
 
 from bears.elm.ElmLintBear import ElmLintBear
 from coalib.results.Result import Result
@@ -143,3 +144,15 @@ class ElmLintBearTest(LocalBearTestHelper):
                 [],
                 filename=fname,
                 create_tempfile=False)
+
+    @patch('bears.elm.ElmLintBear.which')
+    def test_check_prerequisites(self, mock_method):
+        mock_method.return_value = None
+        self.assertEqual(
+            ElmLintBear.check_prerequisites(),
+            'elm-format is missing. Download it from '
+            'https://github.com/avh4/elm-format/blob/master/README.md'
+            '#for-elm-018 and put it into your PATH.')
+
+        mock_method.return_value = 'path/to/elm-format'
+        self.assertEqual(ElmLintBear.check_prerequisites(), True)
--- a/bears/elm/ElmLintBear.py
+++ b/bears/elm/ElmLintBear.py
@@ -24,7 +24,7 @@ class ElmLintBear:
     CAN_DETECT = {'Syntax', 'Formatting'}
 
     @classmethod
-    def check_prerequisites(cls):  # pragma: no cover
+    def check_prerequisites(cls):
         if which('elm-format') is None:
             return ('elm-format is missing. Download it from '
                     'https://github.com/avh4/elm-format/blob/master/README.md'


yash-nisar:
CC @jayvdb @Makman2

tests/elm/ElmLintBearTest.py
```diff
@@ -143,3 +144,18 @@ def test_good_function(self):
                 [],
                 filename=fname,
                 create_tempfile=False)
+
+    def test_check_prerequisites(self):
+        _shutil_which = shutil.which
+        try:
+            shutil.which = lambda *args, **kwargs: None
```
Makman2: use `@patch` on the test method. You can patch the `which` you imported from `shutil`, so the patch-module-string gets `bears.elm.ElmLintBear.which`. Then you don't need to change the import inside `ElmLintBear`, you just have to remove the pragma.

yash-nisar: Cool, on it. :+1:

yash-nisar:
Done @Makman2

Nosferatul:
I think this pr is ready, what do you say @Makman2 ?

Makman2:
yes looks good :)

Makman2:
ack 9d2563a

yash-nisar:
reack b1e3fbd9de6ab719edc30ad4ef344c6e6363b553

yash-nisar:
I'm doing a minor change -> Replacing `assertTrue` with `assertEqual` as in the docs.
>  This method should also be avoided when more specific methods are available (e.g. assertEqual(a, b) instead of assertTrue(a == b)), because they provide a better error message in case of failure.

yash-nisar:
reack 0a25352953ee2b3d91e1350d50153e495fc203f2

yash-nisar:
ack 0a25352953ee2b3d91e1350d50153e495fc203f2

yukiisbored:
@rultor please merge

yash-nisar:
Thanks @yukiisbored :smile:

Makman2:
> I'm doing a minor change -> Replacing assertTrue with assertEqual as in the python docs.
>
> This method should also be avoided when more specific methods are available (e.g. assertEqual(a, b) instead of assertTrue(a == b)), because they provide a better error message in case of failure.

Actually the docs want you to avoid such operators like `==` to use in `assertTrue`. If your result should be true in the end, `assertTrue` is just effective^^ however like this is also okay^^

Makman2:
But if you do a change, don't ack it yourself, it has to be reviewed again by someone else.

yash-nisar:
Sure, will keep that in mind. I felt the need to change it because I missed a bug in my travis PR. Also, I've opened [an issue](https://github.com/coala/coala-bears/issues/1998) with regard to this that will avoid future bugs 😉 :

> In general, self.assertTrue('some_string') does not fail. So even if check_prerequisites() returns 'You are not connected to the internet.', the test will pass.
This can be avoided either by explicitly doing self.assertTrue(URLBear.check_prerequisites() == True) or by self.assertEqual(URLBear.check_prerequisites(), True). The latter one is more preferred.

Makman2:
> Actually the docs want you to avoid such operators like == to use in assertTrue. If your result should be true in the end, assertTrue is just effective^^ however like this is also okay^^

Nvm this, `assertEqual(x, True)` and `assertTrue(x)` are two different things. `assertTrue` invokes `bool` on the given expression, while `assertEqual` does not.

yash-nisar:
Yeah, I agree :+1: . Just wanted to make sure things like these -> `self.assertTrue('some_string')` do not pass. :wink:

Issue #1723, opened 2017-05-13
RSTcheckBear: Allow ignoring roles and directives


jayvdb:
rstcheck has [cmdline args](https://github.com/myint/rstcheck#options) for ignoring directives and ignoring roles.

These need to be exposed via bear settings.  This is blocking a proper fix for https://github.com/coala/coala/pull/4211#issuecomment-301042935

jayvdb:
c.f. https://github.com/coala/coala-bears/pull/1718 for a similar enhancement

mridubhatnagar:
Instead of adding RSTcheckBear here. I can fix this issues in #4211 itself.

jayvdb:
> Instead of adding RSTcheckBear here. I can fix this issues in #4211 itself.

No.  This is an enhancement for RSTcheckBear.

mridubhatnagar:
Ohk so nothing has to be done with .coafile. I will have to go through RSTcheckBear.

mridubhatnagar:
I understood what needs to be done. But for checking if everything works fine. I will have to add rstcheckBear to the .coafile of coala-bears repository.
rstcheckBear is not present in .coafile of coala-bears.

jayvdb:
To test your changes, you need to add tests to `tests/rest/RSTcheckBearTest.py`, and not modify `.coafile`.

mridubhatnagar:
Missing coverage 9-15.

mridubhatnagar:
Can someone please help me with this. I am unable to get 100% code coverage. And not sure if changes are right or not.

Makman2:
@mridubhatnagar it seems you are not working on the issue any more, and @umeshksingla said he has an urgent need for this feature. If you don't mind he would take this issue over :)

@umeshksingla please accept the invitation, then I'll assign the issue to you :)
PS: Hop onto our gitter channel, so you get various help: https://gitter.im/coala/coala

PR #2018, opened 2017-08-24
RSTcheckBear: Allow ignoring directives and roles

umeshksingla:


umeshksingla:
WIP.

umeshksingla:
CircleCI fails due to other Bears and not this one, I suppose. Logs say, 

| Name | Stmts | Miss | Branch | BrPart | Cover | Missing |
| --- | --- | --- | --- | --- | --- | --- |
| bears/rest/RSTcheckBear.py  | 24 | 0 | 12 | 0 | 100% | |
| bears/rest/__init__.py  | 0 | 0 | 0 | 0 | 100% | |
| bears/rest/reSTLintBear.py | 18 | 0 | 2 | 0 | 100% | |

umeshksingla:
@Makman2 please review. Thanks.

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,32 @@ class RSTcheckBear:
...
         :param code_block_language_ignore:
             Comma seperated list for languages of which code blocks
             should be ignored. Code block of following languages
             can be detected: ``bash``, ``c``, ``cpp``, ``json``,
             ``python``, ``rst``.
+
+        :param directive_ignore:
+             Comma separated list of directives to be ignored. Can be
+             used to ignore custom directives as well.
+
```
Makman2: please remove those newlines to conform to the other docstrings in our codebase :)

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,32 @@ class RSTcheckBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
-                         code_block_language_ignore: list=()):
+                         code_block_language_ignore: typed_list(str)=(),
```
Makman2: could you make a new commit out of this change? It's good to use `typed_list(str)` here, however it's a different issue :)

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,32 @@ class RSTcheckBear:
...
+             to ignore custom roles.
         """
         args = ()
         if code_block_language_ignore:
             args = ('--ignore-language=' +
                     ','.join(code_block_language_ignore),)
-        return args + (filename,)
+        if directive_ignore:
+            directives = ','.join(part.strip() for part in directive_ignore)
+            args = ('--ignore-directives=' + directives,)
```
Makman2: you overwrite the `args` if `code_block_language_ignore` was set.

Makman2: you need to append to the `args`: `args += (...)`

I would suggest you to use a list, as appending to lists is usually faster than for tuples :+1:

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,32 @@ class RSTcheckBear:
...
+        if role_ignore:
+            roles = ','.join(part.strip() for part in role_ignore)
+            args = ('--ignore-roles=' + roles,)
```
Makman2: also here, you overwrite `args` if one of the settings before was set :+1:

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,32 @@ class RSTcheckBear:
...
         """
         :param code_block_language_ignore:
             Comma seperated list for languages of which code blocks
             should be ignored. Code block of following languages
             can be detected: ``bash``, ``c``, ``cpp``, ``json``,
             ``python``, ``rst``.
+
+        :param directive_ignore:
+             Comma separated list of directives to be ignored. Can be
+             used to ignore custom directives as well.
```
Makman2: "as well"? :)

Makman2:
Could you squash all commits together (except the change that I mentioned belonging to an own commit)? :)
Please keep the `Closes XXX` reference in the commit body like @mridubhatnagar's commit already has :+1:

bears/rest/RSTcheckBear.py
umeshksingla: removed.

bears/rest/RSTcheckBear.py
umeshksingla: removed.

bears/rest/RSTcheckBear.py
umeshksingla: done.

bears/rest/RSTcheckBear.py
umeshksingla: oops

bears/rest/RSTcheckBear.py
umeshksingla: and then convert it back to tuple while returning?

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,30 @@ class RSTcheckBear:
...
+             Comma separated list of roles to be ignored. Can be used
+             to ignore custom roles.
         """
         args = ()
         if code_block_language_ignore:
             args = ('--ignore-language=' +
                     ','.join(code_block_language_ignore),)
-        return args + (filename,)
+        if directive_ignore:
+            directives = ','.join(part.strip() for part in directive_ignore)
```
Makman2: is stripping actually necessary?

umeshksingla: It is necessary when you pass it to rstcheck, but here we don't know in what way user will pass the arguments

Makman2:
Could you give some credit in the commit message to @mridubhatnagar ? Seems you have used at least a bit of her code :)

umeshksingla:
I'll add "Thanks @mridubhatnagar for helping", works?
Also, CircleCI is failing every time? Do I need to check my tests?

Makman2:
This is one of our dependencies, `language-check`. This happens sometimes. Try to rebase tomorrow and let's see what happens. If this persists we need to take a closer look. But this problem is very likely not caused by you :)

mridubhatnagar:
Thanks @makman2 and @umeshksingla for taking this forward. I was stuck on the tests part becuase of which it was pending since long. 
Once, this gets approved. There is one of the issues blocked because of this. It can also be closed then.

Makman2:
@mridubhatnagar which one actually?

Makman2:
After `language-check` could be installed after some rebuilds, this needs fixing:
```
Project wide:
**** GitCommitBear [Section: commit | Severity: NORMAL] ****
!    ! Invalid full issue reference: #1723
```
Please provide a full link inside the commit message instead the short `#1723` variant.

mridubhatnagar:
@Makman2 https://github.com/coala/coala/pull/4211.

umeshksingla:
coala/coala#4211 looks like updating coala repository to use RSTcheckBear, is it? But either this shouldn't be blocking that. This PR is simply an enhancement.

tests/rest/RSTcheckBearTest.py
```diff
@@ -17,3 +17,60 @@
...
+:RFC:`793`
+'''
+
+bad_role = '''
+Bar
+===
+
+:bad:`345`
+'''
+
```
mridubhatnagar: Just to make the code more readable. We can good_directive1, good_role just after bad file and good file example. Than maybe have all the 3 tests together.
code_block_language_ignore
ignoredirective
ignorerole

mridubhatnagar:
@umeshksingla  Yes. It is for updating the repo to use rstCheckbear. But adding rstcheckbear in coafile raised error related to directive and roles. Hence, for resolving that error. A seperate issue for roles and directives was raised.

tests/rest/RSTcheckBearTest.py
umeshksingla: I just thought I'll add my stuff at the end, but yes it is a good idea and if that's what other test files follow in general in coala.

bears/rest/RSTcheckBear.py
Makman2: maybe coala strips such whitespaces automatically away (also trailing ones behind a comma), we should test that before doing another strip (which would be just waste of execution time and if the user explicitly wants to set whitespaces using escapes - although unlikely here - we would properly fail)

Makman2: Okay I tested it, you can safely leave the `strip()`. Whitespaces inside the lists you pass via a coafile are automatically stripped from each element, so these are the same for example:
```
directive_ignore = a,b      , c     ,     d, e
directive_ignore = a, b, c, d, e
```

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,30 @@ class RSTcheckBear:
...
+            args += ('--ignore-directives=' + directives,)
+        if role_ignore:
+            roles = ','.join(part.strip() for part in role_ignore)
+            args += ('--ignore-roles=' + roles,)
+        return args + (filename, )
```
Makman2: redundant space after comma :+1:

bears/rest/RSTcheckBear.py
umeshksingla: Okay, but we want is to have "a,b,c,d,e" and not "a, b, c, d, e". `rstcheck` doesn't allow spaces between the names of the directives. Also, I saw stripping done in many bear files.

umeshksingla: ^^ doesn't strip() only remove the whitespaces at the end/beginning of a string? A user would not most likely have an escaped whitespace at the end of a string.

Makman2: yes, however coala already does this for you, you don't need to strip again :+1: As you've specified a `typed_list(str)`

jayvdb:
Please rebase.  This commit has a parent from May!

umeshksingla:
@jayvdb I simply did: 
```
git fetch origin
git rebase origin/master
git pull --rebase
git push mycopy
```
Is it okay to show 112 commits?

aviaryan:
I didn't ask to be assigned, did I? 😯

umeshksingla:
@aviaryan I think I screwed while rebasing and the bot took all the labels and assignees from the commits merged before and pounded them over here :/

Makman2:
wee looks like a bug in gitmate-bot :3

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,30 @@ class RSTcheckBear:
...
+        :param role_ignore:
+             Comma separated list of roles to be ignored. Can be used
+             to ignore custom roles.
         """
         args = ()
         if code_block_language_ignore:
             args = ('--ignore-language=' +
                     ','.join(code_block_language_ignore),)
+        if directive_ignore:
+            directives = ','.join(part.strip() for part in directive_ignore)
```
Makman2: like said, you don't need `part.strip()`, just do `','.join(directive_ignore)`.
You can even inline it into the next line:
```python
args += ('--ignore-directives=' + ','.join(directive_ignore),)
```

bears/rest/RSTcheckBear.py
```diff
@@ -32,16 +33,30 @@ class RSTcheckBear:
...
+            args += ('--ignore-directives=' + directives,)
+        if role_ignore:
+            roles = ','.join(part.strip() for part in role_ignore)
```
Makman2: see above

bears/rest/RSTcheckBear.py
umeshksingla: Sure.

Makman2:
@umeshksingla what about https://github.com/coala/coala-bears/pull/2018/commits/20cb78faca1941f8d37d373eeaeb178f961c72f1#r135389951 ? :)

Makman2:
Though I don't mind, imo it's okay like this. If you want to change it like @mridubhatnagar suggested or if you, @mridubhatnagar, want it changed, feel free to file an issue. Imo both styles are okay: A. mixing test data and tests or B. test data first and then tests :smile:

Makman2:
ack 20cb78f

umeshksingla:
@Makman2 I saw both kinds in the existing tests for bears, so..

Makman2:
@gitmate-bot rebase

Makman2:
@rultor merge

umeshksingla:
@Makman2 Thanks :) how soon is this going to get included in your next release? :P

Makman2:
Could take a while. We try to do releases quarterly, however you can already use the feature by installing the latest development-version: `pip3 install coala-bears --pre`

umeshksingla:
Thanks again, @Makman2, though I use coala through tox.

PR #2023, opened 2017-08-28
MypyBear: Use ``language`` correctly

adhikasp:
check done:
- [x] running pytest locally

Fixes https://github.com/coala/coala-bears/issues/2022


tests/python/MypyBearTest.py
```diff
@@ -50,7 +52,8 @@ def test_py2_version(self):
                             ['print 123'])
 
     def test_bad_language(self):
```
jayvdb: You can simply remove this test ; bad languages are not a bear problem now.

adhikasp: Hm... actually this also test if `language` is not supported by the bear (required for 100% cov too)

jayvdb:
ack 08cb3b1 52d75b5 2bff583

jayvdb:
@rultor merge

PR #2037, opened 2017-09-12
CPPCleanBear.py: Add include_paths for header file

stellargo:
A new setting is added for the CPPCleanBear for allowing include_paths for header files to be added in the .coafile. 'include_paths' is a list of paths(strings) for header files which have been included in the cpp file on which we desire to use cppclean. Two tests, namely 'CPPCleanBearIncludePathTest' and 'CPPCleanBearIncludePathTest2' are added. In test_files, CPPCleanBearHeaderTest.h is added which is needed for performing 'CPPCleanBearIncludePathTest2'.

Closes: https://github.com/coala/coala-bears/issues/2026

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


stellargo:
@palash25 Thanks for the input :)

stellargo:
Here are some before and after screenshots:
Before:
cppcleanbefore
After:
cppcleanafter

stellargo:
@Makman2 :)

Makman2:
Tests fail because the commit body contains too long lines. We limit them to 72 chars each :+1:

Makman2:
And also please remove the colon after "Closes" :)

Makman2:
Oh and the shortlog should rather say: "Add include_paths setting"
Then a person directly gets that a setting was added :)

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,12 @@ class CPPCleanBear:
     CAN_DETECT = {'Smell', 'Unused Code', 'Security'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file):
-        return filename,
+    def create_arguments(filename,
+                         file,
+                         config_file,
+                         include_paths:
```
Makman2: please put the type annotation onto the same line as `include_paths`, that reads way better

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,12 @@ class CPPCleanBear:
...
+                         typed_list(str)=()):
+        include = ' '.join('--include-path='+part.strip()
+                           for part in include_paths)
```
Makman2: is joining with spaces correct? Or does cppclean need a different separator like comma? (I'm asking as bash/shells don't support spaces in a single argument except when quoting. Although executables in linter are not run in shell and you don't have to quote arguments in here, when writing this equivalently in bash using spaces is rather unusual when specifying multiple arguments like here)

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,12 @@ class CPPCleanBear:
...
+        return (filename,
+                include)
```
Makman2: single line please, then you also don't have to use braces

bears/c_languages/CPPCleanBear.py
Makman2: nvm misread code.
However, as stated, this is not shell, so you have to put each of those `--include-path=...` into a list :)
Also please add a test using multiple include paths, to test support for this + to avoid the error you already have here when specifying multiple include-paths

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,33 @@
...
+                                     tempfile_kwargs={'suffix': '.cpp'})
+
+CPPCleanBearIncludePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(),
+    invalid_files=(test_file,),
+    tempfile_kwargs={'suffix': '.cpp'}
+    )
+
+CPPCleanBearIncludePathTest2 = verify_local_bear(
```
Makman2: hmm bad name... Rather call it `CPPCleanBearValidIncludePathTest` (and above `CPPCleanBearNoIncludePathTest) or similar. Test names should always reflect their purpose / functionality tested.

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -1,5 +1,6 @@
 from bears.c_languages.CPPCleanBear import CPPCleanBear
 from coalib.testing.LocalBearTestHelper import verify_local_bear
+import os
```
Makman2: please put the import at top of this file, although we can't enforce it yet (with coala), our yet inofficial rule is:
```
system-imports

own imports

other package imports
```

bears/c_languages/CPPCleanBear.py
stellargo: @Makman2 Yeah I get the problem but am not so clear as to how to solve it. I want each of those --include-path=... seperated with a space, for e.g. `cppclean --include-path=path1 --include-path=path2`. So what you are suggesting is to use a list instead of a typed_list? I would still have to separate them with a space

stellargo:
@Makman2 Still failing even after limiting no of characters in each line to less than 72 :(

bears/c_languages/CPPCleanBear.py
Makman2: Effectively those arguments are concatted again by a space.

Linter doesn't use a shell. Without a shell arguments get passed inside a sequence. Actually when you write a command into a shell like bash, this happens inside:
```
myprogram arg1 "some other arg"
```
gets parsed to something like this internally:
```python
['myprogram', 'arg1', 'some other arg']
```
and is passed to the program `myprogram`. Now that we don't use a shell, we have to do that argument splitting manually inside a bear (which is more error-proof and easier).

Makman2:
Not only, but yes, AppVeyor fails at least because of that.

bears/c_languages/CPPCleanBear.py
stellargo: @Makman2 Thank you so much for taking the time to explain this to me :) My latest PR is working perfectly when including multiple include-paths and I have appended the test to use multiple include-paths. When you have time  would be great if you could review the changes :)

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,14 @@ class CPPCleanBear:
...
 
     @staticmethod
-    def create_arguments(filename, file, config_file):
-        return filename,
+    def create_arguments(filename,
+                         file,
+                         config_file,
+                         include_paths: typed_list(str)=()):
+        args = []
+        args.append(filename)
```
Makman2: you can shorten this to
```
args = [filename]
```
:)

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,14 @@ class CPPCleanBear:
...
+        for part in include_paths:
+            args.append('--include-path')
+            args.append(part.strip())
```
Makman2: you don't need to `strip`, coala will do this for you when parsing settings :)

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,14 @@ class CPPCleanBear:
...
     @staticmethod
-    def create_arguments(filename, file, config_file):
-        return filename,
+    def create_arguments(filename,
+                         file,
+                         config_file,
+                         include_paths: typed_list(str)=()):
+        args = []
+        args.append(filename)
+        for part in include_paths:
```
Makman2: just call this variable `path` :)
(or `include_path`. Some optional opinion from me: I usually try to name variables, that contain lists, tuples, sequences with a plural-"s", e.g. `things = []` and not `thing = []`. When I iterate over them, I just leave the `s` for the iteration variable: `for thing in things`. Just some inspirations ;D)

bears/c_languages/CPPCleanBear.py
```diff
@@ -22,5 +23,14 @@ class CPPCleanBear:
...
+            args.append('--include-path')
+            args.append(part.strip())
+
+        return (args)
```
Makman2: redundant braces :)

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,39 @@
...
+    return 0;
+}
+"""
+
+
 CPPCleanBearTest = verify_local_bear(CPPCleanBear,
                                      valid_files=(good_file,),
-                                     invalid_files=(bad_file,))
+                                     invalid_files=(bad_file,),
+                                     tempfile_kwargs={'suffix': '.cpp'})
```
Makman2: unnecessary change, if you like to have it please separate commit :+1:

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,39 @@
 """
 
 
+test_file = """
```
Makman2: could you name the variable after its intention (testing additional include paths)?

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,39 @@
...
+CPPCleanBearValidIncludePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(test_file,),
+    invalid_files=(),
+    settings={'include_paths':  os.path.join(os.path.dirname(__file__),
+                                             'test_files',
+                                             'testCPPHeader1')+','
+              + os.path.join(os.path.dirname(__file__),
+                             'test_files',
+                             'testCPPHeader2')},
```
Makman2: please write a function for that (best is you call it `get_testfile_name` because other tests do the same), so you don't have to repeat yourself:
```
'include_paths': get_testfile_name('testCPPHeader1') + ',' + get_testfile_name(...)
```
Or call it `get_testdirectory_name` or so.

Makman2:
Seems you've stored some MAC-file: DS_Store, please remove it :)

tests/c_languages/CPPCleanBearTest.py
Makman2: Could you maybe also make the test directory names simpler? They look so complicated :3 Maybe just `headers1`, `headers2` directories and `foo.h`, `bar.h` inside them respectively? Imo they are a bit too long 😅

Makman2: Ah and the functions could also then be called `void foo()` and `void bar()` :P

Makman2:
About the commit message body: Please don't break lines inside words 😅 Put the word onto the next line if it doesn't fit 😅

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,39 @@
...
 CPPCleanBearTest = verify_local_bear(CPPCleanBear,
                                      valid_files=(good_file,),
-                                     invalid_files=(bad_file,))
+                                     invalid_files=(bad_file,),
+                                     tempfile_kwargs={'suffix': '.cpp'})
+
+CPPCleanBearNoIncludePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(),
+    invalid_files=(test_file,),
```
Makman2: you could effectively add this case to `CPPCleanBearTest` above :+1:

tests/c_languages/CPPCleanBearTest.py
stellargo: @Makman2 I added the tempfile_kwargs to ensure that the file is .cpp. I could remove that should i?

Makman2: it's okay, as cppclean works on cpp files, but it isn't necessary to be inside this commit (it did work before)

stellargo: Ok I will remove it :) Also appveyor is failing due to an XMLBearTest, that is some other issue right? not related to this PR?

bears/c_languages/CPPCleanBear.py
stellargo: Thank you so much @Makman2 yeah I do realize that improves code readability by a great extent :)

tests/c_languages/CPPCleanBearTest.py
Makman2: yes not related, I've created a PR that fixes / suppresses the error. As long as nothing else fails, your PR is fine :+1: we won't merge it until the fix is on master, but after a rebase your build should go immediately green :+1:

stellargo: Ok I am working on improving my PR, one other thing I experienced travis and circle failing too because of some sourceforge ssh connection issue.

Makman2: hopefully that just goes away :3 timeouts occur quite often in coala as we install many dependencies :+1:

stellargo: They failed :( same error.

stellargo: They did fail again with the same error. Seems some issue with the sourceforge link i guess 👍

stellargo:
@Makman2 Incorporated all your comments into my PR :) Again, thanks a lot :)

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,33 @@
...
-                                     invalid_files=(bad_file,))
+                                     invalid_files=(bad_file,
+                                                    test_include_paths_file,),
+                                     tempfile_kwargs={'suffix': '.cpp'})
+
+CPPCleanBearValidIncludePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(test_include_paths_file,),
+    invalid_files=(),
+    settings={'include_paths': get_testdirectory_name('headers1')+',' +
```
Makman2: be consistent with spaces around operators ;)
either `a + b` or `a+b` :)

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,33 @@
...
+def get_testdirectory_name(test_file):
+    return os.path.join(os.path.dirname(__file__), 'test_files', test_file)
+
+
 CPPCleanBearTest = verify_local_bear(CPPCleanBear,
                                      valid_files=(good_file,),
-                                     invalid_files=(bad_file,))
+                                     invalid_files=(bad_file,
+                                                    test_include_paths_file,),
+                                     tempfile_kwargs={'suffix': '.cpp'})
```
Makman2: please move this change in L41 into an own commit ;)

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,33 @@
...
+
+CPPCleanBearValidIncludePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(test_include_paths_file,),
+    invalid_files=(),
+    settings={'include_paths': get_testdirectory_name('headers1')+',' +
+              get_testdirectory_name('headers2')},
+    tempfile_kwargs={'suffix': '.cpp'}
+    )
```
Makman2: to keep the same style, I would move the closing bracket `)` to the previous line^^

Makman2:
hmm just thinking, maybe we should also test the case when we miss one of the additional include paths? Your decision @stellargo 👍

tests/c_languages/CPPCleanBearTest.py
stellargo: So you asked me to move 'test_include_paths_file' to the CPPCleanBearTest. It doesn't work without explicitly having L41. I can either have a different test as I had before for it or I will have to include it here

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,33 @@
...
+
+int main() {
+    foo();
+    bar();
+    return 0;
+}
+"""
+
+
+def get_testdirectory_name(test_file):
```
Makman2: hm shouldn't the parameter be called `test_directory`? :D

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,33 @@
...
+                                                    test_include_paths_file,),
+                                     tempfile_kwargs={'suffix': '.cpp'})
+
+CPPCleanBearValidIncludePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(test_include_paths_file,),
+    invalid_files=(),
+    settings={'include_paths': get_testdirectory_name('headers1')+',' +
+              get_testdirectory_name('headers2')},
+    tempfile_kwargs={'suffix': '.cpp'}
```
Makman2: maybe you can also omit this here in this commit, and move it to the same commit when you introduce the suffix for the above test? So you have 1. commit: Add additional include support 2. commit: Use .cpp ending for tested files

stellargo: Tests fail when i dont add .cpp ending for files :P

stellargo: @Makman2  I could reverse the order of the steps tho :P First introduce .cpp ending and then add additional include support :)

tests/c_languages/CPPCleanBearTest.py
Makman2: it doesn't work without? okay then nvm :)

tests/c_languages/CPPCleanBearTest.py
Makman2: hm interesting why it fails though :3 but then you don't have to use 2 different commits, totally legit to keep this here :+1:

stellargo: @Makman2 Ok :) Could you have a look at the failed travis and circle builds?
Appear to have a common issue of sourceforge ssh connection not being possible. If you feel I could raise an issue for the same then :)

stellargo:
@Makman2 I have added test for case when we miss only one of the additional include paths :) Plus other suggestions incorporated too :)

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +18,41 @@
...
+    invalid_files=(),
+    settings={'include_paths': get_testdirectory_name('headers1') + ',' +
+              get_testdirectory_name('headers2')},
+    tempfile_kwargs={'suffix': '.cpp'})
+
+CPPCleanBearInvalidIncludeSinglePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(),
+    invalid_files=(test_include_paths_file,),
+    settings={'include_paths': get_testdirectory_name('headers1')},
```
Makman2: ah I just remember another PR, where this has lead to confusion: Please add this to your test:
```
from coala_utils.string_processing import escape
```
and use `escape` on each of those directory names:
```
```
coala preprocesses escape characters. Paths in Windows use backslashes, which yields to test fails because files can't be found (`C:\xyz\abc` gets `C:xyzabc`). To prevent that, we have to escape the backslashes, and `escape` can do that for us:
```
escape(get_testdirectory_name(...), '\\')
```
We have to do something about it in the future so we don't get sudden test fails on Windows due to this^^

stellargo: @Makman2 Ok I can do that, just one thing is this like the same issue xmlbeartest is facing? I remember while looking into it there was problem with tests where directories were being used, just something i remembered could be totally wrong :P

stellargo:
@Makman2 added escape on the directories :)

tests/c_languages/CPPCleanBearTest.py
Makman2: yes it was one of the issues :+1:

stellargo:
@gitmate-bot rebase

tests/c_languages/CPPCleanBearTest.py
```diff
@@ -17,6 +19,44 @@
...
+                                     tempfile_kwargs={'suffix': '.cpp'})
+
+CPPCleanBearValidIncludeMultiplePathTest = verify_local_bear(
+    CPPCleanBear,
+    valid_files=(test_include_paths_file,),
+    invalid_files=(),
+    settings={'include_paths': escape(
+                get_testdirectory_name('headers1'), '\\') + ',' +
+              escape(
+                get_testdirectory_name('headers2'), '\\')},
```
Makman2: 4 spaces indent relative to opening expression:
```
escape(
    get_testdirectory_name(...), ...)  # <-- note: 4 spaces indent at start of line
```
same above
(I would maybe reformat to
```
    settings={'include_paths':
                  escape(get_testdirectory_name('headers1'), '\\') + ',' +
                  escape(get_testdirectory_name('headers2'), '\\')},
```
)

stellargo: @Makman2 So there is some PEP8 issue with that 4 spaces indent. Its forcing me to do something like this : 
'''
settings={'include_paths':
                 escape(get_testdirectory_name('headers1'), '\\') + ',' +
                 escape(get_testdirectory_name('headers2'), '\\')},
'''

stellargo: @Makman2 done:)

Makman2:
ack 720d654

Makman2:
@gitmate-bot rebase

Makman2:
@rultor merge

PR #2060, opened 2017-09-29
.ci/deps.opam.sh: Pin easy-format to v1.2.0

jayvdb:
Infer compilation fails when using easy-format v1.3.0

Fixes https://github.com/coala/coala-bears/issues/2059

### DIFF
--- a/.ci/deps.opam.sh
+++ b/.ci/deps.opam.sh
@@ -10,8 +10,16 @@ if [ ! -e ~/infer-linux64-v0.7.0/infer/bin ]; then
   cd ~/infer-linux64-v0.7.0
   opam init --y
   opam update
+
+  # See https://github.com/coala/coala-bears/issues/1763
   opam pin add --yes --no-action atdgen 1.10.0
+
+  # See https://github.com/coala/coala-bears/issues/2059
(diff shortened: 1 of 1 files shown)


jayvdb:
This passed on my fork https://travis-ci.org/jayvdb/coala-bears/builds/281128565 , which was a complete rebuild.

.ci/deps.opam.sh
```diff
@@ -11,6 +11,7 @@ if [ ! -e ~/infer-linux64-v0.7.0/infer/bin ]; then
   opam init --y
   opam update
   opam pin add --yes --no-action atdgen 1.10.0
+  opam pin add --yes --no-action easy-format 1.2.0
```
sils: a comment about that would be good IMO so it can be removed when infer fixes that and people know what this is about

sils:
marking WIP as per review

.ci/deps.opam.sh
jayvdb: infer is at version 0.13.0
They have fixed it.
This is yet another patch to keep alive a horribly old version.

Mixih:
ack 71ed4df

PR #2063, opened 2017-10-03
HTTPSBear.py: New Bear Added

stellargo:
HTTPS bear can detect URL links which can be changed to https
from http and allows to change them automatically.

Closes https://github.com/coala/coala-bears/issues/1900

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+from coalib.settings.Setting import typed_dict
+from coala_utils.decorators import (enforce_signature, generate_ordering,
+                                    generate_repr)
+
+# class HTTPSResult(HiddenResult):
+#     @enforce_signature
+#     def __init__(self, origin, affected_code,
+#                  link: str,
+#                  http_status_code: (int, None),
+#                  link_context: LINK_CONTEXT):
```
Makman2: commented code ;)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+from bears.general.URLBear import URLBear
+from coalib.bears.LocalBear import LocalBear
+from dependency_management.requirements.PipRequirement import PipRequirement
+from coalib.results.Result import Result
+from aenum import Flag
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+
+from coalib.bears.LocalBear import LocalBear
+from dependency_management.requirements.PipRequirement import PipRequirement
+from coalib.bearlib import deprecate_settings
```
Makman2: unused import

Makman2: Generally many unused imports...

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+                                                else URLBear.DEFAULT_TIMEOUT
+                                                )
+                if httpscode and 200 <= httpscode < 300:
+                    yield Result.from_values(
+                        origin=self,
+                        message=('You could use https instead of http at - '
+                                 '{url}').format(url=link),
+                        file=filename,
+                        line=line_number,
+                        severity=RESULT_SEVERITY.NORMAL)
```
Makman2: we definitely want a patch ;)

bears/general/HTTPSBear.py
stellargo: @Makman2 Yeah still working on it :P I still have to write tests and get it to solve the problem on its own , also remove unnecessary heaers :P

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+                                                if host in network_timeout
+                                                else network_timeout.get('*')
+                                                if '*' in network_timeout
+                                                else URLBear.DEFAULT_TIMEOUT
+                                                )
+                if httpscode and 200 <= httpscode < 300:
+                    yield Result.from_values(
+                        origin=self,
+                        message=('You could use https instead of http at - '
+                                 '{url}').format(url=link),
```
Makman2: with a patch you don't have to include the link in the message ;)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+        A link is considered valid if the server responds with a 2xx code.
+
+        Warning: This bear will make HEAD requests to all URLs mentioned in
+        your codebase, which can potentially be destructive. As an example,
+        this bear would naively just visit the URL from a line that goes like
+        `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
+        all your data.
+
+        :param dependency_results: Results given by URLBear.
+        :param follow_redirects: Set to true to autocorrect redirects.
```
Makman2: -->
```
:param a:
    ABC
:param b:
    DEF
```

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+    LANGUAGES = {'All'}
+    REQUIREMENTS = {PipRequirement('requests', '2.12')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Documentation'}
+    BEAR_DEPS = {URLBear}
+
+    def run(self, filename, file, dependency_results=dict(),
+            follow_redirects: bool=False,
```
Makman2: unused parameter

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,85 @@
...
+                                   if not url == '*' else '*': timeout
+                                   for url, timeout in network_timeout.items()}
+                httpscode = URLBear.get_status_code(
+                                                httpslink,
+                                                network_timeout.get(host)
+                                                if host in network_timeout
+                                                else network_timeout.get('*')
+                                                if '*' in network_timeout
+                                                else URLBear.DEFAULT_TIMEOUT
+                                                )
```
Makman2: bad indent ;)

stellargo:
@Makman2 So the bear is working as intended, but I am facing a little problem while writing tests. So basically should I take a link and test if its https version works or do I need to implement a dummy kind of function which returns a response code that is like the last three numbers like http:random.com/200. Because this is how InvalidLinkBearTest is working :)

Makman2:
Yes mock the components to simulate a webpage returning error code 200 (and error codes to see if the bear then doesn't suggest a switch to https protocol), like `InvalidLinkBearTest` ;)

stellargo:
@Makman2 Ok, but i am still not clear as to how the https mock would work. Basically the bear will just check if the link after changing to https is working or not, how do make a mock for that? I was tinkering with the idea around something that maybe if there is a particular letter, say 'v' in https://www.something.com/v200 then its valid and https://www.something.com/200 is not :)

Makman2:
yeah sounds good, you could generalize that a bit to ease tests :+1: :)

tests/general/HTTPSBearTest.py
```diff
@@ -0,0 +1,138 @@
...
+        status_code = 200
+    else:
+        try:
+            status_code = int(request.path_url[-3:])
+        except ValueError:
+            raise requests.exceptions.RequestException
+    resp = requests.Response()
+    resp.raw = io.BytesIO()
+    resp.status_code = status_code
+    return resp
```
blazeu: Isn't this has the same as [`custom_matcher` in InvalidLinkBear](https://github.com/coala/coala-bears/blob/master/tests/general/InvalidLinkBearTest.py#L17-L56) ?

If so, you could import that instead.

tests/general/HTTPSBearTest.py
```diff
@@ -0,0 +1,138 @@
...
+                            'You are not connected to the internet.')
+
+
+class HTTPSBearTest(LocalBearTestHelper):
+    """
+    The tests are mocked (don't actually connect to internet) and
+    return the int conversion of the last three chars of
+    the URL as status code.
+
+    Check ``custom matcher`` for more info on implementation.
```
blazeu: `custom_matcher_https` ?

tests/general/HTTPSBearTest.py
```diff
@@ -0,0 +1,138 @@
...
+
+            self.assertTrue(URLBear.check_prerequisites() ==
+                            'You are not connected to the internet.')
+
+
+class HTTPSBearTest(LocalBearTestHelper):
+    """
+    The tests are mocked (don't actually connect to internet) and
+    return the int conversion of the last three chars of
+    the URL as status code.
```
blazeu: ... and explain your `custom_matcher_https` behaviour.

blazeu:
unack https://github.com/coala/coala-bears/pull/2063/commits/f0267020124b0eb7fba4c4c23710045b60ee63b9

stellargo:
@blazeu I have incorporated the requested changes :) I believe I need to do a rebase so let me know if this is fine so I can go forward with the rebase. (Please don't use gitmate for that, I would like to do it manually :P )

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+        code.
+
+        Warning: This bear will make HEAD requests to all URLs mentioned in
+        your codebase, which can potentially be destructive. As an example,
+        this bear would naively just visit the URL from a line that goes like
+        `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
+        all your data.
+
+        :param dependency_results:
+            Results given by URLBear.
```
Makman2: Up to you to keep this `:param ...:`, as this param is a special one ;)
However if you keep it, enclose `URLBear` in double backticks so it gets rendered as code ;)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Documentation'}
+    BEAR_DEPS = {URLBear}
+
+    def run(self, filename, file, dependency_results=dict(),
+            network_timeout: typed_dict(str, int, DEFAULT_TIMEOUT)=dict()):
+        """
+        Find http links in any text file and check if the https version of
+        link is valid. If so, provide option for replacing it with https.
```
Makman2: -->
If so, an option is provided for ...

(I think passive phrasing fits here best :+1:)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+        Warning: This bear will make HEAD requests to all URLs mentioned in
+        your codebase, which can potentially be destructive. As an example,
+        this bear would naively just visit the URL from a line that goes like
+        `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
+        all your data.
+
+        :param dependency_results:
+            Results given by URLBear.
+        """
+        for result in dependency_results.get(URLBear.name, []):
```
Makman2: why `.get` instead of `operator []`?

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+            line_number, link, code, context = result.contents
+            if (link[0:5] != 'https'):
```
Makman2: use `str.startswith()` instead

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+                httpslink = 'https' + link[4:]
```
Makman2: to avoid maintenance problems later, I would advise to put the `https` string into a constant, e.g. `HTTPS_PREFIX`. You've also hardcoded the number `4` now that comes from the length of `http`, so maybe we also want a `HTTP_PREFIX`. So in the end having something like this avoids problems:
```
httpslink = HTTPS_PREFIX + link[len(HTTP_PREFIX):]
```
what do you think? ;)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+                host = urlparse(httpslink).netloc
+                network_timeout = {urlparse(url).netloc
+                                   if not url == '*' else '*': timeout
+                                   for url, timeout in network_timeout.items()}
```
Makman2: style recommendation:
```
network_timeout = {
    urlparse(url).netloc is not url =='*' else '*': timeout
    for url, timeout in network_timeout.items()}
```
then you have the item generation and the for-loop part a bit better distinct ;)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+                httpscode = URLBear.get_status_code(
+                    httpslink,
+                    network_timeout.get(host)
+                    if host in network_timeout
+                    else network_timeout.get('*')
+                    if '*' in network_timeout
+                    else URLBear.DEFAULT_TIMEOUT)
```
Makman2: would add an empty line here ;) reads a bit better

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+                if httpscode and 200 <= httpscode < 300:
+                    diff = Diff(file)
+                    current_line = file[line_number - 1]
+                    start = current_line.find(link)
```
Makman2: the link position is not passed by `URLBear`? In that case alright, but let's file an issue afterwards to improve `URLBear` so it also gives us those (as it probably does the same procedure here to actually find links).

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+                    end = start + len(link)
+                    replacement = current_line[:start] + \
+                        'https' + link[4:] + current_line[end:]
```
Makman2: use braces instead of escape characters:
```
replacement = (current_line[...
    ....)
```

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,72 @@
...
+                    diff.change_line(line_number,
+                                     current_line,
+                                     replacement)
+
+                    yield Result.from_values(
+                        origin=self,
+                        message=('https can be used instead of http for '
+                                 '{url}').format(url=link),
```
Makman2: up to you: `url` inside `{}` can be omitted, and also inside the `format()` call:
```
... {}').format(link)
```

stellargo:
@Makman2 incorporated all your suggested changes :) and done the rebase as well

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Documentation'}
+    BEAR_DEPS = {URLBear}
+
+    def run(self, filename, file, dependency_results=dict(),
+            network_timeout: typed_dict(str, int, DEFAULT_TIMEOUT)=dict()):
+        """
+        Find http links in any text file and check if the https version of
+        link is valid. If so, an option is provided for replacing it with
+        https.
```
Makman2: I think you can remove the last sentence, as coala always tries to give patches if possible :)
However if you don't want to remove it:
... for replacing ~~it~~ **them** with...
As it refers to link**s** ;)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+        code.
+
+        Warning: This bear will make HEAD requests to all URLs mentioned in
+        your codebase, which can potentially be destructive. As an example,
+        this bear would naively just visit the URL from a line that goes like
+        `do_not_ever_open = 'https://api.acme.inc/delete-all-data'` wiping out
+        all your data.
+
+        :param dependency_results:
+            Results given by ``URLBear``.
```
Makman2: Optional: Doesn't have to be documented, as it's a built-in feature parameter.

Makman2: But documentation for `network_timeout` is missing :)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        HTTPS_PREFIX = 'https'
+        HTTP_PREFIX = 'http'
```
Makman2: can be moved to class scope

Makman2: but only do that if there's a little performance gain, otherwise can be kept here for better encapsulation

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+        for result in dependency_results.get(URLBear.name, []):
+            line_number, link, code, context = result.contents
+            if (not link.startswith(HTTPS_PREFIX)):
+                httpslink = HTTPS_PREFIX + link[len(HTTP_PREFIX):]
```
Makman2: what if the link has a prefix that's not `http`? Or if it doesn't have a prefix at all (relative links)?

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+                host = urlparse(httpslink).netloc
+                network_timeout = {
+                    urlparse(url).netloc if not url == '*' else '*': timeout
+                    for url, timeout in network_timeout.items()}
```
Makman2: You should especially document the feature to set a default timeout in parallel to specific timouts for certain URLs :+1:

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+                httpscode = URLBear.get_status_code(
+                    httpslink,
+                    network_timeout.get(host)
+                    if host in network_timeout
+                    else network_timeout.get('*')
+                    if '*' in network_timeout
+                    else URLBear.DEFAULT_TIMEOUT)
+
+                if httpscode and 200 <= httpscode < 300:
```
Makman2: just curious, when can httpscode be `None`?

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+                                     current_line,
+                                     replacement)
+
+                    yield Result.from_values(
+                        origin=self,
+                        message=('https can be used instead of http'),
+                        diffs={filename: diff},
+                        file=filename,
+                        line=line_number,
+                        severity=RESULT_SEVERITY.NORMAL)
```
Makman2: could you also pass `affected_code`? :)

bears/general/HTTPSBear.py
stellargo: Then nothing should happen because the purpose of HTTPSBear was just to find those links whose http can be updated to https, I mean then it should be under the scope of URLBear to take care.

bears/general/HTTPSBear.py
stellargo: When the website doesn't have ssl certificate (i.e. only http works)

tests/general/HTTPSBearTest.py
stellargo: This is not the same custom_matcher, it has been modified as per the https bear's requirement

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,75 @@
...
+                network_timeout = {
+                    urlparse(url).netloc if not url == '*' else '*': timeout
+                    for url, timeout in network_timeout.items()}
+                httpscode = URLBear.get_status_code(
+                    httpslink,
+                    network_timeout.get(host)
+                    if host in network_timeout
+                    else network_timeout.get('*')
+                    if '*' in network_timeout
+                    else URLBear.DEFAULT_TIMEOUT)
```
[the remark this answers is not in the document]
stellargo: I think yes because '*' was the wildcard for all urls whose timeout we don't have. Supposing it isn't present in network_timeout, then we need to consider DEFAULT_TIMEOUT. URLBear has also used this second if.

bears/general/HTTPSBear.py
stellargo: `affected_code` isnt a param of `from_values`, so where should I pass it? Also InvalidLinkBear is also not passing it.

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,81 @@
...
+                                      the same host as that of URLs provided
+                                      will be passed that timeout. It can also
+                                      contain a wildcard timeout entry with key
+                                      '*'. The timeout of all the websites not
+                                      in the dict will be the value of the key
+                                      '*'.
+        """
+        for result in dependency_results.get(URLBear.name, []):
+            line_number, link, code, context = result.contents
+            if (not link.startswith(self.HTTPS_PREFIX)):
```
refeed: I think we can change this line to:
```
if link.startswith(self.HTTPS_PREFIX):
    continue
```
Then un-indent the lines below this line by 1 indent, so we don't get too many indentations here.

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,81 @@
...
+                    urlparse(url).netloc if not url == '*' else '*': timeout
+                    for url, timeout in network_timeout.items()}
+                httpscode = URLHeadBear.get_head_response(
+                    httpslink,
+                    network_timeout.get(host)
+                    if host in network_timeout
+                    else network_timeout.get('*')
+                    if '*' in network_timeout
+                    else URLHeadBear.DEFAULT_TIMEOUT).status_code
+                print(httpscode)
```
jayvdb: `print`?

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,81 @@
...
+            httpscode = URLHeadBear.get_head_response(
+                httpslink,
+                network_timeout.get(host)
+                if host in network_timeout
+                else network_timeout.get('*')
+                if '*' in network_timeout
+                else URLHeadBear.DEFAULT_TIMEOUT).status_code
+
+            if not httpscode or not 200 <= httpscode < 300:
+                return
```
jayvdb: shouldnt this be `continue`? dont just copy what I say. :P

tests/general/HTTPSBearTest.py
```diff
@@ -0,0 +1,109 @@
...
+        self.ub_check_prerequisites = URLHeadBear.check_prerequisites
+        self.section = Section('')
+        URLHeadBear.check_prerequisites = lambda *args: True
+        self.uut = HTTPSBear(self.section, Queue())
+
+    def tearDown(self):
+        URLHeadBear.check_prerequisites = self.ub_check_prerequisites
+
+    def test_valid_https(self):
+        file = """
```
jayvdb: `file` is a keyword in Python.  avoid.

also this isnt a file. it is a array of string.

bears/general/HTTPSBear.py
stellargo: ooooops :(  missed the loop, sorry

stellargo: For `CAN_FIX` is 'URL links' fine?

stellargo: Or 'HTTP Links'

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,82 @@
...
+                                      the same host as that of URLs provided
+                                      will be passed that timeout. It can also
+                                      contain a wildcard timeout entry with key
+                                      '*'. The timeout of all the websites not
+                                      in the dict will be the value of the key
+                                      '*'.
+        """
+        for result in dependency_results.get(URLHeadBear.name, []):
+            line_number, link, code, context = result.contents
+            if (link.startswith(self.HTTPS_PREFIX)):
```
jayvdb: unnecessary extra brackets

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,82 @@
...
+                continue
```
jayvdb: add a blank line under each continue.

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,82 @@
...
+            network_timeout = {
+                urlparse(url).netloc if not url == '*' else '*': timeout
+                for url, timeout in network_timeout.items()}
+            httpscode = URLHeadBear.get_head_response(
+                httpslink,
+                network_timeout.get(host)
+                if host in network_timeout
+                else network_timeout.get('*')
+                if '*' in network_timeout
+                else URLHeadBear.DEFAULT_TIMEOUT).status_code
```
jayvdb: `HTTPSBear.DEFAULT_TIMEOUT` ?

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,82 @@
...
+
+            if not httpscode or not 200 <= httpscode < 300:
+                continue
+            diff = Diff(file)
+            current_line = file[line_number - 1]
+            start = current_line.find(link)
+            end = start + len(link)
+            replacement = current_line[:start] + 'https' + link[
```
jayvdb: using `[` as a line continuation like this is terrible.

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,89 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Documentation'}
+    CAN_FIX = {'HTTP Links'}
+    BEAR_DEPS = {URLHeadBear}
+    HTTPS_PREFIX = 'https'
+    HTTP_PREFIX = 'http'
+
+    def run(self, filename, file, dependency_results=dict(),
+            network_timeout: typed_dict(str, int, DEFAULT_TIMEOUT)=dict()):
```
jayvdb: style change requires spaces around `=`

stellargo:
@jayvdb So basically this bear needs to use URLHeadBear to get a response for the new https link obtained from http. Problem is that when there is no such link it returns a `ConnectTimeout` Exception. This obviously does not have a status_code attribute. Hence i have used the try except block. But this is making testing extremely problematic because the `custom_matcher` first checks prerequisites and the LocalBearTestHelper immediately raises AssertionError. This was not the problem earlier when the URLBear hadn't been split into the two bears as then it would just return None for the attribute `status_code`. So either we could re change the URLHeadBear's `get_head_response` method to the way it was, or use the pragma no cover which I have added right now.

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,89 @@
...
+            line_number, link, code, context = result.contents
+            if link.startswith(self.HTTPS_PREFIX):
+                continue
+
+            httpslink = self.HTTPS_PREFIX + link[len(self.HTTP_PREFIX):]
+            host = urlparse(httpslink).netloc
+            network_timeout = {
+                urlparse(url).netloc if not url == '*' else '*': timeout
+                for url, timeout in network_timeout.items()}
+            httpsresponse = URLHeadBear.get_head_response(
```
jayvdb: `snake_case` (everywhere)

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,89 @@
...
+                continue
+
+            if not httpscode or not 200 <= httpscode < 300:
+                continue
+
+            diff = Diff(file)
+            current_line = file[line_number - 1]
+            start = current_line.find(link)
+            end = start + len(link)
+            replacement = current_line[:start] + 'https' + \
```
jayvdb: dont use `\`

bears/general/HTTPSBear.py
```diff
@@ -0,0 +1,89 @@
...
+                link[len(self.HTTP_PREFIX):] + current_line[end:]
+            diff.change_line(line_number, current_line, replacement)
+
+            yield Result.from_values(
+                origin=self,
+                message=('https can be used instead of http'),
```
jayvdb: why `()`?

jayvdb:
ack https://github.com/coala/coala-bears/pull/2063/commits/bbfb1168edbb3d4b37c11ee2927d35fa627e22ab

PR #2086, opened 2017-10-27
Upgrade to ESLint version 3

blazeu:
It also upgrade other dependencies (`babel-eslint`,
`eslint-plugin-import`, `happiness` and `standard`) so that the
dependencies are compatible with ESLint 3.

Closes https://github.com/coala/coala-bears/issues/872

https://github.com/coala/coala-bears/pull/2098 needs to be merged first to pass `standard` & `happiness` tests.


IpshitaC:
Hi!
Circle CI is failing because the code coverage of ESLintBear is not 100%. There are two branches being missed out.
Make sure the tests cover these lines 
https://codecov.io/gh/coala/coala-bears/src/6a6e355fed38666510ec5f0bd00c597f68906f99/bears/js/ESLintBear.py#L62

bears/js/ESLintBear.py
```diff
@@ -58,15 +58,17 @@ def generate_config(filename, file):
...
             return
 
-        output = json.loads(output[0])
+        # ESLint throws error inside stdout
+        try:
+            output = json.loads(output[0])
+        except ValueError:
+            self.warn('While running {0}, some issues were found:'
+                      .format(self.__class__.__name__))
+            self.warn(output[0])
```
Makman2: this seems to be a bit different from above, you do the stuff from above for `output[0]` instead of `output[1]`^^

Makman2: btw, if you don't need `output[1]` any more, you can omit `use_stderr` and `use_stdout` in `@linter` altogether, and just use `output` on its own.

blazeu: Okay, thanks :)

bears/js/ESLintBear.py
```diff
@@ -58,15 +57,17 @@ def generate_config(filename, file):
         return '{"extends": "eslint:recommended"}'
 
     def process_output(self, output, filename, file):
-        if output[1]:
+        if not file or not output:
+            return
+
+        # ESLint throws error inside stdout
```
Makman2: what kind of errors? Would be good if you be as clear as possible here so people reading the code know what's meant ;)

Makman2:
I assume all new versions are correct ;) Commit looks good, just some minor change for the comment :)

bears/js/ESLintBear.py
```diff
@@ -8,8 +8,7 @@
 
 
 @linter(executable='eslint',
-        use_stdin=True,
-        use_stderr=True)
```
jayvdb: Is it possible to keep `use_stderr=True`?

If that was needed for ESLint 2, it would be helpful to keep it so that anyone with ESLint 2 installed will still see any errors that were emitted by their eslint.

You could include a comment that stderr is only needed for backwards compatibility with ESLint 2, even though it is no longer supported.

Or, does `use_stderr=True` mean that some junk is captured that is difficult to exclude?

tests/js/JSStandardBearTest.py
```diff
@@ -17,7 +17,7 @@
   bar = bar === 1
     ? bar
     : 1
-  if ((x = 33)) {
+  if ((x === 33)) {
```
jayvdb: These changes to the tests should be a separate commit, which goes before your 'upgrade to eslint 3' commit, or even a separate PR.

If they are a separate PR we know that these test changes worked on ESLint 2 as well, before we upgrade to ESLint 3.  (Imagine that we need to revert your commit ... we would also revert these test changes, which isnt necessary.)

The extra commit / PR does not need a separate issue.  You can use `Related to https://github.com/coala/coala-bears/issues/872` for the last line of the commit.

blazeu:
Tests PR @https://github.com/coala/coala-bears/pull/2098

bears/js/ESLintBear.py
```diff
@@ -58,15 +58,28 @@ def generate_config(filename, file):
         return '{"extends": "eslint:recommended"}'
 
     def process_output(self, output, filename, file):
+        # Taking output from stderr for ESLint 2 program errors.
+        # ESLint 2 is no longer supported, but it is here so that anyone
+        # with ESLint 2 will still see any errors that were emitted.
         if output[1]:
+            output = output[1]
+        else:
+            output = output[0]
```
Makman2: but we require ESLint 3 now, don't we?

blazeu: Do you mean ESLint 2 ?

Unfortunately we need ESLint 2 just to test this block :(

blazeu: See @jayvdb review

blazeu:
IMO no one uses ESLint 2 and it is [no longer updated](https://github.com/eslint/eslint/tree/2.x).

bears/js/ESLintBear.py
jayvdb: @blazeu , this is a good case to use `# pragma: no cover` (on [the branch](http://coverage.readthedocs.io/en/coverage-4.3.4/excluding.html) which will not be covered by the tests.)

bears/js/ESLintBear.py
```diff
@@ -58,15 +58,28 @@ def generate_config(filename, file):
         return '{"extends": "eslint:recommended"}'
 
     def process_output(self, output, filename, file):
-        if output[1]:
+        # Taking output from stderr for ESLint 2 program errors.
+        # ESLint 2 is no longer supported, but it is here so that anyone
+        # with ESLint 2 will still see any errors that were emitted.
+        if output[1]:  # pragma: no cover
+            output = output[1]
```
jayvdb: I think you can put the `# pragma` here, making it much more specific.

However, the previous code allowed for errors on stderr, which were reported with `self.warn`, *and* real results in json on stdout which were reported as `Result` objects.
Your modified approach only allows stderr _or_ stdout to be processed.

bears/js/ESLintBear.py
Makman2: > But, unfortunately we need ESLint 2 just to test this block 

Yeah but I think because of ESLint 3 we don't need this block at all :O

bears/js/ESLintBear.py
blazeu: I think if there's an error in stderr there shouldn't be any result?

But I'll modify it anyway just for precaution.

Makman2: From before:

> But, unfortunately we need ESLint 2 just to test this block

Yeah but I think because of the ESLint 3 requirement we don't need this block at all :O

jayvdb: > Yeah but I think because of the ESLint 3 requirement we don't need this block at all

Everyone currently *is* using ESLint 2.x, with explicit pins.  either they are doing `npm install -g eslint=2.x`, or unlikely they are fetching and using the `package.json` out of the coala-bears repo , which isnt an easy option as it also requires a custom `NODE_PATH` (a problem which stumped us for months).

If someone upgrades their CI env to using coala 0.12 (i.e. with this patch in it), and doesnt update the eslint version in their CI, any eslint 2 errors would be silently discarded if stderr is disregarded.

This is why it is so important that coala get non-Python dependency management sorted out properly, so that it bombs out at the beginning of CI with a version error.

Makman2: Ahh ugh okay... :+1:

bears/js/ESLintBear.py
```diff
@@ -58,15 +58,31 @@ def generate_config(filename, file):
...
+            self.warn(message)
 
-        if not file or not output[0]:
+        # Taking output from stderr for ESLint 2 program errors.
+        # ESLint 2 is no longer supported, but it is here so that anyone
+        # with ESLint 2 will still see any errors that were emitted.
+        if output[1]:  # pragma: no cover
+            warn_issue(output[1])
+
+        output = output[0]
```
jayvdb: Can you remove this line.
If you keep using `output[0]` , the diff will be smaller, and git will recognise that `output = json.loads(output[0])` was indented.

jayvdb:
ack https://github.com/coala/coala-bears/pull/2086/commits/2c361cf2770b454bc6a507efabbb3af4bc2e6af7

jayvdb:
@rultor merge

PR #2096, opened 2017-10-29
WIP: AnnotationBear: Use Pygments for file parsing

userzimmermann:
And here comes the PR for https://github.com/coala/coala-bears/issues/2092

The only problem is that `Pygments` does not report (yet) any errors on missing closing delimiters :/ That's the reason for the 2 **FAIL**ing test cases. Maybe we can add some custom logic on top of `Pygments` for that...  So I'm marking this PR `WIP` for now

At least this PR shows how much we can simplify the `AnnotationBear` or similar cases of custom code parsing

The 2 changes to expected single line comment end positions in the tests are due to `Pygments` not interpreting the newline as part of the comment

### DIFF
--- a/bears/general/AnnotationBear.py
+++ b/bears/general/AnnotationBear.py
@@ -1,3 +1,9 @@
+import re
+
+from pygments.lexers import get_lexer_by_name
+from pygments.token import Token
+from pygments.util import ClassNotFound
+
 from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.HiddenResult import HiddenResult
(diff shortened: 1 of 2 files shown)


bears/general/AnnotationBear.py
```diff
@@ -31,274 +37,50 @@ def run(self, filename, file, language: str, coalang_dir: str = None):
...
-        multiline_string_delimiters = dict(
-            lang_dict['multiline_string_delimiters'])
-        multiline_comment_delimiters = dict(
-            lang_dict['multiline_comment_delimiters'])
-        comment_delimiter = dict(lang_dict['comment_delimiter'])
-        string_ranges = comment_ranges = ()
+        # the HiddenResult's content dict of SourceRange objects
+        content = {'comments': (), 'strings': ()}
+        # the pygments token types for the different types of source ranges
+        token_types = {'comments': Token.Comment, 'strings': Token.String}
```
aptrishu: what we needed from the new annotationbear is that it should separately return ranges for singleline/multiline strings/comments.

userzimmermann: @aptrishu Can be next step... ;)

userzimmermann:
@aptrishu @Makman2 I also added separate `multiline_comment`, `singleline_comment`, `multiline_string`, and `singleline_string` range categories

Makman2:
Closing due to inactivity.

PR #2105, opened 2017-11-01
CPPLintBear.py: Verify indentation and spaces

swapagarwal:
`cpplint` only performs indentation checking using 2 spaces, according
to https://google.github.io/styleguide/cppguide.html#Spaces_vs._Tabs
Add check_invalid_configuration function to CPPLintBear.py.
Add tests for invalid configurations.

Fixes https://github.com/coala/coala-bears/issues/896

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/c_languages/CPPLintBear.py
+++ b/bears/c_languages/CPPLintBear.py
@@ -3,6 +3,12 @@ from dependency_management.requirements.PipRequirement import PipRequirement
 from coalib.settings.Setting import typed_list
 
 
+def check_invalid_configuration(indent_size, use_spaces):
+    if indent_size != 2 or not use_spaces:
+        raise ValueError('CPPLint only supports an indentation size of 2, '
+                         'using only spaces.')
+
+
 @linter(executable='cpplint',
         use_stdout=False,
         use_stderr=True,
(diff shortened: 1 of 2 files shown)


bears/c_languages/CPPLintBear.py
```diff
@@ -26,12 +32,17 @@ class CPPLintBear:
     def create_arguments(filename, file, config_file,
                          max_line_length: int=79,
                          cpplint_ignore: typed_list(str)=(),
-                         cpplint_include: typed_list(str)=()):
+                         cpplint_include: typed_list(str)=(),
+                         indent_size: int=2,
```
sims1253: Do we want these as defaults or do we want the user to fail without any given argument. With a default value, users might think 4 just works. But I guess they should read the docs and then they'd see.

swapagarwal: Case A: Fails without any given argument. Then maybe they set this argument to 4, which fails again. Then they read the docs and set it to 2.
Case B: Works by default. If user wants to override, docs clearly call out the permitted value.
Thoughts?

bears/c_languages/CPPLintBear.py
```diff
@@ -26,12 +32,17 @@ class CPPLintBear:
...
+                         use_spaces: bool=True):
         """
```
aashraybhandar1: Is there any other way to check what kind of indent_size the user is using than relying on the user himself to give that information? Also does this fix it in case the indent_size is not matching or the user is using tabs?

PR #2121, opened 2017-11-09
GitCommitBear: Fix behaviour for empty body

Makman2:
The bear raised an exception when only the shortlog was included
in the commit message. This commit is meant to fix this bug.

Fixes https://github.com/coala/coala-bears/issues/2031

### DIFF
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -299,7 +299,8 @@ class GitCommitBear(GlobalBear):
             return
 
         if body_close_issue_on_last_line:
-            body = body.splitlines()[-1]
+            if body:
+                body = body.splitlines()[-1]
             result_message = ('Body of HEAD commit does not contain any {} '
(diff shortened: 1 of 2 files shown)


prnvdixit:
ack 419fb094367953585b5f2428ab1855b3657e8b96

prnvdixit:
BTW don't we always add a issue reference to commit? (that shouldn't keep the commit body empty). I mean - In which cases. can the commit body be empty? :/

meetmangukiya:
> BTW don't we always add a issue reference to commit?

No, we don't always add a issue reference to a commit

meetmangukiya:
@rultor merge

Makman2:
@prnvdixit this usually happens on bigger PRs addressing more difficult issues. To get some issue closed, more changes in code are needed which can often packed into several atomic commits instead of a single one (makes review easier and code more intuitive). Although those commits relate to the issue, this is sometimes not mentioned (it's currently not always mandatory).

Issue #2078, opened 2017-10-24
Make CppCheckBear a GlobalBear


Psirus:
Let's say I have a function `f`, defined in `f.cpp` and used in `g.cpp`. If I run `cppcheck` just on `f.cpp`, it will tell me `[f.cpp:1]: (style) The function 'f' is never used.`. If I run cppcheck on the whole directory, that message disappears, because it sees that the function is being used after-all. However, in coala, these messages also appear, because, presumably, coala runs cppcheck on each individual file. Can I tell it to run on the whole directory?

Makman2:
`CppCheckBear` is currently configured as a `LocalBear`, meaning it works like you said. We can change it to be a `GlobalBear` and adapt it a bit to pass all files at once.

Psirus:
If that doesn't break the usecase of someone else, great!

urohit011:
I would like to work on this.

Makman2:
Then, get yourself assigned ;)

khanchi97:
@urohit011 are you still working on it?

Makman2:
Unassigning due to inactivity.

PR #2124, opened 2017-11-12
CPPCheckBear.py: Make CPPCheckBear a GlobalBear

khanchi97:
It solves the following problem #2078 
>Let's say I have a function f, defined in f.cpp and used in g.cpp. If I run cppcheck just on f.cpp, it will tell me [f.cpp:1]: (style) The function 'f' is never used.. If I run cppcheck on the whole directory, that message disappears, because it seems that the function is being used after all. However, in coala, these messages also appear, because, presumably, coala runs cppcheck on each individual file. Can I tell it to run on the whole directory?

### DIFF
--- a/bears/c_languages/CPPCheckBear.py
+++ b/bears/c_languages/CPPCheckBear.py
@@ -8,6 +8,7 @@ from coalib.settings.Setting import typed_list
 @linter(executable='cppcheck',
         use_stdout=False,
         use_stderr=True,
+        global_bear=True,
         output_format='regex',
         output_regex=r'(?P<line>\d+):(?P<severity>[a-zA-Z]+):'
                      r'(?P<origin>[a-zA-Z]+):(?P<message>.*)',
@@ -27,8 +28,7 @@ class CPPCheckBear:
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Security', 'Unused Code', 'Unreachable Code', 'Smell'}
 
-    @staticmethod
-    def create_arguments(filename, file, config_file,
+    def create_arguments(self, config_file,
                          enable: typed_list(str)=[]):
         """
         :param enable:
@@ -38,8 +38,9 @@ class CPPCheckBear:
             missingInclude
         """
         args = ('--template={line}:{severity}:{id}:{message}',)
+        files = tuple(self.file_dict.keys())
 
         if enable:
             args += ('--enable=' + ','.join(enable),)
 
-        return args + (filename,)
+        return args + files
--- a/tests/c_languages/CPPCheckBearTest.py
+++ b/tests/c_languages/CPPCheckBearTest.py
@@ -1,37 +1,35 @@
+import os
+from queue import Queue
+
 from bears.c_languages.CPPCheckBear import CPPCheckBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
-
-good_file = """
-using namespace std;
-int main() {
-    cout << "Hello, world!" << endl;
-    return 0;
-}"""
-
-warn_file = """
-void f1(struct fred_t *p)
-{
-    int x;
-    if (p)
-        do_something(x);
-}"""
-
-bad_file = """
-#define f(c) { \
-    char s[10]; \
-    s[c] = 42; \
-}
-int main() {
-    f(100);
-    return 0;
-}"""
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
 
 
-CPPCheckBearTest1 = verify_local_bear(CPPCheckBear,
-                                      valid_files=(good_file, warn_file),
-                                      invalid_files=(bad_file,))
+def get_absolute_test_path(file):
+    return os.path.join(os.path.dirname(__file__),
+                        'cppcheck_test_files', file)
 
-CPPCheckBearTest2 = verify_local_bear(CPPCheckBear,
-                                      valid_files=(good_file,),
-                                      invalid_files=(warn_file, bad_file),
-                                      settings={'enable': 'unusedFunction'})
+
+@generate_skip_decorator(CPPCheckBear)
+class CPPCheckBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('cppcheck')
+        self.uut = CPPCheckBear(self.section, Queue())
+        self.good_file = get_absolute_test_path('good_file.cpp')
+        self.bad_file = get_absolute_test_path('bad_file.cpp')
+        self.warn_file = get_absolute_test_path('warn_file.cpp')
+
+    def test_default(self):
+        self.check_validity(self.uut, [], self.good_file)
+        self.check_invalidity(self.uut, [], self.bad_file)
+        self.check_validity(self.uut, [], self.warn_file)
+
+    def test_enable(self):
+        self.section.append(Setting('enable', 'unusedFunction'))
+        self.check_validity(self.uut, [], self.good_file)
+        self.check_invalidity(self.uut, [], self.bad_file)
+        self.check_invalidity(self.uut, [], self.warn_file)
--- a/tests/c_languages/CPPCheckBearTest.py
+++ b/tests/c_languages/CPPCheckBearTest.py
@@ -1,9 +1,9 @@
 import os
+import unittest
 from queue import Queue
 
 from bears.c_languages.CPPCheckBear import CPPCheckBear
 from coalib.testing.BearTestHelper import generate_skip_decorator
-from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
 
@@ -14,22 +14,35 @@ def get_absolute_test_path(file):
 
 
 @generate_skip_decorator(CPPCheckBear)
-class CPPCheckBearTest(LocalBearTestHelper):
+class CPPCheckBearTest(unittest.TestCase):
 
     def setUp(self):
         self.section = Section('cppcheck')
-        self.uut = CPPCheckBear(self.section, Queue())
-        self.good_file = get_absolute_test_path('good_file.cpp')
-        self.bad_file = get_absolute_test_path('bad_file.cpp')
-        self.warn_file = get_absolute_test_path('warn_file.cpp')
+        self.file_dict = {}
+        self.queue = Queue()
+        self.test_files = ['good_file.cpp', 'bad_file.cpp', 'warn_file.cpp']
 
-    def test_default(self):
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
-        self.check_validity(self.uut, [], self.warn_file)
+    def get_results(self, files_to_check):
+        files = [get_absolute_test_path(file) for file in files_to_check]
+        for filename in files:
+            with open(filename, 'r', encoding='utf-8') as content:
+                self.file_dict[filename] = tuple(content.readlines())
+        self.uut = CPPCheckBear(self.file_dict,
+                                self.section,
+                                self.queue)
+        return list(self.uut.run_bear_from_section([], {}))
 
-    def test_enable(self):
+    def test_results_complete(self):
         self.section.append(Setting('enable', 'unusedFunction'))
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
-        self.check_invalidity(self.uut, [], self.warn_file)
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertEqual(len(messages), 2)
+        self.assertRegex(messages[0], 'Array .+ out of bounds')
+        self.assertRegex(messages[1], "function 'f1' .+ never used")
+
+    def test_no_enable_entered(self):
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertEqual(len(messages), 1)
+        self.assertRegex(messages[0], 'Array .+ out of bounds')
+        self.assertNotRegex(messages[0], 'function .+ never used')
--- a/tests/c_languages/cppcheck_test_files/bad_file.cpp
+++ b/tests/c_languages/cppcheck_test_files/bad_file.cpp
@@ -1,0 +1,8 @@
+#define f(c) { \
+    char s[10]; \
+    s[c] = 42; \
+}
+int main() {
+    f(100);
+    return 0;
+}
--- a/tests/c_languages/cppcheck_test_files/good_file.cpp
+++ b/tests/c_languages/cppcheck_test_files/good_file.cpp
@@ -1,0 +1,5 @@
+using namespace std;
+int main() {
+    cout << "Hello, world!" << endl;
+    return 0;
+}
--- a/tests/c_languages/cppcheck_test_files/warn_file.cpp
+++ b/tests/c_languages/cppcheck_test_files/warn_file.cpp
@@ -1,0 +1,6 @@
+void f1(struct fred_t *p)
+{
+    int x;
+    if (p)
+        do_something(x);
+}


khanchi97:
@Makman2 why the test build is failing? Locally all the tests are passing.This is what logs are saying https://travis-ci.org/coala/coala-bears/jobs/307038389#L3825

khanchi97:
Hello @NiklasMM @adtac 
would you please help me with this.I have written test against `CPPCheckBear` which is now changed to GlobalBear by me in this issue.But tests are failing on CI and working fine locally.I'm trying to solve this issue from last few days 😞

NiklasMM:
Take a look at the output of the CI steps, they tell you exactly what's wrong. 

In this case the tests are failing because you don't have 100% coverage. You're missing the jump from line 45 to line 48 in `CPPCheckBear.py`, i.e. there is no test that does not enter the `if` branch. Add one that does and you should be ok.

khanchi97:
Thank so much 😃

khanchi97:
Hello @NiklasMM 
Now test coverage is 100% but still there are some issues
1. In appveyor : it is showing that `cppcheck` is not installed.How this problem can be raised.
2. In CircleCI and TravisCI : both have same issue i.e `timeout`.Any hint how I can solve this problem.

NiklasMM:
Regarding 1: Appveyor runs Windows, where apparently cpplint isn't available. You'll need pytest to skip these tests. You can probably do that using the [`generate_skip_decorator`](https://github.com/coala/coala/blob/master/coalib/testing/BearTestHelper.py). Look around the coala-bears codebase to find out how it works.

khanchi97:
Okay, I have added `generate_skip_decorator`.What do you say about the 2nd point.Or should I first check it by pushing it again.

khanchi97:
@NiklasMM It persists.I have added `generate_skip_decorator`But still that problem remains same.

NiklasMM:
Ok, the timeout problem persists with all of your tests. I don't know cppcheck, so I can only guess what's happening. A timeout means the Python process is waiting for the other process (cppcheck) to terminate and it never does. This is often the case, when the other process is waiting for user input. Could that be the case?

khanchi97:
>This is often the case when the other process is waiting for user input. Could that be the case?

No, there is no any case where it is waiting for user input.But one thing that I got is timeout is because of using `os.getcwd()` because then bear have to collect all the `C/C++` files from that directory and analyze them.And when I use this locally it works but in TravisCI it gives timeout error.I think it is probably the main reason for timeout error on CI.

khanchi97:
Now two tests are passing but one is not passing which is the coverage of `os.getcwd()`

khanchi97:
Hello Niklas! 
I couldn't understand how in CircleCI result messages are slightly different as expected and just because of this, CircleCI is failing and TravisCI is passed.

NiklasMM:
Looks like you got it now :)

bears/c_languages/CPPCheckBear.py
```diff
@@ -28,18 +29,21 @@ class CPPCheckBear:
...
+                         path_to_dir: str,
                          enable: typed_list(str)=[]):
         """
         :param enable:
             Choose specific issues to report. Issues that can be
             reported are: all, warning, style, performance,
             portability, information, unusedFunction,
             missingInclude
+        :param path_to_dir:
+            Path to the directory where analysis to be done.
```
Makman2: nope, please look at the signatures of other global-bear-linters, coala feeds in all files in a file-dict. Having a setting for this is definitely bad design, then you have to do
```
# in a coafile
files = **
path_to_dir = this_dir  # <-- hmm that's what actually "files" is for
```

khanchi97: I have tried this
`files = this_dir`
then coala says
No files matching `this_dir` were found.

khanchi97: @Makman2 
Because --files parameter need direct path to files not to directory.May be it can be done using glob pattern like this
`files = this_dir/**.cpp, this_dir/**.c`
Is it right to do like this?

newbazz: **.c will assure that all the c files in any depth will be taken care of so, IMO this_dir/**.c didn't make much sense, still lets see what makman says

khanchi97: @newbazz this linter analyze C and CPP files.So If this way that I propose is right then I think there is no problem in 
`files = this_dir/**.cpp, this_dir/**.c`

Makman2: that's the way we go, the user should specify all C/C++ files and you pass all of them to the linter.

khanchi97: @Makman2 But the problem is how I can get the value of --files argument in `create_argument` method because while using linter for writing globalbear then only allowed option are config_file, and extra setting that we add.

khanchi97: @NiklasMM ^ any help, please

NiklasMM: Well, by design a global bear does not need to be told which files it should look at, because it looks at the whole repo. That's what defines a global bear. You should be fine with just using the current directory `.`. Otherwise using a global bear is not the correct solution.

khanchi97:
I think now it is fine.

Makman2:
Why do you move the test file? Please keep it at the same location :+1:

Makman2:
unack 19d18ea

RaiVaibhav:
@Makman2  If I get the things right then @khanchi97 ,is also taking a case in which if one file is extended into another (that `#include` thing)(that is the issue), so to test that it have to move the test file.

RaiVaibhav:
ok, got it know what you are trying to say :sweat_smile: , silly things

khanchi97:
@Makman2 I have moved the test files.

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,37 +1,50 @@
...
+                self.file_dict[filename] = tuple(_file.readlines())
+        self.uut = CPPCheckBear(self.file_dict,
+                                self.section,
+                                self.queue)
+        return list(self.uut.run_bear_from_section([], {}))
+
+    def test_results_complete(self):
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertTrue("The function 'fun' is never used" in messages[1])
```
Makman2: 1. there's `assertIn`
2. Why not asserting for the complete message?
   ```
   self.assertEqual(message, '...')
   ```

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,37 +1,50 @@
...
-
-
-CPPCheckBearTest1 = verify_local_bear(CPPCheckBear,
-                                      valid_files=(good_file, warn_file),
-                                      invalid_files=(bad_file,))
-
-CPPCheckBearTest2 = verify_local_bear(CPPCheckBear,
-                                      valid_files=(good_file,),
-                                      invalid_files=(warn_file, bad_file),
-                                      settings={'enable': 'unusedFunction'})
```
Makman2: If you are converting tests to your "new" style, I would request another commit prepended to this one that does the refactor. Makes review easier^^

Makman2:
Note: Don't do it in this PR, but I've tested that using `list` instead of `tuple` to construct the arguments with is faster (10-20%). Let's file an issue after this PR to research on the speed optimization and implement it.

tests/c_languages/CPPCheckBearTest.py
khanchi97: Can not use asserting for the complete message because there is little bit difference between `result.messages` given by TravisCI and CircleCI.In this case, there is a difference of trailing period in the messages.

khanchi97: >1. there's `assertIn`

Do you want me to use `assertIn` in place of `assertTrue`

khanchi97:
>Note: Don't do it in this PR, but I've tested that using list instead of tuple to construct the arguments with is faster (10-20%). Let's file an issue after this PR to research on the speed optimization and implement it.

okay :) @Makman2

tests/c_languages/CPPCheckBearTest.py
khanchi97: @Makman2 I have made the required changes.

Makman2:
Do the commits pass the tests on their own? I meant to convert only old tests to the "new" style in the first commit, additional tests and changes due to the consequence of making CPPCheckBear a global bear should go into the second bear (that also does the refactor of CPPCheckBear).

khanchi97:
>I meant to convert only old tests to the "new" style in the first commit

@Makman2  Do you want me to change these tests https://pastebin.com/raw/hSeMzr0E into this https://pastebin.com/raw/BdwWAxTm style.

Makman2:
Yes exactly :)

(sorry for the late reply)

khanchi97:
@Makman2 is this commit is right for changing tests to the new style?Should I go for next commit?

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,37 +1,34 @@
+import os
+
```
Makman2: you can leave this newline, `queue` is a system import ;)

tests/c_languages/cppcheck_test_files/bad_file.cpp
```diff
@@ -0,0 +1,8 @@
+#define f(c) { 
+    char s[10];
+    s[c] = 42;
+}
```
Makman2: have you seen that the original file has escaped the `\n`, so the first `#define` directive appears only on the first line? (Otherwise it would be invalid C++ code anyway)

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,37 +1,34 @@
...
+        self.section = Section('cppcheck')
+        self.uut = CPPCheckBear(self.section, Queue())
+        test_files = os.path.join(os.path.dirname(__file__),
+                                  'cppcheck_test_files')
+        self.good_file = os.path.join(test_files, 'good_file.cpp')
+        self.bad_file = os.path.join(test_files, 'bad_file.cpp')
+        self.warn_file = os.path.join(test_files, 'warn_file.cpp')
+
+    def test_run(self):
+        self.check_validity(self.uut, [], self.good_file)
```
Makman2: does `check_validity` takes the file path or its content? Can't recall...

tests/c_languages/CPPCheckBearTest.py
khanchi97: okay, will remove that

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,37 +1,34 @@
...
+        self.check_validity(self.uut, [], self.warn_file)
+        self.check_invalidity(self.uut, [], self.bad_file)
+        self.section.append(Setting('enable', 'unusedFunction'))
+        self.check_invalidity(self.uut, [], self.warn_file)
```
Makman2: please test also `good_file` and `bad_file` again like the test before ;)

You can maybe split up the test into two functions, `test_default` and `test_enable` or so. Up to you though :+1: (but I would recommend it ;D)

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,37 +1,34 @@
...
+class CPPCheckBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('cppcheck')
+        self.uut = CPPCheckBear(self.section, Queue())
+        test_files = os.path.join(os.path.dirname(__file__),
+                                  'cppcheck_test_files')
+        self.good_file = os.path.join(test_files, 'good_file.cpp')
+        self.bad_file = os.path.join(test_files, 'bad_file.cpp')
+        self.warn_file = os.path.join(test_files, 'warn_file.cpp')
```
Makman2: I think you can use `get_absolute_test_path` here instead of `os.path.join`, then you also save the line with 
```
test_files = os.path.join(...)
```

tests/c_languages/CPPCheckBearTest.py
khanchi97: yes, it takes list of strings as `lines` and filename

tests/c_languages/CPPCheckBearTest.py
khanchi97: okay, I will do that.

tests/c_languages/CPPCheckBearTest.py
khanchi97: okay :)

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -15,20 +15,36 @@ def get_absolute_test_path(file):
 
 
 @generate_skip_decorator(CPPCheckBear)
-class CPPCheckBearTest(LocalBearTestHelper):
+class CPPCheckBearTest(unittest.TestCase):
```
Makman2: Why not keeping `LocalBearTestHelper`? Then you can also leave the tests from the earlier commit :+1:

tests/c_languages/CPPCheckBearTest.py
Makman2: ahh then you have to fill them in from the test file, passing `[]` means empty file :+1:

Makman2: otherwise the content of the test files won't be used ;)

tests/c_languages/CPPCheckBearTest.py
khanchi97: How I can keep `LocalBearTestHelper` as I am writing test for a `GlobalBear`

tests/c_languages/cppcheck_test_files/bad_file.cpp
khanchi97: @Makman2 I didn't get you.Can I change this file to some other example?

khanchi97: >have you seen that the original file has escaped the \n, so the first #define directive appears only on the first line? (Otherwise it would be invalid C++ code anyway)

okay I got it :)
now I think I should escape these lines so that it appears only on the first line.

tests/c_languages/CPPCheckBearTest.py
Makman2: ahhh yeah good point sorry :D

khanchi97: I have made the required changes.
I'm trying to write `GlobalBearTestHelper` because now I have some knowledge about how `LocalBearTestHelper` is working ;) Should I go for it or will it be too hard for me now.

khanchi97:
ping ^ @Makman2

shreyateeza:
@khanchi97 Squash your commits.

Makman2:
@khanchi97 he doesn't have to, I asked him to separate his original commit into these two

tests/c_languages/cppcheck_test_files/good_file.cpp
```diff
@@ -1,5 +1,11 @@
...
-    cout << "Hello, world!" << endl;
+
+int main()
+  {
+    int a;
+    a = f(3,5);
+    cout << a <<endl;
     return 0;
-}
+  }
```
Makman2: hm this is "good" C++? ;D

tests/c_languages/cppcheck_test_files/bad_file.cpp
```diff
@@ -1,8 +0,0 @@
-#define f(c) { \
```
Makman2: why removing this file?

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -1,9 +1,10 @@
 import os
+import unittest
+
```
Makman2: Hm now there's an unnecessary newline at this place 😅 probably GitHub didn't recognize it properly ;)

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +15,36 @@ def get_absolute_test_path(file):
...
         self.section.append(Setting('enable', 'unusedFunction'))
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
-        self.check_invalidity(self.uut, [], self.warn_file)
+        self.file_dict = {}
+        self.queue = Queue()
+        self.test_files = ['good_file.cpp', 'bad_file.c', 'warn_file.cpp']
+
+    def get_results(self, files_to_check):
+        self.files = [get_absolute_test_path(file) for file in files_to_check]
```
Makman2: this can be made a local variable, no need to assign to `self`

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +15,36 @@ def get_absolute_test_path(file):
...
-        self.check_validity(self.uut, [], self.warn_file)
-
-    def test_enable(self):
         self.section.append(Setting('enable', 'unusedFunction'))
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
-        self.check_invalidity(self.uut, [], self.warn_file)
+        self.file_dict = {}
+        self.queue = Queue()
+        self.test_files = ['good_file.cpp', 'bad_file.c', 'warn_file.cpp']
```
Makman2: Don't you want to not aggregate them here, so you can individually test files and make tests more reliable?

tests/c_languages/cppcheck_test_files/bad_file.cpp
khanchi97: I have replaced this `bad_file.cpp` with `bad_file.c` and made some changes to ensure that `cppcheck` working on both C and  C++ files :)
Do you want me to keep the same content in `bad_file.c`?

tests/c_languages/CPPCheckBearTest.py
khanchi97: @Makman2 here we don't have to test individual files, this is a `Globalbear ` and we can/should pass all the files at once.Shouldn't we?

khanchi97:
@Makman2 I have made the required changes

tests/c_languages/cppcheck_test_files/bad_file.cpp
Makman2: if possible yes, please keep old test files and create new ones if you want to test new aspects :+1:

khanchi97: Done! Is it fine now?I have removed `bad_file.c` because it had the same aspect as `bad_file.cpp`

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +14,36 @@ def get_absolute_test_path(file):
...
-        self.check_invalidity(self.uut, [], self.warn_file)
+        self.file_dict = {}
+        self.queue = Queue()
+        self.test_files = ['good_file.cpp', 'bad_file.cpp', 'warn_file.cpp']
+
+    def get_results(self, files_to_check):
+        files = [get_absolute_test_path(file) for file in files_to_check]
+        for filename in files:
+            with open(filename, 'r', encoding='utf-8') as _file:
+                self.file_dict[filename] = tuple(_file.readlines())
```
Makman2: underscore prefixes usually only occur in classes for members that are meant to be "protected". Please don't use the underscore as a prefix for local variables :+1:

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +14,36 @@ def get_absolute_test_path(file):
...
+                                self.section,
+                                self.queue)
+        return list(self.uut.run_bear_from_section([], {}))
+
+    def test_results_complete(self):
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertIn("The function 'fun' is never used", messages[1])
+        self.assertIn("Array 's[10]' ", messages[0])
+        self.assertIn('out of bounds', messages[0])
```
Makman2: I think you should assert with a regex, instead of puzzling here with `assertIn` ;)

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +14,36 @@ def get_absolute_test_path(file):
...
+
+    def test_no_enable_entered(self):
+        self.section.delete_setting('enable')
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertNotIn("The function 'fun' is never used.", messages)
+        self.assertIn("Array 's[10]' ", messages[0])
+        self.assertIn('out of bounds', messages[0])
```
Makman2: same here ;)

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +14,36 @@ def get_absolute_test_path(file):
...
-        self.bad_file = get_absolute_test_path('bad_file.cpp')
-        self.warn_file = get_absolute_test_path('warn_file.cpp')
-
-    def test_default(self):
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
-        self.check_validity(self.uut, [], self.warn_file)
-
-    def test_enable(self):
         self.section.append(Setting('enable', 'unusedFunction'))
```
Makman2: I think you should move this one to `test_results_complete`, then you don't have to delete this setting again in `test_no_enable_entered`

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +14,36 @@ def get_absolute_test_path(file):
...
+        messages = [result.message for result in results]
+        self.assertIn("The function 'fun' is never used", messages[1])
+        self.assertIn("Array 's[10]' ", messages[0])
+        self.assertIn('out of bounds', messages[0])
+
+    def test_no_enable_entered(self):
+        self.section.delete_setting('enable')
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertNotIn("The function 'fun' is never used.", messages)
```
Makman2: Above you do `messages[1]`. It seems the strings are longer, so I think doing this doesn't make much sense. You should do something like
```
for message in messages:
    self.assertNotIn(..., message)
```
to be sure it's definitely not appearing.

tests/c_languages/CPPCheckBearTest.py
Makman2: You should assert for the length of messages, as well as all remaining messages in the tests.

tests/c_languages/cppcheck_test_files/good_file.cpp
Makman2: bump

tests/c_languages/cppcheck_test_files/good_file.cpp
```diff
@@ -1,5 +1,11 @@
...
+
```
Makman2: I'm still confused why you change this, really create new test files, even though they are similar so we can be sure we don't have regressions.

tests/c_languages/CPPCheckBearTest.py
khanchi97: Done!

tests/c_languages/CPPCheckBearTest.py
```diff
@@ -14,22 +14,36 @@ def get_absolute_test_path(file):
...
+        self.assertRegexpMatches(messages[0], "Array .+ out of bounds")
+        self.assertRegexpMatches(messages[1], "function 'f1' .+ never used")
+
+    def test_no_enable_entered(self):
+        results = self.get_results(self.test_files)
+        messages = [result.message for result in results]
+        self.assertEqual(len(messages), 1)
+        self.assertRegexpMatches(messages[0], "Array .+ out of bounds")
+        for message in messages:
+            self.assertNotRegexpMatches(message, 'function .+ never used')
```
khanchi97: Are these asserts using regex are fine now? @Makman2

Makman2:
ack c7e4e0b 9f76a58

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

PR #2134, opened 2017-11-19
Build setup.py and requirements.txt using moban

jayvdb:
This depends on cloning https://github.com/jayvdb/coala-mobans beside `coala-bears` , and https://github.com/moremoban/moban/pull/12


.ci/generate_bear_requirements.py
```diff
@@ -121,64 +115,60 @@ def get_all_requirements(bears):
...
-        req_version = requirement.version
         package_name = requirement.package
+        req_version = requirement.version
         if req_version:
-            if req_version[0] in ('<', '>', '~', '='):
-                npm_dependencies[package_name] = req_version
-            else:
-                npm_dependencies[package_name] = "~" + req_version
-        else:
-            npm_dependencies[package_name] = "*"
```
Monal5031: Did we miss this in PR?

Monal5031:
ack 803db03 
ack d8bb8fc

bear-requirements.yaml
```diff
@@ -0,0 +1,187 @@
+gem_requirements:
```
yukiisbored: Hey, for consistency sake, Use .yml instead of yaml

.ci/generate_bear_requirements.py
jayvdb: This was from when `package.json` had items without any version specified, and isnt needed any longer.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # NOTE: This file is parsed by .ci/generate_bear_requirements.py
-# Use >= for development versions so that source builds always work
+# Edit coala_dependency in coala-build.yaml
```
jayvdb: this should be called `coalib_dependency`, and should reside in the local yaml, not in the global yaml.

jayvdb: The intention of using a global `coala_dependency` is that at release time all repos should switch to using the latest, and not have a local override pinning them to an old version.

The best way to do that is to update the global `coala_dependency` to a pre-release , watch stuff break, fix stuff, and iterate until nothing is broken.

The problem is that the bears repo often needs to use a more recent coala pre-release , and that change would flow to all other repos.  I guess that isnt so bad, as it ensures all repos are in sync more often, and coala core hasnt broken something drastic.

yukiisbored:
ack https://github.com/coala/coala-bears/commit/07304a444f90ef651c69bfc6de514c2064118e3d

jayvdb:
@gitmate-bot ff

PR #2160, opened 2017-12-10
FilenameBear.py: Add filename length check

damian1996:
Adds the check if filename has length more than 4096 on UNIX
systems or more than 260 on Windows.
In that case, user will get message about it. 

Closes https://github.com/coala/coala-bears/issues/2003

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/FilenameBear.py
+++ b/bears/general/FilenameBear.py
@@ -26,6 +26,7 @@ class FilenameBear(LocalBear):
                                   }
 
     def run(self, filename, file,
+            max_filename_length: int=260,
             file_naming_convention: str=None,
             ignore_uppercase_filenames: bool=True,
             filename_prefix: str='',
@@ -42,6 +43,8 @@ class FilenameBear(LocalBear):
             - ``pascal`` (``ThisIsPascalCase``)
             - ``snake`` (``this_is_snake_case``)
             - ``space`` (``This Is Space Case``)
+        :param max_filename_length:
+            Maximum filename length on both Windows and Unix-like systems.
         :param ignore_uppercase_filenames:
             Whether or not to ignore fully uppercase filenames completely,
             e.g. COPYING, LICENSE etc.
@@ -97,6 +100,12 @@ class FilenameBear(LocalBear):
             messages.append(
                 'Filename does not use the suffix {!r}.'.format(
                     filename_suffix))
+
+        if len(filename) > max_filename_length:
+            messages.append(
+                'Filename is too long ({} > {}).'.format(
+                    len(filename),
+                    max_filename_length))
 
         if ignore_uppercase_filenames and filename_without_extension.isupper():
             return
--- a/tests/general/FilenameBearTest.py
+++ b/tests/general/FilenameBearTest.py
@@ -151,3 +151,42 @@ class FilenameBearTest(LocalBearTestHelper):
                                 severity=RESULT_SEVERITY.NORMAL,
                                 file='filename.xyz')],
             filename='filename.xyz')
+
+    def test_file_with_too_long_filename(self):
+        msg = 'Filename is too long ({} > {}).'
+        filename_test1 = '_filenamewhichistoolong'*15 + '.xyz'
+        filename_test2 = '_filenamewhichistoolong'*10 + '.xyz'
+        filename_test3 = '_validfilenamelength'*13
+        max_filename_length = 260
+        self.check_invalidity(
+            self.uut, [''], filename=filename_test1)
+        self.check_validity(
+            self.uut, [''], filename=filename_test2)
+        self.check_validity(
+            self.uut, [''], filename=filename_test3)
+        self.check_results(
+            self.uut,
+            [''],
+            [Result.from_values('FilenameBear',
+                                msg.format(len(filename_test1),
+                                           max_filename_length),
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file=filename_test1)],
+            filename=filename_test1)
+
+    def test_message_too_long_file_with_other_errors(self):
+        self.section['filename_prefix'] = 'pre'
+        filename_test1 = '_filenamewhichistoolong'*15 + '.xyz'
+        max_filename_length = 260
+        msg = ("- Filename does not use the prefix 'pre'.\n"
+               '- Filename is too long ({} > {}).'
+               )
+        self.check_results(
+            self.uut,
+            [''],
+            [Result.from_values('FilenameBear',
+                                msg.format(len(filename_test1),
+                                           max_filename_length),
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file=filename_test1)],
+            filename=filename_test1)


Makman2:
Commit message:
I would slightly rephrase the shortlog: `Add filename length check`
And the body is wrong: If the length has less than 4096, it's actually okay. And nothing gets shortened there from coala side, the user has to do it.

bears/general/FilenameBear.py
```diff
@@ -98,6 +98,11 @@ def run(self, filename, file,
                 'Filename does not use the suffix {!r}.'.format(
                     filename_suffix))
 
+        if len(new_name) > 4096-len(extension):
```
Makman2: Have you thought about making this a setting instead of hardcoding the value? Also what about the 260 char limit of windows? Shall the default limit be maybe system dependent?

Makman2:
unack 58ddd11

bears/general/FilenameBear.py
damian1996: Additionally, you metioned about system dependent limit and I tried to check limits on Windows, Linux and OS X. I found that limit on Windows is really 260, limit on Linux is 4096 and also limit on OS X is different than previous metioned (probably 255). So do you think that better idea would be to check the current system or set default length as minimum of all systems?

Makman2: Probably, at least the default should may be variable. However for the sake of cross-compatibility this might be less good.

> OS X is different than previous metioned (probably 255)

Should be something Linux-like, as OSX is more or less BSD.

> I found that limit on Windows is really 260

And I believe it's possible to change the limit in Windows. 260 chars is for compatibility with older software (and sadly still default). Maybe we should detect the actual compatibility mode? Not sure, sounds like work 😅

damian1996: Should I consider that on Linux (and probably OSX) maximum length of filename is 255, however maximum path length is even 4096?

damian1996: Additionally, [here](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx#maxpath) is metioned that maximum path length is 260, however.. 
"The Windows API has many functions that also have Unicode versions to permit an extended-length path for a maximum total path length of 32,767 characters." 
Seems be that this is various in dependent of functions, however this can be hard to detection, so better choice should be take minimum, which is 260.

Makman2: > so better choice should be to take minimum, which is 260.

sounds good :)

bears/general/FilenameBear.py
```diff
@@ -26,6 +27,8 @@ class FilenameBear(LocalBear):
                                   }
 
     def run(self, filename, file,
+            max_len_path_unix: int=4096,
+            max_len_path_windows: int=260,
```
Makman2: Ah interesting, separating this for different systems. I'm just considering if it still can be made a single settings, I'm thinking of
```
max_path_length = 4096  # <-- affect all systems
max_path_length = {'windows': 100, 'unix': 100} # <-- different settings for different OSes, but this is probably too complicated from user perspective
```
However I think this can be a new issue.

Makman2: Just thinking, do you mean path length or filename length? Paths can be different on different environments.

Also:
Change name pattern to `max_path_length_...` instead of `max_len_path_...`

bears/general/FilenameBear.py
```diff
@@ -98,6 +101,13 @@ def run(self, filename, file,
                 'Filename does not use the suffix {!r}.'.format(
                     filename_suffix))
 
+        p = platform.system()
+        max_len = max_len_path_windows if p == 'Windows' else max_len_path_unix
+        if len(new_name) > max_len-len(extension):
+            messages.append(
+                'Filename length is more than {}, so it is too long.'.format(
```
Makman2: 1. you should be precise with path vs filename
2. incorporate the actual length, I would recommend
    ```
    Filename length ({}) is more than {}.
    ```
    or even better imo
    ```
    'Filename is too long ({} > {})'
    ```
    but up to you to phrase a nice message ;)
3. "so it is too long" is redundant ;)

Makman2: Oh maybe you should also incorporate the system the user is on, so he isn't confused when he sees "Wait, my filename is too long? But `filename_length_unix` is 1000", and because he's on Windows he oversees that the windows-default is used ;) (unlikely, but just a suggestion to provide more information in the message)

tests/general/FilenameBearTest.py
```diff
@@ -151,3 +152,25 @@ def test_file_prefix_suffix(self):
...
+        if platform.system() == 'Windows':
+            filename_testWin1 = '_filenamewhichistoolong'*15+'.xyz'
+            filename_testWin2 = '_filenamewhichistoolong'*10+'.xyz'
+            filename_testWin3 = '_validfilenamelength'*13
+            self.check_invalidity(
+                self.uut, [''], filename=filename_testWin1)
+            self.check_validity(
+                self.uut, [''], filename=filename_testWin2)
+            self.check_validity(
+                self.uut, [''], filename=filename_testWin3)
```
Makman2: could you use `check_results`? I want to see tests for the message :+1:

Makman2:
unack e9db5e7

bears/general/FilenameBear.py
damian1996: Should I check only filename length or filename and path length?

damian1996:
I think, that limit filename length should be 255 bytes on Linux, because 4096 is other limit (path length).

bears/general/FilenameBear.py
damian1996: I think, that this message is great :) 
> 'Filename is too long ({} > {})'

bears/general/FilenameBear.py
damian1996: What do you think about such situation:
When we have different limits on different systems, then when project is moved to other system, then this code wouldn't be useful. Because on first system limit was 260, but on second system is only 255. 
In my opinion setting the minimum of these limits as default should resolve it.

damian1996:
What do you think about such situation:
When we have different limits on different systems, then when project is moved to other system, then this code wouldn't be useful. Because on first system limit was 260, but on second system is only 255.
In my opinion setting the minimum of these limits as default should resolve it.

Makman2:
> In my opinion setting the minimum of these limits as default should resolve it.

agree :+1: :)

bears/general/FilenameBear.py
```diff
@@ -42,6 +43,9 @@ def run(self, filename, file,
             - ``pascal`` (``ThisIsPascalCase``)
             - ``snake`` (``this_is_snake_case``)
             - ``space`` (``This Is Space Case``)
+        :param max_filename_length:
+            Minimum of maximum filename length on Windows and maximum filename
+            length pn Unix-like systems. Default value is 260 characters.
```
Makman2: typo: pn -> on
\+
No need to mention the default value, coala automatically does that.

bears/general/FilenameBear.py
```diff
@@ -42,6 +43,9 @@ def run(self, filename, file,
             - ``pascal`` (``ThisIsPascalCase``)
             - ``snake`` (``this_is_snake_case``)
             - ``space`` (``This Is Space Case``)
+        :param max_filename_length:
+            Minimum of maximum filename length on Windows and maximum filename
```
Makman2: > Minimum of maximum

that sounds confusing :3 Either minimum or maximum (I tend to 2nd ;D)

bears/general/FilenameBear.py
```diff
@@ -54,7 +58,6 @@ def run(self, filename, file,
         """
         head, tail = os.path.split(filename)
         filename_without_extension, extension = os.path.splitext(tail)
-
```
Makman2: unnecessary change

bears/general/FilenameBear.py
```diff
@@ -98,14 +101,19 @@ def run(self, filename, file,
...
+                    max_filename_length))
+
         if ignore_uppercase_filenames and filename_without_extension.isupper():
             return
 
         if messages:
             diff = Diff(file, rename=os.path.join(head, new_name + extension))
             message = ('\n'.join('- ' + mes for mes in messages)
                        if len(messages) > 1 else messages[0])
-
```
Makman2: unnecessary change

bears/general/FilenameBear.py
```diff
@@ -98,14 +101,19 @@ def run(self, filename, file,
                 'Filename does not use the suffix {!r}.'.format(
                     filename_suffix))
 
+        if len(new_name+extension) > max_filename_length:
```
Makman2: I prefer spaces around operators, but optional :+1:

Makman2: Hm okay you are analyzing the new name generated. We should be careful not to confuse the user here, because he thinks "huh, but my file isn't called like that / the length of the name doesn't match up :O".

bears/general/FilenameBear.py
```diff
@@ -98,14 +101,19 @@ def run(self, filename, file,
...
+            messages.append(
+                'Filename is too long ({} > {}).'.format(
+                    len(new_name+extension),
```
Makman2: also here but optional: spaces around operator

bears/general/FilenameBear.py
damian1996: What do you think about adding whole name of file (I mean new_name + extension) in error message? :)

Makman2: still not sure...

Imo I would use the "old" filename for this check, I think even displaying `new_name + extension` would confuse the user. For now I would say we just use the old name, if users start to complain about this setting, we think about it again ;)

damian1996: I'm still not sure. You should imagine such scenario : 
If we'd check filename on the start, then we aren't catch too long filename after adding required suffix or prefix (this is possible).
So in my opinion checking this on new_name is better solution.

Makman2:
ack bd08e1b

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

PR #2173, opened 2017-12-16
SpaceConsistencyBear: Add blanklines_at_EOF option

khanchi97:
Add allow_trailing_blanklines option for whether to allow
trailing blanklines at EOF.

Closes https://github.com/coala/coala-bears/issues/1793

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/.ci/package.json.jinja2
+++ b/.ci/package.json.jinja2
@@ -3,4 +3,3 @@
   "version": "{{version}}",
   "dependencies": {{dependencies | jsonify(indent=4, sort_keys=True)}}
 }
-
--- a/bears/general/SpaceConsistencyBear.py
+++ b/bears/general/SpaceConsistencyBear.py
@@ -18,6 +18,7 @@ class SpaceConsistencyBear(LocalBear):
             file,
             use_spaces: bool,
             allow_trailing_whitespace: bool=False,
+            allow_trailing_blanklines: bool=False,
             indent_size: int=SpacingHelper.DEFAULT_TAB_WIDTH,
             enforce_newline_at_EOF: bool=True):
         '''
@@ -29,6 +30,8 @@ class SpaceConsistencyBear(LocalBear):
             True if spaces are to be used instead of tabs.
         :param allow_trailing_whitespace:
             Whether to allow trailing whitespace or not.
+        :param allow_trailing_blanklines:
+            Whether to allow trailing blanklines at EOF or not.
         :param indent_size:
             Number of spaces per indentation level.
         :param enforce_newline_at_EOF:
@@ -38,7 +41,34 @@ class SpaceConsistencyBear(LocalBear):
         result_texts = []
         additional_info_texts = []
 
-        for line_number, line in enumerate(file, start=1):
+        if not allow_trailing_blanklines:
+            line_nr_start = next(
+                    (i
+                     for i, line in reversed(list(enumerate(file, start=0)))
+                     if line.strip() != ''), -1) + 1
+            if line_nr_start < len(file):
+                additional_info_text = (
+                    'Your source code contains trailing blanklines at EOF.'
+                    'Those usually have no meaning. Please consider'
+                    'removing them.')
+                diff = Diff(file)
+                diff.delete_lines(line_nr_start + 1, len(file))
+                yield Result.from_values(
+                    self,
+                    'Trailing blanklines found.',
+                    diffs={filename: diff},
+                    file=filename,
+                    line=line_nr_start + 1,
+                    column=1,
+                    end_line=len(file),
+                    end_column=len(file[-1]),
+                    additional_info=additional_info_text)
+
+        else :
+            line_nr_start = len(file)
+
+        for line_number, line in enumerate(file[:line_nr_start],
+                                           start=1):
             replacement = line
 
             if enforce_newline_at_EOF:
--- a/tests/general/SpaceConsistencyBearTest.py
+++ b/tests/general/SpaceConsistencyBearTest.py
@@ -3,6 +3,8 @@ from queue import Queue
 from bears.general.SpaceConsistencyBear import (
     SpaceConsistencyBear, SpacingHelper)
 from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
 
@@ -44,15 +46,18 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
         self.section.append(Setting('use_spaces', 'false'))
         self.section.append(Setting('allow_trailing_whitespace', 'true'))
         self.section.append(Setting('enforce_newline_at_EOF', 'false'))
+        self.section.append(Setting('allow_trailing_blanklines', 'false'))
 
         self.check_invalidity(self.uut, ['    t'])
         self.check_validity(self.uut, ['t \n'])
         self.check_validity(self.uut, ['\tt\n'])
+        self.check_validity(self.uut, [])
 
     def test_enforce_newline_at_eof(self):
         self.section.append(Setting('use_spaces', 'true'))
         self.section.append(Setting('allow_trailing_whitespace', 'true'))
         self.section.append(Setting('enforce_newline_at_EOF', 'true'))
+        self.section.append(Setting('allow_trailing_blanklines', 'true'))
 
         self.check_validity(self.uut,
                             ['hello world  \n'],
@@ -60,7 +65,9 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
         self.check_validity(self.uut,
                             ['def somecode():\n',
                              "    print('funny')\n",
-                             "    print('funny end.')\n"],
+                             "    print('funny end.')\n",
+                             '  \n',
+                             '\n'],
                             force_linebreaks=False)
         self.check_invalidity(self.uut,
                               [' no hello world'],
@@ -70,3 +77,35 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
                                "    print('funny')\n",
                                "    print('the result is not funny...')"],
                               force_linebreaks=False)
+
+    def test_trailing_blanklines(self):
+
+        self.maxDiff = None
+        self.section['use_spaces'] = 'true'
+        self.section['allow_trailing_whitespace'] = 'false'
+        self.section['enforce_newline_at_EOF'] = 'true'
+        self.section['allow_trailing_blanklines'] = 'false'
+        self.check_results(self.uut,
+                           ['  \n'],
+                           [Result.from_values('SpaceConsistencyBear',
+                                               'Trailing blanklines found.',
+                                               file='xyz.py',
+                                               line=1,
+                                               column=1,
+                                               end_line=1,
+                                               end_column=3)],
+                           filename = 'xyz.py')
+#        self.check_results(self.uut,
+#                           ['  \n',
+#                            '     \n',
+#                            '\n'],
+#                           [Result.from_values('SpaceConsistencyBear',
+#                                               'Trailing blanklines found.',
+#                                               file='xyz.py',
+#                                               line=1,
+#                                               column=1,
+#                                               end_line=3,
+#                                               end_column=1,
+#                                               severity=RESULT_SEVERITY.NORMAL
+#                                               )],
+#                           filename='xyz.py')


bears/general/SpaceConsistencyBear.py
```diff
@@ -38,48 +41,84 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
+        def get_blanklines_nr_start():
+
```
Makman2: though not violating PEP8, I usually prefer no blanklines between a `def` and code, this kinda interrupts reading flow imo. Optional though if you see it differently ;)

bears/general/SpaceConsistencyBear.py
```diff
@@ -29,6 +30,8 @@ def run(self,
                                           of tabs.
         :param allow_trailing_whitespace: Whether to allow trailing whitespace
                                           or not.
+        :param allow_trailing_blanklines: Whether to allow trailing blanklines
+                                          at the EOF or not.
```
Makman2: at ~~the~~ EOF

khanchi97: okay

Makman2:
The code looks way too complicated, I have the feeling you're touching too much old code. Is it possible to perform the check at the beginning, and then slice the lines in the file with `[x:y]` in `enumerate` in the `for`-loop?
So two cases:
1. don't allow trailing blanklines:
    analyze how many trailing blanklines are there, and suggest a patch. Modify `y` to make following analysis skip those lines.
2. allow trailing blanklines:
    Set `y` to end of file, and skip this analysis.

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,52 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
-        for line_number, line in enumerate(file, start=1):
+        def get_blanklines_nr_start():
+            line_nr_start = False
+            if not allow_trailing_blanklines:
```
ishanSrt: I think this should be `if allow_trailing_blanklines:` because only then you will have to check `if replacement.strip() == ''` otherwise `line_nr_start` [which I am guessing to be the *start from where blank lines or in case of trailing spaces allowed blank lines (with spaces) start]

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,52 @@ def run(self,
...
+                rev_enumerated_zip_obj = zip(reversed(range(1, len(file) + 1)),
+                                             reversed(file))
+                rev_enumerated_tuple = tuple(rev_enumerated_zip_obj)
```
ishanSrt: kind of defeats the purpose of having a `zip/generator` like object which is highly efficient than a `list like` object

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,52 @@ def run(self,
...
+                for line_number, line in rev_enumerated_tuple:
+                    replacement = line
+                    if replacement.strip() == '':
+                        line_nr_start = line_number
+                    else:
+                        break
+
+            return line_nr_start
+
+        if allow_trailing_blanklines:
```
ishanSrt: I think there should be a `not` in here. Please correct me if I am wrong

bears/general/SpaceConsistencyBear.py
ishanSrt: I stand a little bit corrected, if you `if` condition is true `line_nr_start` would be just the last line or False (in case there is no new line at EOF), if your `if` condition is incorrect then your `line_nr_start` would again be just the last line or False [in case there is no new line at EOF (even with spaces)] so don't quite get what you are trying to do over here

ishanSrt: note there are 5 cases in which the file can end in case of trailing spaces are allowed

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,52 @@ def run(self,
...
+                    'Line contains following spacing inconsistencies:'
+                    + inconsistencies,
+                    diffs={filename: diff},
+                    file=filename,
+                    line=end_line + 1,
+                    additional_info='\n\n'.join(additional_info_texts))
+                result_texts = []
+                additional_info_texts = []
+
+
```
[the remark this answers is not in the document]
ishanSrt: this is not the end of a class so only 1 newline is expected here

bears/general/SpaceConsistencyBear.py
khanchi97: No.If trailing blanklines are allowed than then we don't to do any analysis for checking blanklines.

khanchi97:
@ishanSrt Please don't review until I make required changes.It's not completed yet.I have to do some minor changes in bear and test as well.

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,48 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
-        for line_number, line in enumerate(file, start=1):
+        def get_blanklines_nr_start():
+            line_nr_start = False
+            rev_enumerated_zip_obj = zip(reversed(range(1, len(file) + 1)),
+                                         reversed(file))
+            rev_enumerated_tuple = tuple(rev_enumerated_zip_obj)
```
ishanSrt: Explain the reason for converting it into a tuple. You are losing the efficiency of a zip/generator like object by converting it into a list like object

khanchi97:
>Then also your end_line_of_file is set to be the last line containing characters other that spaces.

No,it will set to the last non-blankline as `allow_trailing_blanklines` is not allowed.(acc. to you) ;)

ishanSrt:
@khanchi97 
>allow_trailing_blanklines is not allowed

as in `allow_trailing_whitespace` and `allow_trailing_blankline` are both `False` and you have given a file which 
>Let us suppose there are lines in the end with only spaces.

and since to get val of `line_nr_start` you are stripping spaces and then checking, I'm afraid that 
>No,it will set to the last non-blankline

this may not hold true

bears/general/SpaceConsistencyBear.py
khanchi97: because zip return an iterator, so after zip return an iterator of tuples of `line and line no.` then to iterate over that iterator in `for loop` 👇 I converted it to a tuple.

khanchi97: Can you a little bit explain how I'm losing the efficiency of zip by converting it to a tuple?

ishanSrt: I think you can iterate normally in a for loop over an iterator, no need to convert it to tuple. Plus using iterators/generator like objects is highly efficient

ishanSrt: >Can you a little bit explain how I'm losing the efficiency of zip by converting it to a tuple?

 list comprehension creates the entire list in memory first while the generator expression will create the items on the fly, so you are able to use it for very large (and also infinite!) sequences.

ishanSrt: so if you only have to loop once then generator like objects are considered far more efficient and faster than list like objects. I think you can read the entire explanation on stackoverflow. there are many questions regarding this

khanchi97: I think I've tried this when I was making changes but that didn't work.
btw, let me try this again.May be I was making some mistake ;)

khanchi97:
oh, it will be set to `first_blankline_in_file` i.e `last_non-blankline +1`

bears/general/SpaceConsistencyBear.py
khanchi97: it working 👍 let see it won't error in CI

khanchi97: thanks for your explanation :) @ishanSrt

.ci/package.json.jinja2
```diff
@@ -3,4 +3,3 @@
   "version": "{{version}}",
   "dependencies": {{dependencies | jsonify(indent=4, sort_keys=True)}}
 }
-
```
newbazz: Why this change?

bears/general/SpaceConsistencyBear.py
```diff
@@ -29,6 +30,8 @@ def run(self,
                                           of tabs.
         :param allow_trailing_whitespace: Whether to allow trailing whitespace
                                           or not.
+        :param allow_trailing_blanklines: Whether to allow trailing blanklines
+                                          at EOF or not.
```
newbazz: Btw this is not the correct docstring indentation ;P

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
-        for line_number, line in enumerate(file, start=1):
+        def get_blanklines_nr_start():
+            line_nr_start = False
```
newbazz: variable names are not good ;P and can you please explain what exactly are you doing.

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
...
+            rev_enumerated_zip_obj = zip(reversed(range(1, len(file) + 1)),
+                                         reversed(file))
+            for line_number, line in rev_enumerated_zip_obj:
+                replacement = line
+                if replacement.strip() == '':
+                    line_nr_start = line_number
+                else:
+                    break
```
newbazz: both `if` and `else` can be written inline.

.ci/package.json.jinja2
khanchi97: this patch is applied by CI using this same bear.
I think it is applied because of the option that I've implemented i.e `allow_trailing_blanklines` which has default value `false`
@Makman2 ^ am i right?

bears/general/SpaceConsistencyBear.py
khanchi97: Well for now, it is 😉

bears/general/SpaceConsistencyBear.py
khanchi97: what do you suggest?

bears/general/SpaceConsistencyBear.py
khanchi97: I am not confirm we should do this or not?

bears/general/SpaceConsistencyBear.py
ishanSrt: what constitutes a good variable name?

ishanSrt:
@khanchi97 I am not sure what you meant by https://github.com/coala/coala-bears/pull/2173#issuecomment-355223942 
Can you explain again why https://github.com/coala/coala-bears/pull/2173#issuecomment-355220629 is justified? This won't set to the last non blank line. As is said if the last lines only have spaces, it will completely ignore that and set it to the last line containing characters other than spaces. so the user won't even be prompted that he has trailing spaces in his file. Also since you are not evaluating that part of code, those lines will be left untouched and since the option of allow trailing spaces is false, and the final code will still contain trailing spaces, this is wrong output and you need to fix this. Again correct me if I am wrong. 🙂

bears/general/SpaceConsistencyBear.py
khanchi97: the name which is well suited for what it does and not so long and from it's name we can have a idea that what value it is holding(I guess 😉 )

khanchi97:
@ishanSrt 
Is this type of file what you are asking about ?
```
file="int main(){\n"
      "a = 5;\n"
      "printf("%d",a);\n"
      "}\n"
      "     "
```

ishanSrt:
more like
```
file="int main(){\n"
      "a = 5;\n"
      "printf("%d",a);\n"
      "}\n"
      "     \n"
      "\n"
```

khanchi97:
okay so now as I have reversed all the file before start to check trailing blanklines.So now we will be checking the file in this order:
```
file=6.   "\n"
      5.    "     \n"
      4.    "}\n"
      3.    "printf("%d",a);\n"
      2.    "a = 5;\n"
      1.    "int main(){\n"
```
So now after exit from that `for loop` what do you think what would be the value of `line_nr_start`.
I think it should be 5.Now try to iterate this file into that `for loop` and tell me where do you have doubt?




      "\n"

ishanSrt:
@khanchi97 Ok consider this file instead:
```
file="int main(){\n"
      "a = 5;\n"
      "printf("%d",a);\n"
      "}\n"
      "     \n"
      "   \n"
      "         \n"
      "\n"
```
so the reversed file is:
```
8      "\n"
7      "         \n"
6      "   \n"
5      "     \n"
4      "}\n"
3      "printf("%d",a);\n"
2      "a = 5;\n"
1      "int main(){\n"
```
now 
I think `line_nr_start` will return 5, so you are skipping lines 6 and 7

khanchi97:
No.We are not skipping any line and it should give 5, shouldn't it?

ishanSrt:
@khanchi97 so `end_line_of_file` would be set to **4** and your loop on line 84 will completely ignore lines 5,6, and 7 and thus the user won't be notified that there are trailing spaces

khanchi97:
What do you think line 5,6,7 are not blanklines?

ishanSrt:
won't the lines 5,6,7 would be left as it is in the file with no change? while the user selected the option to not allow trailing spaces and he is left with sort of trailing spaces in the final output file

khanchi97:
No. the lines 5,6,7 would be deleted if `allow_trailing_blanklines` is FALSE and `allow_trailing_whitespace` is also FALSE.
And if `allow_traling_blanklines` is TRUE and `allow_trailing_whitespace` is FALSE then from lines 5,6,7 spaces will be removed.Fwew I think now you got it :)

khanchi97:
And if you are not getting it then please once try to run it yourself and see what is the result :)

ishanSrt:
@khanchi97 I stand a bit corrected. So the lines won't be added as it is as you have considered lines with spaces in the end to be `trailing newlines`. So the user isn't getting notified that he has trailing spaces. 

So, finally what should be done in this case? Shouldn't the user be notified that he has both trailing whitespaces and trailing newlines in the code rather than just removing those lines under trailing newlines. Not sure on this.

.ci/package.json.jinja2
newbazz: I dont think so, coz CI needs to be configured in a way that no new file is being edited or made by it, and coala might have confgured CI correctly ;P. Still i am not sure.

bears/general/SpaceConsistencyBear.py
newbazz: does it violate any bear?

newbazz: as far i remember when theres only one line in if statement it is good to inline it ;)

newbazz:
ok thats what you have done ;P my bad

khanchi97:
@ishan I think we should consider a line containing only a newline character "\n" and a line with whitespace also "   \n" to be a `blankline` and prompt the user that there are some trailing blanklines in your source code.

khanchi97:
@Makman2 review needed 😄

.ci/package.json.jinja2
Makman2: Yes it does immediately, due to the default value. Imo this is okay :+1: It's probably worth though to mention this inside the commit body :+1:

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
...
+
+            return line_nr_start
```
Makman2: 1. You don't need this code to reside inside a function as you only use it once.
2. This can be simplified to a single expression:
    ```
    line_nr_start = next(i for i, line in reversed(list(enumerate(file))) if line.strip() != '')
    ```

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
...
+
+        if allow_trailing_blanklines:
+            end_line_of_file = len(file)
```
Makman2: what about `blanklines_start_line`? I think that would be a good name :)
(yeah I finally got a name for it :D)

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
...
+
+        else:
+            blanklines_nr_start = get_blanklines_nr_start()
+            end_line_of_file = len(file)
+            if blanklines_nr_start:
```
Makman2: Instead of letting the `get_blanklines_nr_start` function return `None` if no blanklines found, rather compare it against `len(file)`, makes things simpler :+1:

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
...
+                    break
+
+            return line_nr_start
+
+        if allow_trailing_blanklines:
+            end_line_of_file = len(file)
+
+        else:
+            blanklines_nr_start = get_blanklines_nr_start()
+            end_line_of_file = len(file)
```
Makman2: seems this happens in both branches of the `if`

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,47 @@ def run(self,
...
+                                          for string in result_texts)
+                yield Result.from_values(
+                    self,
+                    'Line contains following spacing inconsistencies:'
+                    + inconsistencies,
+                    diffs={filename: diff},
+                    file=filename,
+                    additional_info='\n\n'.join(additional_info_texts))
+                result_texts = []
+                additional_info_texts = []
```
Makman2: See above, then you don't have to do this ;)

Makman2:
unack 01f2a9dde9b5e5f6e7dc27825057f78a16c24a9c

khanchi97:
@Makman2 thank you very much for your review 😄 
will do the required changes by tomorrow.
I was a bit busy in organizing #LinuxWeek for our juniors.
(going to sleep 😛 )

khanchi97:
I have made the required changes :)

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,36 @@ def run(self,
...
+        else:
+            blanklines_start_line = line_nr_start + 1 \
+                    if line_nr_start < len(file) else False
+
+        if not allow_trailing_blanklines:
+            if blanklines_start_line:
+                additional_info_text = (
+                        'Your source code contains trailing blanklines at EOF.'
+                        'Those usually have no meaning. Please consider'
+                        'removing them.')
```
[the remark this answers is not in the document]
Makman2: Ah nvm that's the *additional*-info-text

Makman2: Indentation: 4 spaces relative to statement

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,36 @@ def run(self,
...
-        for line_number, line in enumerate(file, start=1):
+        line_nr_start = next((i for i, line in reversed(list(
+            enumerate(file, start=1))) if line.strip() != ''), 0)
+
+        if len(file) == 0:
+            blanklines_start_line = False
+
+        else:
+            blanklines_start_line = line_nr_start + 1 \
+                    if line_nr_start < len(file) else False
```
Makman2: I have the feeling you are unnecessary trying to use special values (--> `False`). 

You can easily check if you have to yield a result if `line_nr_start < len(file)`.
The whole stuff can be refactored to roughly this:
```python
if not allow_trailing_blanklines:
    line_nr_start = next((i for i, line in reversed(list(
        enumerate(file, start=1))) if line.strip() != ''), 0)
    # TODO: Maybe more performance: Don't increment from start=1, but start=0 and
    #       the next-default value = -1, and increment in the end: next(..., -1) + 1

    if line_nr_start < len(file):
        # Note: You have to provide line_nr_start + 1 to refer to the right lines.
        # line_nr_start is right now designed to give the starting line referring
        # to zero-based indices.
        yield Result(...)
else:
    line_nr_start = len(file)

for line_number, line in enumerate(file[:line_nr_start])
```

Makman2: Note also how I've put the relevant analysis inside the if, to save performance in case someone wants not to take care of blanklines

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,36 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
-        for line_number, line in enumerate(file, start=1):
+        line_nr_start = next((i for i, line in reversed(list(
+            enumerate(file, start=1))) if line.strip() != ''), 0)
```
Makman2: Optional:
Is it maybe possible to wrap the statement somehow so it's more readable?

Makman2: Something in the direction of
```python
line_nr_start = next(
    (i
     for i, line in reversed(list(enumerate(file, start=1)))
     if line.strip() != ''), 0)
```

tests/general/SpaceConsistencyBearTest.py
```diff
@@ -70,3 +75,22 @@ def test_enforce_newline_at_eof(self):
...
+                               "  print('Is it funny?')\n",
+                               "  print('Yeah, funny')\n",
+                               '  \n',
+                               '\n'],
+                              force_linebreaks=False)
+        self.check_validity(self.uut,
+                            ['def funny_code():\n',
+                             "  print('Is it funny?')\n",
+                             "  print('Yeah, funny')\n"],
+                            force_linebreaks=False)
```
Makman2: I want to see tests for
- an empty file
- a file with only a single blank line
- a file with only multiple blank lines
- a file with leading and trailing blank lines

Makman2: And please use `check_results` to assert for the bear results :+1:

Makman2:
unack 2e2eaaceb73a96fae24e89919ca02bd828729109

bears/general/SpaceConsistencyBear.py
khanchi97: yeah, I wanted to do the same.But I was finding it difficult to cover the case when file is empty.That's why I used that (--> False), but now it will work as we are using `start=0` and next default value `-1`.

Makman2:
please fix the gitmate issues ;)

jayvdb:
Conflicts to be resolved.

khanchi97:
@Makman2 @jayvdb Can I ask for adding someone as collaborator to my PRs(which are almost completed) to do remaining work. ( I am busy in some other work :) )

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,7 +41,34 @@ def run(self,
...
+                    'Trailing blanklines found.',
+                    diffs={filename: diff},
+                    file=filename,
+                    line=line_nr_start + 1,
+                    column=1,
+                    end_line=len(file),
+                    end_column=len(file[-1]),
+                    additional_info=additional_info_text)
+
+        else :
```
sangamcse: Fix this whitespace issue @khanchi97. Run `coala` over your local repository before pushing it to remote 😉

bears/general/SpaceConsistencyBear.py
```diff
@@ -18,6 +18,7 @@ def run(self,
             file,
             use_spaces: bool,
             allow_trailing_whitespace: bool=False,
+            allow_trailing_blanklines: bool=False,
```
sangamcse: Alphabetical order 😉

tests/general/SpaceConsistencyBearTest.py
```diff
@@ -44,23 +46,28 @@ def test_data_sets_tabs(self):
         self.section.append(Setting('use_spaces', 'false'))
         self.section.append(Setting('allow_trailing_whitespace', 'true'))
         self.section.append(Setting('enforce_newline_at_EOF', 'false'))
+        self.section.append(Setting('allow_trailing_blanklines', 'false'))
```
sangamcse: This line is not required.

tests/general/SpaceConsistencyBearTest.py
```diff
@@ -70,3 +77,35 @@ def test_enforce_newline_at_eof(self):
...
                               force_linebreaks=False)
+
+    def test_trailing_blanklines(self):
+
+        self.maxDiff = None
+        self.section['use_spaces'] = 'true'
+        self.section['allow_trailing_whitespace'] = 'false'
+        self.section['enforce_newline_at_EOF'] = 'true'
+        self.section['allow_trailing_blanklines'] = 'false'
+        self.check_results(self.uut,
```
sangamcse: What do you say about this?
```
    def test_trailing_blanklines(self):
        self.maxDiff = None
        self.section.append(Setting('use_spaces', 'true'))
        self.section.append(Setting('allow_trailing_whitespace', 'false'))
        self.section.append(Setting('enforce_newline_at_EOF', 'true'))
        self.section.append(Setting('allow_trailing_blanklines', 'false'))
        self.check_validity(self.uut,
                            ['sangam\n'],
                            force_linebreaks=False)
        self.check_invalidity(self.uut,
                              [' \n'],
                              force_linebreaks=False)
```

PR #2176, opened 2017-12-17
Makman2/buildfix

Makman2:


Makman2:
Or as pylint ~1.6 did work too, maybe I should do `pylint>=1.6,<1.8`?

jayvdb:
I think you'll need to update the `Requirement` in each bear to be the same.

jayvdb:
ack https://github.com/coala/coala-bears/pull/2176/commits/f1258c7e07dc01cdb035df489b5a3f35ad6331d5 https://github.com/coala/coala-bears/pull/2176/commits/090edb323041fb8da35ae11027090eed1b4b26d2

bears/python/PyLintBear.py
```diff
@@ -23,7 +25,8 @@ class PyLintBear:
     separately.
     """
     LANGUAGES = {'Python', 'Python 2', 'Python 3'}
-    REQUIREMENTS = {PipRequirement('pylint', '1.6')}
+    REQUIREMENTS = {AnyOneOfRequirements([PipRequirement('pylint', '1.6'),
```
jayvdb: no the voodoo parser doesnt work like that.  I am not sure what you intend to do.  just use '1.6.0' if you are trying to prevent 1.7...?  and then `pylint~=1.6.0` in the `bear-requirements.txt` (run https://github.com/coala/coala-bears/blob/master/.ci/generate_bear_requirements.py to generate this file)

jayvdb:
unack f1258c7 090edb3

bears/python/PyLintBear.py
Makman2: alright :+1:

Makman2: so we just have to prevent 1.8, 1.7.5 is working :+1:

jayvdb:
ack https://github.com/coala/coala-bears/pull/2176/commits/cc1ccd743bc032630df42b75e27fd2e931130841

jayvdb:
ack f1258c7e07dc0

PR #2203, opened 2017-12-29
SpaceConsistencyBear.py: Add aspects

### LINKED ISSUE #2181: Aspectize SpaceConsistencyBear
Make SpaceConsistencyBear could be used through coala's aspects. The bear's should have the following aspects:
- `Root.Formatting.Spacing.Indentation` - the taste mapped to `use_spaces` and `indent_size`
- `Root.Formatting.Spacing.BlankLine.NewlineAtEOF` - the aspect enabled (mapped) `enforce_newline_at_EOF`
- `Root.Formatting.Spacing.TrailingSpace` - the aspect enabled (mapped) `allow_trailing_whitespace`

You can read a guide to do this in https://api.coala.io/en/latest/Developers/Writing_Native_Bears.html#aspect-bear

Blocked by https://github.com/coala/coala/issues/4982 because `Indentation` aspects is currently non-existant.

biscuitsnake:
Closes #2181

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/SpaceConsistencyBear.py
+++ b/bears/general/SpaceConsistencyBear.py
@@ -3,9 +3,25 @@ from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.Diff import Diff
 from coalib.results.Result import Result
+from coalib.bearlib.aspects import map_setting_to_aspect
+from coalib.bearlib.aspects.Formatting import (
+    Formatting,
+    TrailingSpace,
+    Indentation,
+    NewlineAtEOF,
+)
 
 
-class SpaceConsistencyBear(LocalBear):
+class SpaceConsistencyBear(
+        LocalBear,
+        aspects={
+            'detect': [
+                Indentation,
+                NewlineAtEOF,
+                TrailingSpace,
+            ]},
+        languages=['Python']):
+
     LANGUAGES = {'All'}
     AUTHORS = {'The coala developers'}
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
@@ -13,10 +29,15 @@ class SpaceConsistencyBear(LocalBear):
     CAN_FIX = {'Formatting'}
 
     @deprecate_settings(indent_size='tab_width')
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF=NewlineAtEOF.newline_at_EOF,
+        allow_trailing_whitespace=TrailingSpace.allow_trailing_spaces,
+        indent_size=Indentation.indent_size,
+        use_spaces=Indentation.indent_type == 'space',)
     def run(self,
             filename,
             file,
-            use_spaces: bool,
+            use_spaces: bool=True,
             allow_trailing_whitespace: bool=False,
             indent_size: int=SpacingHelper.DEFAULT_TAB_WIDTH,
             enforce_newline_at_EOF: bool=True):
@@ -37,6 +58,12 @@ class SpaceConsistencyBear(LocalBear):
         spacing_helper = SpacingHelper(indent_size)
         result_texts = []
         additional_info_texts = []
+
+        aspect = (self.section.aspects and (
+                  self.section.aspects.get('Indentation').indent_type
+                  == 'space'))
+        if aspect:
+            use_spaces = True
 
         for line_number, line in enumerate(file, start=1):
             replacement = line
@@ -89,6 +116,7 @@ class SpaceConsistencyBear(LocalBear):
                     diffs={filename: diff},
                     file=filename,
                     line=line_number,
+                    aspect=Formatting('Python'),
                     additional_info='\n\n'.join(additional_info_texts))
                 result_texts = []
                 additional_info_texts = []
--- a/tests/general/SpaceConsistencyBearTest.py
+++ b/tests/general/SpaceConsistencyBearTest.py
@@ -1,7 +1,9 @@
 from queue import Queue
 
+import coalib.bearlib.aspects as coala_aspects
 from bears.general.SpaceConsistencyBear import (
     SpaceConsistencyBear, SpacingHelper)
+from coalib.bearlib.aspects.collections import AspectList
 from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
@@ -11,15 +13,8 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
 
     def setUp(self):
         self.section = Section('test section')
+        self.section.language = 'Python'
         self.uut = SpaceConsistencyBear(self.section, Queue())
-
-    def test_needed_settings(self):
-        self.section.append(Setting('use_spaces', 'true'))
-
-        needed_settings = self.uut.get_non_optional_settings()
-        self.assertEqual(len(needed_settings),
-                         1 + len(SpacingHelper.get_non_optional_settings()))
-        self.assertIn('use_spaces', needed_settings)
 
     def test_defaults(self):
         # use_spaces is no default, need to set it explicitly.
@@ -70,3 +65,42 @@ class SpaceConsistencyBearTest(LocalBearTestHelper):
                                "    print('funny')\n",
                                "    print('the result is not funny...')"],
                               force_linebreaks=False)
+
+    def test_aspects_space(self):
+        self.section.aspects = AspectList([
+            coala_aspects['Indentation']('Python', indent_type='space')
+        ])
+
+        self.check_validity(self.uut, ['    t'])
+        self.check_invalidity(self.uut, ['\tt\n'])
+
+    def test_aspects_tabs(self):
+        self.section.aspects = AspectList([
+            coala_aspects['Indentation']('Python', indent_type='tab')
+        ])
+
+        self.check_invalidity(self.uut, ['   t'])
+        self.check_validity(self.uut, ['\tt\n'])
+
+    def test_aspects_tab_size(self):
+        self.section.aspects = AspectList([
+            coala_aspects['Indentation']('Python',
+                                         indent_size=2, indent_type='space')
+        ])
+
+        self.check_validity(self.uut,
+                            ['def validghost():\n',
+                             "  print('boo!')\n"],
+                            force_linebreaks=False)
+        self.check_invalidity(self.uut,
+                              ['def invalidghost():\n',
+                               "   print('boo.')\n"],
+                              force_linebreaks=False)
+
+    def test_aspects_nondefault(self):
+        self.section.aspects = AspectList([
+            coala_aspects['NewlineAtEOF']('Python', newline_at_EOF='false')
+        ])
+
+        self.check_validity(self.uut, ['t\n'])
+        self.check_validity(self.uut, ['t'])


bears/general/SpaceConsistencyBear.py
```diff
@@ -1,38 +1,56 @@
...
+            NewlineatEOF,
+            TrailingSpace
+        ]
+    },
+    languages=['Python']):
+    
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF = NewlineAtEOF,
+        allow_trailing_spaces = TrailingSpace,
+        indent_size = Indentation,
```
adhikasp: Use the taste, `Indentation.indent_size`

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,38 +1,56 @@
...
+        use_spaces = Indentation
```
adhikasp: Like above `Indentation.indent_type`. But this line should be dropped, see my comment below.

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,38 +1,56 @@
...
+    aspect={
+        'detect': [
+            Indentation, 
+            NewlineatEOF,
+            TrailingSpace
+        ]
+    },
+    languages=['Python']):
+    
+    @map_setting_to_aspect(
```
adhikasp: Use this decorator above the `run(...)` function

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,38 +1,56 @@
...
-
-    @deprecate_settings(indent_size='tab_width')
     def run(self,
             filename,
             file,
-            use_spaces: bool,
-            allow_trailing_whitespace: bool=False,
-            indent_size: int=SpacingHelper.DEFAULT_TAB_WIDTH,
-            enforce_newline_at_EOF: bool=True):
+            enforce_newline_at_EOF: bool=True,
```
adhikasp: You shouldn't change any of the parameter names here. The point of `@map_setting_to_aspect` is to "convert" the value from aspect/taste and insert it to coala's bear setting here. Also no need to add the default value `True`.

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,38 +1,56 @@
...
+            allow_trailing_spaces: bool=False,
```
adhikasp: Same as above, don't rename.

bears/general/SpaceConsistencyBear.py
```diff
@@ -63,8 +81,8 @@ def run(self,
                         'Your source code contains trailing whitespaces. '
                         'Those usually have no meaning. Please consider '
                         'removing them.')
-
-            if use_spaces:
+            
+            if indent_type == 'space':
```
adhikasp: If you has change above correctly, this should be back to `if use_spaces:`

adhikasp (changes requested):

biscuitsnake:
@adhikasp tests can't import Indentation, I've spent all evening trying to figure out why but I can't so please have a look. Hopefully it isn't an error with the Indentation code again.

newbazz (commented):
Run the PEP8 bear on the changed files ;)

newbazz (commented):
image
theres no class named `Indentation`.

newbazz (changes requested):
run coala before making a PR ;P

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,32 +1,49 @@
...
+        enforce_newline_at_EOF=NewlineAtEOF,
+        allow_trailing_whitespace=TrailingSpace,
+        indent_size=Indentation.indent_size)
     def run(self,
             filename,
             file,
             use_spaces: bool,
             allow_trailing_whitespace: bool=False,
-            indent_size: int=SpacingHelper.DEFAULT_TAB_WIDTH,
+            indent_size: int=4,
```
adhikasp: Dont change this

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,32 +1,49 @@
...
             enforce_newline_at_EOF: bool=True):
         '''
         Check and correct spacing for all textual data. This includes usage of
         tabs vs. spaces, trailing whitespace and (missing) newlines before
         the end of the file.
 
-        :param use_spaces:                True if spaces are to be used instead
-                                          of tabs.
+        :param use_spaces:                True if using spaced indents.
```
adhikasp: And this

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,6 +55,10 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
+        if self.section.aspects:
+            indent_type = (Indentation in self.section.aspects and
```
adhikasp: Should be `use_spaces` 😜 

Also this is missing the case if indent is `tab` and then `use_spaces` should became `False`

bears/general/SpaceConsistencyBear.py
```diff
@@ -89,6 +110,7 @@ def run(self,
                     diffs={filename: diff},
                     file=filename,
                     line=line_number,
-                    additional_info='\n\n'.join(additional_info_texts))
+                    additional_info='\n\n'.join(additional_info_texts),
+                    aspect=Formatting('py'))
```
adhikasp: Put `aspect` above the `additional_info` so you don't need to modify unnecessary lines. It make `git blame/bisect` and history much cleaner.

adhikasp (changes requested):

bears/general/SpaceConsistencyBear.py
biscuitsnake: how would the aspect override it then

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,22 +1,40 @@
...
-    @deprecate_settings(indent_size='tab_width')
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF=NewlineAtEOF,
+        allow_trailing_whitespace=TrailingSpace,
+        indent_size=Indentation.indent_size)
     def run(self,
             filename,
             file,
-            use_spaces: bool,
+            use_spaces: bool=,
```
newbazz: Initialise it to false.

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,6 +56,9 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
+        if self.section.aspects and Indentation in self.section.aspects:
+            use_spaces = Indentation.indent_type == 'space'
```
newbazz: you made `use_spaces` to be a bool right?

biscuitsnake: yes, what's the issue

newbazz: so you are assigning it a string

biscuitsnake: no it's True if Indentation.indent_type == 'space' and false otherwise

biscuitsnake: but perhaps writing this line as 
```use_spaces = (Indentation.indent_type == 'space')```
would make things clearer

Comment (newbazz) [reply bears/general/SpaceConsistencyBear.py]: ohh I didnt see the `==` before 'space' my bad ;P

File: bears/general/SpaceConsistencyBear.py [hunk 13]
```diff
@@ -1,22 +1,40 @@
...
-    @deprecate_settings(indent_size='tab_width')
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF=NewlineAtEOF,
+        allow_trailing_whitespace=TrailingSpace,
+        indent_size=Indentation.indent_size)
     def run(self,
             filename,
             file,
-            use_spaces: bool,
+            use_spaces: bool=True,
```
Mixih: If you set a default here, you'll have to adjust `test_needed_settings` to reflect the new setting count.

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,22 +1,40 @@
...
-    @deprecate_settings(indent_size='tab_width')
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF=NewlineAtEOF,
+        allow_trailing_whitespace=TrailingSpace,
+        indent_size=Indentation.indent_size)
     def run(self,
             filename,
             file,
-            use_spaces: bool,
+            use_spaces: bool=False,
```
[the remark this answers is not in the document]
Makman2: (btw I would make this true by default)

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # NOTE: This file is parsed by .ci/generate_bear_requirements.py
 # Use >= for development versions so that source builds always work
-coala>=0.12.0.dev20170828055500
+coala>=0.12.0.dev20180101025653
```
Makman2: necessary?

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,6 +56,11 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
+        if self.section.aspects and Indentation in self.section.aspects:
+            item = [aspect for aspect in self.section.aspects if
+                    isinstance(aspect, Indentation)][0]
```
Makman2: just layout candy:
put the `if` before `isinstance`, so you have
```
<element> <for x in iterable>
<if condition>
```
optional though ;)

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,6 +56,11 @@ def run(self,
...
+            use_spaces = (item.indent_type == 'space')
```
Makman2: redundant braces, but not sure if it improves readability... up to you

bears/general/SpaceConsistencyBear.py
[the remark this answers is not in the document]
biscuitsnake: Think so, yeah

bears/general/SpaceConsistencyBear.py
biscuitsnake: sure, seems logical 👍

tests/general/SpaceConsistencyBearTest.py
```diff
@@ -2,6 +2,8 @@
 
 from bears.general.SpaceConsistencyBear import (
     SpaceConsistencyBear, SpacingHelper)
+from coalib.bearlib.aspects import get as get_aspect
```
userzimmermann: Just `import coalib.bearlib.aspects as coala_aspects`. The module object is a bit magical ;) It has a `__getitem__`. So you can then for example just write `coala_aspects['Indentation']` to get the `Indentation` aspectclass. That way you also get a clear `AspectNotFoundError` when you accidentally mistype the aspectclass name. `.get()` just returns `None` in that case, which makes it harder to track down the actual problem with failing things like `get('Indentayshon')('Python', ...)`

But actually, you could also just directly import the `Indentation` aspectclass here, since it's the only one you are using below...

bears/general/SpaceConsistencyBear.py
```diff
@@ -89,6 +112,7 @@ def run(self,
                     diffs={filename: diff},
                     file=filename,
                     line=line_number,
+                    aspect=Formatting('Python'),
```
userzimmermann: This is not the intended way of adding an aspect to a `Result`. Results should reference the same aspect instances from `self.section.aspects` that were used for performing the analyses leading to the respective results

We might need some little redesign here...

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,6 +56,11 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
+        if self.section.aspects and Indentation in self.section.aspects:
+            item = [aspect for aspect in self.section.aspects
+                    if isinstance(aspect, Indentation)][0]
+            use_spaces = item.indent_type == 'space'
```
userzimmermann: This is an undesired override, ignoring any possibly conflicting explicit `use_spaces` setting, as described in https://coala.zulipchat.com/#narrow/stream/coala/subject/aspects/near/120485662

We should handle such conflicts via `@map_setting_to_aspect`. And we should extend that decorator and the `TasteMeta` class to directly include cases like the above with a syntax like:

```python
@map_setting_to_aspect(
    ...,
    use_spaces=Indentation.indent_type == 'space',
)

**Review by userzimmermann — CHANGES_REQUESTED** [review]
Hey @biscuitsnake :) This is a great beginning of bear aspectizing! But it also reveals some general problems as described below... :/ We definitely need some improvements in the whole aspects system to get this done properly

cc @Mixih @Makman2 @sils @newbazz @coala/aspects-developers

File: bears/general/SpaceConsistencyBear.py [hunk 21]
```diff
@@ -1,22 +1,40 @@
...
     LANGUAGES = {'All'}
     AUTHORS = {'The coala developers'}
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
     CAN_FIX = {'Formatting'}
 
-    @deprecate_settings(indent_size='tab_width')
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF=NewlineAtEOF,
+        allow_trailing_whitespace=TrailingSpace,
```
userzimmermann: The actual value of a setting always depends on a taste and not on the pure existence of an aspect. So either the tastes have to be added above, or we need to somehow define implicit tastes for such single-taste aspects. So the actual problem here is that `map_setting_to_aspect` allows the above, but doesn't handle it properly...

cc @adhikasp

userzimmermann (commented):
And one more problem ;)

bears/general/SpaceConsistencyBear.py
```diff
@@ -1,22 +1,40 @@
...
+                ]},
+            languages=['Python']
+        ):
     LANGUAGES = {'All'}
     AUTHORS = {'The coala developers'}
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
     CAN_FIX = {'Formatting'}
 
-    @deprecate_settings(indent_size='tab_width')
```
jayvdb: why are we removing this?

bears/general/SpaceConsistencyBear.py
biscuitsnake: @userzimmermann Thank you for all the very detailed feedback!
Am I right in saying this PR is now blocked until the above is implemented?

userzimmermann: @biscuitsnake Let's say so :) Do you want to implement those preconditions meanwhile?

userzimmermann: And you're welcome! Independently from your decision regarding the previous question ;)

biscuitsnake: @userzimmermann All done I think 👍

bears/general/SpaceConsistencyBear.py
```diff
@@ -3,20 +3,40 @@
...
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
     CAN_FIX = {'Formatting'}
 
     @deprecate_settings(indent_size='tab_width')
+    @map_setting_to_aspect(
+        enforce_newline_at_EOF=NewlineAtEOF.newline_at_EOF,
+        allow_trailing_whitespace=TrailingSpace.allow_trailing_spaces,
+        indent_size=Indentation.indent_size,
+        use_spaces=Indentation.indent_type == 'space',)
```
userzimmermann: Try to write tests for the above mappings, running some `SpaceConsistencyBear` analysis on test data based on those aspects and tastes. Of course, the `use_spaces` setting won't work yet due to missing support for that mapping expression in `@map_setting_to_aspect`. But better a temporarily failing test due to missing preconditions than no verification at all that the above mapping stuff is actually working. It doesn't make sense to merge this before

cc @coala/aspects-developers

bears/general/SpaceConsistencyBear.py
```diff
@@ -38,6 +58,12 @@ def run(self,
         result_texts = []
         additional_info_texts = []
 
+        aspect = (self.section.aspects and
+                  self.section.aspects.get('Indentation'))
+        if aspect:
+            use_spaces = (self.section.aspects.get('Indentation').indent_type
```
userzimmermann: The result of `self.section.aspects.get('Indentation')` is already referenced by `aspect` at this point. No need to re-evaluate ;)

But anyway... This should be superseded by `@map_setting_to_aspect(..., use_spaces=Indentation.indent_type == 'space')`

But... too many "but"s in here :D ... keep it for now (simplified according to the hint above) to keep the tests working you already wrote

tests/general/SpaceConsistencyBearTest.py
```diff
@@ -70,3 +64,23 @@ def test_enforce_newline_at_eof(self):
                                "    print('funny')\n",
                                "    print('the result is not funny...')"],
                               force_linebreaks=False)
+
+    def test_aspects_space(self):
+        self.section.append(Setting('aspects', 'Indentation'))
```
userzimmermann: This is useless. aspects are only handled via `Section.aspects` attribute

tests/general/SpaceConsistencyBearTest.py
```diff
@@ -70,3 +64,23 @@ def test_enforce_newline_at_eof(self):
...
+        self.section.language = 'Python'
+        self.section.aspects = AspectList([
+            coala_aspects['Indentation']('Python', indent_type='space')
+        ])
+
+        self.check_validity(self.uut, ['    t'])
+        self.check_invalidity(self.uut, ['\tt\n'])
+
+    def test_aspects_tabs(self):
+        self.section.append(Setting('aspects', 'Indentation'))
```
userzimmermann: Same as above :)

bears/general/SpaceConsistencyBear.py
```diff
@@ -3,20 +3,40 @@
...
-class SpaceConsistencyBear(LocalBear):
+class SpaceConsistencyBear(
+        LocalBear,
+        aspects={
+            'detect': [
+                Indentation,
+                NewlineAtEOF,
+                TrailingSpace,
+            ]},
+        languages=['Python']):
```
userzimmermann: Better add a newline below. I wonder why the linters aren't complaining :)

userzimmermann (changes requested):
@biscuitsnake 2nd round :)

bears/general/SpaceConsistencyBear.py
userzimmermann: @biscuitsnake Hmm... Kind of ;)

bears/general/SpaceConsistencyBear.py
biscuitsnake: @userzimmermann Why can't we have `use_spaces` as a taste of the Indentation aspect rather than `indent_type`?  (https://github.com/coala/coala/blob/master/coalib/bearlib/aspects/Formatting.py#L162)
Obviously this wouldn't mean dodging the improvements to the mapping decorator, it just seems like a more logical approach that would save some unnecessary trouble

jayvdb:
Is this still blocked?

bears/general/SpaceConsistencyBear.py
pareksha: And accordingly, the statement ``'Line contains following spacing inconsistencies:' + inconsistencies`` can be molded for that particular aspect like ``'Line contains following indentation ( the aspect asked by the user) inconsistencies:' + inconsistencies``. So, the result should be shown for only those aspects which are asked by the user.

pareksha (changes requested):

KrishMunot (approved):

jayvdb:
This is waiting for a pre-release of coala, for `Indentation` aspect.
https://github.com/coala/coala/commit/a12192a5fcfecfa0f3d79fa0a4e01230319f12e8

pareksha:
If in any case @biscuitsnake wants to leave it, I can do the issue as part of project. What do you say @jayvdb? But in any case it's totally upto you @biscuitsnake.

biscuitsnake:
go ahead @pareshka :) i'd do it but i don't have time atm unfortunately

jayvdb:
It is blocked. See https://github.com/coala/coala-bears/pull/2203#issuecomment-385335285

PR #2227, opened 2018-01-09
LineCountBear.py: Add min_lines_per_file

damian1996:
Adds the lower limit for length of file, which is set to 0 by default.

Closes https://github.com/coala/coala-bears/issues/1477

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/LineCountBear.py
+++ b/bears/general/LineCountBear.py
@@ -31,6 +31,7 @@ class LineCountBear(LocalBear):
         if exclude_blank_lines:
             num_blank_lines = self._get_blank_line_count(file)
             file_length = file_length - num_blank_lines
+
         if file_length > max_lines_per_file:
             yield Result.from_values(
                 origin=self,
--- a/bears/general/LineCountBear.py
+++ b/bears/general/LineCountBear.py
@@ -1,4 +1,5 @@
 import re
+import logging
 
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.Result import Result
@@ -17,17 +18,26 @@ class LineCountBear(LocalBear):
             list(filter(lambda x: re.match(r'^\s*$', x), file)))
         return num_blank_lines
 
-    def run(self, filename, file, max_lines_per_file: int,
+    def run(self, filename, file, min_lines_per_file: int,
+            max_lines_per_file: int,
             exclude_blank_lines: bool = False,
             ):
         """
-        Count the number of lines in a file and ensure that they are
-        smaller than a given size.
+        Count the number of lines in a file and ensure that they lie within
+        the range of given sizes.
 
+        :param min_lines_per_file: Minimum number of lines required per file.
         :param max_lines_per_file: Maximum number of lines allowed per file.
         :param exclude_blank_lines: ``True`` if blank lines are to be excluded.
         """
         file_length = len(file)
+        if min_lines_per_file > max_lines_per_file:
+            logging.error('Allowed maximum lines per file ({}) is smaller '
+                          'than minimum lines per file ({})'
+                          .format(max_lines_per_file,
+                                  min_lines_per_file))
+            return
+
         if exclude_blank_lines:
             num_blank_lines = self._get_blank_line_count(file)
             file_length = file_length - num_blank_lines
@@ -40,4 +50,13 @@ class LineCountBear(LocalBear):
                          .format(count=file_length,
                                  extra=file_length-max_lines_per_file)),
                 severity=RESULT_SEVERITY.NORMAL,
+                file=filename)
+
+        elif file_length < min_lines_per_file:
+            yield Result.from_values(
+                origin=self,
+                message=('This file has {} lines, while {} lines are '
+                         'required.'
+                         .format(file_length,
+                                 min_lines_per_file)),
                 file=filename)
--- a/bears/general/LineCountBear.py
+++ b/bears/general/LineCountBear.py
@@ -24,7 +24,7 @@ class LineCountBear(LocalBear):
         Count the number of lines in a file and ensure that they are
         smaller than a given size.
 
-        :param max_lines_per_file: Number of lines allowed per file.
+        :param max_lines_per_file: Maximum number of lines allowed per file.
         :param exclude_blank_lines: "True" if blank lines are to be excluded.
         """
         file_length = len(file)
--- a/bears/general/LineCountBear.py
+++ b/bears/general/LineCountBear.py
@@ -25,7 +25,7 @@ class LineCountBear(LocalBear):
         smaller than a given size.
 
         :param max_lines_per_file: Maximum number of lines allowed per file.
-        :param exclude_blank_lines: "True" if blank lines are to be excluded.
+        :param exclude_blank_lines: ``True`` if blank lines are to be excluded.
         """
         file_length = len(file)
         if exclude_blank_lines:
--- a/bears/general/LineCountBear.py
+++ b/bears/general/LineCountBear.py
@@ -18,8 +18,8 @@ class LineCountBear(LocalBear):
             list(filter(lambda x: re.match(r'^\s*$', x), file)))
         return num_blank_lines
 
-    def run(self, filename, file, min_lines_per_file: int,
-            max_lines_per_file: int,
+    def run(self, filename, file, min_lines_per_file: int = 1,
+            max_lines_per_file: int = 1000,
             exclude_blank_lines: bool = False,
             ):
         """
--- a/tests/general/LineCountBearTest.py
+++ b/tests/general/LineCountBearTest.py
@@ -69,7 +69,6 @@ class LineCountBearTest(LocalBearTestHelper):
             filename='default')
 
     def test_default_min_lines_per_file(self):
-        self.section.append(Setting('min_lines_per_file', 1))
         self.section.append(Setting('max_lines_per_file', 2))
         self.section.append(Setting('exclude_blank_lines', True))
         self.check_results(
--- a/tests/general/LineCountBearTest.py
+++ b/tests/general/LineCountBearTest.py
@@ -1,7 +1,9 @@
 from queue import Queue
+import logging
 
 from bears.general.LineCountBear import LineCountBear
 from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.testing.LocalBearTestHelper import execute_bear
 from coalib.results.Result import RESULT_SEVERITY, Result
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
@@ -14,6 +16,7 @@ class LineCountBearTest(LocalBearTestHelper):
         self.uut = LineCountBear(self.section, Queue())
 
     def test_run(self):
+        self.section.append(Setting('min_lines_per_file', 0))
         self.section.append(Setting('max_lines_per_file', 1))
         self.check_results(
             self.uut, ['line 1', 'line 2', 'line 3'],
@@ -27,8 +30,9 @@ class LineCountBearTest(LocalBearTestHelper):
         self.check_validity(self.uut, [])  # Empty file
 
     def test_exclude_blank_lines(self):
-        self.section.append(Setting('exclude_blank_lines', True))
+        self.section.append(Setting('min_lines_per_file', 0))
         self.section.append(Setting('max_lines_per_file', 2))
+        self.section.append(Setting('exclude_blank_lines', True))
         self.check_results(
             self.uut, ['line 1', ' ', 'line 2',
                        'line 3', '\n', '\t', ' line 4',
@@ -40,3 +44,52 @@ class LineCountBearTest(LocalBearTestHelper):
                                 severity=RESULT_SEVERITY.NORMAL,
                                 file='default')],
             filename='default')
+
+    def test_range_allowed_for_lines_per_file(self):
+        self.section.append(Setting('min_lines_per_file', 5))
+        self.section.append(Setting('max_lines_per_file', 10))
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.check_results(
+            self.uut, ['line 1', '', 'line 2', '', 'line 3', 'line 4'],
+            [Result.from_values('LineCountBear',
+                                'This file has 4 lines, while 5 lines '
+                                'are required.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+            filename='default')
+
+    def test_one_element_range(self):
+        self.section.append(Setting('min_lines_per_file', 5))
+        self.section.append(Setting('max_lines_per_file', 5))
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.check_results(
+            self.uut, ['line 1', 'line 2', '\tline 3', ' line 4 ', ' line 5',
+                       '\n', ' ', '\t', '\t\n', '\t \n'],
+            [],
+            filename='default')
+
+    def test_default_min_lines_per_file(self):
+        self.section.append(Setting('min_lines_per_file', 1))
+        self.section.append(Setting('max_lines_per_file', 2))
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.check_results(
+            self.uut, [''],
+            [Result.from_values('LineCountBear',
+                                'This file has 0 lines, while 1 lines '
+                                'are required.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+            filename='default')
+
+    def test_min_limit_exceeds_max_limit(self):
+        EXCEPTION_MESSAGE = ('ERROR:root:Allowed maximum lines per file (1) '
+                             'is smaller than minimum lines per file (2)')
+        self.section.append(Setting('min_lines_per_file', 2))
+        self.section.append(Setting('max_lines_per_file', 1))
+        self.section.append(Setting('exclude_blank_lines', True))
+        logger = logging.getLogger()
+
+        with self.assertLogs(logger, 'ERROR') as log:
+            with execute_bear(self.uut, filename='F', file='') as result:
+                self.assertEqual(len(log.output), 1)
+                self.assertEqual(log.output[0], EXCEPTION_MESSAGE)


bears/general/LineCountBear.py
```diff
@@ -18,15 +19,35 @@ def _get_blank_line_count(self, file):
...
-        smaller than a given size.
+        Count the number of lines in a file and ensure that they lie within
+        the range of given sizes. There are files for which lower limit isn't
+        required, but these files still shouldn't be rejected. Some of such
+        files when minimum is 1 are mentioned in ignore_files_regex
+        in the repository .coafile.
 
-        :param max_lines_per_file: Number of lines allowed per file.
-        :param exclude_blank_lines: "True" if blank lines are to be excluded.
+        :param max_lines_per_file: Maximum number of lines allowed per file.
```
newbazz: Indent the docstring properly ;P

tests/general/LineCountBearTest.py
```diff
@@ -40,3 +42,57 @@ def test_exclude_blank_lines(self):
...
+            self.uut, [''],
+            [Result.from_values('LineCountBear',
+                                'This file had 0 lines, which is 1 lines '
+                                'fewer than the minimum required.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+            filename='default')
+        self.check_results(self.uut, [''], [], filename='__init__.py')
+
+    def test_max_limit_is_smaller_than_min_limit(self):
```
newbazz: Write a test case in which you test any of the ignored files ;)

damian1996: I added such test case :) 

`self.check_results(self.uut, [''], [], filename='__init__.py')`

This test case should check that __init__.py file is empty and despite that min_lines_per_file=1 and max_lines_per_file=2 this test case shouldn't write any message.

newbazz: but there is coverage error in the CI.

.coafile
```diff
@@ -48,6 +48,7 @@ body_enforce_issue_reference = False
 enabled = False
 bears = LineCountBear
 max_lines_per_file = 1000
+ignore_files_regex = ^.*__init__.py$, ^.*\.gitkeep$, ^.*\.gitignore$
```
newbazz: Are you sure you need to ignore init files, coz IMO they have some code written in them.

newbazz: Yeah ik but not all init files are empty you know ;)

bears/general/LineCountBear.py
```diff
@@ -39,3 +64,14 @@ def run(self, filename, file, max_lines_per_file: int,
                                  extra=file_length-max_lines_per_file)),
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
+        elif file_length < min_lines_per_file:
+            if not any(regex.search(filename) for regex in ignore_regexes):
+                yield Result.from_values(
+                    origin=self,
+                    message=('This file had {count} lines, which is {missing} '
+                             'lines fewer than the minimum required.'
+                             .format(count=file_length,
```
newbazz: No need of using `count` and `missing`.

.coafile
damian1996: @newbazz But I don't ignore all init.py files :) I will ignore it just in special case, it means when number of lines in init file is smaller than allowed minimum set. Then ignore file, means that for this type of file, I can't write the right message, because init can have less lines than set minimum.

bears/general/LineCountBear.py
newbazz: change this as well.

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
         return num_blank_lines
 
     def run(self, filename, file, max_lines_per_file: int,
-            exclude_blank_lines: bool=False):
+            ignore_files_regex: typed_list(str),
+            exclude_blank_lines: bool=False,
```
newbazz: Change its default value to True ;P

newbazz: btw blank lines as in trailing blank lines?

damian1996: @newbazz Exclude_blank_lines was written by someone else. This is part of code which already was written here when I started doing this issue. :) So I wouldn't want change this :)

damian1996: @newbazz And blank lines are allow by default, so this should be set to True :)

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
         return num_blank_lines
 
     def run(self, filename, file, max_lines_per_file: int,
-            exclude_blank_lines: bool=False):
+            ignore_files_regex: typed_list(str),
```
Makman2: there's `ignore_files` defined by coala's settings system. You shouldn't implement such settings for bears.

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
+            exclude_blank_lines: bool=False,
+            min_lines_per_file: int=1):
         """
-        Count the number of lines in a file and ensure that they are
-        smaller than a given size.
+        Count the number of lines in a file and ensure that they lie within
+        the range of given sizes. There are files for which lower limit isn't
+        required, but these files still shouldn't be rejected. Some of such
```
Makman2: if a limit is not required, then the files won't be rejected all times, right? This is kinda confusing...

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
+        files when minimum is 1 are mentioned in ignore_files_regex
+        in the repository .coafile.
```
Makman2: This is specific to this repo, please don't document such infos in a bear.

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
-        :param max_lines_per_file: Number of lines allowed per file.
-        :param exclude_blank_lines: "True" if blank lines are to be excluded.
+        :param max_lines_per_file:
+            Maximum number of lines allowed per file.
+        :param exclude_blank_lines:
+            ``True`` if blank lines are to be excluded.
+        :param min_lines_per_file:
+            Minimum number of lines required per file.
+        :param ignore_files_regex:
+            Files, for which minimum is not required.
```
Makman2: Please add another commit changing the docstring layout before for `max_lines_per_file` and `exclude_blank_lines`

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
+
+        if min_lines_per_file > max_lines_per_file:
+            yield Result.from_values(
+                origin=self,
+                message=('Allowed maximum lines per file({}) is smaller than '
+                         'minimum lines per file({}) and this is prohibited'
+                         .format(max_lines_per_file,
+                                 min_lines_per_file)),
+                severity=RESULT_SEVERITY.NORMAL,
+                file=filename)
```
Makman2: this is not a `Result`, but an error. Please raise an exception.

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
+            return
         if exclude_blank_lines:
             num_blank_lines = self._get_blank_line_count(file)
             file_length = file_length - num_blank_lines
         if file_length > max_lines_per_file:
             yield Result.from_values(
                 origin=self,
-                message=('This file had {count} lines, which is {extra} '
+                message=('This file had {} lines, which is {} '
```
Makman2: why this change?

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
                          'lines more than the maximum limit specified.'
-                         .format(count=file_length,
-                                 extra=file_length-max_lines_per_file)),
+                         .format(file_length,
+                                 file_length-max_lines_per_file)),
```
Makman2: Definitely unnecessary change. Please file an issue to discuss removing named placeholders if the format-string is simple enough.

bears/general/LineCountBear.py
```diff
@@ -18,24 +19,58 @@ def _get_blank_line_count(self, file):
...
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
+        elif file_length < min_lines_per_file:
+            if not any(regex.search(filename) for regex in ignore_regexes):
+                yield Result.from_values(
+                    origin=self,
+                    message=('This file had {} lines, which is {} '
+                             'lines fewer than the minimum required.'
```
Makman2: than ~~the minimum~~ required.

bears/general/LineCountBear.py
damian1996: @Makman2 This have been suggested by previous reviewer :p

bears/general/LineCountBear.py
damian1996: @Makman2 This have been suggested by previous reviewer :p

Makman2:
unack bd7812f5496150401b468ed59d1a6981a578c588

bears/general/LineCountBear.py
damian1996: @Makman2 Hm, so I shouldn't consider exceptions for required limit? :) I will need ignore_files for check exceptions later :)

bears/general/LineCountBear.py
damian1996: @Makman2 There are files which are below lower limit, but the files are still needed in project. For example init.py files.
Maybe I should set min_lines_per_file to 0 and I won't mention any exceptions? Then user can define his own exceptions, if user will have bigger minimum limit than 0 ? :)

bears/general/LineCountBear.py
damian1996: Ok, I will correct this

bears/general/LineCountBear.py
Makman2: Not part of your desired change :+1: Could be a good newcomer issue anyway :+1:

bears/general/LineCountBear.py
Makman2: `min_lines_per_file` should be disabled by default (--> `=0`), but I rather meant to improve the phrasing here :)

damian1996:
I removed ignore_files, because I set min_lines_per_file to 0 by default. Additionally, I have had problem with raise ValueError (with test it). I tried to do similar to DartLintBear, where contributors used assertRaisesRegex, however I can't use this by other error, so I made it another (like in the newest commit).

bears/general/LineCountBear.py
```diff
@@ -17,16 +17,37 @@ def _get_blank_line_count(self, file):
...
+            min_lines_per_file: int=0):
         """
-        Count the number of lines in a file and ensure that they are
-        smaller than a given size.
+        Count the number of lines in a file and ensure that they lie within
+        the range of given sizes.
 
-        :param max_lines_per_file: Number of lines allowed per file.
-        :param exclude_blank_lines: "True" if blank lines are to be excluded.
+        :param max_lines_per_file: Maximum number of lines allowed per file.
```
Makman2: Unnecessary change. You can put this into a separate commit

bears/general/LineCountBear.py
```diff
@@ -17,16 +17,37 @@ def _get_blank_line_count(self, file):
...
+        :param exclude_blank_lines: ``True`` if blank lines are to be excluded.
```
Makman2: Here too, separate commit

.coafile
Makman2: I don't understand this change too, is the `LineCountBear` triggered in our repo?

Makman2: Also, we don't have an `ignore_files_regex` setting :O

bears/general/LineCountBear.py
```diff
@@ -17,16 +17,37 @@ def _get_blank_line_count(self, file):
...
+            raise ValueError('Allowed maximum lines per file({}) is smaller '
+                             'than minimum lines per file({}) and this is '
+                             'prohibited'.format(max_lines_per_file,
+                                                 min_lines_per_file))
+        return
+
     def run(self, filename, file, max_lines_per_file: int,
-            exclude_blank_lines: bool=False):
+            exclude_blank_lines: bool=False,
+            min_lines_per_file: int=0):
```
Makman2: hm, I would argue that the param `min_lines_per_file` should go before `max_lines_per_file` ;)

bears/general/LineCountBear.py
```diff
@@ -17,16 +17,37 @@ def _get_blank_line_count(self, file):
             list(filter(lambda x: re.match(r'^\s*$', x), file)))
         return num_blank_lines
 
+    def _check_correct_range(self, min_lines_per_file, max_lines_per_file):
+        if min_lines_per_file > max_lines_per_file:
+            print(min_lines_per_file, max_lines_per_file)
```
Makman2: debug remain, please remove that

bears/general/LineCountBear.py
```diff
@@ -17,16 +17,37 @@ def _get_blank_line_count(self, file):
...
+            raise ValueError('Allowed maximum lines per file({}) is smaller '
+                             'than minimum lines per file({}) and this is '
+                             'prohibited'.format(max_lines_per_file,
+                                                 min_lines_per_file))
+        return
```
Makman2: Obsolete return

Makman2: Tbh, I wouldn't write a special function for that this is a simple "if condition -> raise"

bears/general/LineCountBear.py
```diff
@@ -17,16 +17,37 @@ def _get_blank_line_count(self, file):
...
         file_length = len(file)
+        try:
+            self._check_correct_range(min_lines_per_file, max_lines_per_file)
+        except ValueError as e:
+            yield Result.from_values(
+                origin=self,
+                message=str(e),
+                severity=RESULT_SEVERITY.NORMAL,
+                file=filename)
+            return
```
Makman2: No that's not a result, let the exception bubble up.

(We maybe have to implement somewhen a special exception for such errors, so the user gets a nice message instead of a traceback... Though probably those checks will then happen inside `__init__` then... Not sure yet)

bears/general/LineCountBear.py
```diff
@@ -39,3 +60,12 @@ def run(self, filename, file, max_lines_per_file: int,
                                  extra=file_length-max_lines_per_file)),
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
```
Makman2: please add another blank line, I believe this improves readability

bears/general/LineCountBear.py
```diff
@@ -39,3 +60,12 @@ def run(self, filename, file, max_lines_per_file: int,
...
+        if file_length < min_lines_per_file:
+            yield Result.from_values(
+                origin=self,
+                message=('This file had {count} lines, which is {missing} '
```
Makman2: It *has* :)

bears/general/LineCountBear.py
```diff
@@ -39,3 +60,12 @@ def run(self, filename, file, max_lines_per_file: int,
...
+                         'lines fewer than required.'
```
Makman2: You don't have to name the placeholders, just use `{}` twice

bears/general/LineCountBear.py
```diff
@@ -39,3 +60,12 @@ def run(self, filename, file, max_lines_per_file: int,
...
+                         .format(count=file_length,
+                                 missing=min_lines_per_file - file_length)),
+                severity=RESULT_SEVERITY.NORMAL,
```
Makman2: Isn't this default?

bears/general/LineCountBear.py
```diff
@@ -39,3 +60,12 @@ def run(self, filename, file, max_lines_per_file: int,
                                  extra=file_length-max_lines_per_file)),
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
+        if file_length < min_lines_per_file:
+            yield Result.from_values(
+                origin=self,
+                message=('This file had {count} lines, which is {missing} '
+                         'lines fewer than required.'
+                         .format(count=file_length,
+                                 missing=min_lines_per_file - file_length)),
```
Makman2: I would include in the message the given min-lines-per-file setting, so the user remembers what he has passed. Consider you have a few files violating this min-lines-per-file, It's a bit confusing if both numbers change in every result (because all flagged files have a different number of lines)

tests/general/LineCountBearTest.py
```diff
@@ -9,12 +9,17 @@
 
 class LineCountBearTest(LocalBearTestHelper):
 
+    EXCEPTION_MESSAGE_RE = ('ValueError: Allowed maximum lines per file(1) '
+                            'is smaller than minimum lines per file(2) and '
+                            'this is prohibited')
```
Makman2: Unused variable

tests/general/LineCountBearTest.py
```diff
@@ -40,3 +45,53 @@ def test_exclude_blank_lines(self):
...
             filename='default')
+
+    def test_of_range_allowed_for_lines_per_file(self):
+        self.section.append(Setting('max_lines_per_file', 10))
+        self.section.append(Setting('exclude_blank_lines', True))
+        self.section.append(Setting('min_lines_per_file', 5))
+        self.check_results(
+            self.uut, ['line 1', '', 'line 2', '', 'line 3', 'line 4'],
+            [Result.from_values('LineCountBear',
+                                'This file had 4 lines, which is 1 lines '
```
Makman2: hmm, this is a good demonstration why the message needs polishing ;)
1 line*s*

Makman2:
unack 5201237eef64dace3109c080415be21250986006

tests/general/LineCountBearTest.py
damian1996: @Makman2 Should I to correct only my tests (for min_lines_per_file)? Maybe should I also to correct tests with max_lines_per_file?

bears/general/LineCountBear.py
damian1996: @Makman2 I have a small problem. I made something like this : 

`
if min_lines_per_file > max_lines_per_file:  
 raise ValueError('Allowed maximum lines per file({}) is smaller than minimum lines per file({}) and this is prohibited'.format(max_lines_per_file, min_lines_per_file)) 
`

And I made similar test to DartLintBear with use of assertRaisesRegex. However I had error.
How can I make test for this? :)

Test look like this : 

`def test_max_limit_is_smaller_than_min_limit(self):`

        EXCEPTION_MESSAGE = ('ValueError: Allowed maximum lines per '
                                'file(1) is smaller than minimum lines per '
                                'file(2) and this is prohibited')
        self.section.append(Setting('max_lines_per_file', 1))
        self.section.append(Setting('min_lines_per_file', 2))
        self.section.append(Setting('exclude_blank_lines', True))
        with self.assertRaisesRegex(AssertionError, EXCEPTION_MESSAGE):
            self.check_validity(self.uut, [])`

Makman2: Tbh I just read a different bear already using `logging.err` + `return`. I've thought about this again, and in fact the logging approach works too, even with caching. So take what you like, I'm gonna think about a uniform way to handle this later...

tests/general/LineCountBearTest.py
Makman2: Ah it's the same? Then keep it, we can file an issue afterwards to correct both messages and use a grammatical-proof message :)

tests/general/LineCountBearTest.py
```diff
@@ -1,20 +1,22 @@
...
 from bears.general.LineCountBear import LineCountBear
 from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.testing.LocalBearTestHelper import execute_bear
 from coalib.results.Result import RESULT_SEVERITY, Result
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
 
 
 class LineCountBearTest(LocalBearTestHelper):
-
```
jayvdb: please restore this blank line

tests/general/LineCountBearTest.py
```diff
@@ -1,20 +1,22 @@
...
     def setUp(self):
         self.section = Section('name')
         self.uut = LineCountBear(self.section, Queue())
 
     def test_run(self):
         self.section.append(Setting('max_lines_per_file', 1))
+        self.section.append(Setting('min_lines_per_file', 0))
```
jayvdb: why is this necessary?  it has a default

bears/general/LineCountBear.py
```diff
@@ -17,19 +18,29 @@ def _get_blank_line_count(self, file):
...
             exclude_blank_lines: bool=False):
         """
-        Count the number of lines in a file and ensure that they are
-        smaller than a given size.
+        Count the number of lines in a file and ensure that they lie within
+        the range of given sizes.
 
         :param max_lines_per_file: Number of lines allowed per file.
         :param exclude_blank_lines: "True" if blank lines are to be excluded.
+        :param min_lines_per_file: Minimum number of lines required per file.
```
jayvdb: keep the order of parameters the same as in the function declaration

damian1996:
@jayvdb I made corrections in my commit :)

bears/general/LineCountBear.py
```diff
@@ -17,19 +18,29 @@ def _get_blank_line_count(self, file):
             list(filter(lambda x: re.match(r'^\s*$', x), file)))
         return num_blank_lines
 
-    def run(self, filename, file, max_lines_per_file: int,
+    def run(self, filename, file, min_lines_per_file: int=0,
+            max_lines_per_file: int=1000,
```
Makman2: adding a default here is a separate change :+1:

bears/general/LineCountBear.py
```diff
@@ -17,19 +18,29 @@ def _get_blank_line_count(self, file):
...
+            logging.error('Allowed maximum lines per file({}) is smaller than'
+                          ' minimum lines per file({}) and this is '
+                          'prohibited.'.format(max_lines_per_file,
+                                               min_lines_per_file))
+            return
+
         if exclude_blank_lines:
             num_blank_lines = self._get_blank_line_count(file)
             file_length = file_length - num_blank_lines
+
```
Makman2: separate commit

bears/general/LineCountBear.py
```diff
@@ -39,3 +50,12 @@ def run(self, filename, file, max_lines_per_file: int,
                                  extra=file_length-max_lines_per_file)),
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
+
+        if file_length < min_lines_per_file:
```
Makman2: you can actually make this an `elif` :+1:

Makman2:
unack 7dccef61282c5b88ac744c3fdfe342c6230124cb

bears/general/LineCountBear.py
```diff
@@ -17,16 +18,25 @@ def _get_blank_line_count(self, file):
...
+        the range of given sizes.
 
+        :param min_lines_per_file: Minimum number of lines required per file.
         :param max_lines_per_file: Number of lines allowed per file.
         :param exclude_blank_lines: "True" if blank lines are to be excluded.
         """
         file_length = len(file)
+        if min_lines_per_file > max_lines_per_file:
+            logging.error('Allowed maximum lines per file({}) is smaller than'
+                          ' minimum lines per file({})'
```
Makman2: add a space between `file` and the braces, looks like some code term otherwise.

bears/general/LineCountBear.py
```diff
@@ -17,16 +18,25 @@ def _get_blank_line_count(self, file):
             list(filter(lambda x: re.match(r'^\s*$', x), file)))
         return num_blank_lines
 
-    def run(self, filename, file, max_lines_per_file: int,
+    def run(self, filename, file, min_lines_per_file: int,
+            max_lines_per_file: int=1000,
```
Makman2: Still, separate commit for this one

bears/general/LineCountBear.py
```diff
@@ -39,3 +49,12 @@ def run(self, filename, file, max_lines_per_file: int,
                                  extra=file_length-max_lines_per_file)),
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
+
+        elif file_length < min_lines_per_file:
+            yield Result.from_values(
+                origin=self,
+                message=('This file has {} lines, which is {} lines fewer '
```
Makman2: fewer -> less

bears/general/LineCountBear.py
```diff
@@ -39,3 +49,12 @@ def run(self, filename, file, max_lines_per_file: int,
...
+                         'than required.'
+                         .format(file_length,
+                                 min_lines_per_file - file_length)),
```
Makman2: maybe instead of saying "this is 3 lines less than required", say "x lines are required". That makes more results from the same kind better to read, as this number won't change.

bears/general/LineCountBear.py
damian1996: @Makman2 What do you understand by new commit? :) My second commit in this PR or new issue for newcomers ? :)

bears/general/LineCountBear.py
```diff
@@ -39,3 +50,12 @@ def run(self, filename, file, max_lines_per_file: int,
...
                 severity=RESULT_SEVERITY.NORMAL,
                 file=filename)
+
+        elif file_length < min_lines_per_file:
+            yield Result.from_values(
+                origin=self,
+                message=('This file has {} lines, while {} lines are '
+                         'required.'
+                         .format(file_length,
+                                 min_lines_per_file - file_length)),
```
HizkiFW: I think the message could be rephrased to:

> This file has `file_length` lines, while `min_lines_per_file` lines are required.

The phrase "n lines are required" specifies the minimum number of lines required, not the number of lines to be added to meet the requirement.

damian1996: @HizkiFW Thanks, I forgot about change this :) Corrected :)

tests/general/LineCountBearTest.py
```diff
@@ -40,3 +42,52 @@ def test_exclude_blank_lines(self):
                                 severity=RESULT_SEVERITY.NORMAL,
                                 file='default')],
             filename='default')
+
+    def test_of_range_allowed_for_lines_per_file(self):
```
palash25: IMO you should drop the `of` from the function name and rename it to `test_range_allowed_for_lines_per_file`
`test_of` didn't sound like a test function name while I was reading it.

tests/general/LineCountBearTest.py
```diff
@@ -40,3 +42,52 @@ def test_exclude_blank_lines(self):
...
+        self.check_results(
+            self.uut, ['line 1', '', 'line 2', '', 'line 3', 'line 4'],
+            [Result.from_values('LineCountBear',
+                                'This file has 4 lines, while 5 lines '
+                                'are required.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+            filename='default')
+
+    def test_of_one_element_range(self):
```
palash25: `test_of_one_element_range`  -->  `test_one_element_range`
:+1:

tests/general/LineCountBearTest.py
```diff
@@ -40,3 +42,52 @@ def test_exclude_blank_lines(self):
...
+        self.check_results(
+            self.uut, [''],
+            [Result.from_values('LineCountBear',
+                                'This file has 0 lines, while 1 lines '
+                                'are required.',
+                                severity=RESULT_SEVERITY.NORMAL,
+                                file='default')],
+            filename='default')
+
+    def test_max_limit_is_smaller_than_min_limit(self):
```
palash25: `def test min_limit_exceeds_max_limit` might be better.

damian1996:
@javydb 
Why default for min_lines_per_file should be 1? In my opinion 0 is good choice, user can change this value later.

jayvdb:
1 and 1000 are a sensible range.
files with 0 lines usually shouldnt be part of a project.  It typically indicates the file is wrong.

There is a special case for python `__init__.py` files, as they often are zero lines.
But that is a python specific thing; most other languages do not have anything similar, and many language actually barf on a zero line file.

jayvdb:
You have a conflict, because some whitespace was added to the args.  You will need to rebase manually.

damian1996:
@jayvdb Yes, thanks :) Can you look at travis failure in my last commit? I didn't find anything in the log. :/

bears/general/LineCountBear.py
```diff
@@ -17,20 +18,30 @@ def _get_blank_line_count(self, file):
             list(filter(lambda x: re.match(r'^\s*$', x), file)))
         return num_blank_lines
 
-    def run(self, filename, file, max_lines_per_file: int,
+    def run(self, filename, file, min_lines_per_file: int=1,
+            max_lines_per_file: int=1000,
```
jayvdb: you need to amend the commits so that each default has spaces around the `=`

jayvdb:
ack https://github.com/coala/coala-bears/pull/2227/commits/52b516deb4bb0c63acef61f61207779fc2a22326 https://github.com/coala/coala-bears/pull/2227/commits/869e210eff72d6f524f2255b7c6e7a2d6d126d33 https://github.com/coala/coala-bears/pull/2227/commits/a48861b8c93c2f33bc5e8532affaa9f6e4e1a91e https://github.com/coala/coala-bears/pull/2227/commits/c086e4ab248c54962ba9aca8f50102e9c2205064 https://github.com/coala/coala-bears/pull/2227/commits/20cdaf4e249cfd41029aaa804c7c58b747b05146

jayvdb:
@gitmate-bot ff

PR #2233, opened 2018-01-11
circle.yml : Add script to kill developer builds

kriti21:
This ensures that circle will be used on PRs,
branch master and those matching regex coala/* 
The added script checks for 'Circle' in the commit message and returns exit status
based on which the build is killed or not

Closes https://github.com/coala/coala-bears/issues/1946

### DIFF
--- a/.ci/buildci.sh
+++ b/.ci/buildci.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+# this script check for 'Circle' in the commit message
+
+MESSAGE=$(git log -1 HEAD --pretty=format:%s)
+
+if [[ "$MESSAGE" == *Circle* ]]; then
+    exit 0
+fi
+
+exit 1
--- a/circle.yml
+++ b/circle.yml
@@ -54,6 +54,15 @@ machine:
   java:
     version: oraclejdk8
 
+jobs:
+  build:
+    branches:
+      only:
+        - master
+        - coala/*
+    command:
+      - bash .ci/buildci.sh
+
 test:
   override:
     - python setup.py bdist_wheel:


sangamcse:
Change your commit :)

newbazz:
@kriti21 you can hit `enter` while writing commit body ;P

.ci/buildci.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# this script check for 'Circle' in the commit message
+
+MESSAGE=$(git log -1 HEAD --pretty=format:%s)
+
+if [[ "$MESSAGE" == *Circle* ]]; then
```
ishanSrt: So are you checking if there is `Circle` anywhere in the commit message?
https://github.com/coala/coala-bears/issues/1946#issuecomment-317456449
https://github.com/coala/coala-bears/issues/1946#issuecomment-317467859
I don't think the above comments suggest that. Maybe they didn't reach a decision over there. You should discuss it with them.

ishanSrt:
Git mate is failing your reviews because you can have maximum 70 characters in a line in commit body. your single line is too long. Add a line break somewhere. 😀

.ci/buildci.sh
kriti21: It is just a way to limit circle CI builds on the basis of condition which could be any message or word (Circle for now) in the commit. I did ask for guidance before starting on this issue. 
For now i think i should wait for them to review it first and then I would discuss further. Your thoughts about this?
And thanks a lot for the suggestion on commit message :)

Makman2:
Redundant space before colon in shortlog ;)

Makman2:
I've actually enabled "Only build Pull Requests" on the CircleCI settings, maybe this already solves the problem. Could you check this @kriti21 if this option takes effect?

screenshot from 2018-02-07 23-10-04

kriti21:
@Makman2 Can you please give me some more info on how should i test this option. Thanks :)

Makman2:
You can open wip-PRs (just title them with "WIP", and maybe mention this PR so everybody understands why you are opening stub-PRs :smile:)

kriti21:
@Makman2 I think i am not getting this right. I opened a PR adding WIP tag but could not understand how to avoid circle ci tests from running on it. Here is the PR https://github.com/coala/coala-bears/pull/2288

Makman2:
Nvm, I have thought nonsense. I mean the feature explicitly states that PRs are built (and the issue wants that too :P).
I think we don't have to modify our CI-scripts to implement this, the option I've turned on is exactly what we need. Sorry for wasting your time @kriti21 :/

Issue #337, opened 2016-04-22
GitCommitBear: Detect revert commits


sils1297:
And make sure they have a reason! Reverts *always must have a reason*.

difficulty/low

ujas800:
I would like to work on this issue.

abhsag24:
@ujas800 here you go :)

Nosferatul:
@ujas800 are you working on it?

Nosferatul:
Unassigned because of inactivity

kriti21:
Please assign this issue to me.

jayvdb:
a `git revert` creates a commit with a very specific structure of the commit message which needs to be precisely checked.

```
Revert "<reverted shortlog>"
  
This reverts commit <sha>.
```

A revert commit needs to have that format.

A 'reason' needs to be added after the above, with a blank line separating it from the above.

jayvdb:
@kriti21 , please dont rebase until you've fixed the problems identified in the code review.

jayvdb:
git revert shortlog messages still cause errors when they exceed 50 chars

PR #139, opened 2016-03-01
bears/bash: Add ShellCheckBear

tushar-rishav:
The ShellCheckBear uses `shellcheck` to parse the file and returns
the issues suggested by the linter.

Fixes https://github.com/coala-analyzer/coala-bears/issues/20

PR #136, opened 2016-03-01
README: Use pip3 instead of pip

aditigupta96:
Fixes https://github.com/coala-analyzer/coala-bears/issues/125

PR #2241, opened 2018-01-17
GitCommitBear.py: Detect revert commits

kriti21:
This detects revert commits
and checks they have a reason as well.

Closes https://github.com/coala/coala-bears/issues/337
### Checklist

      them.
      pass on every single commit!

### DIFF
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -68,6 +68,12 @@ class GitCommitBear(GlobalBear):
             return True
 
     @classmethod
+    def get_revert_checks_metadata(cls):
+        return FunctionMetadata.from_function(
+            cls.check_revert,
+            omit={'self', 'shortlog', 'body'})
+
+    @classmethod
     def get_shortlog_checks_metadata(cls):
         return FunctionMetadata.from_function(
             cls.check_shortlog,
@@ -91,6 +97,7 @@ class GitCommitBear(GlobalBear):
             FunctionMetadata.from_function(
                 cls.run,
                 omit={'self', 'dependency_results'}),
+            cls.get_revert_checks_metadata(),
             cls.get_shortlog_checks_metadata(),
             cls.get_body_checks_metadata(),
             cls.get_issue_checks_metadata())
@@ -141,6 +148,10 @@ class GitCommitBear(GlobalBear):
                 yield Result(self, 'HEAD commit has no message.')
             return
 
+        yield from self.check_revert(
+            shortlog,
+            body,
+            **self.get_revert_checks_metadata().filter_parameters(kwargs))
         yield from self.check_shortlog(
             shortlog,
             **self.get_shortlog_checks_metadata().filter_parameters(kwargs))
@@ -150,6 +161,59 @@ class GitCommitBear(GlobalBear):
         yield from self.check_issue_reference(
             body,
             **self.get_issue_checks_metadata().filter_parameters(kwargs))
+
+    def check_revert(self, shortlog, body,
+                     allow_revert_commits_without_reason: bool = False):
+        """
+        Validates revert commits.
+
+        :param shortlog:
+            The shortlog message string.
+        :param body:
+            The body of the commit message of HEAD.
+        :param allow_revert_commits_without_reason:
+            Whether reason is needed for revert commits.
+        """
+        if not shortlog.startswith('Revert '):
+            return
+
+        shortlog_start_index = shortlog.find('"')
+        shortlog_end_index = shortlog.rfind('"')
+        reverted_shortlog = shortlog[shortlog_start_index +
+                                     1: shortlog_end_index]
+
+        body = body.strip('\n')
+        blankline = body.find('\n\n')
+        headline = body[:blankline] if blankline != -1 else body
+
+        if not headline.startswith('This reverts '):
+            yield Result(self, 'Invalid revert commit.')
+            return
+
+        sha_regex = re.compile(r'([0-9a-f]{40})')
+        commit_sha = re.findall(sha_regex, headline)[0]
+        command = ('git log -n 1 ' +
+                   str(commit_sha) + ' --pretty=%B')
+        previous_commit_stdout, previous_commit_stderr = run_shell_command(
+            command)
+
+        if previous_commit_stderr:
+            yield Result(self, 'Invalid sha for reverted commit.')
+            return
+
+        previous_commit_stdout = previous_commit_stdout.rstrip('\n')
+        position = previous_commit_stdout.find('\n')
+        old_shortlog = previous_commit_stdout[:position] if position != - \
+            1 else previous_commit_stdout
+
+        if old_shortlog != reverted_shortlog:
+            yield Result(self, 'Shortlog of revert commit does '
+                         'not match the original commit.')
+
+        reason = body[blankline+1:] if blankline != -1 else ''
+
+        if not allow_revert_commits_without_reason and len(reason) == 0:
+            yield Result(self, 'Revert commit does not have a reason.')
 
     def check_shortlog(self, shortlog,
                        shortlog_length: int = 50,
--- a/tests/vcs/git/GitCommitBearTest.py
+++ b/tests/vcs/git/GitCommitBearTest.py
@@ -95,11 +95,13 @@ class GitCommitBearTest(unittest.TestCase):
         metadata = GitCommitBear.get_metadata()
         self.assertEqual(
             metadata.name,
-            "<Merged signature of 'run', 'check_shortlog', 'check_body'"
-            ", 'check_issue_reference'>")
+            "<Merged signature of 'run', 'check_revert', 'check_shortlog'"
+            ", 'check_body', 'check_issue_reference'>")
 
         # Test if at least one parameter of each signature is used.
         self.assertIn('allow_empty_commit_message', metadata.optional_params)
+        self.assertIn('allow_revert_commits_without_reason',
+                      metadata.optional_params)
         self.assertIn('shortlog_length', metadata.optional_params)
         self.assertIn('body_line_length', metadata.optional_params)
         self.assertIn('body_close_issue', metadata.optional_params)
@@ -123,6 +125,68 @@ class GitCommitBearTest(unittest.TestCase):
 
         self.assertEqual(self.run_uut(allow_empty_commit_message=True),
                          [])
+        self.assert_no_msgs()
+
+    def test_revert_checks(self):
+        self.git_commit('Test commit')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+        self.git_commit('Revert "Test commit"\n\n'
+                        'Abcdef blablabla commit ' + str(commit_hash) + '\n\n'
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(self.run_uut(),
+                         ['Invalid revert commit.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Test commit1\n\n'
+                        'This is body of test commit1\n')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+
+        self.git_commit('Revert "Another Test commit2"\n\n'
+                        'This reverts commit ' + str(commit_hash) + '\n\n'
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(
+            self.run_uut(),
+            ['Shortlog of revert commit does not match the original commit.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Test commit3')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+        self.git_commit('Revert "Test commit3"\n\n'
+                        'This reverts commit ' + str(commit_hash) + '\n\n')
+        self.assertEqual(self.run_uut(),
+                         ['Revert commit does not have a reason.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Revert "Test commit3"\n\n'
+                        'This reverts commit ' + str(commit_hash) + '\n\n')
+        self.assertEqual(
+            self.run_uut(allow_revert_commits_without_reason=True),
+            [])
+        self.assert_no_msgs()
+
+        self.git_commit('Revert "Test commit3"\n\n'
+                        'This reverts commit ' + str(commit_hash) + '\n\n'
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(self.run_uut(), [])
+        self.assert_no_msgs()
+
+        invalid_commit_hash = '0000000000000000000000000000000000000000'
+        self.git_commit('Revert "Test commit3"\n\n'
+                        'This reverts commit ' + invalid_commit_hash + '\n\n'
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(self.run_uut(),
+                         ['Invalid sha for reverted commit.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Test commit4\n\n'
+                        'This is body of test commit1\n')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+        self.git_commit('Revert "Another Test commit5"\n\n'
+                        'This reverts commit ' + str(commit_hash) + '\n\n'
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(
+            self.run_uut(),
+            ['Shortlog of revert commit does not match the original commit.'])
         self.assert_no_msgs()
 
     @unittest.mock.patch('bears.vcs.git.GitCommitBear.run_shell_command',


bears/vcs/git/GitCommitBear.py
```diff
@@ -123,6 +123,9 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
                 yield Result(self, 'HEAD commit has no message.')
             return
 
+        if shortlog.startswith('Revert') and len(body) == 0:
```
ishanSrt: So you are not detecting revert commits if they do have a reason?

kriti21: I don't think revert commits with reason have any problem.

kriti21: When revert commits don't have any reason then they are wrong. Hence, detected only then.

ishanSrt: I am not sure what needs to be done here. On what was meant by the issue, whether you have to detect revert commits and an additional check if they don't have a reason, or for only when they don't have a reason. You should discuss this with someone

kriti21: I did mention @sils once on gitter asking something like this. Never got any response.

ishanSrt: you can discuss with any other maintainer 👍

kriti21: Okay. I will write tests for this first and then discuss about detecting revert commits in general. Thanks :)

newbazz: Then you need to change tests again accordingly, so its better to ask someone what to do first ;)

nalinbhardwaj: I don’t particularly like this solution since reason could be given in the shortlog as well, and failing coala on a repository due to that is not good.
Also, is there a better way to detect reverting of commits? I couldn’t find one myself, but it seems this might over-trigger quite easily.

Makman2: I would also say: Simple by default, powerful if needed.
We should implement other settings controlling what is considered a revert commit. But that can be part of a new issue.

Makman2: So I would remove the `len(body)` part

Makman2: And btw: I want to see a setting to enable looking for revert-commits :+1:

Makman2:
Commit message shortlog: Please no space before colon ;)

Makman2:
unack b4e0f48e98797965a2bea01c5482b4bda1c7b697

Makman2:
Comments not all addressed :)

Makman2:
unack 0105d2b7ccf2198170c634506a98f62e6139b844

bears/vcs/git/GitCommitBear.py
```diff
@@ -123,6 +127,9 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
                 yield Result(self, 'HEAD commit has no message.')
             return
 
+        if shortlog.startswith('Revert') and check_revert_commits:
```
Makman2: For performance reasons, switch both conditions:
```python
if check_revert_commits and shortlog.startswith('Revert'):
```

bears/vcs/git/GitCommitBear.py
```diff
@@ -96,7 +96,9 @@ def get_host_from_remotes():
             netloc = urlparse(url)[1]
         return netloc.split('.')[0]
 
-    def run(self, allow_empty_commit_message: bool = False, **kwargs):
+    def run(self, allow_empty_commit_message: bool = False,
+            check_revert_commits: bool = True,
```
Makman2: maybe to fit the naming scheme: `allow_revert_commits`? And it's by default `False`? (tbh I'm not sure what should be default for this anyway...)

bears/vcs/git/GitCommitBear.py
```diff
@@ -105,6 +107,8 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
 
         :param allow_empty_commit_message: Whether empty commit messages are
                                            allowed or not.
+        :param check_revert_commits: whether revert commits are to be
+                                            checked or not.
```
Makman2: 1. Bad docstring indentation.
2. First letter should be upper-case.

bears/vcs/git/GitCommitBear.py
kriti21: Ideally it should be true because this not only checks for revert commits but also that they have a reason too.

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,6 +125,14 @@ def test_empty_message(self):
                          [])
         self.assert_no_msgs()
 
+    def test_revert_commit(self):
+        self.git_commit('Revert : identifies revert commit\n')
```
Makman2: Actually, you can maybe try to invoke `git revert` itself. That way we can see if we are really compliant with git's revert commits. So commit some arbitrary stuff, and then revert and test.

Makman2:
unack 28aa0304b341dd596741acd9389e5dc8cf593a50

kriti21:
@Makman2 I have made all the suggested changes. Please review :)

bears/vcs/git/GitCommitBear.py
```diff
@@ -123,6 +127,10 @@ def run(self, allow_empty_commit_message: bool = False, **kwargs):
                 yield Result(self, 'HEAD commit has no message.')
             return
 
+        if not allow_revert_commits and shortlog.startswith('Revert'):
```
jayvdb: this isnt sufficient.  it is a good heuristic to start the checking process, but it needs to be more detailed.  See https://github.com/coala/coala-bears/issues/337#issuecomment-372045150

jayvdb:
@kriti21 , ping.  This is quite urgent.  Only need to do a little more improvement (see comments above) , and the rest of the problem will be done in future PRs.

kriti21:
On it @jayvdb

kriti21:
@jayvdb Please check if this does what is asked in the comment  [#337 (comment)](https://github.com/coala/coala-bears/issues/337#issuecomment-372045150)

bears/vcs/git/GitCommitBear.py
jayvdb: Ping @kriti21 , please fix this first.  Until you're correctly verifying that it is a revert commit, everything else in the PR is not worth reviewing.

kriti21: According to https://github.com/coala/coala-bears/issues/337#issuecomment-372045150 , the shortlog must be of the format `Revert "<reverted shortlog>"` implying that it must begin with `Revert` and the following string could be anything. This is the basis of line #136 above. 

In the line #139 below, the expected first line of commit body i.e. `This reverts commit <sha>.` is separated and the rest should be reason. 

I am not getting why and how to change this line (#136) above. If a check is applied to verify that the headline starts with `This reverts commit ` and contains a valid commit sha, this should work fine IMO. Could you please correct me where I am getting this wrong? @jayvdb

jayvdb: Because `shortlog.startswith('Revert')` is insufficient for the shortlog.  Please precisely check the shortlog.

kriti21:
@nalinbhardwaj Based on https://github.com/coala/coala-bears/pull/2241#issuecomment-376489510 , I have made some changes in my code that do the first part of it [link](https://pastebin.com/R5iWKxks) . Could you please provide me some more detail on how to implement the second part of your comment.

nalinbhardwaj:
@kriti21 : The simplest method to implement that is using a library like [GitPython](https://github.com/gitpython-developers/GitPython), which offers a [commit object](http://gitpython.readthedocs.io/en/stable/tutorial.html#the-commit-object) making it as simple as `repo.commit(<hash>)` to get a commit. I suspect a large part of your implementation for first part would also be rendered redundant by it.

In future, please don't use pastebin etc., try to stick to git, makes it much easier to review code in it's entirety :)

jayvdb:
@kriti21 , as mentioned on gitter, CI is broken atm, so pushing is strongly discouraged.  I am cancelling all of your CI jobs.

kriti21:
@jayvdb My PRs were failing because of some error in my code as well. That's why I pushed.

jayvdb:
But it cant be reviewed atm, so it doesnt need to be pushed.

jayvdb:
Cancelling your build.
Rebase before you push for review.

bears/vcs/git/GitCommitBear.py
```diff
@@ -131,6 +133,51 @@ def run(self,
                 yield Result(self, 'HEAD commit has no message.')
             return
 
+        revert_shortlog_regex = re.compile(r'^Revert\s"[\S+\s*]*"')
```
virresh: I think this regex match should be ``` r'^Revert\s"([\S+\s*]*)"' ```

bears/vcs/git/GitCommitBear.py
```diff
@@ -131,6 +133,51 @@ def run(self,
...
+        if not allow_revert_commits and revert_shortlog_regex.match(shortlog):
+
+            reverted_shortlog = shortlog.lstrip('Revert ')
+
+            if len(body) != 0:
+                body = body.strip('\n')
+                blankline = body.find('\n\n')
+                headline = body[:blankline] if blankline != -1 else body
+
+                if headline_regex.match(headline):
```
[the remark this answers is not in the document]
kriti21: I have used the regex to match the format of complete revert commit, not just for getting sha. That's why, I think regex is fine.

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,6 +127,45 @@ def test_empty_message(self):
                          [])
         self.assert_no_msgs()
 
+    def test_revert_checks(self):
+        self.run_git_command('commit', stdin='Test commit')
```
virresh: Please use ```self.run_git_command``` for committing only after doing a ```git add```.
Otherwise this will fail due to empty diffs and as a result no commit would be made. 
And a ```git rev-parse HEAD``` on a repository with no commit will result in failures.

Consider using ```self.git_commit``` instead as it allows empty commits and can be done even without doing a ```git add```

Rather if you want to use ```run_git_command```, use it to generate a revert commit 
For e.g ```self.run_git_command('revert', 'head')```

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +161,53 @@ def run(self,
...
+        :param shortlog:                 The shortlog message string.
+        :param body:          The body of the commit message of HEAD.
+        :param allow_revert_commits_without_reason: Whether reason is
+                                           needed for revert commits.
+        """
+        revert_shortlog_regex = re.compile(r'^Revert\s"([\S+\s*]*)"$')
+        headline_regex = re.compile(r'^This reverts commit [0-9a-f]{40}')
+        sha_regex = re.compile(r'\b[0-9a-f]{40}\b')
+
+        if (not allow_revert_commits_without_reason and
```
virresh: Question:
What will happen in case the user allows revert commits without reason and while generating a revert commit, say the shortlog doesn't match or the commit sha is altered for some reason ? (via a ```git commit --amend``` for example)

kriti21:
Please review @jayvdb @virresh

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,6 +127,64 @@ def test_empty_message(self):
...
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(self.run_uut(),
+                         ['Invalid revert commit.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Test commit1\n\n'
+                        'This is body of test commit1\n')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+
+        self.run_git_command('revert', 'head')
```
virresh: These revert commits are redundant since you are already making another commit on top of these, so Commit Bear is never run on them
(Similarly on line 179)

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,6 +127,64 @@ def test_empty_message(self):
...
+        self.assertEqual(
+            self.run_uut(),
+            ['Shortlog of revert commit does not match the original commit.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Test commit3')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+        self.run_git_command('revert', 'head')
+
+        self.git_commit('Revert "Test commit3"\n\n'
```
virresh: This is the default format of a revert commit, you can omit creating this one manually and just use the commit generated in line 155

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +161,49 @@ def run(self,
...
+        :param allow_revert_commits_without_reason: Whether reason is
+                                           needed for revert commits.
+        """
+        revert_shortlog_regex = re.compile(r'^Revert\s"([\S+\s*]*)"$')
+        headline_regex = re.compile(r'^This reverts commit [0-9a-f]{40}')
+        sha_regex = re.compile(r'\b[0-9a-f]{40}\b')
+
+        if revert_shortlog_regex.match(shortlog):
+            reverted_shortlog = shortlog.lstrip('Revert ')
+            reverted_shortlog = reverted_shortlog.strip('"')
```
virresh: Consider having a look at capturing and non-capturing regexes

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +161,47 @@ def run(self,
...
+            headline = body[:blankline] if blankline != -1 else body
+
+            if not HEADLINE_REGEX.match(headline):
+                yield Result(self, 'Invalid revert commit.')
+                return
+
+            commit_sha = re.search(HEADLINE_REGEX, headline).group(2)
+            command = ('git log -n 1 ' +
+                       str(commit_sha) + ' --pretty=%B')
+            rstdout, _ = run_shell_command(command)
```
virresh: Question : What would happen if the commit_sha is an invalid number (or a commit_hash that doesn't exist) ?
[Refer](https://github.com/coala/coala-bears/pull/2241#issuecomment-376489510).

kriti21:
Changes done. Please review @jayvdb  @virresh

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -95,11 +95,13 @@ def test_get_metadata(self):
         metadata = GitCommitBear.get_metadata()
         self.assertEqual(
             metadata.name,
-            "<Merged signature of 'run', 'check_shortlog', 'check_body'"
-            ", 'check_issue_reference'>")
+            "<Merged signature of 'run', 'check_revert', 'check_shortlog'"
+            ", 'check_body', 'check_issue_reference'>")
```
damngamerz: please don't change existing test unless there's a reason. I don't see any such change in your patch which might affect this test.
Maybe write a new separate test for this?

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,6 +127,68 @@ def test_empty_message(self):
                          [])
         self.assert_no_msgs()
 
+    def test_revert_checks(self):
+        self.git_commit('Test commit')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+        self.git_commit('Revert "Test commit"\n\n'
+                        'Abcdef blablabla commit ' + str(commit_hash) + '\n\n'
```
damngamerz: Test content should be self explanatory about its existence.
Please use proper sentence in place of
`Abcdef blablabla commit ` :P

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +162,59 @@ def run(self,
...
+
+        sha_regex = re.compile(r'([0-9a-f]{40})')
+        commit_sha = re.findall(sha_regex, headline)[0]
+        command = ('git log -n 1 ' +
+                   str(commit_sha) + ' --pretty=%B')
+        previous_commit_stdout, previous_commit_stderr = run_shell_command(
+            command)
+
+        if previous_commit_stderr:
+            yield Result(self, 'Invalid sha for reverted commit.')
```
ksdme: sha -> SHA

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +162,59 @@ def run(self,
...
+            The body of the commit message of HEAD.
+        :param allow_revert_commits_without_reason:
+            Whether reason is needed for revert commits.
+        """
+        if not shortlog.startswith('Revert '):
+            return
+
+        shortlog_start_index = shortlog.find('"')
+        shortlog_end_index = shortlog.rfind('"')
+        reverted_shortlog = shortlog[shortlog_start_index +
```
ksdme: shortlog_start_index, shortlog_end_index and reverted_shortlog are confusing names in this context as, in other places of the code shortlog refers to the entire string including the subject and the message.

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +162,59 @@ def run(self,
...
+        if previous_commit_stderr:
+            yield Result(self, 'Invalid sha for reverted commit.')
+            return
+
+        previous_commit_stdout = previous_commit_stdout.rstrip('\n')
+        position = previous_commit_stdout.find('\n')
+        old_shortlog = previous_commit_stdout[:position] if position != - \
+            1 else previous_commit_stdout
+
+        if old_shortlog != reverted_shortlog:
```
ksdme: This could be a problem if shortlog for the previous commit barely fit within the length limit as, to be a valid revert commit, it'll have to be rewritten with extra characters in the new commit and thereby cause an impossible situation.

I think the solution would be to rethink about the shortlog structure you require for a valid revert commit. I propose it should be written as
```
08b54a: Revert Commit

This reason to revert would be...

Reference if any
```

and this structure is also more aligned with the general guidelines.

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +162,59 @@ def run(self,
...
+
+        body = body.strip('\n')
+        blankline = body.find('\n\n')
+        headline = body[:blankline] if blankline != -1 else body
+
+        if not headline.startswith('This reverts '):
+            yield Result(self, 'Invalid revert commit.')
+            return
+
+        sha_regex = re.compile(r'([0-9a-f]{40})')
```
ksdme: As the intent is also to maintain readability why not support short SHA1 hashes?

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +162,59 @@ def run(self,
...
+            The shortlog message string.
+        :param body:
+            The body of the commit message of HEAD.
+        :param allow_revert_commits_without_reason:
+            Whether reason is needed for revert commits.
+        """
+        if not shortlog.startswith('Revert '):
+            return
+
+        shortlog_start_index = shortlog.find('"')
```
ksdme: Why can't I use single quotes?

bears/vcs/git/GitCommitBear.py
```diff
@@ -151,6 +162,59 @@ def run(self,
...
+        previous_commit_stdout, previous_commit_stderr = run_shell_command(
+            command)
+
+        if previous_commit_stderr:
+            yield Result(self, 'Invalid sha for reverted commit.')
+            return
+
+        previous_commit_stdout = previous_commit_stdout.rstrip('\n')
+        position = previous_commit_stdout.find('\n')
+        old_shortlog = previous_commit_stdout[:position] if position != - \
```
seblat: this linebreak looks pretty confusing try breaking before the `if`

bears/vcs/git/GitCommitBear.py
jayvdb: `git revert` doesnt do that?

jayvdb:
Lots of manual conflict resolution needed now that https://github.com/coala/coala-bears/pull/2310 has landed.

tests/vcs/git/GitCommitBearTest.py
damngamerz: Okay it all cool :+1: you just added `check_revert`

bears/vcs/git/GitCommitBear.py
kriti21: The format followed is one that is generated by git itself for revert commits. This PR only checks if a reason for revert is mentioned or not. See [link](https://github.com/coala/coala-bears/issues/337#issuecomment-372045150) 
I don't think changing shortlog structure would be most apt solution. One possible solution could be to increase shorlog length limit for revert commits (to maybe 60). @ksdme @jayvdb @damngamerz

bears/vcs/git/GitCommitBear.py
kriti21: `git revert` doesnt do that

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -125,6 +127,68 @@ def test_empty_message(self):
...
+                        'Reason for the revert commit is explained here.\n')
+        self.assertEqual(self.run_uut(),
+                         ['Invalid revert commit.'])
+        self.assert_no_msgs()
+
+        self.git_commit('Test commit1\n\n'
+                        'This is body of test commit1\n')
+        commit_hash, _ = run_shell_command('git rev-parse HEAD')
+
+        self.git_commit('Revert "Another Test commit2"\n\n'
```
refeed: This test method contains many testcases of revert features. I think it will be better if we can split this into several test method like `test_revert_check_success`, `test_revert_check_commit_no_reason`, `test_revert_check_invalid_sha_commit`, etc

bears/vcs/git/GitCommitBear.py
jayvdb: > One possible solution could be to increase shorlog length limit for revert commits

Yes, whenever a revert commit is detected, and revert commits have been permitted in the repo, other rules which might reject the commit must be turned off.

jayvdb:
Duplicate of https://github.com/coala/coala-bears/pull/2652

PR #2242, opened 2018-01-17
GherkinLintBear.py: Add new lint bear

anctartica:
Add bear using gherkin-lint

Closes https://github.com/coala/coala-bears/issues/2214

### DIFF
--- a/bears/gherkin/GherkinLintBear.py
+++ b/bears/gherkin/GherkinLintBear.py
@@ -1,0 +1,121 @@
+import json
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.NpmRequirement import NpmRequirement
+from coalib.results.Result import Result
+
+_setting_map = {True: 'on',
+                False: 'off'}
+
+
+@linter(executable='gherkin-lint',
+        use_stderr=True,
+        use_stdout=False,
+        global_bear=True)
+class GherkinLintBear:
+    """
+    Use Gherkin to run linting on feature files
+    """
+
+    LANGUAGES = {'Gherkin'}
+    REQUIREMENTS = {NpmRequirement('gherkin-lint', '2')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala_devel@googlegroups.com'}
+    LICENSE = 'AGPL_3.0'
+    CAN_DETECT = {'Formatting', 'Syntax'}
+
+    def process_output(self, output, filename, file):
+        stderr = json.loads(output)
+        for fileErr in stderr:
+            filePath = fileErr['filePath']
+            for err in fileErr['errors']:
+                yield Result.from_values(
+                    origin='{} ({})'.format(
+                        self.__class__.__name__,
+                        err['rule']),
+                    message=err['message'],
+                    line=int(err['line']),
+                    file=filePath)
+
+    @staticmethod
+    def generate_config(filename, file,
+                        allow_trailing_whitespace: bool = True,
+                        allow_dupe_feature_names: bool = True,
+                        allow_dupe_scenario_names: bool = True,
+                        allow_duplicate_tags: bool = True,
+                        allow_empty_file: bool = True,
+                        allow_files_without_scenarios: bool = True,
+                        allow_homogenous_tags: bool = True,
+                        allow_multiple_empty_lines: bool = True,
+                        allow_scenario_outlines_without_examples: bool = True,
+                        allow_superfluous_tags: bool = True,
+                        allow_unnamed_features: bool = True,
+                        allow_unnamed_scenarios: bool = True,
+                        one_space_between_tags: bool = False,
+                        use_and: bool = False):
+        """
+        :param allow_trailing_whitespace:
+            Allow trailing spaces
+        :param allow_dupe_feature_names:
+            Disallows duplicate Feature names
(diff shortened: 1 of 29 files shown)


anctartica:
gherkin-lint is not installed which is causing appveyor build to fail. I added the package name to package.json file. How can this be installed? @Makman2

anctartica:
I will try and fix this @newbazz

anctartica:
no, i just changed the version number for gherkin-lint. The appveyor tests are still failing. I see that appveyor uses a cache of node_modules. Do we need to add gherkin-lint to this cache maybe?

newbazz:
image
After the dependencies from `e` there directly comes `h`.

anctartica:
if you look in files changed it shows gherkin-lint was added.

newbazz:
ohh ya my bad ;P

bears/gherkin/GherkinLintBear.py
```diff
@@ -0,0 +1,133 @@
...
+                    origin='{} ({})'.format(
+                        self.__class__.__name__,
+                        err['rule']),
+                    message=err['message'],
+                    line=int(err['line']),
+                    file=filePath)
+
+    @staticmethod
+    def generate_config(filename, file,
+                        indentation: bool = False,
```
jayvdb: please use the standard setting name for this.

bears/gherkin/GherkinLintBear.py
```diff
@@ -0,0 +1,133 @@
...
+                        name_length: bool = False,
+                        new_line_at_eof: bool = False,
```
jayvdb: is this the same name as used by SpaceConsistencyBear?

bears/gherkin/GherkinLintBear.py
```diff
@@ -0,0 +1,133 @@
...
+                        no_dupe_feature_names: bool = False,
+                        no_dupe_scenario_names: bool = False,
+                        no_duplicate_tags: bool = False,
+                        no_empty_file: bool = False,
```
jayvdb: We have a PR about this for LineCountBear

bears/gherkin/GherkinLintBear.py
```diff
@@ -0,0 +1,133 @@
...
+                        no_files_without_scenarios: bool = False,
+                        no_homogenous_tags: bool = False,
+                        no_multiple_empty_lines: bool = False,
```
jayvdb: this would be an interesting bear; most languages have rules around how many consecutive blank lines are permitted.  Python's rules is 1 or 2 depending on context.

bears/gherkin/GherkinLintBear.py
```diff
@@ -0,0 +1,133 @@
...
+                        no_restricted_tags: bool = False,
+                        no_scenario_outlines_without_examples: bool = False,
+                        no_superfluous_tags: bool = False,
+                        no_trailing_spaces: bool = False,
```
jayvdb: see SpaceConsistencyBear

anctartica: changed to `no_trailing_whitespace`. How did you remember this detail?

bears/gherkin/GherkinLintBear.py
anctartica: changed to `enforce_newline_at_EOF`

bears/gherkin/GherkinLintBear.py
anctartica: Sorry, but I couldn't find the standard setting name... i found `indent_size` but that specifies the size of the indent. Should it be `allow_indent`?

bears/gherkin/GherkinLintBear.py
anctartica: changed to  `allow_trailing_whitespace`

bears/gherkin/GherkinLintBear.py
jayvdb: @anctartica , does it have a fixed indentation size?

If so, use `indent_size = <x>` and allow only the fixed indentation size or `0`

jayvdb: I dont see any of the test files which dont use indentation.

anctartica: `indentation` has to be configured like this 
`"indentation" : [
    "on", {
      "Feature": 0,
      "Background": 0,
      "Scenario": 0,
      "Step": 2,
      "Examples": 0,
      "example": 2,
      "given": 2,
      "when": 2,
      "then": 2,
      "and": 2,
      "but": 2,
      "feature tag": 0,
      "scenario tag": 0
    }
  ]`

anctartica: should it be removed from here? @jayvdb

jayvdb: yup ; remove options that we cant easily support.

they can be added later.

bears/gherkin/GherkinLintBear.py
jayvdb: IMO these `no_foo` should be `foo` with the default inverted

jayvdb:
ack https://github.com/coala/coala-bears/pull/2242/commits/3208610a9fb93ab5aa43c8c377feb7041275a4e5

jayvdb:
@gitmate-bot ff

PR #2253, opened 2018-01-24
Jinja2Bear: Add support for whitespace control

kriti21:
This adds support for whitespace control for
statement tags and when trim_blocks are used
with statement or control tags.

Closes https://github.com/coala/coala-bears/issues/2002

### DIFF
--- a/bears/jinja2/Jinja2Bear.py
+++ b/bears/jinja2/Jinja2Bear.py
@@ -112,11 +112,17 @@ class Jinja2Bear(LocalBear):
 
     VARIABLE_REGEX = re.compile(
         r'(?P<open>{{)(?P<content>.*?)(?P<close>}})')
+    STATEMENT_REGEX = re.compile(
+        r'(?P<open>{%[+-]?)'
+        r'(?P<content>(?![+-]?\s*for)'
+        r'(?![+-]?\s*if)(?![+-]?\s*endfor)'
+        r'(?![+-]?\s*endif).*?)'
+        r'(?P<close>[+-]?%})')
     CONTROL_START_REGEX = re.compile(
-        r'(?P<open>{%)(?P<content>\s*(for|if).*?)(?P<close>%})')
+        r'(?P<open>{%[+-]?)(?P<content>\s*(for|if).*?)(?P<close>[+-]?%})')
     CONTROL_END_REGEX = re.compile(
-        r'(?P<open>{%)(?P<content>\s*end(for|if)\s*?)'
-        r'(?P<close>%})(?P<label>{#.*?#})?')
+        r'(?P<open>{%[+-]?)(?P<content>\s*end(for|if)\s*?)'
+        r'(?P<close>[+-]?%})(?P<label>{#.*?#})?')
 
     def handle_control_spacing_issue(self, file, filename, line, line_number,
                                      control_spacing, match_object):
@@ -182,6 +188,44 @@ class Jinja2Bear(LocalBear):
                     column=m.start(0) + 1,
                     end_line=line_number,
                     end_column=m.end(0) + 1,
+                    diffs=diff)
+
+    def check_for_statement_spacing_issues(self,
+                                           file,
+                                           filename,
+                                           line,
+                                           line_number,
+                                           statement_spacing):
+        """
+        Checks any statement in the given line for spacing issues.
+
+        :param file:
+            The content of the file currently being inspected.
+        :param filename:
+            The name of the file currently being inspected.
+        :param line:
+            The content of the line currently being inspected.
+        :param line_number:
+            The current line number.
+        :param statement_spacing:
+            The number of spaces required on each side of a statement tag.
+        """
+        for match in self.STATEMENT_REGEX.finditer(line):
+            content = match.group('content')
+            if not has_required_spacing(content, statement_spacing):
+                diff = generate_spacing_diff(
+                    file, filename, line, line_number, match,
+                    statement_spacing)
+                yield Result.from_values(
+                    origin=self,
+                    message='Statement block not spaced with '
+                            '{} spaces on each side.'.format(
+                                statement_spacing),
+                    file=filename,
+                    line=line_number,
+                    column=match.start() + 1,
+                    end_line=line_number,
+                    end_column=match.end() + 1,
                     diffs=diff)
 
     def check_control_start_tags(self,
@@ -310,6 +354,7 @@ class Jinja2Bear(LocalBear):
             filename,
             file,
             variable_spacing: int=1,
+            statement_spacing: int=1,
             control_spacing: int=1):
         """
         Check `Jinja2 templates <http://jinja.pocoo.org>`_ for syntax,
@@ -321,6 +366,10 @@ class Jinja2Bear(LocalBear):
             this: ``{{ var_name }}``. This can be set to any number of spaces
             via the setting ``variable_spacing``.
             Malformatted variable tags are detected and fixes suggested.
+        * Statement spacing:
+            Statement tags should be padded with one space on each side, like
+            this: ``{% statement %}``. This can be set to any number of spaces
+            via the setting ``statement_spacing``.
         * Control spacing:
             Like variable spacing, but for control blocks, i.e. ``if`` and
             ``for`` constructs. Looks at both start and end block.
@@ -345,6 +394,9 @@ class Jinja2Bear(LocalBear):
         :param variable_spacing:
             The number of spaces a variable block should be spaced with.
             Default is 1.
+        :param statement_spacing:
+            The number of spaces a statement block should be spaced with.
+            Default is 1.
         :param control_spacing:
             The number of spaces a control block should be spaced with.
             Default is 1.
@@ -359,6 +411,9 @@ class Jinja2Bear(LocalBear):
 
             yield from self.check_for_variable_spacing_issues(
                 file, filename, line, line_number, variable_spacing)
+
+            yield from self.check_for_statement_spacing_issues(
+                file, filename, line, line_number, statement_spacing)
 
             yield from self.check_control_start_tags(
                 file, filename, line, line_number, control_spacing)
--- a/tests/jinja2/Jinja2BearTest.py
+++ b/tests/jinja2/Jinja2BearTest.py
@@ -23,6 +23,41 @@ Jinja2BearCustomVariableSpacingTest = verify_local_bear(
                    r'{{  var }}'),
     settings={'variable_spacing': '0'})
 
+Jinja2BearStatementSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r'foo {% statement1 %} bar',
+                 r'foo {% statement2 foobar %} bar',
+                 r'foo {%+ statement3 %} bar',
+                 r'foo {%+ statement4 -%} bar',
+                 r'foo {% statement5 -%} bar',
+                 r'foo {%- statement1 +%} bar'),
+    invalid_files=(r'foo {%foo bar %} bar',
+                   r'foo {% foo bar%} bar',
+                   r'foo {%foo bar%} bar',
+                   r'foo {%   foo bar  %} bar',
+                   r'foo {%+   statement3     %} bar',
+                   r'foo {%+statement4    -%} bar',
+                   r'foo {%    statement5-%} bar',
+                   r'foo {%-  statement1    +%} bar'))
+
+Jinja2BearCustomStatementSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r'foo {%statement1%} bar',
+                 r'foo {%statement2 foobar%} bar',
+                 r'foo {%+statement3%} bar',
+                 r'foo {%+statement4-%} bar',
+                 r'foo {%statement5-%} bar',
+                 r'foo {%-statement1+%} bar'),
+    invalid_files=(r'foo {%foo bar %} bar',
+                   r'foo {% foo bar%} bar',
+                   r'foo {% foo bar %} bar',
+                   r'foo {%   foo bar  %} bar',
+                   r'foo {%+   statement3     %} bar',
+                   r'foo {%+statement4    -%} bar',
+                   r'foo {%    statement5-%} bar',
+                   r'foo {%-  statement1    +%} bar'),
+    settings={'statement_spacing': '0'})
+
 
 class Jinja2BearSpacingDiffTest(unittest.TestCase):
 
@@ -38,6 +73,16 @@ class Jinja2BearSpacingDiffTest(unittest.TestCase):
                              '@@ -1 +1 @@\n'
                              '-foo {{var }} bar\n'
                              '+foo {{ var }} bar')
+
+    def test_statement_spacing(self):
+        content = (r'foo {%-statement1 %} bar',)
+        with execute_bear(self.uut, 'F', content) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-foo {%-statement1 %} bar\n'
+                             '+foo {%- statement1 %} bar')
 
     def test_control_spacing(self):
         content = [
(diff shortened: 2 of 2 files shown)


tests/jinja2/Jinja2BearTest.py
```diff
@@ -23,6 +23,25 @@
...
+
+Jinja2BearCustomStatementSpacingTest = verify_local_bear(
+    Jinja2Bear,
+    valid_files=(r'{%statement1%}',
+                 r'{%statement2 foobar%}'),
+    invalid_files=(r'{%foo bar %}',
+                   r'{% foo bar%}'
+                   r'{% foo bar %}'
+                   r'{%   foo bar  %}'),
+    settings={'statement_spacing': '0'})
```
khanchi97: I think you can put this zero without single quotes.

kriti21: @khanchi97 That's true but I don't think it creates much of a difference here. Other similar tests have 0 within single quotes.

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +186,45 @@ def check_for_variable_spacing_issues(self,
                     end_column=m.end(0) + 1,
                     diffs=diff)
 
+    def check_for_statement_spacing_issues(self,
+                                           file,
+                                           filename,
+                                           line,
+                                           line_number,
+                                           statement_spacing):
```
newbazz: You have to give default value here right? and datatype isnt defined as well ;)

kriti21: Default values and their data type has been defined in the run function already.

kriti21:
@Makman2 your guidance really made my day. This one finally passed all the tests. Please review :)

bears/jinja2/Jinja2Bear.py
```diff
@@ -112,11 +112,16 @@ class Jinja2Bear(LocalBear):
 
     VARIABLE_REGEX = re.compile(
         r'(?P<open>{{)(?P<content>.*?)(?P<close>}})')
+    STATEMENT_REGEX = re.compile(
+        r'(?P<open>{%[+-]?)'
+        # Ignore PycodestyleBear (E501) LineLengthBear
```
Makman2: nah nah, please break the string properly ;)

bears/jinja2/Jinja2Bear.py
```diff
@@ -294,7 +338,6 @@ def check_for_empty_control_stack(self, filename):
         """
         Checks if the control stack is empty and yields a Result for each
         item left on the stack.
-
```
Makman2: unnecessary change

bears/jinja2/Jinja2Bear.py
```diff
@@ -321,6 +365,11 @@ def run(self,
             this: ``{{ var_name }}``. This can be set to any number of spaces
             via the setting ``variable_spacing``.
             Malformatted variable tags are detected and fixes suggested.
+        * Statement spacing:
+            Statement tags should be padded with one space on each side, like
+            this: ``{% statement %}``. This can be set to any number of spaces
+            via the setting ``statement_spacing``.
+            Malformatted statement tags are detected and fixes suggested.
```
Makman2: I think you can leave this sentence, it's repeated from before. I think users understand that coala could propose changes :)

bears/jinja2/Jinja2Bear.py
```diff
@@ -345,6 +394,9 @@ def run(self,
         :param variable_spacing:
             The number of spaces a variable block should be spaced with.
             Default is 1.
+        :param statement_spacing:
+            The number of spaces a statement block should be spaced with.
+            Default is 1.
```
Makman2: hm it's actually redundant to mention the default value, coala displays them automatically when querying bear infos. But due to consistency you can keep it. We file an issue after this PR to remove this.

bears/jinja2/Jinja2Bear.py
```diff
@@ -112,11 +112,16 @@ class Jinja2Bear(LocalBear):
 
     VARIABLE_REGEX = re.compile(
         r'(?P<open>{{)(?P<content>.*?)(?P<close>}})')
+    STATEMENT_REGEX = re.compile(
+        r'(?P<open>{%[+-]?)'
+        # Ignore PycodestyleBear (E501) LineLengthBear
+        r'(?P<content>(?![+-]?\s*for)(?![+-]?\s*if)(?![+-]?\s*endfor)(?![+-]?\s*endif).*?)'
+        r'(?P<close>[+-]?%})')
```
Makman2: Isn't this covered by `CONTROL_START_REGEX` and `CONTROL_END_REGEX`?

tests/jinja2/Jinja2BearTest.py
```diff
@@ -39,6 +58,16 @@ def test_variable_spacing(self):
...
 
+    def test_statement_spacing(self):
+        content = (r'foo {%-statement1 %} bar',)
+        with execute_bear(self.uut, 'F', content) as result:
+            self.assertEqual(result[0].diffs['F'].unified_diff,
+                             '--- \n'
+                             '+++ \n'
+                             '@@ -1 +1 @@\n'
+                             '-foo {%-statement1 %} bar\n'
+                             '+foo {%- statement1 %} bar')
```
Makman2: could you also test with `{%-statement1%}` and `{%- statement1%}` and `{%-     statement1   %}` ? :)

Makman2: In fact you can move your `verify_local_bear` tests here ;)

Makman2: though... actually nvm, I think your tests using `verify_local_bear` are actually quite good.

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +189,45 @@ def check_for_variable_spacing_issues(self,
...
+                yield Result.from_values(
+                    origin=self,
+                    message='Statement blocks should be spaced with '
+                            '`{0}` spaces on each side.'.format(
+                                statement_spacing),
+                    file=filename,
+                    line=line_number,
+                    column=m.start(0) + 1,
+                    end_line=line_number,
+                    end_column=m.end(0) + 1,
```
Makman2: you don't have to specify `0` for `start()` and `end()`, it's default.

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +189,45 @@ def check_for_variable_spacing_issues(self,
...
+        for m in self.STATEMENT_REGEX.finditer(line):
+            match = m.group('content')
+            if not has_required_spacing(match, statement_spacing):
+                diff = generate_spacing_diff(
+                    file, filename, line, line_number, m,
+                    statement_spacing)
+                yield Result.from_values(
+                    origin=self,
+                    message='Statement blocks should be spaced with '
+                            '`{0}` spaces on each side.'.format(
```
Makman2: unnecessary `0` inside format placeholder :)

Makman2: I would rephrase the message a bit:
```
Statement blocks not spaces with {} spaces on each side.
```

Makman2: So it gets a more passive voice. Using `should` could sound like some recommendation from our side, although the coala-user was the one specifying the option ;)

Makman2:
@kriti21 what about
> - End tag label does not match expected label
> - Control start tag has no corresponding end

Is this included? If not I would anyway file new issues for them, they can be easily separated.

Makman2:
unack 2834786b4b1bad86866d42cdf50c9dbb96c43c1c

bears/jinja2/Jinja2Bear.py
kriti21: @Makman2 Please provide some more detail on how exactly should i improve this. I am not able to get it clearly.

bears/jinja2/Jinja2Bear.py
kriti21: Will keep in mind to open an issue for this once this PR approved. :)

bears/jinja2/Jinja2Bear.py
kriti21: No this was not covered b `CONTROL_START_REGEX` and `CONTROL_END_REGEX`. In fact if I do not include this, all my tests for statement blocks would fail.

tests/jinja2/Jinja2BearTest.py
kriti21: >could you also test with {%-statement1%} and {%- statement1%} and {%- statement1 %}


Done !
> In fact you can move your verify_local_bear tests here

Do i need to move the tests in `verify_local_bear` also?

kriti21:
>- End tag label does not match expected label
>- Control start tag has no corresponding end

These are already detected by `Jinja2Bear` . The only issue that is still to be addressed is that some tests for control start tags also apply for statement tags  (if i write a more general regex). So perhaps we could file two issues once this is approved.

1. To correct mistakes in the control start tags and
2. And then improve regex for statement tags as well.

bears/jinja2/Jinja2Bear.py
Makman2: -->
```
(r'string-that-is-ver'
 r'y-long')
```
:)

tests/jinja2/Jinja2BearTest.py
Makman2: no really keep them as they are 😅 (I should have removed my comments 😅 )

bears/jinja2/Jinja2Bear.py
Makman2: Can you reuse the `CONTROL_..._REGEX` variables in your `STATEMENT_REGEX`?

tests/jinja2/Jinja2BearTest.py
kriti21: Okay Got it :)

bears/jinja2/Jinja2Bear.py
```diff
@@ -112,11 +112,17 @@ class Jinja2Bear(LocalBear):
 
     VARIABLE_REGEX = re.compile(
         r'(?P<open>{{)(?P<content>.*?)(?P<close>}})')
+    STATEMENT_REGEX = re.compile(
+        r'(?P<open>{%[+-]?)'
+        r'(?P<content>(?![+-]?\s*for)'
+        r'(?![+-]?\s*if)(?![+-]?\s*endfor)'
+        r'(?![+-]?\s*endif).*?)'
```
Makman2: hm can you maybe simplify this to `\s(?!if).*?` or so? You repeat `[+-]` very often and it's already being matched in `<open>`

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +190,45 @@ def check_for_variable_spacing_issues(self,
...
 
+    def check_for_statement_spacing_issues(self,
+                                           file,
+                                           filename,
+                                           line,
+                                           line_number,
+                                           statement_spacing):
+        """
+        Checks any statement in the given line for spacing issues.
+        Yields a Result for each issue found.
```
Makman2: You can remove that sentence ;) That's obvious :)

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +190,45 @@ def check_for_variable_spacing_issues(self,
...
+            The name of the file currently being inspected.
+        :param line:
+            The content of the line currently being inspected.
+        :param line_number:
+            The current line number.
+        :param statement_spacing:
+            The number of spaces required on each side of a statement tag.
+        """
+        for m in self.STATEMENT_REGEX.finditer(line):
+            match = m.group('content')
```
Makman2: optional: Maybe call `m` --> `match`, and `match` --> `content`

Makman2:
unack c0c86ab1917539ec8c2671e5faec6fe64e336a74

bears/jinja2/Jinja2Bear.py
```diff
@@ -112,11 +112,17 @@ class Jinja2Bear(LocalBear):
 
     VARIABLE_REGEX = re.compile(
         r'(?P<open>{{)(?P<content>.*?)(?P<close>}})')
+    STATEMENT_REGEX = re.compile(
+        r'(?P<open>{%[+-]?)'
+        r'(?P<content>(?!for)'
+        r'(?!if)(?!endfor)'
+        r'(?!endif).*?)'
```
Makman2: I think it would be good for the sake of readability to do it like
```python
r'(?P<open>{%[+-]?)'
r'(?P<content>'
r'(?!for)(?!if)(?!endfor)(?!endif)'
r'.*?)'
```
Is it also possible to use the `|` operator?
```
(?!for|if|endfor|endif)
```

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +190,44 @@ def check_for_variable_spacing_issues(self,
...
+        """
+        for match in self.STATEMENT_REGEX.finditer(line):
+            content = match.group('content')
+            if not has_required_spacing(content, statement_spacing):
+                diff = generate_spacing_diff(
+                    file, filename, line, line_number, match,
+                    statement_spacing)
+                yield Result.from_values(
+                    origin=self,
+                    message='Statement blocks not spaces with '
```
Makman2: 1. block~~s~~
2. space~~s~~**d**
3. Why "blocks" actually? Not just "Statement not spaced with ... spaces"?

bears/jinja2/Jinja2Bear.py
kriti21: Nope this did not work. I tried it.

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +190,44 @@ def check_for_variable_spacing_issues(self,
...
+                            '`{}` spaces on each side.'.format(
```
Makman2: Hm it's a number, I think you shouldn't enclose it with backticks inside the message, I mean it appear in the console :)

bears/jinja2/Jinja2Bear.py
kriti21: `!` needs to applied for each one separately.

bears/jinja2/Jinja2Bear.py
```diff
@@ -184,6 +190,44 @@ def check_for_variable_spacing_issues(self,
...
+                                statement_spacing),
+                    file=filename,
+                    line=line_number,
+                    column=match.start() + 1,
+                    end_line=line_number,
+                    end_column=match.end() + 1,
+                    diffs=diff)
```
Makman2: >         :param diffs:
>            A dictionary with filename as key and ``Diff`` object
>            associated with it as value.

Makman2: Nvm, the variable name is just a bit misleading :) But it's okay, as the existing code calls it the same.

Makman2:
unack ca364f7cedbb5d62e5c15906042294d8a5a4151f

bears/jinja2/Jinja2Bear.py
kriti21: `diff` is not a parameter for the function I have written here. It is being passed to Result class. Should i still define this as well?

Makman2: No no keep it as is, `generate_spacing_diff` wraps everything properly inside a dict like required :)

tests/jinja2/Jinja2BearTest.py
```diff
@@ -88,6 +133,20 @@ def test_control_spacing(self):
...
 good_file3 = '{% for x in y %} one liner needs no label {% endfor %}'
 
+good_file4 = """
+foo
+{%+ for x in something -%}
+render stuff
+{%+ endfor -%}{# for x in something #}
+"""
+
+good_file4 = """
```
Makman2: --> `good_file5` ;)

tests/jinja2/Jinja2BearTest.py
```diff
@@ -88,6 +133,20 @@ def test_control_spacing(self):
...
+foo
+{%- for x in something +%}
+render stuff
+{%- endfor +%}{# for x in something #}
+"""
```
Makman2: However, these are all unused variables, aren't they?

Makman2:
Now this is an extensive test set ;) However some final polishing needed :P

Makman2:
unack 99c8da0409a41c81c4273adaeaec174dede816bf

kriti21:
@Makman2 The tests i added yesterday were not completely correct and so the simplified regex gave no error. As you can see, the tests now are better and more correct but this is the only regex with which `pytest` runs successfully. I tried a lot to simplify it, but even removing minor things like `[+-]` cause `pytest` to fail. I do have certain proposals to improve control_start regex like - 
1. control_start_regex should start with `for` or `if` or `endfor` or `endif` and therefore,
2. control_start tags test should be applied to tags beginning with `for` and `if` which is not the case right now. Also another option is to segregate handling tags and white spaces for control tags which is (to some extent) contained in one function only. (This might make the process slightly redundant though)
I want to open these issue ones this PR is approved. Review this one please :)

Makman2:
ack c1fb2f47c694292e7a44257b535b86a8e30300db

Makman2:
@gitmate-bot rebase

Makman2:
Marking approved, unrelated CI fail.

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

PR #2257, opened 2018-01-25
GoVetBear: Implement all flags

manankalra:
Implements all the flags supported by Vet.
Flags can only be used with `tool vet` as
`vet` is commonly used to examine packages.

Closes https://github.com/coala/coala-bears/issues/358

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


manankalra:
I've implemented it according to the default behavior of Vet because of the following reasons:

- We will always want Vet to check for everything that is not good with the source code and this is what Vet does by default i.e. enables all of the flags.
- If a user wants to just check for a particular type of problem, he can have that functionality using the added behavior i.e. of this PR.

bears/go/GoVetBear.py
```diff
@@ -25,5 +25,130 @@ class GoVetBear:
...
+                         range_loops: bool=False,
+                         shadow_variables: bool=False,
+                         shifts: bool=False,
+                         struct_tags: bool=False,
+                         tests_and_docs: bool=False,
+                         unreachable: bool=False,
+                         unsafe_pointers: bool=False,
+                         unused_result: bool=False):
+        """
+        :param assembly_declarations:
```
newbazz: Also add a deascription here of what the bear does.

manankalra:
@newbazz I've made the changes but don't you think the lines I've now added are a repetition of the docstring under the class definition? I guess we don't have to explain what the bear does under `create_arguments()`.

newbazz:
Yeah remove that :P

bears/go/GoVetBear.py
```diff
@@ -25,5 +25,130 @@ class GoVetBear:
...
+        if struct_tags:
+            args += ('-structtags=true',)
+        if tests_and_docs:
+            args += ('-tests=true',)
+        if unreachable:
+            args += ('-unreachable=true',)
+        if unsafe_pointers:
+            args += ('-unsafeptr=true',)
+        if unused_result:
+            args += ('-unusedresult=true',)
```
Makman2: You should use a list and not tuples, this is way faster with that huge amount of args ;)

Makman2: and I recommend using `itertools.compress`, this compresses the code a bit :) But optional

tests/go/GoVetBearTest.py
```diff
@@ -20,7 +21,39 @@ def setUp(self):
...
+        self.section.append(Setting('printf', True))
+        self.section.append(Setting('range_loops', True))
+        self.section.append(Setting('shadow_variables', True))
+        self.section.append(Setting('shifts', True))
+        self.section.append(Setting('struct_tags', True))
+        self.section.append(Setting('unreachable', True))
+        self.section.append(Setting('unsafe_pointers', True))
+        self.section.append(Setting('unused_result', True))
+        self.check_validity(self.uut, [], self.good_file)
+        self.check_invalidity(self.uut, [], self.bad_file)
```
Makman2: Maybe use `check_results`?

bears/go/GoVetBear.py
Makman2: This is already done in the class docstring ;)

bears/go/GoVetBear.py
```diff
@@ -25,5 +25,130 @@ class GoVetBear:
...
+            or signatures and example tests that document identifiers not in
+            the package.
+        :param unreachable:
+            Unreachable code.
+        :param unsafe_pointers:
+            Likely incorrect uses of unsafe.Pointer to convert integers to
+            pointers.
+        :param unused_result:
+            Calls to well-known functions and methods that return a value that
+            is discarded
```
Makman2: The setting names should be a bit clearer, like when you enable the setting, what constructs are allowed / forbidden.

bears/go/GoVetBear.py
```diff
@@ -25,5 +25,130 @@ class GoVetBear:
...
+            reflect.StructTag.Get. Well-known encoding struct tags (json, xml)
+            used with unexported fields.
+        :param tests_and_docs:
+            Mistakes involving tests including functions with incorrect names
+            or signatures and example tests that document identifiers not in
+            the package.
+        :param unreachable:
+            Unreachable code.
+        :param unsafe_pointers:
+            Likely incorrect uses of unsafe.Pointer to convert integers to
```
Makman2: Missing space after period

Makman2: Oh it's a code-term, then please use double backticks for `unsafe.Pointer`

Makman2:
unack 17d8c9d570dc9a3db87f5c1ef54ca78d9e9b934f

tests/go/GoVetBearTest.py
manankalra: I implemented it in a way in which the already written test for GoVetBear was implemented.

bears/go/GoVetBear.py
manankalra: I've used the same names that are a part of Vet itself. So, someone familiar with Vet will be able to map his understanding with the functionalities of this bear. Using different names IMO can prove to be conflicting for users with the working knowledge of only Vet or both Vet and GoVetBear. Thoughts?

tests/go/GoVetBearTest.py
Makman2: kk :+1:

bears/go/GoVetBear.py
Makman2: It should resemble coala's style, as those are then part of coala's native settings set. If the names are well-phrased, they can be used across different bears for the same things.

bears/go/GoVetBear.py
```diff
@@ -25,5 +25,130 @@ class GoVetBear:
...
+        if struct_tags:
+            args.append('-structtags=true')
+        if tests_and_docs:
+            args.append('-tests=true')
+        if unreachable:
+            args.append('-unreachable=true')
+        if unsafe_pointers:
+            args.append('-unsafeptr=true')
+        if unused_result:
+            args.append('-unusedresult=true')
```
Makman2: you don't want to use `itertools.compress`? :) Just asking.

manankalra: done ;)

bears/go/GoVetBear.py
manankalra: I've changed the names of almost all of the params. Still, if they seem to be less self-explanatory then that may be because most of them are Go specific. If you go through their explanations, they look justified. Let me know if anything else needs to be changed. ;)

bears/go/GoVetBear.py
```diff
@@ -25,5 +28,96 @@ class GoVetBear:
...
+                         range_loop_variable: bool=False,
+                         shadow_variable: bool=False,
+                         shift: bool=False,
+                         struct_tag_format: bool=False,
+                         unreachable_code: bool=False,
+                         unsafe_pointer: bool=False,
+                         unused_result: bool=False):
+        """
+        :param assembly_go_mismatch:
+            Mismatches between assembly files and Go function declarations.
```
Makman2: Allowed or disallowed when this setting is `True`?

bears/go/GoVetBear.py
```diff
@@ -25,5 +28,96 @@ class GoVetBear:
     CAN_DETECT = {'Unused Code', 'Smell', 'Unreachable Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file):
-        return 'vet', filename
+    def create_arguments(filename, file, config_file,
+                         assembly_go_mismatch: bool=False,
+                         useless_assignment: bool=False,
```
Makman2: We should check for useless assignments by default :)
Also -> `useless_assignments` (plural s)

Makman2: Maybe add `check_` prefix or `allow_`/`disallow_` prefix to many settings to clarify when something is allowed or disallowed.

bears/go/GoVetBear.py
```diff
@@ -25,5 +28,96 @@ class GoVetBear:
...
+                         atomic: bool=False,
```
Makman2: many things can be atomic ;)

bears/go/GoVetBear.py
manankalra: @Makman2 https://github.com/coala/coala-bears/pull/2257#issuecomment-360585568
Every setting is true by default. ;)

bears/go/GoVetBear.py
```diff
@@ -25,5 +28,96 @@ class GoVetBear:
...
+                         boolean: bool=False,
```
Makman2: And that's truly bad, you have a setting, that takes `True`/`False` and is called boolean. Nobody understands what this means :D

Makman2: For example use the name `check_boolean_operator_problems` or similar

bears/go/GoVetBear.py
manankalra: This is a type of check. When it is set to True, this check will be performed and the mismatches will not be allowed.

manankalra: Do I need to prefix *disallow* to every param name? :|

bears/go/GoVetBear.py
```diff
@@ -25,5 +28,96 @@ class GoVetBear:
...
+                       'shadow_variable': '-shadow=true',
+                       'shift': '-shift=true',
+                       'struct_tag_format': '-structtags=true',
+                       'unreachable_code': '-unreachable=true',
+                       'unsafe_pointer': '-unsafeptr=true',
+                       'unused_result': '-unusedresult=true'}
+        frame = inspect.currentframe()
+        options, _, _, value_dict = inspect.getargvalues(frame)
+        values = [value_dict[i] for i in options[3:]]
+        valid_options = [vet_options[i] for i in options[3:]]
```
Makman2: uh looks like nice magic :) But I would maybe do it simpler:
```
vet_options = {
    '-asmdecl': assembly_go_mismatch,
    ...}

# Just a prototype, a better name would be nicce.
def to_vet_bool(val):
    return 'true' if val else 'false'

# Not sure actually what's better now... compress or dict-comprehension.
options = ('{}={}'.format(argname, to_vet_bool(value)) for argname, value in vet_options.items())
```

bears/go/GoVetBear.py
manankalra: Actually, atomic here refers to the sync/atomic package in Go.
https://golang.org/pkg/sync/atomic/

bears/go/GoVetBear.py
manankalra: Yeah, I think those prefixes should be added to all of them.

bears/go/GoVetBear.py
manankalra: As I've already mentioned, Vet works like this:
- If a flag is set to `true`, check corresponding to it will be performed and all other checks will be disabled.
- If a flag is set to `false`, only that check will be disabled and all other checks will be performed.

And because of https://github.com/coala/coala-bears/pull/2257#issuecomment-360585568, the flags not provided by the user should not be explicitly set to `false`.

So, I'm going with:
```python
def to_vet_bool(val):
    return 'true' if val
```

manankalra: As I've already mentioned, Vet works like this:

-If a flag is set to true, check corresponding to it will be performed and all other checks will be disabled.
- If a flag is set to false, only that check will be disabled and all other checks will be performed.
And because of #2257 (comment), the flags not provided by the user should not be explicitly set to false.

So, ```options = ('{}={}'.format(argname, to_vet_bool(value)) for argname, value in vet_options.items())``` won't work. `compress()` will do the workin this case.

bears/go/GoVetBear.py
```diff
@@ -25,5 +27,93 @@ class GoVetBear:
     CAN_DETECT = {'Unused Code', 'Smell', 'Unreachable Code'}
 
     @staticmethod
-    def create_arguments(filename, file, config_file):
-        return 'vet', filename
+    def create_arguments(filename, file, config_file,
+                         disallow_assembly_go_mismatches: bool=False,
```
jayvdb: If every check is enabled by default by go vet, then all of these need to be defaulting to `True` to replicate existing behaviour of the bear.  Otherwise, checks which were enabled before your change will be deactivated by your change, leaving a repository with no active checking.

bears/go/GoVetBear.py
```diff
@@ -25,5 +27,93 @@ class GoVetBear:
...
+            pointers.
+        :param disallow_unused_results:
+            Calls to well-known functions and methods that return a value that
+            is discarded.
+        """
+        # Every check in Vet is true by default.
+        # Adding a flag set to True enables that check and disables all other
+        # checks.
+        # Adding a flag set to False disables that check and runs all
+        # other checks.
```
manankalra: @jayvdb  I've added a comment here for clarification. 
- Actually, there's also a flag called `-all` in `vet` which is by default set to `true`. Therefore, all checks are performed.
- In this bear, there's no behavior where the defaults i.e. `bool=False` is explicitly setting those flags to `false`.

bears/go/GoVetBear.py
manankalra: Quoting from the docs:
>By default the -all flag is set so all checks are performed. If any flags are explicitly set to true, only those tests are run. Conversely, if any flag is explicitly set to false, only those tests are disabled. Thus -printf=true runs the printf check, -printf=false runs all checks except the printf check.

bears/go/GoVetBear.py
manankalra: Maybe, the comment is a little misleading. Multiple flags set to true do not hinder with each other. I should update that. :)

manankalra: I meant all **flags** are enabled by default. The ones which are added as a part of this PR.

manankalra: Should I add a comment stating - `multiple checks can be enabled and/or disabled at once`?

tests/go/GoVetBearTest.py
```diff
@@ -24,3 +25,25 @@ def setUp(self):
...
+        self.section.append(Setting('check_method_signatures', True))
+        self.section.append(Setting('disallow_nil_function_comparisons', True))
+        self.section.append(Setting('check_printf_calls', True))
+        self.section.append(Setting('check_range_loop_variables', True))
+        self.section.append(Setting('check_shadow_variables', True))
+        self.section.append(Setting('check_shifts', True))
+        self.section.append(Setting('check_struct_tag_format', True))
+        self.section.append(Setting('disallow_unreachable_code', True))
+        self.section.append(Setting('disallow_unsafe_pointers', True))
+        self.section.append(Setting('disallow_unused_results', True))
```
sangamcse: I think you have enabled all checks. Could you add more tests by disabling some of the checks.

manankalra: Actually, all of these checks are already enabled in `vet`. This PR doesn't add the functionality to disable any check explicitly. The way to disable them is by setting the other flags to `true`. That is if you enable a particular flag explicitly, that check is performed and all others are neglected. Note: They do not hinder with each other (multiple flags can be set to `true`).

manankalra: Another test can be added by enabling a subset of them, but IMO will not add any value. Your say?

sangamcse: I'm aware of it.
> Actually, all of these checks are already enabled in vet.

This is what IMO should be tested too. We add tests to check all the cases which we are going to provide via the Bear.
> The way to disable them is by setting the other flags to `true`

Yes, It will not add any further value but it will check other possibilities of user's settings. I hope you are understanding my concern.
> Another test can be added by enabling a subset of them, but IMO will not add any further value.

PR #2264, opened 2018-01-29
PycodestyleBear.py: Add ignore regex option

anctartica:
Add settings for ignoring certain regex expression while considering line length (E501 error rule).


jayvdb:
See issue. This would be better implemented in our generic line length bear, possibly using URLBear.
The user can turn off this pycodestyle rule.

PR #2282, opened 2018-02-16
DocumentationStyleBear: Corrects generated patch in python

kriti21:
This removes the extra 'r' added in raw docstrings
in generated patch in python by DocumentationStyleBear.

Fixes https://github.com/coala/coala-bears/issues/2010


bears/documentation/DocumentationStyleBear.py
```diff
@@ -97,6 +99,13 @@ def process_documentation(self,
 
             new_metadata.append(m._replace(desc=new_desc.lstrip(' ')))
 
+        # Check for raw doctrings in python
+        if re.match(r'\s*r', doc_comment.indent):
+            doc_comment.marker = ('r"""', '', '"""')
+            doc_comment.indent = ' ' * (len(doc_comment.indent)-1)
+        else:
+            doc_comment.indent = ' ' * len(doc_comment.indent)
+
```
sks444: @kriti21 , Is this solution working locally?

kriti21: Yes absolutely! @sks444

sks444: I tested it locally, It is showing the same patches as previous

kriti21: Can you share the file you are testing it on? I will test it on that and get back to you. I tested these changes locally. I am pretty sure they should work.

sks444: @kriti21 , Sorry, my bad. I pasted your code at the wrong place.
It is absolutely working fine :+1:

tests/documentation/test_files/DocumentationStyleBear/bad_file3.py.test
```diff
@@ -6,14 +6,14 @@ def docstring_missing_description(dummy):
...
-    """
+    r"""
     Example docstring class.
     :param dummy:dummy description
     :return:nothing
     """
 
     def docstring_memberfunction(self, dummy):
-        """
+        r"""
```
Makman2: could you rather add new functions with new docstrings? Maybe inside a new test file. Then we keep the "old" tests still intact.

bears/documentation/DocumentationStyleBear.py
```diff
@@ -97,6 +99,13 @@ def process_documentation(self,
 
             new_metadata.append(m._replace(desc=new_desc.lstrip(' ')))
 
+        # Check for raw doctrings in python
+        if re.match(r'\s*r', doc_comment.indent):
```
Makman2: I'm not sure if this is so super-nice exploiting the indentation... This highly depends on parsing of indentation... Can we maybe introduce new markers for python also including `r"""`, `r"`, `r'` and `r'''`?

Makman2:
unack c7dda9a300eb7db603f079a6d4d50eedaa5a4cac

bears/documentation/DocumentationStyleBear.py
kriti21: I tried that but that did not solve the problem here( I really don't know why). The first method i tried was to change markers in https://github.com/coala/coala/blob/master/coalib/bearlib/languages/documentation/default.coalang and https://github.com/coala/coala/blob/master/coalib/bearlib/languages/documentation/doxygen.coalang  Plus then this issue would better be in `coala/coala` repo. What should I do? @Makman2

kriti21:
@Makman2 @jayvdb I did mention regarding these failures on gitter long time back but just to add a reminder here.
I tried everything mentioned in the pip documentation and many more solutions but was unable to get it right. With all possible formats to install a commit in requirements.txt file, different kind of errors occur. If https://github.com/coala/coala/pull/5184/ could be merged in the main branch before, it might of some help.

Also please look at coala/coala-bears#1161 . I think this might be the reason for the error I am stuck with here.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 # NOTE: This file is parsed by .ci/generate_bear_requirements.py
 # Use >= for development versions so that source builds always work
 coala>=0.12.0.dev20170828055500
+-e git+git://github.com/kriti21/coala.git#egg=b62f5a98da19cb99a429a32e194d002e9947fc87
```
jayvdb: blocked because that needs to get merged first.

PR #2295, opened 2018-02-26
PyTBear.py: New Bear

MacBox7:
Uses pyt to perform security analysis
Link: https://github.com/python-security/pyt

Closes: https://github.com/coala/coala-bears/issues/1637


anctartica:
are your tests passing locally?

bears/python/PyTBear.py
```diff
@@ -0,0 +1,76 @@
...
+                file=variable_tainted.groupdict()['filename'],
+                line=int(variable_tainted.groupdict()['line']),
+                severity=self.severity_map['Weak'],
+                message='Variable {} is being modified using {}'.format(
+                         variable_tainted.groupdict()['modified_variable'],
+                         variable_tainted.groupdict()['modifier']))
+
+        for tainted_variable_execution in re.finditer(
+                regex_tainted_variable_execution, output):
+            yield Result.from_values(
```
RaiVaibhav: Yielding 3 results at a same time line-52,60,71 :sweat_smile:  don't get the point what you are doing here, if you want all 3 at a same time is yes then try append and use `check_result` and create 3 different test to check whether they are outputting same result.

I think that's why test are failing.

MacBox7:
@anctartica the tests pass locally because I have pyt module installed. But I am having hard time installing pyt in travis and ci since pyt isn't available in pypi repository. (Though I am working on it by collaborating in pyt).

bears/python/PyTBear.py
MacBox7: @RaiVaibhav No, tests are not failing because of that. You can go through https://github.com/coala/coala-bears/issues/1637, I have explained my approach. The single analysis provides with 3 main information as I have explained and since `Result` is not much customizable, the only option is to extract that 3 information and display them separately. You can go through pyt output and if there is any better approach to display that result in coala. I'll be happy to change.

RaiVaibhav: I understand, thanks for clearification.

MacBox7:
Yes, even I feel that some more changes have to be done. As per my request, they have added support for returning information in JSON. So, I'll be modifying my code a bit to incorporate the changes.

jayvdb:
Did they release a new version with the json support?

MacBox7:
No, not yet. But I'll ask them if they could. The pypi repo of pyt is currently controlled by me. They allowed me to. They are very helpful when it comes to fixing bugs and adding enhancements in upstream.

jayvdb:
python-taint 0.30 was released a few days ago, however I see more fixes to json occurring on master, so making this blocked for now.

MacBox7:
Yes, I totally agree. The built is broken since I last checked. I have reported the same about 2 days back. I am following the repo and would resume the work as soon as I see that it has been stabilised and ready to be integrated.

PR #2335, opened 2018-03-13
GoMetaLintBear: Add bear implementation

manankalra:
Lints Go code.
Runs a set of linters to check for all possible
issues.

Closes https://github.com/coala/coala-bears/issues/972


manankalra:
WIP: I still need to add tests.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,66 @@
...
+from dependency_management.requirements.GoRequirement import GoRequirement
+from coalib.settings.Setting import typed_list
+
+
+@linter(executable='gometalinter',
+        output_format='regex',
+        output_regex=r'(?P<filename>\w+\.go):'
+                     r'(?P<line>\d+):'
+                     r'(?P<column>\d*):'
+                     r'(?P<severity>[a-z]+): '
```
MacBox7: Hey, wouldn't it be great if you add severity_map parameter of `@linter` annotation. Refer this http://api.coala.io/en/latest/Developers/Writing_Linter_Bears.html#using-severities.

manankalra: No, it won't be great because:
- Unlike whatever is mentioned in the tutorials, there is a clear indication of what the severity level is when gometalinter is invoked. Not just letters.
- There are many severity levels in gometalinter because it's a combination of every Go bear that we currently support and jolting them down to mere three levels vandalizes its purpose.

Still, I'll look into it. ;)

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,66 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.GoRequirement import GoRequirement
+from coalib.settings.Setting import typed_list
+
+
+@linter(executable='gometalinter',
+        output_format='regex',
+        output_regex=r'(?P<filename>\w+\.go):'
```
jayvdb: capturing filename this emits a warning, because the filename is always constant

bears/go/GoMetaLintBear.py
MacBox7: Well, I went through the code and I saw that all the messages were mapped to essentially 2 severities, warning and error. https://github.com/alecthomas/gometalinter/blob/master/issue.go. Are there really other severity levels besides these two. Have you tested the linter?

manankalra: Oh, right. I think I confused it with the other linter bear I was working on. :laughing: 
Will add the severity map thing.;)

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,66 @@
...
+            default, if present.
+        :param ignore_config:
+            Ignores configuration files, if present.
+            Can't be used together with ``gometalinter_config_file``.
+        """
+        # Arguments are parsed in order.
+        args = (filename,)
+        if enable:
+            args += tuple('--enable={}'.format(e) for e in enable)
+        if enable_all:
```
anctartica: `enable` and `enable_all` should be mutually exclusive in my opinion. Same goes for `enable_all` and `disable`. If the user is adding any two of these settings we should throw an error.

manankalra: There are ~30 linters that are supported by gometalinter. ~18 of them are enabled by default.
- `enable_all` and `enable`:
There was a reason I left it like this. Can't remember it ATM. Will look into it surely. Plus, gometalinter supports this behavior.
- `enable_all` and `disable`:
For example:  In case you want to enable 25 of them, then you first need to enable them all and then disable the remaining 5. Same goes for `disable_all` and `enable`.

P.S. - Arguments are parsed in order.

anctartica: okay yes just confirm the reason for `enable_all` and `enable` 👍 
btw i don't see the `disable_all` setting

manankalra: I know; the PR is marked as WIP. :smile:

bears/go/GoMetaLintBear.py
jayvdb: You have not created a GlobalBear , so that is all irrelevant.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    SEE_MORE = 'http://gopkg.in/alecthomas/gometalinter.v2'
+
+    def create_arguments(self, filename, file, config_file,
+                         enable: typed_list(str)=(),
+                         enable_all: bool=False,
+                         disable: typed_list(str)=(),
+                         disable_all: bool=False,
+                         gometalinter_config_file: str='',
```
jayvdb: we've had a syntax change in pycodestyle.  Please add spaces around the equals.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+            It overrides ``.gometalinter.json`` which is picked up by
+            default, if present.
+        :param ignore_config:
+            Ignores configuration files, if present.
+            Can't be used together with ``gometalinter_config_file``.
+        """
+        # Arguments are parsed in order.
+        args = (filename,)
+
+        if (enable and enable_all) or \
```
jayvdb: please do not use line continuation symbol `\`.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+           (disable and disable_all) or \
```
jayvdb: subsequent if condition lines should always be indented 8 or more, so they do not line up with the block inside the `if`

jayvdb:
@gitmate-bot rebase

jayvdb:
Two of the tests fail.

manankalra:
Actually 3. 
But https://postimg.cc/image/mr14qsh6f/

manankalra:
I'm using `1.6.1`. Also tried out with `1.10.2` and `1.9.6`. Green light with both. 
There are no recent changes in `gometalinter` too, I'm using the latest version.

jayvdb:
Maybe you need to install other go packages for the linters?

manankalra:
`gometalinter --install` does the same. Otherwise, the tests would not have passed as `gometalinter` issues an error when it tries to run a linter which isn't installed.

Also, the `circle-ci` errors show that different results are being returned in the ci, as compared to my local env. I've also tried executing the bear individually on each test file, the results do fit the expectations of the tests.

.ci/deps.sh
```diff
@@ -61,3 +62,6 @@ wget "https://downloads.sourceforge.net/project/astyle/astyle/astyle%203.0.1/ast
 tar -xvzf ~/astyle.tar.gz -C ~/
 make -C ~/astyle/build/gcc
 sudo make install -C ~/astyle/build/gcc
+
+# gometalinter installation
+gometalinter --install
```
manankalra: this is to be added here, right?

tests/go/GoMetaLintBearTest.py
```diff
@@ -0,0 +1,115 @@
...
+    with open(get_testfile_path(filename)) as file:
+        contents = file.read().splitlines(True)
+    return contents
+
+
+@skipIf(which('go') is None, 'go is not installed')
+class GoMetaLintBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('GoMetaLinter')
```
yukiisbored: Sorry for nitpicking. For consistency with the other test cases, I think it's better to not name the Section. So just `Section()` should be enough.

tests/go/GoMetaLintBearTest.py
```diff
@@ -0,0 +1,115 @@
...
+
+from coalib.testing.LocalBearTestHelper import execute_bear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+
+
+def get_testfile_path(testfile):
```
yukiisbored: Sorry for nitpicking. I think `get_test_file_path(test_file):` is better.

tests/go/GoMetaLintBearTest.py
```diff
@@ -0,0 +1,115 @@
...
+    return os.path.join(os.path.dirname(__file__),
+                        'gometalinter_test_files',
+                        testfile)
+
+
+def load_testfile(filename):
```
yukiisbored: Sorry for nitpicking. I think `load_test_file` is better.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+        severity_map={'error': RESULT_SEVERITY.MAJOR,
+                      'warning': RESULT_SEVERITY.NORMAL})
+class GoMetaLintBear:
+    """
+    Lints your Go files!
+    Concurrently runs a number of Go lint tools.
+    """
+
+    LANGUAGES = {'Go'}
+    REQUIREMENTS = {GoRequirement(
```
jayvdb: put GoRequirement on the next line, indented.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+    LICENSE = 'AGPL-3.0'
+    SEE_MORE = 'http://gopkg.in/alecthomas/gometalinter.v2'
+
+    def create_arguments(self, filename, file, config_file,
+                         enable: typed_list(str) = (),
+                         enable_all: bool = False,
+                         disable: typed_list(str) = (),
+                         disable_all: bool = False,
+                         gometalinter_config_file: str = '',
+                         ignore_config: bool = False):
```
jayvdb: add trailing comma, and put `):` on the next line; see other bears

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+    LANGUAGES = {'Go'}
+    REQUIREMENTS = {GoRequirement(
+        package='github.com/alecthomas/gometalinter', flag='-u')}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+    SEE_MORE = 'http://gopkg.in/alecthomas/gometalinter.v2'
+
+    def create_arguments(self, filename, file, config_file,
+                         enable: typed_list(str) = (),
```
jayvdb: see other bears; `enable` needs some sort of prefix, as is is common to have multiple bears in the same section and a setting `enable` is then quite ambiguous.

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+            Enable all supported linters.
+        :param disable:
+            List of linters to disable.
+        :param disable_all:
+            Disable all supported linters.
+        :param gometalinter_config_file:
+            A JSON configuration file for gometalinter.
+            It overrides ``.gometalinter.json`` which is picked up by
+            default, if present.
+        :param ignore_config:
```
jayvdb: what about using `gometalinter_config_file=False` to mean 'ignore config'

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,86 @@
...
+                         enable: typed_list(str) = (),
+                         enable_all: bool = False,
+                         disable: typed_list(str) = (),
+                         disable_all: bool = False,
+                         gometalinter_config_file: str = '',
+                         ignore_config: bool = False):
+        """
+        :param enable:
+            List of linters to enable.
+            Some linters are disabled by default. Refer goo.gl/gMirRv
```
jayvdb: use full URL.

bears/go/GoMetaLintBear.py
manankalra: all of the Go bears follow this; still want me to change it?

bears/go/GoMetaLintBear.py
manankalra: I'm already using ``gometalinter_config_file``.
Also, when no config file is provided explicitly, ``gometalinter`` automatically takes a file named ``.gometalinter.json`` as a config file. In that case, ``ignore_config`` is used.
I think I should add another config file named ``.gometalinter.json`` to showcase that functionality but it'll then require to turn on ``ignore_config`` in every test.

jayvdb:
@gitmate-bot rebase

.ci/deps.sh
jayvdb: Copied this to https://github.com/coala/coala-bears/pull/2536

.ci/deps.sh
```diff
@@ -20,6 +20,7 @@ go get -u sourcegraph.com/sqs/goreturns
 go get -u golang.org/x/tools/cmd/gotype
 go get -u github.com/kisielk/errcheck
 go get -u github.com/BurntSushi/toml/cmd/tomlv
+go get -u github.com/alecthomas/gometalinter
```
Vamshi99: It is better to install stable version, see https://github.com/alecthomas/gometalinter#installing

bears/go/GoMetaLintBear.py
```diff
@@ -0,0 +1,83 @@
+from coalib.bearlib.abstractions.Linter import linter
+from dependency_management.requirements.GoRequirement import GoRequirement
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.Setting import typed_list
+
+
+@linter(executable='gometalinter',
+        output_format='regex',
+        output_regex=r'\w+\.go:'
```
Vamshi99: Invalid regex, the regex mentioned in their repo is `<file>:<line>:[<column>]: <message> (<linter>)`

PR #2338, opened 2018-03-14
GoVetBearTest.py: Change go test files

### LINKED ISSUE #2337: GoVetBear throwing error on appveyor
GoVetBearTest failing with error `The local bear 'GoVetBear' yields a result although it shouldn't.`
Link: https://ci.appveyor.com/project/coala/coala-bears/build/1.0.9263/job/4xamw8qmesan9aat

MacBox7:
The previous test files had invalid syntax and thus
resulting into failure

Fixes #2337

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/tests/go/GoVetBearTest.py
+++ b/tests/go/GoVetBearTest.py
@@ -14,13 +14,19 @@ class GoVetBearTest(LocalBearTestHelper):
     def setUp(self):
         self.section = Section('test section')
         self.uut = GoVetBear(self.section, Queue())
-        self.good_file = os.path.join(os.path.dirname(__file__),
-                                      'test_files',
-                                      'vet_good.go')
-        self.bad_file = os.path.join(os.path.dirname(__file__),
-                                     'test_files',
-                                     'vet_bad.go')
 
-    def test_run(self):
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
+    def test_syntax_error(self):
+        good_file = os.path.join(os.path.dirname(__file__),
+                                 'test_files',
+                                 'vet_good.go')
+        bad_file = os.path.join(os.path.dirname(__file__),
+                                'test_files',
+                                'vet_bad_semantics.go')
+        self.check_validity(self.uut, [], good_file)
+        self.check_invalidity(self.uut, [], bad_file)
+
+    def test_semantic_error(self):
+        bad_file = os.path.join(os.path.dirname(__file__),
+                                'test_files',
+                                'vet_bad_syntax.go')
+        self.check_invalidity(self.uut, [], bad_file)
--- a/tests/go/test_files/vet_bad.go
+++ b/tests/go/test_files/vet_bad.go
@@ -1,4 +1,0 @@
-package main
-func_misspelt main() {
-    return 1 || 1
-}
--- a/tests/go/test_files/vet_bad_semantics.go
+++ b/tests/go/test_files/vet_bad_semantics.go
@@ -1,0 +1,8 @@
+package main
+
+import "fmt"
+
+func main() {
+	str := "hello world!"
+	fmt.Printf("%d\n", str)
+}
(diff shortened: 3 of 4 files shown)


MacBox7:
@manankalra `vet` checks for `semantic` errors and in here both the files are `syntactically` incorrect. The purpose of `vet_bad.go` is to generate `semantic` error and not `syntax` error. Hope you understand.

jayvdb:
@MacBox7 , it is still useful to give `go vet` syntax errors to ensure it finds and reports an error correctly.

You can add lots of tests cases; that is better than few test cases.

More analysis needed.  Specifically why were our tests passing yesterday, and are failing now.  What changed?  Please explain that on the issue https://github.com/coala/coala-bears/issues/2337  Also, why are the tests passing on Travis/Circle, and not Appveyor?  Likely it is related to a go/govet version change, and caching on the VMs.

MacBox7:
@jayvdb added seperate test cases for checking `syntax` and `semantic` errors. Is there anything else that needs to be taken care of?

MacBox7:
We can install the stable version from https://golang.org/dl/ but it would pin our go version to 1.10. Shouldn't the version update be handled in a different PR? I don't think installing from ppa would be a good idea still there is no official ppa that go lang supports. You can check this out https://github.com/golang/go/wiki/Ubuntu they clearly mention that they don't support it.

virresh:
Since the current method of go is also from installing ppa's, and the issue is due to golang's version update, I thought PPA update would fit it in the PR fixing the bug, although yes probably we can have a separate issue for that :+1:

tests/go/test_files/vet_good.go
```diff
@@ -1,5 +1,5 @@
 package main
 
 func main() {
-	return 1
+
```
virresh: Probably consider placing the sample hello world program of go in here
Better than a blank function :sweat_smile: 
```
package main
import "fmt"
func main() {
    fmt.Printf("%q","hello world")
}
```

Rest looks good to go

MacBox7: That's an unnecessary request, don't you think :wink:.

virresh: Just a suggestion :smiley: 
My two cents for making coala better :)

anctartica:
Change the commit message using `git commit --amend` and then push again

tests/go/GoVetBearTest.py
```diff
@@ -14,13 +14,19 @@ class GoVetBearTest(LocalBearTestHelper):
...
-    def test_run(self):
-        self.check_validity(self.uut, [], self.good_file)
-        self.check_invalidity(self.uut, [], self.bad_file)
+    def test_syntax_error(self):
+        good_file = os.path.join(os.path.dirname(__file__),
+                                 'test_files',
+                                 'vet_good.go')
+        bad_file = os.path.join(os.path.dirname(__file__),
+                                'test_files',
+                                'vet_bad_semantics.go')
```
sangamcse: In my opinion, it's a bad practice of writing test. Why not using `setUp` to set abs path of test_files and then call it in each test?
I mean you can do below code in setUp
`self.test_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'test_files'))`
And call it in each test by using `good_file = os.path.join(self.test_path, 'vet_good.go')`.

MacBox7: Yes, I felt the same. But followed the same practice which was used before. Would update it soon :smile;

tests/go/test_files/vet_good.go
jayvdb: It is a reasonable request, given that our previous example good go file wasnt very good.
It is possible that a future version of go disallows an empty function.

jayvdb:
ack https://github.com/coala/coala-bears/pull/2338/commits/f8b779f1c33e4db41e184dc2b9df1ee5e969698c

jayvdb:
@gitmate-bot ff

PR #2364, opened 2018-03-18
CMakeLintBear: Implement line length

kriti21:
Add infinite line length feature when
max_line_length is zero.

Closes https://github.com/coala/coala-bears/issues/2329


bears/cmake/CMakeLintBear.py
```diff
@@ -21,10 +23,17 @@ class CMakeLintBear:
...
+                         max_line_length: int=79,
                          cmakelint_config: path=''):
         """
+        :param max_line_length:
+             Maximum number of characters for a line.
++            When set to 0 allows infinite line length.
         :param cmakelint_config: The location of the cmakelintrc config file.
         """
+        if not max_line_length:
+            max_line_length = sys.maxsize
```
anctartica: you are not adding the setting to the args

rohanrk:
I just had this issue with CircleCI. There's been a recent patch to upstream/master to fix the coverage bug. Rebase and recommit.

bears/cmake/CMakeLintBear.py
jayvdb: Still not done.

jayvdb:
@kriti21, I repeat, please do not do pushes unless you have address all of the review points made by others.

tests/cmake/CMakeLintBearTest.py
```diff
@@ -6,6 +6,9 @@
 
 good_file = 'project(FooBar C)\nset(VERSION 0)\n'
 
+test_file = ('target_compile_options(example ' +
```
jayvdb: `test_file` is not an informative variable name.

tests/cmake/CMakeLintBearTest.py
```diff
@@ -6,6 +6,9 @@
...
+             'PUBLIC -std=c++1y -Wall -Wfloat-conversion)')
```
jayvdb: this is only 75 chars, which means it is not testing the line length check at all

bears/cmake/CMakeLintBear.py
```diff
@@ -21,11 +23,19 @@ class CMakeLintBear:
...
+
         args = ()
         if cmakelint_config:
-            args += ('--config=' + cmakelint_config,)
+            args += ('--config=' + cmakelint_config,
+                     '--linelength=' + str(max_line_length),)
```
jayvdb: your logic isnt working yet, because you have not tried to *prove* that this feature is actually working correctly.

bears/cmake/CMakeLintBear.py
```diff
@@ -21,11 +23,19 @@ class CMakeLintBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
+                         max_line_length: int=79,
```
jayvdb: before your change, the default line length check performed by `cmakelint` was infinity.  It wasnt checking line length.

You are introducing that feature, and it should not be defaulting to 79.  That will mean anyone who was using this bear and had lines longer than 79 chars will have their CI break when they upgrade to the next version of coala.  That isnt nice.

bears/cmake/CMakeLintBear.py
jayvdb: Still not really done - wasnt tested.

bears/cmake/CMakeLintBear.py
jayvdb: still not working.

bears/cmake/CMakeLintBear.py
```diff
@@ -21,11 +23,19 @@ class CMakeLintBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
+                         max_line_length: int=sys.maxsize,
```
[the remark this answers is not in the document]
newbazz: as u need to keep the default value that the line length parameter takes in case not given.

bears/cmake/CMakeLintBear.py
newbazz: @kriti21 have you checked that the argument to be given for this is `linelength` . Just asking coz its not working.

kriti21: `linelength` is a valid argument. https://github.com/richq/cmake-lint/#14

bears/cmake/CMakeLintBear.py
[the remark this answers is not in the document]
newbazz: change this to the actual default value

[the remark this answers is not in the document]
kriti21: Tried it locally. Does not make any difference.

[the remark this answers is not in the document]
newbazz: yeah it will not make any difference but what the default value is sud be the default value here as well ;)

kriti21:
@jayvdb I have tried with a number of changes locally and according to https://github.com/richq/cmake-lint/#14 the code seems fine too. Even the tests are not showing any errors. Could you please guide me on what might be missing here to make it work.

bears/cmake/CMakeLintBear.py
jayvdb: read your code changes.
carefully.

newbazz: IMO this sud come outside the if condition.

sangamcse:
```
        if cmakelint_config:
            args += ('--config=' + cmakelint_config,)

        if max_line_length:
            args += ('--linelength=' + str(max_line_length),)
```
Why I am suggesting to break and make it in 2 conditions is because if a user does not give `cmakelint_config` then your code has no use of `max_line_length` that you have given [here](https://github.com/coala/coala-bears/pull/2364/files#diff-f5c369821ed5f2660ab333edc21ed4c7R35).

anctartica: there is no need for `if max_line_length`. Consider what happens when `max_line_length` is 0.

newbazz: @sangamcse thats what is written in the above review :P

newbazz: just do it like ` return args + (--linelength.... , filename, )`

sangamcse: Then `max_line_length = sys.maxsize` @anctartica

newbazz: that condition `if not max_line_length` will get triggered as it already might have changed to sys.maxsize ;P

jayvdb:
needs a rebase

PR #2366, opened 2018-03-18
Add BlackBear

underyx:
Closes https://github.com/coala/coala-bears/issues/2340

### Checklist

      them.
      pass on every single commit!

### DIFF
--- a/bear-requirements.txt
+++ b/bear-requirements.txt
@@ -4,6 +4,7 @@ apertium-lint~=0.29
 autoflake~=0.7
 autopep8~=1.2
 bandit~=1.2
+black==18.4a4 ; python_version >= "3.6"
 cmakelint~=1.3
 cppclean~=0.12.0
 cpplint~=1.3
--- a/bear-requirements.yaml
+++ b/bear-requirements.yaml
@@ -127,6 +127,8 @@ pip_requirements:
     version: ~=1.2
   bandit:
     version: ~=1.2
+  bandit:
+    version: 18.4a4
   cmakelint:
     version: ~=1.3
   cppclean:
--- a/bears/python/BlackBear.py
+++ b/bears/python/BlackBear.py
@@ -1,0 +1,39 @@
+from coalib.bears.LocalBear import LocalBear
+from dependency_management.requirements.DistributionRequirement import DistributionRequirement
+from dependency_management.requirements.PipRequirement import PipRequirement
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+
+
+class BlackBear(LocalBear):
+    LANGUAGES = {'Python', 'Python 2', 'Python 3'}
+    REQUIREMENTS = {
+        DistributionRequirement('python3', version='3.6'),
+        PipRequirement('black', '18.4a4'),
+    }
+    AUTHORS = {'Bence Nagy'}
+    AUTHORS_EMAILS = {'bence@underyx.me'}
+    LICENSE = 'AGPL'
+    CAN_FIX = {'Formatting'}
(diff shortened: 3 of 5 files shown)


tests/python/BlackBearTest.py
```diff
@@ -0,0 +1,28 @@
+from queue import Queue
+
+from bears.python.BlackBear import BlackBear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+
+
+class BlackBearTest(LocalBearTestHelper):
+    """Test simple invalidity and validity results of BlackBear."""
```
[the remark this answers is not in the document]
underyx: I guess I'll just remove the docstring 🙄

underyx:
The bear is Python 3.6+ only, so we have to skip the test on other Python version, which makes CI fail on <100% coverage; how do we fix this?

bears/python/BlackBear.py
```diff
@@ -0,0 +1,34 @@
+from coalib.bears.LocalBear import LocalBear
+from dependency_management.requirements.PipRequirement import PipRequirement
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+
+
+class BlackBear(LocalBear):
+    LANGUAGES = {'Python', 'Python 2', 'Python 3'}
```
jayvdb: well `'Python 2'` is wrong if it is Python 3.6 only.

probably put only `'Python 3.6'` in here?

bears/python/BlackBear.py
```diff
@@ -0,0 +1,34 @@
...
+    REQUIREMENTS = {PipRequirement('black', '18.3a2')}
```
jayvdb: we could add a `DistributionRequirement('python3', version='3.6')` and then our build parsers could be enhanced to detect that and skip the bear tests on lower versions.

A quick hack is to edit the build scripts to delete the bear & its tests when the job is < 3.6

.gitignore
```diff
@@ -43,3 +43,4 @@ Gemfile.lock
 venv
 bears/upload/
 rultor_secrets.sh
+.pytest_cache
```
rohanrk: Add "Closes https://github.com/coala/coala-bears/issues/2326" to your second commit. Also I recommend being asked to be assigned to that issue or dropping this change because a lot of newcomers want to be assigned to the issue

bear-requirements.yaml
```diff
@@ -127,6 +127,8 @@ pip_requirements:
     version: ~=1.2
   bandit:
     version: ~=1.2
+  bandit:
```
anctartica: this should be black??

bears/python/BlackBear.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class BlackBear(LocalBear):
+    LANGUAGES = {'Python', 'Python 2', 'Python 3'}
+    REQUIREMENTS = {
+        DistributionRequirement('python3', version='3.6'),
+        PipRequirement('black', '18.4a4'),
+    }
+    AUTHORS = {'Bence Nagy'}
+    AUTHORS_EMAILS = {'bence@underyx.me'}
```
anctartica: Look at the this metadata from other bears... it should be `The coala developers` :wink:
also change the email accordingly

bear-requirements.txt
```diff
@@ -4,6 +4,7 @@ apertium-lint~=0.29
 autoflake~=0.7
 autopep8~=1.2
 bandit~=1.2
+black==18.4a4 ; python_version >= "3.6"
```
anctartica: this does not fit the moban template; probably add this to the ci

PR #2367, opened 2018-03-18
QuotesBear.py: Add preferred quote in Result

rohanrk:
Closes https://github.com/coala/coala-bears/issues/2327

### DIFF
--- a/bears/general/QuotesBear.py
+++ b/bears/general/QuotesBear.py
@@ -56,7 +56,9 @@ class QuotesBear(LocalBear):
         diff.change_line(sourcerange.start.line,
                          file[sourcerange.start.line - 1],
                          replacement)
-        yield Result(self, 'You do not use the preferred quotation marks.',
+        yield Result(self,
+                     ('You do not use the preferred quotation marks.'
+                      ' Preferred mark: {}'.format(preferred_quotation)),
                      diff.affected_code(filename), diffs={filename: diff})
 
     def run(self, filename, file, dependency_results,


bears/general/QuotesBear.py
```diff
@@ -56,7 +56,8 @@ def correct_single_line_str(self, filename, file, sourcerange,
         diff.change_line(sourcerange.start.line,
                          file[sourcerange.start.line - 1],
                          replacement)
-        yield Result(self, 'You do not use the preferred quotation marks.',
+        yield Result(self, 'You do not use the preferred quotation marks.'
+                     + 'Preferred mark: ' + preferred_quotation,
```
sangamcse: I think one space should be better before `Preferred` as there is no space after 1st sentence. I mean `'Preferred mark: '`--> `' Preferred mark: '`.

rohanrk:
The commit guidelines all have colons after the filenames in their shortlog.

sangamcse:
@anctartica I didn't get your point. Which colon?

anctartica:
my bad @sangamcse @rohanrk

bears/general/QuotesBear.py
```diff
@@ -56,7 +56,8 @@ def correct_single_line_str(self, filename, file, sourcerange,
         diff.change_line(sourcerange.start.line,
                          file[sourcerange.start.line - 1],
                          replacement)
-        yield Result(self, 'You do not use the preferred quotation marks.',
+        yield Result(self, 'You do not use the preferred quotation marks.'
+                     + ' Preferred mark: ' + preferred_quotation,
```
jayvdb: the strings on the two lines are not aligned together.

Python automatically concatenates strings.

And please use `%s` or `{0}` variable substitution instead of `+`

bears/general/QuotesBear.py
```diff
@@ -56,7 +56,8 @@ def correct_single_line_str(self, filename, file, sourcerange,
         diff.change_line(sourcerange.start.line,
                          file[sourcerange.start.line - 1],
                          replacement)
-        yield Result(self, 'You do not use the preferred quotation marks.',
+        yield Result(self, 'You do not use the preferred quotation marks.'
+                     ' Preferred mark: {}'.format(preferred_quotation),
```
refeed: The strings are still not aligned

you can do something like:
```
       yield Result(self,
                    ('You do not use the preferred quotation marks.'
                     ' Preferred mark: {}'.format(preferred_quotation)),
```

jayvdb:
@gitmate-bot ff

PR #2390, opened 2018-03-30
BanditBear: Add `bandit_selected_tests` option

sangamcse:
This adds an additional feature to run the bear
using a list of selected tests IDs.

Closes https://github.com/coala/coala-bears/issues/2386

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/python/BanditBear.py
+++ b/bears/python/BanditBear.py
@@ -1,5 +1,9 @@
 import json
+import re
 
+from bears.python.ChecksSelector.ChecksSelector import set_checks
+
+from coalib.bearlib import deprecate_settings
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.PipRequirement import PipRequirement
 from coalib.results.Result import Result
@@ -7,7 +11,9 @@ from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.settings.Setting import typed_list
 
 
-@linter(executable='bandit')
+@linter(executable='bandit',
+        use_stdout=True,
+        use_stderr=True)
 class BanditBear:
     """
     Performs security analysis on Python source code, utilizing the ``ast``
@@ -21,21 +27,28 @@ class BanditBear:
     CAN_DETECT = {'Security'}
 
     @staticmethod
+    @deprecate_settings(bandit_skip_tests='bandit_skipped_tests')
     def create_arguments(filename, file, config_file,
-                         bandit_skipped_tests: typed_list(str) = (
+                         bandit_skip_tests: typed_list(str) = (
                              'B105', 'B106', 'B107', 'B404', 'B603', 'B606',
                              'B607'),
-                         ):
+                         bandit_select_tests: typed_list(str) = ()):
         """
-        :param bandit_skipped_tests:
-            The IDs of the tests ``bandit`` shall not perform. You can get
-            information about the available builtin codes at
+        :param bandit_skip_tests:
+            The IDs of the tests ``bandit`` shall not perform.
+        :param bandit_select_tests:
+            The IDs of the tests ``bandit`` shall perform.
+            You can get information about the available builtin codes at
             https://github.com/openstack/bandit#usage.
         """
         args = (filename, '-f', 'json')
 
-        if bandit_skipped_tests:
-            args += ('-s', ','.join(bandit_skipped_tests))
+        select = set_checks(ignore='--skip',
+                            ignore_check=bandit_skip_tests,
+                            select='--tests',
+                            select_check=bandit_select_tests)
+        if select:
+            args += select
 
         return args
 
@@ -48,7 +61,23 @@ class BanditBear:
                       'LOW': 50}
 
     def process_output(self, output, filename, file):
-        output = json.loads(output)
+        def err_issue(message):
+            self.err('While running {0}, some issues were found:'
+                     .format(self.__class__.__name__))
+            self.err(message)
+
+        stdout, stderr = output
+        # Taking output from stderr in case bandit shows errors
+        # such as selected test ID and skipped test ID are same.
+        err_pattern = re.compile(r'ERROR\s(?P<error>.*)')
+        match = err_pattern.search(stderr)
+        if match:
+            err_issue(match.group('error'))
+
+        if not stdout:
+            return
+
+        output = json.loads(stdout)
 
         for error in output['errors']:
             yield Result.from_values(
--- a/bears/python/ChecksSelector/ChecksSelector.py
+++ b/bears/python/ChecksSelector/ChecksSelector.py
@@ -1,0 +1,23 @@
+from coalib.settings.Setting import typed_list
+
+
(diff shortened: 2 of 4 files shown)


li-boxuan:
IMO ```bandit_selected_tests``` might be a better name.

li-boxuan:
I'm just curious about what would happen if we skip and select same test IDs.

stellargo:
Instead of `-s` and `-t` isnt something along the lines of `--skip-test` `--perform-test` better? :p

li-boxuan:
@stellargo ``-s`` and ``-t`` are used by bandit, not BanditBear. :stuck_out_tongue:

stellargo:
Ahhhh sorry nvm then :)

bears/python/BanditBear.py
```diff
@@ -24,18 +26,24 @@ class BanditBear:
...
+                          'B607'),
+                         bandit_selected_tests: typed_list(str)=()):
         """
-        :param bandit_skipped_tests:
-            The IDs of the tests ``bandit`` shall not perform. You can get
-            information about the available builtin codes at
-            https://github.com/openstack/bandit#usage.
+        :param bandit_skipped_tests: The IDs of the tests ``bandit`` shall
+                                     not perform.
+        :param bandit_selected_tests:  The IDs of the tests ``bandit`` shall
```
li-boxuan: A redundant space here

sangamcse: Thank you @li-boxuan. Any other review? 😛

bears/python/BanditBear.py
```diff
@@ -47,7 +55,22 @@ def create_arguments(filename, file, config_file,
...
+        # such as selected test ID and skipped test ID are same.
+        err_pattern = re.compile(r'ERROR.*')
+        match = err_pattern.search(output[1])
+        if match:
+            warn_issue(match.group())
+
+        if not file or not output[0]:
+            return
+
+        output = json.loads(output[0])
```
virresh: The try, catch block was removed from here ?
(Not sure though if bandit will ever give incorrect json, so just asking :sweat_smile: )

sangamcse: try, catch block was never here, though I tried generating one incorrect json using 6/7 test files and in all the case, it either gave ERROR (which was not in json) or json output. So, IMO it is not required.

virresh: Ah
Cool :+1:

li-boxuan:
IMO generate_skip_decorator should be added, but seems that's not related to this issue.

bears/python/BanditBear.py
```diff
@@ -47,7 +55,22 @@ def create_arguments(filename, file, config_file,
...
-        output = json.loads(output)
+        def warn_issue(message):
+            self.warn('While running {0}, some issues were found:'
+                      .format(self.__class__.__name__))
+            self.warn(message)
+
+        # Taking output from stderr in case bandit shows errors
+        # such as selected test ID and skipped test ID are same.
+        err_pattern = re.compile(r'ERROR.*')
+        match = err_pattern.search(output[1])
```
li-boxuan: Can you add a line like ```stdout, stderr = output``` otherwise ```output[1]``` and ```output[0]``` are hard to understand

bears/python/BanditBear.py
```diff
@@ -47,7 +55,22 @@ def create_arguments(filename, file, config_file,
...
+        if match:
+            warn_issue(match.group())
+
+        if not file or not output[0]:
```
li-boxuan: I'm not sure here, but why is ```not file``` necessary?

bears/python/BanditBear.py
```diff
@@ -7,7 +8,8 @@
 from coalib.settings.Setting import typed_list
 
 
-@linter(executable='bandit')
+@linter(executable='bandit',
+        use_stderr=True)
```
li-boxuan: I understand the default value of ```use_stdout``` is ```True```. Not sure here, but maybe explicitly adding ```use_stdout=True``` makes the code easier to understand? (ref: http://api.coala.io/en/latest/Developers/Linter_Bears_Advanced.html)

sangamcse: It doesn't matter as `use_stdout` is True by default. If I add it here, it will be redundant. And What's the point of adding one extra line. To understand this, one should have understanding of `linter` class and if anyone understand `linter` class, then IMO he should also know about default `use_stdout`.

bears/python/BanditBear.py
```diff
@@ -22,19 +27,26 @@ class BanditBear:
 
     @staticmethod
     def create_arguments(filename, file, config_file,
-                         bandit_skipped_tests: typed_list(str)=
+                         bandit_skip_tests: typed_list(str)=
```
jayvdb: need to deprecate this setting.
(and fix spaces around `=` and rebase...)

bears/python/ChecksSelector/ChecksSelector.py
```diff
@@ -0,0 +1,23 @@
+from coalib.settings.Setting import typed_list
```
jayvdb: no; we have one module per bear.

bears/python/ChecksSelector/ChecksSelector.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def set_checks(ignore: str = '--ignore',
```
jayvdb: This method belongs in your bear, until it is migrated into coalib.

bears/python/BanditBear.py
```diff
@@ -21,21 +27,28 @@ class BanditBear:
     CAN_DETECT = {'Security'}
 
     @staticmethod
+    @deprecate_settings(bandit_skip_tests='bandit_skipped_tests')
     def create_arguments(filename, file, config_file,
-                         bandit_skipped_tests: typed_list(str) = (
+                         bandit_skip_tests: typed_list(str) = (
```
jayvdb: "tests" is bandit terminology.
And it is confusing terminology.

We are using `foo_ignore` and `foo_select`

bears/python/ChecksSelector/ChecksSelector.py
```diff
@@ -0,0 +1,23 @@
...
+               ignore_check: typed_list(str) = (),
+               select: str = '--select',
```
jayvdb: `ignore` & `select` are confusing names. `ignore_cli_arg` and `select_cli_arg` are more clear.

tests/python/BanditBearTest.py
```diff
@@ -24,31 +26,38 @@ def load_testfile(name, splitlines=False):
...
+@generate_skip_decorator(BanditBear)
 class BanditBearTest(LocalBearTestHelper):
-    test_assert = test(
+
+    def setUp(self):
+        self.section = Section('bandit')
+        self.queue = Queue()
+        self.uut = BanditBear(self.section, self.queue)
+
+    def execute_test(testfilename, expected_results):
```
jayvdb: why did you move the `test` function code here?

PR #2401, opened 2018-04-02
.coafile: Replace Default section with all 

### LINKED ISSUE #2397: .coafile: Replace default section with all
Since **Default** section is deprecated, it should be replaced with **all**.
Many people might take `.coafile` of coala as a reference for their project, so IMO it should be updated.

rammanoj:
As Default section is deprecated, it is replaced with all in .coafile

Closes #2397


Vamshi99:
@rammanoj You have two commits, squash them into one commit.

PR #2416, opened 2018-04-09
GitCommitBear: verify issue exists and is open

MacBox7:
Checks if the issue is valid, open and not a PR.

Closes https://github.com/coala/coala-bears/issues/2369

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


MacBox7:
Needs more work. Task left: 
1. Add more tests.
2. Check if it is a PR or an issue
3. Add setting to get auth details which is required for private repos. (needs discussion)

bears/vcs/git/GitCommitBear.py
```diff
@@ -102,6 +105,70 @@ def get_host_from_remotes():
             netloc = urlparse(url)[1]
         return netloc.split('.')[0]
 
+    @staticmethod
+    def get_owner_from_url(url):
```
virresh: We are looking for a library to parse git urls already,
IMO it would be much better to replace these functions by a library that does this already
For e.g have a look at [this](https://github.com/retr0h/git-url-parse) and [this](https://github.com/nephila/giturlparse)

nalinbhardwaj: Most of these tasks are handled already by [IGitt](https://gitlab.com/gitmate/open-source/IGitt), I recommend using that.

MacBox7: Yes. Could be done not an issue.

bears/vcs/git/GitCommitBear.py
```diff
@@ -269,7 +336,9 @@ def check_issue_reference(self, body,
                               body_close_issue: bool=False,
                               body_close_issue_full_url: bool=False,
                               body_close_issue_on_last_line: bool=False,
-                              body_enforce_issue_reference: bool=False):
+                              body_enforce_issue_reference: bool=False,
+                              body_enforce_issue_open: bool=False,
```
nalinbhardwaj: I would personally like a setting which only enforces this for the last line close/fix reference, since it is not uncommon to include references such as "Related to `old issue`" in the body.

bears/vcs/git/GitCommitBear.py
```diff
@@ -269,7 +336,9 @@ def check_issue_reference(self, body,
...
+                              remote: str=None):
```
virresh: We could possibly be able to do away with this particular setting, and use a brute-force method of checking over all remotes for an open issue with the issue number,
Coming up with an optimal algorithm to find the central repository could be a little troublesome though
(maybe you can think something on the lines of --set-upstream option, but I couldn't develop the idea, so I would recommend bruteforce)

RaiVaibhav:
yes I will rather say easy way :+1: ,

MacBox7:
I don't understand how the tokens being in CI scrips help our case.

RaiVaibhav:
>I don't understand how the tokens being in CI scrips help our case.

Just to solve the issue of rate limiting problem during running the `CI` ( only if needed )

MacBox7:
I think you misunderstand. What will happen when this bear will be run locally by the user. One way to approach this is that we can create our token and use it to query all issue state. But should this be done?

RaiVaibhav:
>So using it would require that we expect our user to have an environment variable storing his token. Is this a recommended way?

direclty import the token from environment (Use IGitt or PyGithub or request anything)
If you want to use the request , import the token from environement pass that to `Authentication` header to request , strore that [json](https://api.github.com/repos/coala/coala-bears/issues/2416)  and check the value of `state key` in json data( consider a case it is open)  and also check if there is something in `pull_request key` if no then this means issue is open and it is not a PR.
Max 5-6 line of code and accordingly create a test.

RaiVaibhav:
If you want you can first `ping` user to `export their token into their environment`

bears/vcs/git/GitCommitBear.py
jayvdb: This setting must be removed.
I've said that many times.
Different people have different remote names for the same thing, and so `remote` can not be a setting which applies to everyone.
We can use `git remote` to get information we need, but not remote names.  if there are multiple remote URLs pointing at different repos, the bear needs to sanely figure out the right one.  Or *fail* because the users remotes are too strange.  Failing is better than allowing a remote name to be in the settings and forcing everyone to use the same remote names (or fail because they choose to have a different setup)

bears/vcs/git/GitCommitBear.py
jayvdb: All of this remote stuff, and issue fetching, is completely off base.

Anything to do with remotes must be removed.  Even guessing what `#1` is not necessarily in scope of this issue.  We have setting `body_close_issue_full_url` which could be required to be enabled for your feature to work. 

As you have now learnt, you can not use the GitLab API for issues.  How are you going to check the URL? :P

bears/vcs/git/GitCommitBear.py
```diff
@@ -102,6 +105,70 @@ def get_host_from_remotes():
...
+    @staticmethod
+    def get_owner_from_url(url):
+        return urlparse(url)[2].split('/')[1]
+
+    @staticmethod
+    def get_repository_from_url(url):
+        return urlparse(url)[2].split('/')[2]
+
+    @staticmethod
+    def get_owner_from_remote(remote):
```
jayvdb: Surely there are libraries that do all of this? :P

We've been talking about libraries for this in the community and coala-bears rooms.

MacBox7: Yes, I am very well aware of this issue. I didn't push new changes since there are no means of checking GitLab repo without auth token as we have discussed [here](https://gitlab.com/gitmate/open-source/IGitt/issues/114). I think this issue should be blocked until environment variables or some other mechanism to read user-specific settings gets introduced in coala.

Issue #11, opened 2016-02-19
SwiftLint for Swift


nishant-mor:
we can use https://github.com/realm/SwiftLint

Adrianzatreanu:
You want a linter bear for swift?

sils1297:
Yeah why not?

Adrianzatreanu:
It wasnt a matter of why, it was a matter of : "is this what you asked?" as this issue was described pretty vague, and i wanted to ask to be sure :)

Adrianzatreanu:
Well id be glad to help if you'd let me then :)

sudheesh001:
Assigned it to you @Adrianzatreanu

Adrianzatreanu:
I was planning on writing this on linux. I didnt expect it not to be able to be installed on it ;p seems all the install methods are for mac os only.. my bad ;( if this is only for mac osx users to fix, please unassign me and sorry :)

sils1297:
this is not doable on linux? Do we want it?

AbdealiJK:
As @sils1297 mentioned in https://github.com/coala-analyzer/coala-bears/issues/336 we could use https://github.com/sleekbyte/tailor if swiftlint is complicated.

AbdealiJK:
@SanketDG why is this blocked ? Please give a reason if you add that label

SanketDG:
@AbdealiJK sorry about that, but we had a discussion on https://github.com/coala-analyzer/coala-bears/pull/177. Looks like `swift` is not ready for `linux`. `tailor` uses java to parse code, so I think that can be used.

sils1297:
@arafsheikh time to look into this? Having a linter for swift using tailor would be neat and easy.

This is not blocked as tailor can be used.

arafsheikh:
I'll look into it first thing tomorrow morning.

PR #2466, opened 2018-05-05
GitCommitBear: Mandate Valid Issue reference

virresh:
Require the issue reference to be valid in accordance with the hoster.
For e.g disallow full issue reference for bitbucket

Closes https://github.com/coala/coala-bears/issues/1134

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


bears/vcs/git/GitCommitBear.py
```diff
@@ -34,14 +34,21 @@ class GitCommitBear(GlobalBear):
             'issue': r'(?:\w+/\w+)?#(\S+)',
             'full issue': r'https?://gitlab\S+/issues/(\S+)',
         },
+        'bitbucket': {
+            'issue': r'(?:\w+/\w+)?#(\S+)',
```
jayvdb: I think the `(?:\w+/\w+)?` here is not supported by bitbucket??

virresh: I think bitbucket allows an issue to be referred to using 
```close #11```
or some word in between like ```close issue #11``` or ```close bug #11``` or ```close ticket #11```
So I am not sure how to go about that, I thought the above regex is for scenarios where it matches a word before issue number (possibly a link to the repo), so maybe for bitbucket we can use it to serve to support these words ?

Please correct me, I am very bad at regexe .

**Comment by virresh** [comment]
I have removed the leading link match, although it doesn't support the ```close issue #11``` kind of style in bitbucket,
I tried adding the matching ```r' ?(?:bug|issue|ticket)?``` to SUPPORTED_HOST_KEYWORD, but inside the check_issue_reference it always ends up picking ```issue``` as the issue reference instead of ```#11``` in case of closes and resolves, but it works for fixes ( [example](https://regex101.com/r/uA2xR5/74) )
And making further capture groups breaks the logic for regexes below

**Comment by virresh** [comment]
Used an ugly hack:
The SUPPORTED_HOST_KEYWORD_REGEX is no longer a proper regex for bitbucket, instead it needs to be wrapped in a enclosing non-capturing clause.

This will work in the current arrangement since we replace SUPPORTED_HOST_KEYWORD_REGEX using a .format() into a template such as ```(?:{0})```

File: bears/vcs/git/GitCommitBear.py [hunk 2]
```diff
@@ -34,14 +34,22 @@ class GitCommitBear(GlobalBear):
...
                    r'|[Rr]esolve[sd]?'
                    r'|[Ff]ix(?:e[sd])?'),
         'gitlab': (r'[Cc]los(?:e[sd]?|ing)'
                    r'|[Rr]esolv(?:e[sd]?|ing)'
-                   r'|[Ff]ix(?:e[sd]|ing)?')
+                   r'|[Ff]ix(?:e[sd]|ing)?'),
+        'bitbucket': (r'[Cc]los(?:e[sd]?|ing)'
+                      r'|[Rr]esolv(?:e[sd]?|ing)'
+                      r'|[Ff]ix(?:e[sd]|ing)?'
+                      r')(?: ?(?:bug|issue|ticket)?'),
```
ksdme: Replace the whitespace  with \s, this is because any kind of whitespace is supported during regex joining.

ksdme: The whitespace shouldn't be optional either if something is to proceed after the reference type. The regex should rather group the space and descriptor and make it optional (https://regex101.com/r/gJ9oKe/1/).

virresh: Thanks @ksdme 

But I am not sure if we want to match with a newline or carriage return (since a ```\n``` or ```\r``` do not result in closing of the issue

Though yes thanks for pointing this out, multiple spaces and tabs do need to be accomodated :sweat_smile:

virresh: Updated @ksdme 
Please review :)

bears/vcs/git/GitCommitBear.py
```diff
@@ -34,14 +34,22 @@ class GitCommitBear(GlobalBear):
...
                    r'|[Rr]esolve[sd]?'
                    r'|[Ff]ix(?:e[sd])?'),
         'gitlab': (r'[Cc]los(?:e[sd]?|ing)'
                    r'|[Rr]esolv(?:e[sd]?|ing)'
-                   r'|[Ff]ix(?:e[sd]|ing)?')
+                   r'|[Ff]ix(?:e[sd]|ing)?'),
+        'bitbucket': (r'[Cc]los(?:e[sd]?|ing)'
+                      r'|[Rr]esolv(?:e[sd]?|ing)'
+                      r'|[Ff]ix(?:e[sd]|ing)?'
+                      r')(?:(?:[ \t]*(?:bug|issue|ticket)?)?'),
```
ksdme: The shortlog should say something like 'Support bitbucket type issue references' or something along those lines since you are specifically aiming to improve those aspects of the bear. The commit message could also be improved by removing the example.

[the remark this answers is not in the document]
virresh: I tried it for [this](https://bitbucket.org/virresh/issueresolvinggit/commits/3fd3a151d17fc77d93fce59b9a776ea48ccf2e2b) and [this](https://bitbucket.org/virresh/issueresolvinggit/commits/055683134fb75ad1dbbb0ea75a6c05de94e6006a)
And no whitespace still works for closing the issue, 

So should I go ahead and check that for one or more spaces then ?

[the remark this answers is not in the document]
ksdme: If you match for zero or more occurrences of ```[ \t]``` then you would essentially be depending upon https://github.com/virresh/coala-bears/blob/4ff7382a06f48267a290fba41196866052b92226/bears/vcs/git/GitCommitBear.py#L38 to find the issue with the the reference. I don't think this should be the case.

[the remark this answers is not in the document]
virresh: So if I get it correct,
we should not allow ```fixesbug#11```, but we should allow ```fixes bug#11``` ?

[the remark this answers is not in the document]
jayvdb: Use https://confluence.atlassian.com/bitbucket/resolve-issues-automatically-when-users-push-code-221451126.html as your guide.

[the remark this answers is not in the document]
virresh: Yep
I read the guidelines there, but they don't seem to emphasize on the whitespaces,
As per the guidelines even ```fixesbug#11```, ```fixesbug #11```, ```fixes bug#11```, all three are valid.

So I accomodated all varieties of whitespaces. Not sure if I should limit it to only ```fixes bug #11```

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -298,16 +298,106 @@ def test_check_issue_reference(self):
...
+                        'Resolving    bug #1112')
+        self.assertEqual(self.run_uut(
+                             body_close_issue=True), [])
+
+        self.git_commit('Shortlog\n\n'
+                        'First line, blablablablablabla.\n'
+                        'Another line, blablablablablabla.\n'
+                        'Closesticket #1112')
+        self.assertEqual(self.run_uut(
+                             body_close_issue=True), [])
```
jayvdb: this should be an error, yes?

virresh: [This](https://bitbucket.org/virresh/issueresolvinggit/commits/3fd3a151d17fc77d93fce59b9a776ea48ccf2e2b) works as per guidelines :sweat_smile:

bears/vcs/git/GitCommitBear.py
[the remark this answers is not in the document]
ksdme: Although the suggested guidelines do not explicitly state anything about whitespaces it does include a space while defining the format as ```<command> <issue id>``` and advertise this behavior in all the examples, also for the case of readability we should take it as an implicit guideline.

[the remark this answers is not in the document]
virresh: Oh
Cool
Thanks @ksdme

tests/vcs/git/GitCommitBearTest.py
ksdme: I do not think anyone would have a reason to write messages in such a skewed way. Maybe we should have an option for this bear indicating whether to stick to core guidelines or readability guidelines.

bears/vcs/git/GitCommitBear.py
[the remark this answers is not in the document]
virresh: Opened https://bitbucket.org/site/master/issues/16240/issue-referencing-on-bitbucket-cloud

And off to support ```fixes bug#11``` and ```fixes bug #11```

bears/vcs/git/GitCommitBear.py
```diff
@@ -34,14 +34,22 @@ class GitCommitBear(GlobalBear):
             'issue': r'(?:\w+/\w+)?#(\S+)',
             'full issue': r'https?://gitlab\S+/issues/(\S+)',
         },
+        'bitbucket': {
+            'issue': r'#(\S+)',
```
ksdme: https://confluence.atlassian.com/sourcetreekb/link-to-bitbucket-issue-tracker-from-commits-296911608.html seems to indicates that the regex required to match an issue would rather be ```#(\d+)```.

bears/vcs/git/GitCommitBear.py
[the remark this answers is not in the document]
ksdme: I think it is very clear from https://confluence.atlassian.com/bitbucket/mark-up-comments-issues-and-commit-messages-321859781.html and https://confluence.atlassian.com/bitbucket/resolve-issues-automatically-when-users-push-code-221451126.html that we should only support the syntax ```<command> <issue id>``` irrespective of what works during commit actions. The command should also maintain readability and thus we should only support references written as ```fixes bug #11```.

bears/vcs/git/GitCommitBear.py
```diff
@@ -34,14 +34,22 @@ class GitCommitBear(GlobalBear):
...
-        'gitlab': (r'[Cc]los(?:e[sd]?|ing)'
+                   r'|[Ff]ix(?:e[sd])?)\s+'),
+        'gitlab': (r'(?:[Cc]los(?:e[sd]?|ing)'
                    r'|[Rr]esolv(?:e[sd]?|ing)'
-                   r'|[Ff]ix(?:e[sd]|ing)?')
+                   r'|[Ff]ix(?:e[sd]|ing)?)\s+'),
+        'bitbucket': (r'(?:(?:[Cc]los(?:e[sd]?|ing)'
+                      r'|[Rr]esolv(?:e[sd]?|ing)'
+                      r'|[Ff]ix(?:e[sd]|ing))'
+                      r'(?:\s+(?:(?:bug|issue|ticket)[ \t]*)?))'),
```
ksdme: The ```\s+``` in ```r'(?:\s+(?:(?:bug|issue|ticket)[ \t]*)?))'),``` should rather be ```[ \t]+``` as an individual reference needs to be a single line.

bears/vcs/git/GitCommitBear.py
```diff
@@ -34,14 +34,22 @@ class GitCommitBear(GlobalBear):
...
             'full issue': r'https?://gitlab\S+/issues/(\S+)',
         },
+        'bitbucket': {
+            'issue': r'#(\S+)',
+            'full issue': None,
+        },
     }
     SUPPORTED_HOST_KEYWORD_REGEX = {
-        'github': (r'[Cc]lose[sd]?'
+        'github': (r'(?:[Cc]lose[sd]?'
```
ksdme: I do not think changes made to the regex of hosts other than BitBucket are either necessary or in the scope of this issue.

bears/vcs/git/GitCommitBear.py
```diff
@@ -337,7 +350,7 @@ def check_issue_reference(self, body,
 
         concat_regex = '|'.join(kw for kw in self.CONCATENATION_KEYWORDS)
         compiled_joint_regex = re.compile(
-            r'(?:{0})\s+'           # match issue related keywords,
+            r'(?:{0})'              # match issue related keywords,
```
ksdme: This change should be reverted too.

bears/vcs/git/GitCommitBear.py
virresh: Oh
Okay
Did that to support ```fixes bug#166``` but not ```fixesbug #11```
Will revert :sweat_smile: 

Thanks :D

virresh:
Supported only the sane version of ```<command> <issue id>```

Please review @ksdme  :)

virresh:
Thanks @ksdme 
The shortlog was exceeding character limit, so replaced support with add :sweat_smile:

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -298,16 +298,140 @@ def test_check_issue_reference(self):
 
         # Adding BitBucket remote for testing
         self.run_git_command('remote', 'add', 'test',
-                             'https://bitbucket.com/user/repo.git')
+                             'https://user@bitbucket.org/user/gittest.git')
```
ksdme: Is there a specific reason why the test repo url needs to be changed to this?

virresh: Oh
I did that to keep the url format consistent with the one suggested by bitbucket on the clone option for a repository

Although the above also redirects fine 

Will change that :sweat_smile:

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -300,14 +300,138 @@ def test_check_issue_reference(self):
         self.run_git_command('remote', 'add', 'test',
                              'https://bitbucket.com/user/repo.git')
 
-        # Unsupported Host - Bitbucket
+        # Unsupported Host and Issue reference combination - Bitbucket
```
ksdme: The host is now supported.

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -300,14 +300,137 @@ def test_check_issue_reference(self):
...
+                             body_close_issue=True,
+                             body_close_issue_full_url=True),
+                         ['Host bitbucket does not support full issue '
+                          'reference.'])
+
+        # Adding BitBucket's ssh remote for testing
+        self.run_git_command('remote', 'set-url', 'test',
+                             'git@bitbucket.org:user/repo.git')
+
+        # Unsupported Host and Issue reference combination - Bitbucket
```
ksdme: You left out this instance of the comment.

PR #2491, opened 2018-05-19
GitCommitBear.py: Modify shortlog tests

### LINKED ISSUE #2471: GitCommitBear: Check for WIP: shortlog
Currently the bear only yields a result in case the shortlog starts with
"[wip] "
It should also check for "WIP: "

rohanrk:
GitCommitBear's imperative check modified shortlog param
so if shortlog contained 'wip:', the wip check would not
catch it. This change checks for wip before checking
imperatives. It also adds a new variable instead of
modifying the shortlog param.

Fixes #2471


palash25:
@rohanrk why two commits? Please squash them into one 

The code changes and the accompanying test changes are always considered as one change

bears/vcs/CommitBear.py
```diff
@@ -203,23 +203,24 @@ def check_shortlog(self, shortlog,
...
+                    'This commit seems to be marked as work in progress and '
+                    'should not be used in production. Treat carefully.')
+
         if shortlog_imperative_check:
             colon_pos = shortlog.find(':')
-            shortlog = (shortlog[colon_pos + 1:]
-                        if colon_pos != -1
-                        else shortlog)
-            has_flaws = self.check_imperative(shortlog)
+            shortlog_imp = (shortlog[colon_pos + 1:]
```
jayvdb: you dont need to change the name of this variable.

If anyone wants to see the real complete shortlog, theor code should be placed above this.

Or they can create a new variable.

jayvdb:
It is correct to just change the order of the checks, so that the most appropriate message is shown.  The shortlog could be a bit clearer that this is an order change.

jayvdb:
ping @rohanrk , could you rebase this and fix the pep8 errors

PR #2543, opened 2018-06-20
VCSCommitMetadataBear: New meta bear for analyzing commits

kriti21:
This adds a new meta bear that returns HiddenResult
of commit analysis for other bears to use.

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/vcs/VCSCommitMetadataBear.py
+++ b/bears/vcs/VCSCommitMetadataBear.py
@@ -1,0 +1,112 @@
+import abc
+
+from aenum import Flag
+
+from coalib.bears.GlobalBear import GlobalBear
+from coalib.results.HiddenResult import HiddenResult
+from coala_utils.decorators import (enforce_signature, generate_ordering,
+                                    generate_repr)
+
+
+class COMMIT_TYPE(Flag):
+    simple_commit = 0
+    merge_commit = 1
+
+
+@generate_repr(('id', hex),
+               'origin',
+               'raw_commit_message',
+               'commit_sha',
+               'parent_commits',
+               'commit_type',
+               'modified_files',
+               'added_files',
+               'deleted_files')
+@generate_ordering('raw_commit_message',
+                   'commit_sha',
+                   'parent_commits',
+                   'commit_type',
+                   'modified_files',
+                   'added_files',
+                   'deleted_files',
+                   'severity',
+                   'confidence',
+                   'origin',
+                   'message_base',
+                   'message_arguments',
+                   'aspect',
+                   'additional_info',
+                   'debug_msg')
+class CommitResult(HiddenResult):
+
+    @enforce_signature
+    def __init__(self, origin,
+                 raw_commit_message: str,
+                 commit_sha: str,
+                 parent_commits: list,
+                 commit_type: COMMIT_TYPE,
+                 modified_files: list,
+                 added_files: list,
+                 deleted_files: list):
+
+        super().__init__(origin, '')
+        self.raw_commit_message = raw_commit_message
+        self.commit_sha = commit_sha
+        self.parent_commits = parent_commits
+        self.commit_type = commit_type
+        self.modified_files = modified_files
+        self.added_files = added_files
+        self.deleted_files = deleted_files
+
+
+class VCSCommitMetadataBear(GlobalBear):
+    __metaclass__ = abc.ABCMeta
+    LANGUAGES = {'Git'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+
+    @abc.abstractmethod
+    def get_head_commit_sha(self):
+        """
+        Return the SHA of the head commit.
+        """
+
+    @abc.abstractmethod
+    def analyze_commit(self, head_commit_sha):
+        """
+        Check the current commit at HEAD.
+
+        Return the commit information such as commit message,
+        type of commit, commit SHA, modified, added and
+        delete files by the commit.
+
+        :param head_commit_sha: The HEAD commit SHA
+        :return:                A tuple comprising of commit
+                                message, commit sha, list of
+                                parent commits, type of commit
+                                viz. simple or merge
+                                commit and lists of modified,
+                                added and deleted files.
+        """
+
+    def run(self, **kwargs):
+        """
+        This bear returns information about the HEAD commit
+        as HiddenResult which can be used for inspection by
+        other bears.
+        """
+        try:
+            head_commit_sha = self.get_head_commit_sha()
+
+        except RuntimeError:
+            return
+
+        for (raw_commit_message, commit_sha, parent_commits,
+             commit_type, modified_files, added_files,
+             deleted_files) in self.analyze_commit(
+                head_commit_sha):
+
+            yield CommitResult(self, raw_commit_message, commit_sha,
+                               parent_commits, commit_type, modified_files,
+                               added_files, deleted_files)
--- a/bears/vcs/git/GitCommitMetadataBear.py
+++ b/bears/vcs/git/GitCommitMetadataBear.py
@@ -1,0 +1,65 @@
+import os
+import shutil
+
+from bears.vcs.VCSCommitMetadataBear import VCSCommitMetadataBear, COMMIT_TYPE
+from coala_utils.ContextManagers import change_directory
+from coalib.misc.Shell import run_shell_command
+
+
+class GitCommitMetadataBear(VCSCommitMetadataBear):
+    LANGUAGES = {'Git'}
+
+    @classmethod
+    def check_prerequisites(cls):
+        if shutil.which('git') is None:
+            return 'git is not installed.'
+        else:
+            return True
+
+    def get_head_commit_sha(self):
+        with change_directory(self.get_config_dir() or os.getcwd()):
+            (stdout, stderr) = run_shell_command('git rev-parse HEAD')
+
+            if stderr:
+                vcs_name = list(self.LANGUAGES)[0].lower()+':'
+                self.err(vcs_name, repr(stderr))
+                raise RuntimeError('The directory is not a git repository.')
+
+            head_commit_sha = stdout.strip('\n')
+            return head_commit_sha
+
+    def analyze_commit(self, head_commit_sha):
+        commit_type = COMMIT_TYPE.simple_commit
+
+        command = 'git log -1 --pretty=%B ' + head_commit_sha
+        head_commit = run_shell_command(command)[0]
+
+        get_parent_commits = 'git log --pretty=%P -n 1 ' + head_commit_sha
+        all_parent_commits = run_shell_command(get_parent_commits)[0]
+        parent_commits_list = all_parent_commits.split(' ')
+
+        if len(parent_commits_list) >= 2:
+            commit_type |= COMMIT_TYPE.merge_commit
+
+        get_all_committed_files = ('git show --pretty="" --name-status ' +
+                                   head_commit_sha)
+        all_committed_files = run_shell_command(get_all_committed_files)[0]
+        all_committed_files = all_committed_files.split('\n')
+
+        modified_files_list = []
+        added_files_list = []
+        deleted_files_list = []
+
+        for line in all_committed_files:
+            pos = line.find('\t')
+            change = line[:pos]
+            if change == 'M':
+                modified_files_list.append(line[pos+1:])
+            elif change == 'A':
+                added_files_list.append(line[pos+1:])
+            elif change == 'D':
+                deleted_files_list.append(line[pos+1:])
+
+        yield (head_commit, head_commit_sha, parent_commits_list,
+               commit_type, modified_files_list, added_files_list,
+               deleted_files_list)
--- a/tests/vcs/VCSCommitMetadataBearTest.py
+++ b/tests/vcs/VCSCommitMetadataBearTest.py
@@ -1,0 +1,112 @@
+import os
+import platform
+import shutil
+import unittest
+import unittest.mock
+from pathlib import Path
+from queue import Queue
+from tempfile import mkdtemp
+
+from bears.vcs.VCSCommitMetadataBear import (VCSCommitMetadataBear,
+                                             CommitResult, COMMIT_TYPE)
+from bears.vcs.git.GitCommitMetadataBear import GitCommitMetadataBear
+from coalib.misc.Shell import run_shell_command
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+
+
+class FakeCommitBear(VCSCommitMetadataBear):
+    @classmethod
+    def check_prerequisites(cls):
+        return True
+
+    def get_head_commit_sha(self):
+        raise RuntimeError('RuntimeError from test')
+
+
+@generate_skip_decorator(VCSCommitMetadataBear)
+class InvalidCommitResultTest(unittest.TestCase):
+
+    def run_uut(self, *args, **kwargs):
+        """
+        Runs the unit-under-test (via `self.uut.run()`) and collects the
+        messages of the yielded results as a list.
+
+        :param args:   Positional arguments to forward to the run function.
+        :param kwargs: Keyword arguments to forward to the run function.
+        :return:       A list of result values that give information related
+                       to head commit.
+        """
+        return list(result.message for result in self.uut.run(*args, **kwargs))
+
+    def setUp(self):
+        self.msg_queue = Queue()
+        self.section = Section('')
+        self.uut = FakeCommitBear(None, self.section, self.msg_queue)
+
+        self._old_cwd = os.getcwd()
+        self.tempdir = mkdtemp()
+        os.chdir(self.tempdir)
+
+    def test_check_prerequisites(self):
+        _shutil_which = shutil.which
+        try:
+            shutil.which = lambda *args, **kwargs: None
+            self.assertEqual(GitCommitMetadataBear.check_prerequisites(),
+                             'git is not installed.')
+
+            shutil.which = lambda *args, **kwargs: 'path/to/git'
+            self.assertTrue(GitCommitMetadataBear.check_prerequisites())
+        finally:
+            shutil.which = _shutil_which
+
+    @staticmethod
+    def _windows_rmtree_remove_readonly(func, path, excinfo):
+        os.chmod(path, stat.S_IWRITE)
+        func(path)
+
+    def tearDown(self):
+        os.chdir(self._old_cwd)
+        if platform.system() == 'Windows':
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.tempdir, onerror=onerror)
+
+    def test_get_head_commit_sha(self):
+        Path('testfile.txt').touch()
+        run_shell_command('git add testfile.txt')
+        run_shell_command('git commit -m "Add testfile"')
+        self.assertEqual(self.run_uut(), [])
+
+
+class CommitResultTest(unittest.TestCase):
+    def setUp(self):
+        self.raw_commit_message = 'raw_commit_message'
+        self.commit_sha = 'commit_sha'
+        self.parent_commits = ['parent_commits']
+        self.modified_files = ['modified_files']
+        self.added_files = ['added_files']
+        self.deleted_files = ['deleted_files']
+
+    def test_commitresult_object_repr(self):
+        repr_result = repr(CommitResult(VCSCommitMetadataBear,
+                                        self.raw_commit_message,
+                                        self.commit_sha,
+                                        self.parent_commits,
+                                        COMMIT_TYPE.simple_commit,
+                                        self.modified_files,
+                                        self.added_files,
+                                        self.deleted_files,))
+
+        repr_regex = (
+            r'<CommitResult object\(id=.+, origin=\'bearclass\', '
+            r'raw_commit_message=\'.+\', '
+            r'commit_sha=\'.+\', '
+            r'parent_commits=\[.+\], '
+            r'commit_type=<COMMIT_TYPE.simple_commit: 0>, '
+            r'modified_files=\[.+\], '
+            r'added_files=\[.+\], '
+            r'deleted_files=\[.+\]\) at .+>'
+            )
+        self.assertRegex(repr_result, repr_regex)
--- a/tests/vcs/git/GitCommitMetadataBearTest.py
+++ b/tests/vcs/git/GitCommitMetadataBearTest.py
@@ -1,0 +1,180 @@
+import os
+import platform
+import shutil
+import stat
+import unittest
+import unittest.mock
+from pathlib import Path
+from queue import Queue
+from tempfile import mkdtemp
+
+from bears.vcs.VCSCommitMetadataBear import COMMIT_TYPE
+from bears.vcs.git.GitCommitMetadataBear import GitCommitMetadataBear
+from coalib.misc.Shell import run_shell_command
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+
+
+@generate_skip_decorator(GitCommitMetadataBear)
+class GetHeadCommitTest(unittest.TestCase):
+
+    def run_uut(self, *args, **kwargs):
+        """
+        Runs the unit-under-test (via `self.uut.run()`) and collects the
+        messages of the yielded results as a list.
+
+        :param args:   Positional arguments to forward to the run function.
+        :param kwargs: Keyword arguments to forward to the run function.
+        :return:       A list of result values that give information related
+                       to head commit.
+        """
+        return list(result.message for result in self.uut.run(*args, **kwargs))
+
+    def setUp(self):
+        self.msg_queue = Queue()
+        self.section = Section('')
+        self.uut = GitCommitMetadataBear(None, self.section, self.msg_queue)
+
+        self._old_cwd = os.getcwd()
+        self.tempdir = mkdtemp()
+        os.chdir(self.tempdir)
+
+    @staticmethod
+    def _windows_rmtree_remove_readonly(func, path, excinfo):
+        os.chmod(path, stat.S_IWRITE)
+        func(path)
+
+    def tearDown(self):
+        os.chdir(self._old_cwd)
+        if platform.system() == 'Windows':
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.tempdir, onerror=onerror)
+
+    def test_get_head_commit_sha(self):
+        exception_msg = r'The directory is not a git repository.'
+        Path('testfile.txt').touch()
+        run_shell_command('git add testfile.txt')
+        run_shell_command('git commit -m "Add testfile"')
+        with self.assertRaisesRegex(RuntimeError, exception_msg):
+            self.uut.get_head_commit_sha()
+
+
+@generate_skip_decorator(GitCommitMetadataBear)
+class AnalyzeGitCommitTest(unittest.TestCase):
+
+    def run_uut(self, *args, **kwargs):
+        """
+        Runs the unit-under-test (via `self.uut.run()`) and collects the
+        messages of the yielded results as a list.
+
+        :param args:   Positional arguments to forward to the run function.
+        :param kwargs: Keyword arguments to forward to the run function.
+        :return:       A list of result values that give information related
+                       to head commit.
+        """
+        return list((result.raw_commit_message,
+                     result.commit_sha,
+                     result.parent_commits,
+                     result.commit_type,
+                     result.added_files,
+                     result.modified_files,
+                     result.deleted_files) for result in self.uut.run(
+                     *args, **kwargs))
+
+    def setUp(self):
+        self.msg_queue = Queue()
+        self.section = Section('')
+        self.uut = GitCommitMetadataBear(None, self.section, self.msg_queue)
+
+        self._old_cwd = os.getcwd()
+        self.gitdir = mkdtemp()
+        os.chdir(self.gitdir)
+        run_shell_command('git init')
+        run_shell_command('git config -- user.email coala@coala.io')
+        run_shell_command('git config -- user.name coala')
+
+    @staticmethod
+    def _windows_rmtree_remove_readonly(func, path, excinfo):
+        os.chmod(path, stat.S_IWRITE)
+        func(path)
+
+    def tearDown(self):
+        os.chdir(self._old_cwd)
+        if platform.system() == 'Windows':
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.gitdir, onerror=onerror)
+
+    def test_analyze_simple_git_commit(self):
+        run_shell_command('git checkout -b testbranch master')
+        Path('testfile1.txt').touch()
+        run_shell_command('git add testfile1.txt')
+        run_shell_command('git commit -m "Add testfile1"')
+        test_sha1 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        get_parents = 'git log --pretty=%P -n 1 ' + test_sha1
+        parents = run_shell_command(get_parents)[0].split(' ')
+        self.assertEqual(self.run_uut(),
+                         [('Add testfile1\n\n', test_sha1, parents,
+                           COMMIT_TYPE.simple_commit, ['testfile1.txt'],
+                           [], [])])
+
+        Path('testfile2.txt').touch()
+        run_shell_command('git add testfile2.txt')
+        run_shell_command('git commit -m "Add testfile2 [ci skip]"')
+        test_sha2 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        get_parents = 'git log --pretty=%P -n 1 ' + test_sha2
+        parents = run_shell_command(get_parents)[0].split(' ')
+        self.assertEqual(self.run_uut(),
+                         [('Add testfile2 [ci skip]\n\n', test_sha2,
+                           parents, COMMIT_TYPE.simple_commit,
+                           ['testfile2.txt'], [], [])])
+
+        with open('testfile2.txt', 'w') as f:
+            f.write('Some text')
+        run_shell_command('git add testfile2.txt')
+        run_shell_command('git commit -m "Modify testfile2"')
+        test_sha3 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        get_parents = 'git log --pretty=%P -n 1 ' + test_sha3
+        parents = run_shell_command(get_parents)[0].split(' ')
+        self.assertEqual(self.run_uut(),
+                         [('Modify testfile2\n\n', test_sha3, parents,
+                           COMMIT_TYPE.simple_commit, [], ['testfile2.txt'],
+                           [])])
+
+    def test_analyze_git_revert_commit(self):
+        Path('testfile3.txt').touch()
+        run_shell_command('git add testfile3.txt')
+        run_shell_command('git commit -m "another commit [skip ci]"')
+        test_sha3 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        run_shell_command('git revert HEAD --no-edit')
+        test_sha4 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        get_parents = 'git log --pretty=%P -n 1 ' + test_sha4
+        parents = run_shell_command(get_parents)[0].split(' ')
+        test_raw_commit_msg = ('Revert "another commit [skip ci]"\n\n'
+                               'This reverts commit %s.\n\n' % (test_sha3))
+        self.assertEqual(
+            self.run_uut(),
+            [(test_raw_commit_msg, test_sha4, parents,
+              COMMIT_TYPE.simple_commit, [], [], ['testfile3.txt'])])
+
+    def test_analyze_git_merge_commit(self):
+        Path('testfile4.txt').touch()
+        run_shell_command('git add testfile4.txt')
+        run_shell_command('git commit -m "Commit in master branch"')
+        run_shell_command('git checkout master')
+        run_shell_command('git checkout -b new-feature master')
+        Path('testfile5.txt').touch()
+        run_shell_command('git add testfile5.txt')
+        run_shell_command('git commit -m "Commit in feature branch"')
+        run_shell_command('git checkout master')
+        run_shell_command('git merge --no-ff new-feature')
+        test_raw_commit_msg = "Merge branch 'new-feature'\n\n"
+        test_sha5 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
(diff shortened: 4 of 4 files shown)


bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,81 @@
...
+        commit_info['commit_sha'] = commit_sha
+        commit_info['commit_type'] = commit_type
+        commit_info['modified_files'] = modified_files_list
+        commit_info['added_files'] = added_files_list
+        commit_info['deleted_files'] = deleted_files_list
+
+        return str(commit_info)
+
+    def run(self):
+        stdout, _ = self.get_head_commit()
```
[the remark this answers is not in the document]
refeed: I also think the `stdout` is more to a `head_commit_message`, use that name instead..

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,81 @@
+import abc
+
+from coalib.bears.GlobalBear import GlobalBear
+from coalib.results.HiddenResult import HiddenResult
+from coalib.results.Result import Result
+from coalib.misc.Shell import run_shell_command
+
+
+class CommitResult(HiddenResult):
+    def __init__(self, origin, commit_info, message):
```
refeed: Add some type annotation and `@enforce_signature` see `URLHeadResult`

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,81 @@
...
+        get_parent_commits = 'git log --pretty=%P -n 1 ' + commit_sha
+        parent_commits, _ = run_shell_command(get_parent_commits)
+        parent_commits_list = parent_commits.split('\n')
+
+        if len(parent_commits_list) > 2:
+            commit_type.append('merge_commit')
+
+        get_all_committed_files = ('git show --pretty="" --name-status ' +
+                                   commit_sha)
+        all_committed_files, _ = run_shell_command(get_all_committed_files)
```
refeed: Remove `_`

bears/vcs/git/GitCommitBear.py
```diff
@@ -2,11 +2,12 @@
 import shutil
 
 from bears.vcs.CommitBear import _CommitBear
+from bears.vcs.VCSCommitBear import VCSCommitBear
 from coala_utils.ContextManagers import change_directory
 from coalib.misc.Shell import run_shell_command
 
 
-class GitCommitBear(_CommitBear):
+class GitCommitBear(_CommitBear, VCSCommitBear):
```
virresh: You need to add 
```
        try:
            super(_CommitBear,self).run()
        except NotImplementedError as e:
            pass
```
within the run method of _CommitBear so that it can call VCSCommitBear's run method as well

bears/vcs/VCSCommitBear.py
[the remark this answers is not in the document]
kriti21: `get_head_commit` returns two values, that's why `_` is needed here.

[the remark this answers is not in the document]
refeed: Ah yup, actually I know about that. I just wasn't agree with  the "_" style. After a while searching in the internet about this style, it turns out this style is [fine](https://softwareengineering.stackexchange.com/questions/139582/which-style-to-use-for-unused-return-parameters-in-a-python-function-call)
I never used this style tbh, since I usually use `_` to access `gettext` in my another community's codebase. But in this codebase, I think it's fine to use this style..

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,107 @@
...
+                   'message_base',
+                   'message_arguments',
+                   'aspect',
+                   'additional_info',
+                   'debug_msg')
+class CommitResult(HiddenResult):
+
+    @enforce_signature
+    def __init__(self, origin, commit_info):
+        Result.__init__(self, origin, 'HEAD commit information', commit_info)
```
refeed: Delete `commit_info` argument from this

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,107 @@
...
+                deleted_files_list.append(file[pos+1:])
+
+        commit_info.raw_commit_message = raw_commit_message
+        commit_info.commit_sha = commit_sha
+        commit_info.commit_type = commit_type
+        commit_info.modified_files = modified_files_list
+        commit_info.added_files = added_files_list
+        commit_info.deleted_files = deleted_files_list
+
+        return (commit_info)
```
refeed: The parentheses are rendundant

bears/vcs/VCSCommitBear.py
damngamerz: Delete the whole `Result` line inside `_init_`. I dont understand its existence. you are already passing `commit_info` as contents right :)

damngamerz: >Well, the reason of why Result.__init__ is there is because HiddenResult is a subclass of Result.
So we need to run the Result.__init__

This makes sense. Well keep it then.

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,179 @@
...
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.gitdir, onerror=onerror)
+
+    def test_analyze_git_commit(self):
+        run_shell_command('echo "abc" >> testfile1.txt ')
+        run_shell_command('git add testfile1.txt')
+        run_shell_command('git commit -m "Add testfile1"')
+        self.assertEqual(self.run_uut(), [])
```
refeed: and also, test the result by using `get_results()` (See `URLHeadBear` for the example) as we're not matching only the message here, we're also matching the contents of the `CommitResult.contents`

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,179 @@
...
+    def run_uut(self, *args, **kwargs):
+        """
+        Runs the unit-under-test (via `self.uut.run()`) and collects the
+        messages of the yielded results as a list.
+
+        :param args:   Positional arguments to forward to the run function.
+        :param kwargs: Keyword arguments to forward to the run function.
+        :return:       A list of the message strings.
+        """
+        return list(result.message for result in self.uut.run(*args, **kwargs))
```
refeed: Return all of the yielded values from `uut.run()`, not only the message.
`return list(self.uut.run(*args, **kwargs))

tests/vcs/VCSCommitBearTest.py
refeed: See https://github.com/coala/coala-bears/pull/2543/files#r197507946

tests/vcs/VCSCommitBearTest.py
kriti21: This also generates an output like 
`
[<CommitResult object(id=0xa4641b38563489790f5f40e78ef1ff3, origin='GitCommitBear', commit_info=(<class 'bears.vcs.VCSCommitBear.Commit_Info'>,), message='HEAD commit information') at 0x7f311e4a30f0>]
`

checking which is similar to https://github.com/coala/coala-bears/pull/2543/files/8484120bc877ab722a9fad4e3bc2f4e9706d43ca#diff-0704b7478b1a43f7ae7b1f2cafe0d8c3R170

Should I still do it this way ?
 Ideally we should check the contents of `<class 'bears.vcs.VCSCommitBear.Commit_Info'>`

refeed: Yes, that's a `CommitResult` object yielded by `VCSBear`. It's intended

You can make an instance of `CommitBear` that you're expecting by yourself in the unittest
```py
        self.assertEqual(self.uut.run(), [CommitResult(self.uut, commit_info)]
``

File: bears/vcs/VCSCommitBear.py [hunk 9]
```diff
@@ -0,0 +1,106 @@
...
+
+        commit_sha = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+
+        commit_type = []
+
+        head_commit = head_commit.strip('\n')
+
+        ciskip_regex = re.compile(r'\[ci skip\]|\[skip ci\]')
+        match = re.search(ciskip_regex, head_commit)
+        if match:
```
refeed: This can be simplified by just using one-line `re.match('(\[ci skip\]|\[skip ci\])', head_commit)`

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,106 @@
...
+from coala_utils.decorators import (enforce_signature, generate_ordering,
+                                    generate_repr)
+
+
+@generate_repr(('id', hex),
+               'origin',
+               'commit_info',
+               'message')
+@generate_ordering('commit_info',
+                   'contents',
```
refeed: Remove this

bears/vcs/git/GitCommitBear.py
```diff
@@ -25,3 +26,10 @@ def get_remotes():
     def get_head_commit(self):
         with change_directory(self.get_config_dir() or os.getcwd()):
             return run_shell_command('git log -1 --pretty=%B')
+
+    def run(self, *args, **kwargs):
+        for x in _CommitBear.run(self, *args, **kwargs):
+            yield x
+        for y in VCSCommitBear.run(self, *args, **kwargs):
```
refeed: Use `commit_result` for this variable name

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,181 @@
...
+    def tearDown(self):
+        os.chdir(self._old_cwd)
+        if platform.system() == 'Windows':
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.gitdir, onerror=onerror)
+
+    def test_analyze_git_commit(self):
+        run_shell_command('touch "abc" >> testfile1.txt ')
```
refeed: `touch testfile.txt` is enough to create a file

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,106 @@
...
+                   'origin',
+                   'message_base',
+                   'message_arguments',
+                   'aspect',
+                   'additional_info',
+                   'debug_msg')
+class CommitResult(HiddenResult):
+
+    @enforce_signature
+    def __init__(self, origin, commit_info):
```
refeed: Give `commit_info` argument a `namedtuple` annotation. `commit_info: namedtuple`

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,106 @@
...
+        get_all_committed_files = ('git show --pretty="" --name-status ' +
+                                   commit_sha)
+        all_committed_files = run_shell_command(get_all_committed_files)[0]
+        all_committed_files = all_committed_files.split('\n')
+
+        modified_files_list = []
+        added_files_list = []
+        deleted_files_list = []
+
+        for file in all_committed_files:
```
refeed: I think we might want to rename it to `line` instead, it's a bit ambiguous if it's `file` as it can also refer to IO objects

bears/vcs/git/GitCommitBear.py
```diff
@@ -25,3 +26,10 @@ def get_remotes():
...
+            if args is None or kwargs is None:
```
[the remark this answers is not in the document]
kriti21: I had to add this because unless this `if` condition is used, many assertion errors are generated in tests of `GitCommitBear`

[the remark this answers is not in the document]
refeed: Ok, then move the `if` above the iteration of `VCSCommitBear.run`, we'll followup this later after the redesign..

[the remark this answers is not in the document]
kriti21: ok

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,181 @@
...
+        self.run_git_command('commit', '-m', '"another commit [skip ci]"')
+        self.run_git_command('revert', 'HEAD', '--no-edit')
+        self.assertEqual(self.run_uut(), [])
+
+        run_shell_command('git checkout -b new-feature master')
+        run_shell_command('touch "some text" >> testfile4.txt ')
+        self.run_git_command('add', 'testfile4.txt')
+        self.run_git_command('commit', '-m', '"Commit in feature branch"')
+        run_shell_command('git checkout master')
+        run_shell_command('git merge new-feature')
```
virresh: Again wrt https://gitter.im/coala/2018-gsoc-all?at=5b2a6f0dad21887018cadd01
This is not even creating a merge commit, it is simply creating a normal commit which the `git merge` will simply fast forward and there will be *only one* parent commit.

kriti21:
@virresh Pushed some changes that I had made locally. I am still working on tests for `VCSCommitBear` but this will give you an idea about the `HiddenResult` being returned.

And yes I was confused about the code changes in `GitCommitBear` but since you said in your last review, I _will_ have to modify the tests in `GitCommitBeatTest`, I have done that too. Hope this helps.

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -87,62 +87,73 @@ def test_git_failure(self):
         git_error = self.msg_queue.get().message
         self.assertEqual(git_error[:4], 'git:')
 
-        self.assert_no_msgs()
```
virresh: Why remove this test ?, why not handle the case when the folder isn't a git repository in the VCSCommitBear itself ?

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,131 @@
...
+                   'aspect',
+                   'additional_info',
+                   'debug_msg')
+class CommitResult(HiddenResult):
+
+    @enforce_signature
+    def __init__(self, origin, raw_commit_message: str,
+                 commit_sha: str, commit_type: list, modified_files: list,
+                 added_files: list, deleted_files: list):
+        HiddenResult.__init__(self, origin, '')
```
virresh: Can you help me understand why did you split all of these, why a named_tuple being passed here wasn't sufficient and why are you setting the origin of the result to be `CommitResult` class rather than the origin parameter ?

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,181 @@
...
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.gitdir, onerror=onerror)
+
+    def test_analyze_git_commit(self):
+        run_shell_command('touch "abc" >> testfile1.txt ')
+        run_shell_command('git add testfile1.txt')
+        run_shell_command('git commit -m "Add testfile1"')
+        self.assertEqual(self.run_uut(), [])
```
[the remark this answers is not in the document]
kriti21: @virresh  This isn't correct. If you run the code now, you will see that the actual `CommitResult` is being checked against `[]` . My concern earlier was that these tests were not even executing but now they are. I am improving them so they would check only the output of `analyze_git_commit` here and not the entire result.

tests/vcs/git/GitCommitBearTest.py
kriti21: I have just removed the `assert_no_msgs()` check because it does get a message now which is `HEAD commit information`

kriti21:
@virresh I am working on 
> but I think the actual contents of the HiddenResult are not being checked, only whether a HiddenResult was generated or not is being checked

I have tried to reply to all your comments. Let me know of any other improvements you suggest

virresh:
@kriti21 
If nothing else, then the first point that I'd question is why does the assertion in GitCommitBear get a `'HEAD commit information'`
while you are checking for the CommitResult Object in the `VCSCommitBearTests`
I know that you are using the same function of self.run_uut which was meant to only return messages,
But checking if the hidden result was generated or not is of no use in the GitCommitBear, either don't check it at all, (using some approach as described in my previous comment) or check the whole commit object like what would be done in the VCSCommitResult (both of them should be done imo, i.e don't check for the hidden results in the existing tests, so they won't change, and make another class to check for the hidden results)

As for the main logic, I'd recommended previously as well to have a look over generators,
The for loop that I had written as an example was simply to illustrate the point why the tests weren't being executed, I expected you to read up a little more on that and use the appropriate things 
What would be apt would be using a `yield from` IMO

I hope I haven't got things wrong here and am clearer now :sweat_smile:

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,131 @@
...
+                   'origin',
+                   'message_base',
+                   'message_arguments',
+                   'aspect',
+                   'additional_info',
+                   'debug_msg')
+class CommitResult(HiddenResult):
+
+    @enforce_signature
+    def __init__(self, origin, raw_commit_message: str,
```
refeed: Style please,
```py
def __init__(self, origin,
             raw_commit_message: str,
             commit_sha: str,
             ...)
```

tests/vcs/git/GitCommitBearTest.py
```diff
@@ -87,62 +87,73 @@ def test_git_failure(self):
...
-
     def test_empty_message(self):
         self.git_commit('')
 
         self.assertEqual(self.run_uut(),
-                         ['HEAD commit has no message.'])
+                         ['HEAD commit information',
```
refeed: We probably might want to modify `self.run_uut()` to just ignore the `CommitResult` in this test, so we don't need to modify these testcases
And `GitCommitBear` is a bear that's used in CLI, which means it just shows the `Result` to the CLI, rather than `HiddenResult`

bears/vcs/VCSCommitBear.py
kriti21: re.match is used only to match the full-string, that's why re.search instead

refeed: yes, you can still shrink it into one line `if re.search('(\[ci skip\]|\[skip ci\])', head_commit):`

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,223 @@
...
+        messages of the yielded results as a list.
+
+        :param args:   Positional arguments to forward to the run function.
+        :param kwargs: Keyword arguments to forward to the run function.
+        :return:       A list of the message strings.
+        """
+        return list((result.message) for result in self.uut.run(*args,
+                                                                **kwargs))
+
+    def assert_no_msgs(self):
```
refeed: This method isn't used anywhere in the class, remove it

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        return list((result.raw_commit_message,
+                     result.commit_sha,
+                     result.commit_type,
+                     result.added_files,
+                     result.modified_files,
+                     result.deleted_files) for result in self.uut.run(
+                     *args, **kwargs))
+
+    def assert_no_msgs(self):
```
refeed: This method isn't used anywhere in the class, remove it

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,223 @@
...
+                               'This reverts commit %s.\n\n' % (test_sha3))
+        self.assertEqual(self.run_uut(),
+                         [(test_raw_commit_msg, test_sha4,
+                           ['ci_skip_commit', 'merge_commit', 'revert commit'],
+                             [], [], ['testfile3.txt'])])
+
+        run_shell_command('git checkout master')
+        run_shell_command('git checkout -b new-feature master')
+        run_shell_command('touch testfile4.txt ')
+        GitCommitBearTest.run_git_command('add', 'testfile4.txt')
```
refeed: I think it will be better if we start to take a consistency here, I saw you use `run_shell_command('git..)` and `GitCommitBear.run_git_command` back and forth, I think just use `run_shell_command` in all of the test would be alright

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,126 @@
...
+
+        get_parent_commits = 'git log --pretty=%P -n 1 ' + commit_sha
+        parent_commits = run_shell_command(get_parent_commits)[0]
+        parent_commits_list = parent_commits.split('\n')
+
+        if len(parent_commits_list) >= 2:
+            commit_type.append('merge_commit')
+
+        if head_commit.startswith('Revert '):
+            commit_type.append('revert commit')
```
refeed: For consistency, use underscore instead of space in `revert commit`

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,189 @@
...
+        run_shell_command('touch testfile2.txt')
+        run_shell_command('git add testfile2.txt')
+        run_shell_command('git commit -m "Add testfile2 [ci skip]"')
+        test_sha2 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        self.assertEqual(self.run_uut(),
+                         [('Add testfile2 [ci skip]\n\n', test_sha2,
+                           ['ci_skip_commit', 'merge_commit'],
+                           ['testfile2.txt'], [], [])])
+
+        run_shell_command(['echo Some text here > testfile2.txt'], shell=True)
```
jayvdb: do not use `echo` & `shell=True`.
Create the file using normal Python file routines.

kriti21:
circleci and travis-ci failing due to sourceforge. Please review @jayvdb @damngamerz @refeed

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,150 @@
...
+        shutil.rmtree(self.gitdir, onerror=onerror)
+
+    def test_analyze_git_commit(self):
+        run_shell_command('git checkout -b testbranch master')
+        run_shell_command('touch testfile1.txt ')
+        run_shell_command('git add testfile1.txt')
+        run_shell_command('git commit -m "Add testfile1"')
+        test_sha1 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        self.assertEqual(self.run_uut(),
+                         [('Add testfile1\n\n', test_sha1, ['merge_commit'],
```
refeed: I'm sure that this commit shouldn't be marked as `merged_commit` type

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,150 @@
...
+                           ['testfile1.txt'], [], [])])
+
+        run_shell_command('touch testfile2.txt')
+        run_shell_command('git add testfile2.txt')
+        run_shell_command('git commit -m "Add testfile2 [ci skip]"')
+        test_sha2 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        self.assertEqual(self.run_uut(),
+                         [('Add testfile2 [ci skip]\n\n', test_sha2,
+                           ['ci_skip_commit', 'merge_commit'],
```
refeed: This commit shouldn't be marked as `merge_commit` either

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,150 @@
...
+
+        file_handle = open('testfile2.txt', 'a+')
+        file_handle.write('Some text')
+        file_handle.close()
+        run_shell_command('git add testfile2.txt')
+        run_shell_command('git commit -m "Modify testfile2"')
+        test_sha3 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        self.assertEqual(self.run_uut(),
+                         [('Modify testfile2\n\n', test_sha3,
+                           ['merge_commit'],
```
refeed: Also this, `merge_commit` shouldn't be there..

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,150 @@
...
+        run_shell_command('git add testfile3.txt')
+        run_shell_command('git commit -m "another commit [skip ci]"')
+        test_sha3 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        run_shell_command('git revert HEAD --no-edit')
+        test_sha4 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        test_raw_commit_msg = ('Revert "another commit [skip ci]"\n\n'
+                               'This reverts commit %s.\n\n' % (test_sha3))
+        self.assertEqual(self.run_uut(),
+                         [(test_raw_commit_msg, test_sha4,
+                           ['ci_skip_commit', 'merge_commit', 'revert_commit'],
```
refeed: `merge_commit` shouldn't be there

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,126 @@
...
+        commit_type = []
+
+        head_commit = head_commit.strip('\n')
+
+        if re.search(r'\[ci skip\]|\[skip ci\]', head_commit):
+            commit_type.append('ci_skip_commit')
+
+        get_parent_commits = 'git log --pretty=%P -n 1 ' + commit_sha
+        parent_commits = run_shell_command(get_parent_commits)[0]
+        parent_commits_list = parent_commits.split('\n')
```
refeed: I think this is the reason why `merge_commit` always appears in the testcase. I think you should split it with a space, not a newline
```
> git log --pretty=%P -n 1 (git rev-parse HEAD)
00c07fb1d4eeaaaa4cd76d6650b5ccfbb0f40be8 38536dfc0a633bb40b99013716ac26642ac17c84
```

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,140 @@
...
+        Return the commit message from the head commit
+        """
+
+    def analyze_git_commit(self, head_commit):
+        """
+        Check the current commit at HEAD.
+
+        Return the commit information such as commit message,
+        type of commit, commit SHA, modified, added and
+        delete files by the commit.
```
refeed: Also explain about the paremeters and the return value, see other bears, c.f `CommitBear.py`

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,145 @@
...
+    @enforce_signature
+    def __init__(self, origin,
+                 raw_commit_message: str,
+                 commit_sha: str,
+                 commit_type: list,
+                 modified_files: list,
+                 added_files: list,
+                 deleted_files: list):
+
+        HiddenResult.__init__(self, origin, '')
```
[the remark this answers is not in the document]
kriti21: Using `super()` gives Runtime error `Queue objects should only be shared between processes through inheritance` and also this part of regex `origin=\'bearclass\'` should be hardcoded to origin=\'CommitResult\' which is unlike other bears that return HiddenResult.

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,147 @@
...
+        os.chdir(self._old_cwd)
+        if platform.system() == 'Windows':
+            onerror = self._windows_rmtree_remove_readonly
+        else:
+            onerror = None
+        shutil.rmtree(self.gitdir, onerror=onerror)
+
+    def test_analyze_git_commit(self):
+        run_shell_command('git checkout -b testbranch master')
+        run_shell_command('touch testfile1.txt ')
```
refeed: As @jayvdb said in https://github.com/coala/coala-bears/pull/2501#discussion_r199858579

Use https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,144 @@
...
+
+@generate_repr(('id', hex),
+               'origin',
+               'raw_commit_message',
+               'commit_sha',
+               'commit_type',
+               'modified_files',
+               'added_files',
+               'deleted_files'
+               )
```
refeed: Move this parentheses to line `18` for consistency

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,147 @@
...
+        run_shell_command('git add testfile1.txt')
+        run_shell_command('git commit -m "Add testfile1"')
+        test_sha1 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        self.assertEqual(self.run_uut(),
+                         [('Add testfile1\n\n', test_sha1, [],
+                           ['testfile1.txt'], [], [])])
+
+        run_shell_command('touch testfile2.txt')
```
refeed: Use https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,147 @@
...
+        with open('testfile2.txt', 'w') as f:
+            f.write('Some text')
+        run_shell_command('git add testfile2.txt')
+        run_shell_command('git commit -m "Modify testfile2"')
+        test_sha3 = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+        self.assertEqual(self.run_uut(),
+                         [('Modify testfile2\n\n', test_sha3,
+                           [], [], ['testfile2.txt'], [])])
+
+        run_shell_command('touch testfile3.txt')
```
refeed: Use https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,147 @@
...
+        test_raw_commit_msg = ('Revert "another commit [skip ci]"\n\n'
+                               'This reverts commit %s.\n\n' % (test_sha3))
+        self.assertEqual(self.run_uut(),
+                         [(test_raw_commit_msg, test_sha4,
+                           ['ci_skip_commit', 'revert_commit'],
+                             [], [], ['testfile3.txt'])])
+
+        run_shell_command('git checkout master')
+        run_shell_command('git checkout -b new-feature master')
+        run_shell_command('touch testfile4.txt ')
```
refeed: Use https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,144 @@
...
+        :param head_commit: The commit message at HEAD
+        :return:            A tuple comprising of commit message,
+                            commit sha, type of commit and lists of
+                            modified, added and deleted files
+        """
+        raw_commit_message = head_commit
+
+        commit_sha = run_shell_command('git rev-parse HEAD')[0].strip('\n')
+
+        commit_type = []
```
refeed: Another good idea for this thing is to use `Flag`. It has several advantages than using this kind of list, such as more little memory usage, the type is constant (which means we can't add unknown `commit_type` in it), and more. See `URLBear.LINK_CONTEXT` for the usage, we already has `aenum` installed in this repo's requirements, so it's worth to use it. It'd be very nice if you can implement it :)

tests/vcs/VCSCommitBearTest.py
```diff
@@ -0,0 +1,151 @@
...
+        self.assertEqual(self.run_uut(),
+                         [(test_raw_commit_msg, test_sha5,
+                           COMMIT_TYPE.merge_commit, [], [], [])])
+
+
+class CommitResultTest(unittest.TestCase):
+    def setUp(self):
+        self.raw_commit_message = 'raw_commit_message'
+        self.commit_sha = 'commit_sha'
+        self.commit_type = ['commit_type']
```
refeed: This is unused

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,152 @@
+import abc
+import re
+
+from aenum import Flag
+
+from coalib.bears.GlobalBear import GlobalBear
+from coalib.results.HiddenResult import HiddenResult
```
refeed: Place this after the `coalib.misc.Shell` import. We usually also wanna sort the imports by the letter as well

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def analyze_git_commit(self, head_commit):
+        """
+        Check the current commit at HEAD.
+
+        Return the commit information such as commit message,
+        type of commit, commit SHA, modified, added and
+        delete files by the commit.
+
+        :param head_commit: The commit message at HEAD
```
refeed: Don't forget the dot (`.`) ;)

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,152 @@
...
+        :return:            A tuple comprising of commit message,
+                            commit sha, type of commit and lists of
+                            modified, added and deleted files
```
refeed: dot

kriti21:
Ping @jayvdb . Review request for this PR.

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,149 @@
...
+        self.parent_commits = parent_commits
+        self.commit_type = commit_type
+        self.modified_files = modified_files
+        self.added_files = added_files
+        self.deleted_files = deleted_files
+
+
+class VCSCommitBear(GlobalBear):
+    __metaclass__ = abc.ABCMeta
+    LANGUAGES = {'Git'}
```
jayvdb: If this is a generic *VCS*CommitBear, it should not mention git anywhere.  At all.

And `CommitResult` should include a field 'repository' which holds a VCSRepository object that contains at least 'vcs_type' ("git", "hg", etc), and maybe an (unique) 'identifier' field which for most VCS would hold the remote name, but no guarantee on which remote name it would hold.  We can then improve the VCSRepository class later.

Also to be more clear, to prevent confusion with our GitCommitBear, lets call these metadata-only worker bears VCSCommitMetadataBear.

So

1. VCSCommitMetadataBear - generic
2. GitCommitMetadataBear - holds git stuff

Then later someone else can create HgCommitMetadataBear .

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,149 @@
...
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+
+    @abc.abstractmethod
+    def get_head_commit(self):
+        """
+        Return the commit message from the head commit.
+        """
+
+    def analyze_git_commit(self, head_commit):
```
jayvdb: remove "git" from this name, as not all vcs are git, and it should be abstract.

Then move git implementation to a new subclass GitCommitMetadataBear

bears/vcs/VCSCommitBear.py
```diff
@@ -0,0 +1,149 @@
...
+        :param head_commit: The commit message at HEAD.
+        :return:            A tuple comprising of commit message,
+                            commit sha, list of parent commits,
+                            type of commit viz. simple or merge
+                            commit and lists of modified,
+                            added and deleted files.
+        """
+        raw_commit_message = head_commit
+
+        commit_sha = run_shell_command('git rev-parse HEAD')[0].strip('\n')
```
jayvdb: use `get_head_commit`

kriti21:
@jayvdb I did something wrong while rebasing this branch by mistake because of which gitmate assigned this PR to some other members. Please unassign them and remove extra labels.

bears/vcs/VCSCommitMetadataBear.py
```diff
@@ -0,0 +1,108 @@
...
+                                added and deleted files.
+        """
+
+    def run(self, **kwargs):
+        """
+        This bear returns information about the HEAD commit
+        as HiddenResult which can be used for inspection by
+        other bears.
+        """
+        head_commit_sha = self.get_head_commit_sha()
```
refeed: Catch the exception and `return` if `RuntimeError` is raised from `get_head_commit_sha(), by using `try: ... except RuntimeError:`

bears/vcs/git/GitCommitMetadataBear.py
```diff
@@ -0,0 +1,65 @@
...
+    LANGUAGES = {'Git'}
+
+    @classmethod
+    def check_prerequisites(cls):
+        if shutil.which('git') is None:
+            return 'git is not installed.'
+        else:
+            return True
+
+    def get_head_commit_sha(self):
```
jayvdb: this must also workaround the GitHub/Travis problem

Add @staticmethod to `GitCommitBear.check_github_pull_request_temporary_merge_commit` and use it here also.

tests/vcs/VCSCommitMetadataBearTest.py
```diff
@@ -0,0 +1,112 @@
...
+    @classmethod
+    def check_prerequisites(cls):
+        return True
+
+    def get_head_commit(self):
+        return ('This is the fake head commit', '')
+
+
+@generate_skip_decorator(VCSCommitMetadataBear)
+class InvalidCommitResultTest(unittest.TestCase):
```
refeed: I still don't understand what's this class testing. Could you explain it? if it's not needed remove it

kriti21: Test for line 102-103 of `VCSCommitMetadataBear` . It gave 95 % coverage otherwise.

kriti21: Required thing in this class is the function `def test_get_head_commit_sha()`

bears/vcs/git/GitCommitMetadataBear.py
refeed: `GitCommitBear.check_github_pull_request_temporary_merge_commit` currently returns the commit message (it doesn't return the commit head's SHA).

I think we can solve it after this PR is merged (?) Smaller patch easier to review

tests/vcs/VCSCommitMetadataBearTest.py
refeed: Ah, okay it makes sense for me now

tests/vcs/VCSCommitMetadataBearTest.py
```diff
@@ -0,0 +1,112 @@
...
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+
+
+class FakeCommitBear(GitCommitMetadataBear):
+    @classmethod
+    def check_prerequisites(cls):
+        return True
+
+    def get_head_commit(self):
```
refeed: The name of this method is `get_head_commit_sha()`

And for this test it should only contains `raise RuntimeError('RuntimeError from test')

bears/vcs/git/GitCommitMetadataBear.py
jayvdb: ok

tests/vcs/VCSCommitMetadataBearTest.py
```diff
@@ -0,0 +1,112 @@
...
+
+from bears.vcs.VCSCommitMetadataBear import (VCSCommitMetadataBear,
+                                             CommitResult, COMMIT_TYPE)
+from bears.vcs.git.GitCommitMetadataBear import GitCommitMetadataBear
+from coalib.misc.Shell import run_shell_command
+from coalib.settings.Section import Section
+from coalib.testing.BearTestHelper import generate_skip_decorator
+
+
+class FakeCommitBear(GitCommitMetadataBear):
```
refeed: Inherit from `VCSCommitMetadataBear`

jayvdb:
ack https://github.com/coala/coala-bears/pull/2543/commits/582e38330d03c696bc7476e102acaa833c9b33af

jayvdb:
@gitmate-bot ff

PR #2597, opened 2018-07-09
Use coalib c31fb2e

jayvdb:


jayvdb:
Circle is failing because `pip` is somehow using `python` and not `python3`. ;-(

jayvdb:
The fix to use master was easy.

Issue #2585, opened 2018-07-06
ClangBear.py: Rename variable range


MacBox7:
`range` is a python builtin and shouldn't be redefined. Rename occurrences at [1](https://github.com/coala/coala-bears/blob/master/bears/c_languages/ClangBear.py#L55) and [2](https://github.com/coala/coala-bears/blob/master/bears/c_languages/ClangBear.py#L103).

Difficulty: Newcomer

kUSHAL0601:
I would like to work on this for a start.

MacBox7:
Leave a comment on the group to get assigned.

sinscary:
@MacBox7  Can i take this up?

MorTD:
Hello

Is this issue available?

MacBox7:
Yes. Ask for assignment on coala chat.

jayvdb:
PR closed and issue unassigned because you've not been following the guideline or reading the bot messages.

AKG001:
Hey, can i work on this issue?

ANURADHAJHA99:
can you please assign  this issue to me?

MacBox7:
Ask on group chat please

jayvdb:
Assigned to @Karna98 as they have already accepted the invite to join @coala

Karna98:
@jayvdb Since the issue states that the defined variable 'range' is coinciding with per-defined variable in python 'range'. So after changing the variable_name how to check whether is it working or not?. Do I have to the whole project in my system?

PR #2624, opened 2018-07-19
Change name of 'range'

MorTD:
Rename 'range' when looping over cindex.SourceRange objects. #2585

### DIFF
--- a/bears/c_languages/ClangBear.py
+++ b/bears/c_languages/ClangBear.py
@@ -56,13 +56,13 @@ def sourcerange_from_clang_range(range):
     """
     Creates a ``SourceRange`` from a clang ``SourceRange`` object.
 
-    :param range: A ``cindex.SourceRange`` object.
+    :param clang_range: A ``cindex.SourceRange`` object.
     """
-    return SourceRange.from_values(range.start.file.name,
-                                   range.start.line,
-                                   range.start.column,
-                                   range.end.line,
-                                   range.end.column)
+    return SourceRange.from_values(clang_range.start.file.name,
+                                   clang_range.start.line,
+                                   clang_range.start.column,
+                                   clang_range.end.line,
+                                   clang_range.end.column)
 
 
 class ClangBear(LocalBear):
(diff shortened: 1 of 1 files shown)


bears/c_languages/ClangBear.py
```diff
@@ -52,17 +52,17 @@ def diff_from_clang_fixit(fixit, file):
     return Diff.from_string_arrays(file, new_file)
 
 
-def sourcerange_from_clang_range(range):
+def sourcerange_from_clang_range(srcrange):
```
MacBox7: It would be better to call the parameter as `clang_range` rather than `srcrange`.

bears/c_languages/ClangBear.py
```diff
@@ -99,8 +99,8 @@ def run(self, filename, file,
                         2: RESULT_SEVERITY.NORMAL,
                         3: RESULT_SEVERITY.MAJOR,
                         4: RESULT_SEVERITY.MAJOR}.get(diag.severity)
-            affected_code = tuple(sourcerange_from_clang_range(range)
-                                  for range in diag.ranges)
```
MacBox7: Same here.

bears/c_languages/ClangBear.py
MorTD: done @MacBox7

bears/c_languages/ClangBear.py
MorTD: As well as here

bears/c_languages/ClangBear.py
```diff
@@ -52,17 +52,18 @@ def diff_from_clang_fixit(fixit, file):
     return Diff.from_string_arrays(file, new_file)
 
 
-def sourcerange_from_clang_range(range):
+
```
jayvdb: this extra blank line needs to be removed per PEP8 standard

MorTD: done

bears/c_languages/ClangBear.py
```diff
@@ -56,13 +56,13 @@ def sourcerange_from_clang_range(range):
...
 
-    :param range: A ``cindex.SourceRange`` object.
+    :param clang_range: A ``cindex.SourceRange`` object.
     """
-    return SourceRange.from_values(range.start.file.name,
-                                   range.start.line,
-                                   range.start.column,
-                                   range.end.line,
-                                   range.end.column)
+    return SourceRange.from_values(clang_range.start.file.name,
```
jayvdb: https://travis-ci.org/coala/coala-bears/jobs/407191026#L3565

PR #2626, opened 2018-07-20
PySafetyBear: Update PySafetyBear

RaiVaibhav:
Update PySafetyBear so that it support the
latest safety version 1.8.2

Related to https://github.com/coala/coala-bears/issues/2511
Related to https://github.com/coala/coala-bears/issues/1684

### DIFF
--- a/bears/python/requirements/PySafetyBear.py
+++ b/bears/python/requirements/PySafetyBear.py
@@ -1,3 +1,4 @@
+import os
 from collections import namedtuple
 import pkg_resources
 import re
@@ -7,11 +8,13 @@ from safety import safety
 from coalib.bears.LocalBear import LocalBear
 from dependency_management.requirements.PipRequirement import PipRequirement
 from coalib.results.Result import Result
+from coalib.settings.Setting import path
 from coalib.results.SourceRange import SourceRange
 from coalib.settings.Setting import typed_list
 
 
-def cve_key_checker(vulnerability):
+# It was for old versions of safety and those versions will be allow in future.
+def cve_key_checker(vulnerability):  # pragma: no cover
     if 'cve' in vulnerability.data:
         if vulnerability.data['cve'] is None:
             return None
@@ -24,6 +27,23 @@ def cve_key_checker(vulnerability):
 # the safety module expects an object that looks like this
 # (not importing it from there because it's in a private-ish location)
 Package = namedtuple('Package', ('key', 'version'))
+
+safety_get_vulnerabilities = safety.get_vulnerabilities
+_insecure_full_json_url = ('https://raw.githubusercontent.com/pyupio/'
+                           'safety-db/master/data/insecure_full.json')
+
+_insecure_json_url = ('https://raw.githubusercontent.com/'
+                      'pyupio/safety-db/master/data/insecure.json')
+
+
+def _get_vulnerabilities(pkg, spec, db):
+    for entry in safety_get_vulnerabilities(pkg, spec, db):
+        entry['cve'] = entry['id'] if entry['cve'] is None else entry['cve']
+        entry['id'] = entry['cve']
+        yield entry
+
+
+safety.get_vulnerabilities = _get_vulnerabilities
 
 
 class PySafetyBear(LocalBear):
@@ -40,15 +60,28 @@ class PySafetyBear(LocalBear):
         'Python 3 Requirements',
     }
     AUTHORS = {'Bence Nagy'}
-    REQUIREMENTS = {PipRequirement('safety', '0.5.1')}
+    REQUIREMENTS = {PipRequirement('safety', '1.8.2')}
     AUTHORS_EMAILS = {'bence@underyx.me'}
     LICENSE = 'AGPL'
     CAN_DETECT = {'Security'}
 
-    def run(self, filename, file):
+    def setup_dependencies(self):
+        file = self.download_cached_file(_insecure_full_json_url,
+                                         'insecure_full.json')
+        self.download_cached_file(_insecure_json_url,
+                                  'insecure.json')
+        type(self).db_path = os.path.dirname(file)
+
+    def run(self, filename, file,
+            db_path: path = '',
+            cve_ignore: typed_list(str) = []):
         """
         Checks for vulnerable package versions in requirements files.
+
+        :param db_path:           Path to a local vulnerability database.
+        :param cve_ignore:        A list of CVE number to be ignore.
         """
+        db_path = self.db_path if not db_path else db_path
         packages = list(
             Package(key=req.key, version=req.specs[0][1])
             for req in self.try_parse_requirements(file)
@@ -58,16 +91,19 @@ class PySafetyBear(LocalBear):
         if not packages:
             return
 
-        for vulnerability in safety.check(packages=packages):
-            if cve_key_checker(vulnerability):
+        for vulnerability in safety.check(packages, key=None,
+                                          db_mirror=db_path, cached=False,
+                                          ignore_ids=cve_ignore):
+            if 'cve' in vulnerability.vuln_id.strip().lower():
                 message_template = (
-                    '{vuln.name}{vuln.spec} is vulnerable to {vuln.cve_id} '
+                    '{vuln.name}{vuln.spec} is vulnerable to {vuln.vuln_id} '
                     'and your project is using {vuln.version}.'
                 )
             else:
                 message_template = (
-                    '{vuln.name}{vuln.spec} is vulnerable and your project is '
-                    'using {vuln.version}.'
+                    '{vuln.name}{vuln.spec} is vulnerable to '
+                    'pyup.io-{vuln.vuln_id} and your project is using '
+                    '{vuln.version}.'
                 )
 
             # StopIteration should not ever happen so skipping its branch
@@ -87,7 +123,7 @@ class PySafetyBear(LocalBear):
             yield Result(
                 self,
                 message_template.format(vuln=vulnerability),
-                additional_info=vulnerability.data['advisory'],
+                additional_info=vulnerability.advisory,
                 affected_code=(source_range, ),
             )
 
--- a/tests/python/requirements/PySafetyBearWithoutMockTest.py
+++ b/tests/python/requirements/PySafetyBearWithoutMockTest.py
@@ -3,6 +3,7 @@ import os
 from queue import Queue
 from bears.python.requirements.PySafetyBear import PySafetyBear
 from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
 from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
 from coalib.results.Result import Result
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
@@ -23,7 +24,8 @@ def load_testfile(name):
 class PySafetyBearTest(LocalBearTestHelper):
 
     def setUp(self):
-        self.uut = PySafetyBear(Section('name'), Queue())
+        self.section = Section('name')
+        self.uut = PySafetyBear(self.section, Queue())
 
     def test_without_vulnerability(self):
         self.check_validity(self.uut, ['lxml==3.6.0'])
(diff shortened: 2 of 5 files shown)


tests/python/requirements/PySafety_test_files/requirement.txt
```diff
@@ -1,2 +1,2 @@
-bottle==0.10.0
+aiocouchdb==0.2
```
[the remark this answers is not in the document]
RaiVaibhav: with the current safety version `bottle==0.10.1` will give 2 result object for same package
https://bpaste.net/show/496974b7c96f

So I changed the package which contains some smaller result object.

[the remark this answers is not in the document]
jayvdb: No.  Add both results.

We need to know that the old tests are not breaking .. more data is natural.

tests/python/requirements/PySafetyBearWithoutMockTest.py
```diff
@@ -24,6 +24,7 @@ class PySafetyBearTest(LocalBearTestHelper):
 
     def setUp(self):
         self.uut = PySafetyBear(Section('name'), Queue())
+        self.maxDiff = None
```
jayvdb: this doesnt belong in the checked in version

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,14 +30,28 @@ class PySafetyBear(LocalBear):
...
-    def run(self, filename, file):
+    def run(self, filename, file,
+            key: str = '',
+            db_mirror: str = '',
+            cached: bool = False,
+            ignore_ids: typed_list(str) = []):
         """
         Checks for vulnerable package versions in requirements files.
+
+        :param key:           API Key for pyup.io's vulnerability database. Can
```
jayvdb: we do not store keys in `.coafile` which are checked into a repository.

If `SAFETY_API_KEY` works, that will have to be how it is provided to the bear, and core will need some enhancement in future to manage bear envvars.

If a key is mandatory, the bear needs to check that `SAFETY_API_KEY` or `db_mirror` is set in an earlier phase of the Bear configuration, so it doesnt `run()` multiple times and get the same error.

tests/python/requirements/PySafetyBearWithoutMockTest.py
```diff
@@ -39,9 +40,9 @@ def test_with_cve_vulnerability(self):
             self.uut,
             file_contents,
             [Result.from_values('PySafetyBear',
-                                'bottle<0.12.10 is vulnerable '
-                                'to CVE-2016-9964 and your project '
```
jayvdb: has the ability to fetch the CVE been removed entirely from `safety`?

bears/python/requirements/PySafetyBear.py
RaiVaibhav: IMO `db_mirror` setting can be there ? or I remove that also?

tests/python/requirements/PySafetyBearWithoutMockTest.py
RaiVaibhav: Currenlty safety return Vulnerability class i.e.,
```
class Vulnerability(namedtuple("Vulnerability",
                               ["name", "spec", "version", "advisory", "vuln_id"])):
    pass
```

Which don't have `cve_key` , One case was I iterate the database one more time to get `cve_key` but seems like one problem here that `safety` also take care that if some user provided a path to database it will fetch from there and these are the logic created inside `safety.check`.

bears/python/requirements/PySafetyBear.py
jayvdb: Lets leave db_mirror, but default None.

And if not provided, use `Bear.download_cached_file` to fetch a copy of the database and store it in the coala bears cache

https://github.com/coala/coala-bears/blob/master/bears/java/CheckstyleBear.py#L58

or, add a dependency on https://pypi.org/project/safety-db/ , and use that dataset.
(even if the data is old; we can fix that later)

tests/python/requirements/PySafetyBearWithoutMockTest.py
jayvdb: Is `advisory` the same as `cve` ?

RaiVaibhav: No , advisory is additional detail

bears/python/requirements/PySafetyBear.py
```diff
@@ -58,10 +88,11 @@ def run(self, filename, file):
         if not packages:
             return
 
-        for vulnerability in safety.check(packages=packages):
-            if cve_key_checker(vulnerability):
+        for vulnerability in safety.check(packages, key, db_mirror, cached,
```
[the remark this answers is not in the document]
jayvdb: `key=False, db=False, cached=False`

bears/python/requirements/PySafetyBear.py
```diff
@@ -11,21 +12,34 @@
...
-def cve_key_checker(vulnerability):
-    if 'cve' in vulnerability.data:
-        if vulnerability.data['cve'] is None:
-            return None
-        else:
-            return True
-    else:
-        return None
+# It was for old versions of safety and those versions will be allow in future.
+# def cve_key_checker(vulnerability):
```
jayvdb: Keep it as it was, uncommented. Sorry if i was confusing.

Use pragma nocover for now.

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,15 +54,36 @@ class PySafetyBear(LocalBear):
...
+        insecure_josn_url = ('https://raw.githubusercontent.com/'
+                             'pyupio/safety-db/master/data/insecure.json')
+        self.download_cached_file(insecure_full_json_url,
+                                  'insecure_full.json')
+        my_file = self.download_cached_file(insecure_josn_url,
+                                            'insecure.json')
+        type(self).db_path = os.path.dirname(my_file)
+
+    def run(self, filename, file,
+            db_path: str = None,
```
jayvdb: See other bears (inc CheckstyleBear ) for the type to use for paths

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,15 +54,36 @@ class PySafetyBear(LocalBear):
...
+            cache_request: bool = False,
```
jayvdb: Why False by default?

Why would someone ever want coala to not cache external resources?

bears/python/requirements/PySafetyBear.py
RaiVaibhav: I don't get what exactly you are saying, when I initially make it to `db_path: str = ''` you check if it is `None` so I make it to None.

RaiVaibhav: got your point :+1:  `from coalib.settings.Setting import path`

bears/python/requirements/PySafetyBear.py
jayvdb: we make our own decisions, for own own users needs, which may be quite different than upstream.

We need to know what it does, especially wrt to a locally cached db.

We would look very silly if this option was meaningless when not using a `key`

bears/python/requirements/PySafetyBear.py
```diff
@@ -1,17 +1,20 @@
 from collections import namedtuple
 import pkg_resources
+import os
```
jayvdb: Import order.

Needs rebase.

And see reply re cache_request.

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,15 +55,33 @@ class PySafetyBear(LocalBear):
...
     AUTHORS_EMAILS = {'bence@underyx.me'}
     LICENSE = 'AGPL'
     CAN_DETECT = {'Security'}
 
-    def run(self, filename, file):
+    def setup_dependencies(self):
+        insecure_full_json_url = ('https://raw.githubusercontent.com/'
+                                  'pyupio/safety-db/master/data/'
+                                  'insecure_full.json')
+        insecure_josn_url = ('https://raw.githubusercontent.com/'
```
jayvdb: josn

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,15 +55,33 @@ class PySafetyBear(LocalBear):
...
+                             'pyupio/safety-db/master/data/insecure.json')
+        self.download_cached_file(insecure_full_json_url,
+                                  'insecure_full.json')
+        my_file = self.download_cached_file(insecure_josn_url,
```
jayvdb: `my_` is not a variable name that belongs in anything except for a hacky POC

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,15 +55,33 @@ class PySafetyBear(LocalBear):
...
+                                            'insecure.json')
+        type(self).db_path = os.path.dirname(my_file)
```
jayvdb: why are you downloading both, and then using the less useful one ?

bears/python/requirements/PySafetyBear.py
```diff
@@ -40,15 +55,33 @@ class PySafetyBear(LocalBear):
...
     AUTHORS = {'Bence Nagy'}
-    REQUIREMENTS = {PipRequirement('safety', '0.5.1')}
+    REQUIREMENTS = {PipRequirement('safety', '1.8.2')}
     AUTHORS_EMAILS = {'bence@underyx.me'}
     LICENSE = 'AGPL'
     CAN_DETECT = {'Security'}
 
-    def run(self, filename, file):
+    def setup_dependencies(self):
+        insecure_full_json_url = ('https://raw.githubusercontent.com/'
```
jayvdb: put these in module level global _private variables so that people can access them easily if they need to.

bears/python/requirements/PySafetyBear.py
```diff
@@ -26,6 +29,18 @@ def cve_key_checker(vulnerability):
...
+def _get_vulnerabilities(pkg, spec, db):
+    for entry in db[pkg]:
+        entry['cve'] = 'None' if entry['cve'] is None else entry['cve']
+        entry['id'] = entry['cve']
+        for entry_spec in entry['specs']:
+            if entry_spec == spec:
+                yield entry
+
+
+safety.get_vulnerabilities = _get_vulnerabilities
```
jayvdb: add the `cve` in a hack of `get_database_file` instead of a hack of `get_vulnerabilities`.

You are already doing a hack on their db loading mechanism, in `setup_dependencies`, so it is much safer and logical to modify their file loading mechanism than to duplicate&modify their function which has comparison logic.

The hack should be in `setup_dependencies` .

And to help with future breakages, you should first do `safety.get_database_file` (EOL) , as that will make Python check that they have a function called `safety.get_database_file`.  If it doesnt, the exception will occur there.  Then you can replace it, knowing the safety code will call it.

Otherwise they may change the name of that function and the bear error will be very confusing.

jayvdb: ah, it is called `fetch_database_file`

bears/python/requirements/PySafetyBear.py
jayvdb: > As far as I know safety needed both 
> https://github.com/pyupio/safety/blob/master/safety/safety.py#L104

No, it needs one or the other.

bears/python/requirements/PySafetyBear.py
```diff
@@ -26,6 +29,18 @@ def cve_key_checker(vulnerability):
 Package = namedtuple('Package', ('key', 'version'))
 
 
+def _get_vulnerabilities(pkg, spec, db):
+    for entry in db[pkg]:
+        entry['cve'] = 'None' if entry['cve'] is None else entry['cve']
+        entry['id'] = entry['cve']
```
jayvdb: you should only overwrite this if it is a cve.  If it isnt, you need to show the pyup vuln id.
find a solution.
many exist.
choose one.

bears/python/requirements/PySafetyBear.py
jayvdb: and alternative is to keep hacking `get_vulnerabilities` , but also use it!

```
safety_get_vulnerabilities = safety.get_vulnerabilities

def _get_vulnerabilities(...)
    for entry in safety_get_vulnerabilities(...):
        hack_entry
        yield entry
```

bears/python/requirements/PySafetyBear.py
```diff
@@ -58,16 +87,19 @@ def run(self, filename, file):
...
-                    '{vuln.name}{vuln.spec} is vulnerable to {vuln.cve_id} '
+                    '{vuln.name}{vuln.spec} is vulnerable to {vuln.vuln_id} '
                     'and your project is using {vuln.version}.'
                 )
             else:
                 message_template = (
-                    '{vuln.name}{vuln.spec} is vulnerable and your project is '
-                    'using {vuln.version}.'
+                    '{vuln.name}{vuln.spec} is vulnerable to '
+                    'pyup.io-{vuln.vuln_id} your project is using '
```
jayvdb: `and` missing

jayvdb:
ack 8faa08e

jayvdb:
@gitmate-bot ff

PR #2648, opened 2018-07-28
generate_package.py: Remove from coala-bears

### LINKED ISSUE #1199: Move generate_package.py to a helper script
As noted on https://github.com/coala/coala-bears/issues/1000 , there is not much in "bears" except plugins.  If we can empty out "bears" of everything not loaded via setuptools entry points, all dependencies belong to each plugin, and there is no developer API in "bears", so no need for API backwards compatibility, only user settings backwards compatibility.

Removing all python files from the top bears directory is also necessary for adopting [PEP 420 -- Implicit Namespace Packages](https://www.python.org/dev/peps/pep-0420/).  

This should probably be done after https://github.com/coala/coala-bears/pull/617 .

anctartica:
Closes #1199


anctartica:
https://github.com/anctartica/bear-packages/pull/3
I have added a positional argument to the script which takes the path to the coala-bears repository. Tested locally.

.ci/generate_package.py
```diff
@@ -95,7 +95,8 @@ def create_upload_parser():
 def main():
     args = create_upload_parser().parse_args()
 
-    os.chdir(os.path.join(os.path.dirname(__file__), '..'))
+    current_dir = os.path.dirname(__file__)
```
jayvdb: this shouldnt need changing.

other than the bears directory, the script shouldnt rely on anything else being in some specific location.

.ci/generate_package.py
```diff
@@ -145,7 +146,7 @@ def main():
                                  'BEAR_NAME': bear_name,
                                  'ENTRY': 'coala' + bear_name}
 
-            create_file_from_template(os.path.join('bears', 'setup.py.in'),
+            create_file_from_template(os.path.join(current_dir, 'setup.py.in'),
```
jayvdb: create an argparse arg for the file filename of the template ; the tests already allow this extra arg to be added once in the setUp .

It is very important that we use explicit paths so it isnt 'magic' - they need to read this file, and hopefully improve it, before uploading crappy packages to pypi with that template.

my preference would be to keep this file in .moban.dt , but I have no objection if you feel it is better in `.ci`

jayvdb:
ack https://github.com/coala/coala-bears/pull/2648/commits/26b696969dd38ca1a1ab91163e0154f6c905e252

jayvdb:
@gitmate-bot ff

PR #2657, opened 2018-08-05
PycodestyleBear: Migrate bear to aspects

pareksha:
This migrates `PycodestyleBear` to its respective
aspects and tastes.

Closes https://github.com/coala/coala-bears/issues/2656

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/python/PycodestyleBear.py
+++ b/bears/python/PycodestyleBear.py
@@ -1,15 +1,21 @@
 import sys
+import re
 
 from coalib.bearlib.abstractions.Linter import linter
 from coalib.settings.Setting import typed_list
+from coalib.results.Result import Result
+from coalib.bearlib.aspects import map_setting_to_aspect
+from coalib.bearlib.aspects.Formatting import LineLength
 
 from dependency_management.requirements.PipRequirement import PipRequirement
 
 
+OUTPUT_REGEX = (r'(?P<line>\d+) (?P<column>\d+) '
+                r'(?P<message>(?P<origin>\S+).*)')
+
+
 @linter(executable='pycodestyle',
-        output_format='regex',
-        output_regex=r'(?P<line>\d+) (?P<column>\d+) '
-                     r'(?P<message>(?P<origin>\S+).*)')
+        )
 class PycodestyleBear:
     """
     A wrapper for the tool ``pycodestyle`` formerly known as ``pep8``.
@@ -21,8 +27,11 @@ class PycodestyleBear:
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting'}
 
-    @staticmethod
+    @map_setting_to_aspect(
+        max_line_length=LineLength.max_line_length,
+    )
     def create_arguments(
+            self,
             filename, file, config_file,
             pycodestyle_ignore: typed_list(str) = (
                 'E121', 'E123', 'E126', 'E133', 'E226',
@@ -61,3 +70,21 @@ class PycodestyleBear:
         arguments.append(filename)
 
         return arguments
+
+    def process_output(self, output, filename, file):
+        result = re.match(OUTPUT_REGEX, output)
+        if not result:
+            return
+        line, column, message, rule = result.groups()
+        if rule == 'E501':
+            aspect = LineLength('py')
+        else:
+            aspect = None
+        yield Result.from_values(
+            origin='{} ({})'.format(self.name, rule),
+            message=message,
+            file=filename,
+            line=int(line),
+            column=int(column),
+            aspect=aspect,
+            )
--- a/tests/python/PycodestyleBearTest.py
+++ b/tests/python/PycodestyleBearTest.py
@@ -1,5 +1,15 @@
+import unittest
+
 from bears.python.PycodestyleBear import PycodestyleBear
-from coalib.testing.LocalBearTestHelper import verify_local_bear
+from coalib.testing.LocalBearTestHelper import verify_local_bear, execute_bear
+from coalib.settings.Section import Section
+from coala_utils.ContextManagers import prepare_file
+from queue import Queue
+from coalib.bearlib.aspects.Formatting import LineLength
+from coalib.bearlib.aspects import (
+    AspectList,
+    get as get_aspect,
+)
 
 
 good_file = '''
@@ -45,6 +55,8 @@ PycodestyleBearConfigSelectTest = verify_local_bear(
 
 long_line = 'a = "{0}"'.format('a' * 100)
 
+small_line = 'Small line.'
+
 PycodestyleBearLineLengthTest = verify_local_bear(
     PycodestyleBear,
     valid_files=(),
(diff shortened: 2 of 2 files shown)


bears/python/PycodestyleBear.py
```diff
@@ -21,9 +23,14 @@ class PycodestyleBear:
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting'}
 
-    @staticmethod
```
jayvdb: great. removing this is necessary, because the decorator uses `self.section.aspects` , and so `self` is needed.

tests/python/PycodestyleBearTest.py
```diff
@@ -63,3 +69,22 @@ def hello():
     valid_files=(file_with_very_long_line,),
     invalid_files=(),
     settings={'max_line_length': 0})
+
+PycodestyleBearAspectsTest = verify_local_bear(
```
jayvdb: We need to verify that the `Result` object returned uses the aspect.
(I suspect that it doesnt.)

If it doesnt, create a bug against `@linter` because its default `run` needs to use aspects for the `Result`, and `map_setting_to_aspect` may also need to be improved in the process.

And we may just merge this as-is if the bug is in Linter.

If the Result objects correctly use aspects, then this test class needs to be a more detailed test class ... which checks that the Result object reports the correct aspect.

pareksha: Yes, it doesn't use the aspect. 

We need to modify `@linter` to use the aspect just as the `Result` module does.

jayvdb: Ya, I expected so.  Thanks.

What you could try is add a `run` method to this bear, and it would do

```
for result in super().run(*args, **kwargs):
   if (max_line_length is from aspect) and (result is line length problem):
        result.aspect = LineLength('Python').max_line_length
```

bears/python/PycodestyleBear.py
```diff
@@ -61,3 +70,24 @@ def create_arguments(
...
+        if not result:
+            return
+        line, column, message, origin = result.groups()
+        yield Result.from_values(
+            origin=self,
+            message=message,
+            file=filename,
+            line=int(line),
+            column=int(column),
+            aspect=Formatting('py'),
```
jayvdb: is it possible / sensible to set this to the taste for LineLenght , if the aspect was used to determine the max line length?

jayvdb: See the docstring for `Result.aspect`.

> An aspectclass instance which this result is associated to.
>  Note that this should be a leaf of the aspect tree!

Formatting('py') is not the leaf of the aspect tree.

pareksha: Oh, didn't see that one, I'll update it to `LineLength`

bears/python/PycodestyleBear.py
```diff
@@ -61,3 +67,24 @@ def create_arguments(
...
+        if not result:
+            return
+        line, column, message, origin = result.groups()
+        yield Result.from_values(
+            origin=self,
+            message=message,
+            file=filename,
+            line=int(line),
+            column=int(column),
+            aspect=LineLength('py'),
```
jayvdb: not all pycodestyle errors are LineLength('py').

Please decode the pycodestyle errors and use appropriate aspects

bears/python/PycodestyleBear.py
```diff
@@ -1,15 +1,16 @@
...
+from coalib.bearlib.aspects.Formatting import LineLength
 
 from dependency_management.requirements.PipRequirement import PipRequirement
 
 
-@linter(executable='pycodestyle',
-        output_format='regex',
-        output_regex=r'(?P<line>\d+) (?P<column>\d+) '
-                     r'(?P<message>(?P<origin>\S+).*)')
+@linter(executable='pycodestyle',)
```
jayvdb: a trailing comma is useless if followed by a bracket.

bears/python/PycodestyleBear.py
```diff
@@ -1,15 +1,16 @@
...
+from coalib.results.Result import Result
+from coalib.bearlib.aspects import map_setting_to_aspect
+from coalib.bearlib.aspects.Formatting import LineLength
 
 from dependency_management.requirements.PipRequirement import PipRequirement
 
 
-@linter(executable='pycodestyle',
-        output_format='regex',
-        output_regex=r'(?P<line>\d+) (?P<column>\d+) '
```
jayvdb: keep the `output_regex` here as a module level variable.

the diff will then show that the regex hasnt changed.

tests/python/PycodestyleBearTest.py
jayvdb: Now we need a test class which verifies that the aspect is used when the line length is exceeded.

bears/python/PycodestyleBear.py
```diff
@@ -61,3 +72,26 @@ def create_arguments(
...
+        result = re.match(OUTPUT_REGEX, output)
+        if not result:
+            return
+        line, column, message, origin = result.groups()
+        if origin == 'E501':
+            aspect = LineLength('py')
+        else:
+            aspect = None
+        yield Result.from_values(
+            origin=self,
```
jayvdb: this is a regression.
Previously the origin was the error code.
you are discarding that

pareksha: origin is to be kept `PycodestyleBear` only as that when we want to ignore PycodestyleBear, we have to do `# Ignore Pycodestyle` and not `# Ignore E501(or any other rule)`.

I have changed it to `rule` so that there's no confusion.

jayvdb: it was previously `origin` ... was it being ignored previously?

if the `Result` objects were previously setting origin to `PycodestyleBear`, then why was the regex capturing `origin` .. what was happening with it previously?

wrt `# Ignore ...` if the bear was previously using `origin` of `E501`, then users *must* be using `# Ignore E501` currently, and if you change it then their CI will break.

So we need to be careful.

We can still change it, but we need to ensure that change isnt hidden in an 'aspectification' commit.  It needs to be a separate issue which is classified as a bug, and properly analysed.

pareksha: Currently, it uses `# Ignore Pycodestyle` or `# Ignore Pycodestyle (E501)` (an example is https://github.com/coala/coala-bears/blob/e46d352d30f5895b4deead2bcc174a50f480aade/tests/js/PrettierLintBearTest.py#L11).

I see what `@linter` was doing previously, see https://github.com/coala/coala/blob/d936d1624a8d37b4be96d023079c39f244a351c3/coalib/bearlib/abstractions/Linter.py#L320

Now I have changed this to what was being done previously in the linter.

jayvdb: Thank you -- that is better.

bears/python/PycodestyleBear.py
```diff
@@ -21,9 +27,14 @@ class PycodestyleBear:
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Formatting'}
 
-    @staticmethod
+    @map_setting_to_aspect(
+        max_line_length=LineLength.max_line_length,
+    )
     def create_arguments(
-            filename, file, config_file,
```
jayvdb: this line does not need to change.

you only need to add a line above with `self,`

bears/python/PycodestyleBear.py
```diff
@@ -61,3 +72,26 @@ def create_arguments(
         arguments.append(filename)
 
         return arguments
+
+    def process_output(
```
jayvdb: all args fit on one line, and should be on one line.  The signature of this method does not change.

bears/python/PycodestyleBear.py
```diff
@@ -1,15 +1,21 @@
...
 from coalib.settings.Setting import typed_list
+from coalib.results.Result import Result
+from coalib.bearlib.aspects import map_setting_to_aspect
+from coalib.bearlib.aspects.Formatting import LineLength
 
 from dependency_management.requirements.PipRequirement import PipRequirement
 
 
+OUTPUT_REGEX = (r'(?P<line>\d+) (?P<column>\d+) '
+                r'(?P<message>(?P<rule>\S+).*)')
```
jayvdb: please restore this to 'origin'

jayvdb:
ack https://github.com/coala/coala-bears/pull/2657/commits/c9e8df083c5cae5651b81c0efdc9da3d0f79cd0b

jayvdb:
@gitmate-bot ff

tests/python/PycodestyleBearTest.py
KVGarg: > Ya, I expected so. Thanks.
> 
> What you could try is add a `run` method to this bear, and it would do
> 
> ```
> for result in super().run(*args, **kwargs):
>    if (max_line_length is from aspect) and (result is line length problem):
>         result.aspect = LineLength('Python').max_line_length
> ```
I don't think so, defining such if-else for each aspect will be a nice idea. There will be a lot of if-else then because there exists many leaf-aspects for which we will have to define a rule then.
Can we use nltk package? that can somewhere help us to extract meaningful words from result message and then check whether that result belongs to that leaf-aspect or not?

KVGarg: > Yes, it doesn't use the aspect.
> We need to modify `@linter` to use the aspect just as the `Result` module does.
> 
> I have created an issue [coala/coala#5694](https://github.com/coala/coala/issues/5694)

Do you mean here by LocalBear and GlobalBear which are having a `aspects` argument? And something like that we need to implement for `@linter` also?

PR #2685, opened 2018-09-01
Fix master build

Makman2:


jayvdb:
Dup of https://github.com/coala/coala-bears/pull/2680 , not reviewed for two weeks.

PR #2702, opened 2018-09-27
PEP8NoteBear: Add infinite line length

rishabhgarg25699:
Allows infinite line length when
max_line_length is set to 0.

Closes https://github.com/coala/coala-bears/issues/2443

### DIFF
--- a/bears/python/PEP8NotebookBear.py
+++ b/bears/python/PEP8NotebookBear.py
@@ -1,5 +1,6 @@
 import autopep8
 import nbformat
+import sys
 
 from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
 from coalib.bears.LocalBear import LocalBear
@@ -84,6 +85,7 @@ class PEP8NotebookBear(LocalBear):
         will not change functionality of the code in any way.
 
         :param max_line_length:   Maximum number of characters for a line.
+                                  When set to 0 allows infinite line length.
         :param indent_size:       Number of spaces per indent level.
         :param pep_ignore:        A list of errors/warnings to ignore.
         :param pep_select:        A list of errors/warnings to exclusively
@@ -91,6 +93,9 @@ class PEP8NotebookBear(LocalBear):
         :param local_pep8_config: Set to true if autopep8 should use a config
                                   file as if run normally from this directory.
         """
+        if not max_line_length:
+            max_line_length = sys.maxsize
+
         options = {'ignore': pep_ignore,
                    'select': pep_select,
                    'max_line_length': max_line_length,
--- a/tests/python/PEP8NotebookBearTest.py
+++ b/tests/python/PEP8NotebookBearTest.py
@@ -22,6 +22,20 @@ good_file = r"""{
    "source": [
     "a markdown cell is not a code cell"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "options": {
+    "max_line_length": 0
+   },
+   "outputs": [],
+   "source": [
+    "'x =' + ' 1 +' * 100 + ' 1'"
+   ]
   }
  ],
  "metadata": {},


bears/python/PEP8NotebookBear.py
```diff
@@ -1,4 +1,5 @@
 import autopep8
+import sys
```
apb7: Don't we prefer to sort imports alphabetically, @rishabhgarg25699 and @jayvdb? Although PEP8 does not explicitly mention about this (it only mentions about the grouping of imports), I think alphabetically sorted imports are more readable and searchable. Thanks!

bears/python/PEP8NotebookBear.py
```diff
@@ -84,13 +85,17 @@ def run(self, filename, file,
...
         :param indent_size:       Number of spaces per indent level.
         :param pep_ignore:        A list of errors/warnings to ignore.
         :param pep_select:        A list of errors/warnings to exclusively
                                   apply.
         :param local_pep8_config: Set to true if autopep8 should use a config
                                   file as if run normally from this directory.
         """
+        if not max_line_length:
+            max_line_length = sys.maxsize
+
```
Naveenaidu: Is this extra line really necessary?

rishabhgarg25699: I think yes because line no. 120 in this program has also the same concept . It has also the gap of one line .

rishabhgarg25699:
@Naveenaidu I see the test of both the PR .In both there is a function made for the test but the test file for this i.e https://github.com/coala/coala-bears/blob/master/tests/python/PEP8NotebookBearTest.py  I am unable to understand In which language it is written. I have see this type of python program first time with meta data.

bears/python/PEP8NotebookBear.py
Naveenaidu: > @Naveenaidu I see the test of both the PR .In both there is a function made for the test but the test file for this i.e https://github.com/coala/coala-bears/blob/master/tests/python/PEP8NotebookBearTest.py I am unable to understand In which language it is written. I have see this type of python program first time with meta data.

@rishabhgarg25699 Python has been used to write the test file. It's similiar to GnuIndentBearTest. It would also help you if you understand what the role of `PEP8NotebookBear` is. Just go through the [doc-string](https://github.com/coala/coala-bears/blob/559c953e1b5787df9dddf96ece8449e50c945716/bears/python/PEP8NotebookBear.py#L14) of the file, the test file will make more sense to you.

> I have also to made only function for test at last of program ?

Please try to understand the process. You could actually get enough idea on writing the tests on seeing the two PR which have been attached.
I'll give you few hints:-
1. You need to make a file which has more than 80 characters. Have a look at the `source field` in the `good file`.
2. Once you create the above file you can then create the function( take reference from the PR's mentioned by @newbazz in the issue) in which you will have to append the settings of maxlength and voila you would have successfully written your tests. :)

Hope you could solve this issue now. Good Luck :)

rishabhgarg25699: @Naveenaidu I have to add code in good_file as well as in bad_file and then at last i have to add a function  in the test i.e https://github.com/coala/coala-bears/blob/master/tests/python/PEP8NotebookBearTest.py

rishabhgarg25699: I have to add this 
{
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "a markdown cell is not a code cell"
   ]
  },

in good_file

Naveenaidu: @rishabhgarg25699 That's for you to figure it out :) There won't be much fun left for you if I do tell.
A little suggestion, which I follow myself. Why not try out your approach in your local computer first, in this way you would not have to wait for someone to look at your approach and tell you if it's right or wrong. This will also help you to eliminate your approaches and arrive at a right solution. And moreover it will also help you to ask the right questions :)

rishabhgarg25699:
@Naveenaidu Now please see and suggest some changes. Now I have to add a function also? Or i have to make changes in the present function ?

rishabhgarg25699:
When I am running the PEP8NotebookBearTest.py locally from the coala-bears/tests/python directory then the result is coming is https://pastebin.com/xJwkfT2g . Please see this and suggest some changes in PEP8NotebookBearTest.py

rishabhgarg25699:
@li-boxuan we have to add this  ['a =' + ' 1 +' * 100 + ' 1'] in source of good file .

rishabhgarg25699:
@Naveenaidu  Now please see my changes and suggest suggestions. Also when I am running PEP8NotebookBearTest.py locally from coala-bears/tests directory by command pytest -k PEP8NotebookBearTest.py  it is showing error i.e https://pastebin.com/fzckQstk

tests/python/PEP8NotebookBearTest.py
```diff
@@ -22,6 +22,19 @@
...
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [
+    "'x =' + ' 1 +' * 100 + ' 1'"
+   ],
+   "source": [
+    "max_line_length = 0"
```
jayvdb: this 'source' doesnt match the 'outputs' above, making this an invalid file.

however more importantly, this does not test your code changes.

your changes to the tests need to *prove* that your code changes work correctly.  See examples that other people have given.

jayvdb:
Unassigning as this is going nowhere

Issue #2703, opened 2018-09-29
PEP8NotebookBear.py: Correct the spelling mistake in docstring


Naveenaidu:
There is a spelling mistake at [Line 11](https://github.com/coala/coala-bears/blob/559c953e1b5787df9dddf96ece8449e50c945716/bears/python/PEP8NotebookBear.py#L11).

```
# Comments regarind Jupyter Notebooks:
```

This can be a `difficulty/newcomer` issue.

Daronion:
I would like to be assigned to this issue.

shivank1234:
Is this issue open?

AditiKatiyar:
I would like to work on this issue.

PR #2704, opened 2018-09-29
PEP8NotebookBear.py: Correct the spelling mistake in docstring #2703

flamespt:
Fixed spelling mistake, error issue #2703. 
Sorry if this is not the right standard to do a pull request, but is my first one so bear with me please :)

### DIFF
--- a/bears/python/PEP8NotebookBear.py
+++ b/bears/python/PEP8NotebookBear.py
@@ -8,7 +8,7 @@ from coalib.results.Diff import Diff
 from coalib.results.Result import Result
 from coalib.settings.Setting import typed_list
 
-# Comments regarind Jupyter Notebooks:
+# Comments regarding Jupyter Notebooks:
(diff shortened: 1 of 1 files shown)


Naveenaidu:
@flamespt The changes look fine. There are few problems with the way you opened a PR
1. Please always ask the issue to be assigned to you. You can request for the assignment of the issue in the issue page. This is done because there might be other people who might be already working on the issue. That's what happens in this case too. Someone had already claimed the issue and even you worked on it. This leads to duplication of the results without others knowing it.  This has been mentioned in the documentation. Please always ask for the issue to be assigned to you before working on it. Sorry! to say but I don't think this PR would be accepted by maintainers since you did not claim it first. But I leave it up to them to decide.
2.  The commit message is not in proper format. Please read through the documentation about writing proper commit message.

Hope you find yourself a better newcomer issue :)

PR #2759, opened 2018-11-08
LineContinuationBear.py: Add line continuation bear

abhishalya:
Closes https://github.com/coala/coala-bears/issues/2427

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/LineContinuationBear.py
+++ b/bears/general/LineContinuationBear.py
@@ -1,0 +1,49 @@
+import logging
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result
+from coalib.bearlib.languages.Language import Language, UnknownLanguageError
+
+
+class LineContinuationBear(LocalBear):
+    def run(self,
+            filename,
+            file,
+            language,
+            ):
+        """
+        It is a generic bear which will look for any lines which
+        end with the explicit use of line continuation operators,
+        which is considered as a bad practice. The solution
+        would be to use implicit type of line continuation methods.
+
+        Note that this requires the coalang to know the tokens
+        used for line continuation and hence any language not
+        having the line continuation attribute will result in an error.
+
+        :param filename:
+            Name of the file that needs to be checked.
+        :param file:
+            File that needs to be checked in the form of a list of strings.
+        :param language:
+            Language to be used for finding the line continuation tokens.
+        """
+        try:
+            lang = Language[language]
+            line_continuation = lang.get_default_version().line_continuation
+        except (UnknownLanguageError, AttributeError):
+            logging.error('Language {} is not yet supported.'.format(language))
+            return
+
+        for line_number, line in enumerate(file):
+            if len(line) > 1:
+                if line[-2] in line_continuation:
+                    yield Result.from_values(
+                        origin=self,
+                        message='Explicit line continuation is not allowed.',
+                        file=filename,
+                        line=line_number + 1,
+                        column=len(line) - 1,
+                        end_line=line_number + 1,
+                        end_column=len(line),
+                        )
--- a/tests/general/LineContinuationBearTest.py
+++ b/tests/general/LineContinuationBearTest.py
@@ -1,0 +1,46 @@
+import logging
+from queue import Queue
+
+from bears.general.LineContinuationBear import LineContinuationBear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.testing.LocalBearTestHelper import execute_bear
+from coalib.results.Result import Result
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+
+
+good_file = """
+a = some_function(
+    '1' + '2')
+"""
+
+
+class LineContinuationBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('name')
+        self.uut = LineContinuationBear(self.section, Queue())
+
+    def test_good_file(self):
+        self.section.append(Setting('language', 'Python'))
+        self.check_validity(self.uut, good_file.splitlines())
+
+    def test_bad_file(self):
+        self.section.append(Setting('language', 'Python'))
+        self.check_results(
+            self.uut, ['a = 1 + \\', '2'],
+            [Result.from_values('LineContinuationBear',
+                                'Explicit line continuation is not allowed.',
+                                line=1, column=9, end_line=1, end_column=10,
+                                file='default')],
+            filename='default')
+
+    def test_lang_exception(self):
+        self.section.append(Setting('language', 'BlaBlaBla'))
+        ERROR_MESSAGE = 'ERROR:root:Language BlaBlaBla is not yet supported.'
+        logger = logging.getLogger()
+
+        with self.assertLogs(logger, 'ERROR') as log:
+            with execute_bear(self.uut, filename='F', file='') as result:
+                self.assertEqual(len(log.output), 1)
+                self.assertEqual(log.output[0], ERROR_MESSAGE)
--- a/bear-languages.yaml
+++ b/bear-languages.yaml
@@ -221,6 +221,7 @@ LatexLintBear:
   - Tex
 LicenseCheckBear:
 LicenseHeaderBear:
+LineContinuationBear:
 LineCountBear:
 LineLengthBear:
 LuaLintBear:


bears/general/LineContinuationBear.py
```diff
@@ -0,0 +1,32 @@
...
+            ):
+        try:
+            lang = Language[language]
+            line_continuation = lang.get_default_version().line_continuation
+        except Exception:
+            logging.error('Language {} is not yet supported.'.format(language))
+            return
+
+        for line_number, line in enumerate(file):
+            if len(line) - 1:
```
li-boxuan: What if `len(line) == 0`?
Maybe that would never happen - but it's better to be explicit IMO, e.g. `len(line) > 1`

bears/general/LineContinuationBear.py
```diff
@@ -0,0 +1,32 @@
...
+                if line[-2] in line_continuation:
```
li-boxuan: If there is trailing whitespace after line_continuation delimiter, your bear cannot detect

tests/general/LineContinuationBearTest.py
```diff
@@ -0,0 +1,20 @@
...
+a = '1' \
+    + '2' \
+    + '3'
+"""
+
+LineContinuationBearPythonTest = verify_local_bear(LineContinuationBear,
+                                                   valid_files=(good_file,),
+                                                   invalid_files=(bad_file,),
+                                                   settings={'language':
+                                                             'Python'})
```
li-boxuan: Can you add `self.assertEqual` methods (e.g. KeywordBearTest) to verify the behavior of your bear?

bears/general/LineContinuationBear.py
li-boxuan: Actually given my comment below,  `len(line) > whatever` here is a bad condition.

abhishalya: Yeah, any other options to solve it?

bears/general/LineContinuationBear.py
abhishalya: Yeah, but I guess that is already handled by another bear. So I guess these two conditions will never overlap, even if they do, one of the bears will throw an error.

Besides, as stated here https://github.com/coala/coala-bears/issues/2427#issuecomment-434286503 the initial idea of this bear is same.

tests/general/LineContinuationBearTest.py
abhishalya: Hmm, can you give me an example?

bears/general/LineContinuationBear.py
li-boxuan: Well, IMO it is invalid to have other characters following line continuation delimiter, so it might be fine.

tests/general/LineContinuationBearTest.py
li-boxuan: For example, how do we know the line number and column number you generate is correct? You have sth like `+1`, etc., which may or may not be true. Adding tests help reviewers check your bear has the correct behavior.

bears/general/LineContinuationBear.py
```diff
@@ -0,0 +1,32 @@
...
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result
+from coalib.bearlib.languages.Language import Language
+
+
+class LineContinuationBear(LocalBear):
+    def run(self,
+            filename,
+            file,
+            language,
```
ishanSrt: this must be a keyword argument 😉

bears/general/LineContinuationBear.py
```diff
@@ -0,0 +1,49 @@
...
+            Name of the file that needs to be checked.
+        :param file:
+            File that needs to be checked in the form of a list of strings.
+        :param language:
+            Language to be used for finding the line continuation tokens.
+        """
+        try:
+            lang = Language[language]
+            line_continuation = lang.get_default_version().line_continuation
+        except Exception:
```
ishanSrt: The exact exception like the `KeyError` or whatever pops up would be nice over here

tests/general/LineContinuationBearTest.py
```diff
@@ -0,0 +1,48 @@
...
+        self.section = Section('name')
+        self.uut = LineContinuationBear(self.section, Queue())
+
+    def test_run(self):
+        self.section.append(Setting('language', 'Python'))
+        self.check_results(
+            self.uut, bad_file.splitlines(),
+            [Result.from_values('LineContinuationBear',
+                                'Explicit line continuation is not allowed.',
+                                line=2, column=9, end_line=2, end_column=10,
```
ishanSrt: shouldn't this point to line 1 ? As the token for line continuation is on Line 1 and if only one line is shown in diff i.e.
```
    '2'
```
This would certainly be confusing for the user

ishanSrt: or the starting should be from line 1 and the end_line could be the next line

bears/general/LineContinuationBear.py
abhishalya: Hmm, true. Will find which type of exceptions I will get.

tests/general/LineContinuationBearTest.py
abhishalya: Umm, no since you see my triple `"` start from line `1` and the line continuation delimiter is on line `2`.

bears/general/LineContinuationBear.py
ishanSrt: what if the language is supported but the token for line continuation isn't there? A different error message or the same one for the both? Gonna keep this open for other reviewers ;)

tests/general/LineContinuationBearTest.py
abhishalya: I initially thought of starting the code from line 1 itself but it was looking weird :sweat_smile: 

But I can move it to line 1, no problems there :)

bears/general/LineContinuationBear.py
abhishalya: Yeah, I had implemented two different error messages. But I felt it was redundant, so just kept one for both since only one language supports line_continuation, Python.

ishanSrt: It still would be better to have both kinds of exceptions listed here. Everything else looks great!

abhishalya: Thank you for the help. I will definitely improve this :smile:

tests/general/LineContinuationBearTest.py
li-boxuan: Why not put test files into a folder?

tests/general/LineContinuationBearTest.py
```diff
@@ -0,0 +1,48 @@
...
+from coalib.settings.Setting import Setting
+
+
+good_file = """
+a = some_function(
+    '1' + '2')
+"""
+
+bad_file = r"""
+a = '1' \
```
jayvdb: this bear is about banning use of line continuation in our codebase.  that includes test modules.  Please remove the use of line continuation here.

This was previously mentioned at https://coala.zulipchat.com/#narrow/stream/101940-coala-bears/subject/line.20continuation.20bear/near/147283234

It can be used in external sample files which are not our codebase.

abhishalya: Oh, okay. I understand it now @jayvdb :+1:

tests/general/LineContinuationBearTest.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+
+class LineContinuationBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('name')
+        self.uut = LineContinuationBear(self.section, Queue())
+
+    def test_run(self):
```
jayvdb: give this a meaningful name

tests/general/LineContinuationBearTest.py
```diff
@@ -0,0 +1,43 @@
...
+        self.check_results(
+            self.uut, ['a = 1 + \\', '2'],
+            [Result.from_values('LineContinuationBear',
+                                'Explicit line continuation is not allowed.',
+                                line=1, column=9, end_line=1, end_column=10,
+                                file='default')],
+            filename='default')
+        self.check_validity(self.uut, good_file.splitlines())
+
+    def test_min_limit_exceeds_max_limit(self):
```
jayvdb: this doesnt describe the test being performed

tests/general/LineContinuationBearTest.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+
+class LineContinuationBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('name')
+        self.uut = LineContinuationBear(self.section, Queue())
+
+    def test_bear_run(self):
```
jayvdb: split into test_good_file and test_bad_file , so when these fail we have a better idea of what has happened

jayvdb:
ack https://github.com/coala/coala-bears/pull/2759/commits/0b20f85be35b75a22ccac4b35756bb563a2c5d08

jayvdb:
@gitmate-bot ff

PR #2764, opened 2018-11-10
PyStringConcatBear.py: Add Py string concat bear

samuelpiltch:
Detect use of explicit string concatenation in Python.
The bear checks if a line ends in a `+` and if so checks if the following line starts with a string enclosing character (', ", `).

Closes https://github.com/coala/coala-bears/issues/2434

### DIFF
--- a/bears/python/PyStringConcatBear.py
+++ b/bears/python/PyStringConcatBear.py
@@ -1,0 +1,37 @@
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result
+
+
+class PyStringConcatBear(LocalBear):
+    LANGUAGES = {'Python', 'Python 2', 'Python 3'}
+    CAN_DETECT = {'Formatting'}
+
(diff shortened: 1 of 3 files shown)


bears/python/PyStringConcatBear.py
```diff
@@ -0,0 +1,30 @@
...
+    CAN_DETECT = {'Formatting'}
+
+    def run(self, filename, file):
+        """
+        Detects use of `+` to concatenate two strings, which should be avoided.
+        """
+        string_enclosings = ['\"', '`', '\'']
+
+        for line_number, line in enumerate(file):
+            if len(line) - 1:
```
bkhanale: Improve this by `len(line) > 0`

jayvdb:
ack https://github.com/coala/coala-bears/pull/2764/commits/ed75df6464e8b306c897ab8accd904fe05060c73

jayvdb:
@gitmate-bot ff

PR #2779, opened 2018-11-13
Support Tslint 5

jayvdb:
Sits on top of https://github.com/coala/coala-bears/issues/2778

### DIFF
--- a/bears/typescript/TSLintBear.py
+++ b/bears/typescript/TSLintBear.py
@@ -37,7 +37,7 @@ class TSLintBear:
         args = ('--format', 'json')
         if tslint_config:
             args += ('--config', tslint_config)
-        if rules_dir:
+        if rules_dir:  # pragma: no cover
             args += ('--rules-dir', rules_dir)
         return args + (filename,)
 
--- a/bears/typescript/TSLintBear.py
+++ b/bears/typescript/TSLintBear.py
@@ -17,7 +17,7 @@ class TSLintBear:
     """
 
     LANGUAGES = {'TypeScript'}
-    REQUIREMENTS = {NpmRequirement('tslint', '3'),
+    REQUIREMENTS = {NpmRequirement('tslint', '>=3.0 <6.0'),
                     NpmRequirement('typescript', '>=1.7.3')}
     AUTHORS = {'The coala developers'}
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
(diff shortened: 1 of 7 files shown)


bears/typescript/TSLintBear.py
```diff
@@ -37,7 +37,7 @@ def create_arguments(filename, file, config_file,
         args = ('--format', 'json')
         if tslint_config:
             args += ('--config', tslint_config)
-        if rules_dir:
+        if rules_dir:  # pragma: no cover
```
jayvdb: Worth noting that since tslint requires a `tslint.json` otherwise it does nothing if no rules are selected , `rules_dir` is a bit redundant as `tslint.json` will usually contain the `rulesDirectory`.

It might be simpler to deprecated/decommission rules_dir, or at least not enhance it until someone does some analysis to show it is useful.

li-boxuan: If we don't provide `rules_dir` option in the future, that means `tslint_config` option becomes a must?

jayvdb: If there is no value for `tslint_config` , tslint may find tslint.json in the top level directory where coala is run.  If there is no tslint.json, there are no rules being used, especially not any custom rules which might reside in a custom directory.

If there is a `tslint.json`, why wouldnt it include `rulesDirectory` ?  It seems silly to not use `rulesDirectory`.

It would seem to me that a `rules_dir` setting only makes sense if the bear also had a setting for custom rule names to enforce, in which case a directory to get the rules from is needed.

But more analysis may expose a reason for `rules_dir`.

jayvdb:
Created https://github.com/coala/coala-bears/issues/2783

jayvdb:
ack https://github.com/coala/coala-bears/pull/2779/commits/ce194a6e1adce1e40bd8d9665f10110719481679 https://github.com/coala/coala-bears/pull/2779/commits/38da4f51539546c2f365049e91e3dba655fc5b09 https://github.com/coala/coala-bears/pull/2779/commits/96eed925aa31b8e7fc1562529c1616eaab60f7db https://github.com/coala/coala-bears/pull/2779/commits/5b981463f03ae98d86ecedd01f6a2912707aeb01

jayvdb:
@gitmate-bot ff

jayvdb:
@gitmate-bot ff

PR #2800, opened 2018-11-29
AnnotationBear: Remove LanguageDefinition

Kawzeg:
LanguageDefinition is deprecated and should not be used.
Now uses Language instead.

Fixes https://github.com/coala/coala-bears/issues/2799

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/AnnotationBear.py
+++ b/bears/general/AnnotationBear.py
@@ -1,4 +1,5 @@
-from coalib.bearlib.languages.LanguageDefinition import LanguageDefinition
+from coalib.bearlib.languages.Language import Language, UnknownLanguageError
+from coalib.bearlib import deprecate_settings
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.HiddenResult import HiddenResult
 from coalib.results.Result import Result, RESULT_SEVERITY
@@ -12,6 +13,7 @@ class AnnotationBear(LocalBear):
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
 
+    @deprecate_settings(coalang_dir='coalang_dir')
     def run(self, filename, file, language: str, coalang_dir: str = None):
         """
         Finds out all the positions of strings and comments in a file.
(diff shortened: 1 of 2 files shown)


bears/general/AnnotationBear.py
```diff
@@ -12,16 +12,14 @@ class AnnotationBear(LocalBear):
     AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
     LICENSE = 'AGPL-3.0'
 
-    def run(self, filename, file, language: str, coalang_dir: str = None):
+    def run(self, filename, file, language: str):
```
jayvdb: keep the setting `coalang_dir` , but 'deprecate' it (see other deprecated settings in bears repo), and also dont attempt to re-implement it at all.

We might be able to re-implement this later, when coAST is usable.

bears/general/AnnotationBear.py
```diff
@@ -290,6 +295,17 @@ def get_singleline_comment(file, filename, text, comment, position):
                 end_position)
 
 
+def get_if_present(language, attribute, default={}):
```
jayvdb: `def _get_lang_attr` and invoke it with `_get_lang_attr(.., ..., default=..)` which automatically implies that default will apply if the attr is not present.

tests/general/AnnotationBearTest.py
```diff
@@ -54,6 +71,36 @@ def test_single_line_comment(self):
...
 
+    def test_string_comment_delim(self):
+        text = ['Line with //comment\n',
+                'Second line with //different comment\n']
+        joined = ''.join(text)
+        first_comment_start = joined.find('//')
+        first_comment_end = joined.find('\n', first_comment_start + 1)
+        second_comment_start = joined.find('//', first_comment_start + 1)
+        second_comment_end = joined.find('\n', second_comment_start + 1)
+        expected = (SourceRange.from_absolute_position(
```
jayvdb: this line makes the indentation on the next line incorrect.  this fixes it:

```py

        expected = (
            SourceRange.from_absolute_position(
```

bears/general/AnnotationBear.py
sils: uhm why don't you just use `getattr(language, attribute, default)` in the first place? No need to define this function

sils: https://docs.python.org/3/library/functions.html#getattr

PR #2827, opened 2018-12-13
JSONFormatBear.py : Fix erroneous message

naveentvelu:
This fixes the erroneous message which
was omitting the json_sort value.

Fixes https://github.com/coala/coala-bears/issues/2503


bears/js/JSONFormatBear.py
```diff
@@ -90,9 +90,13 @@ def run(self, filename, file,
...
         if len(diff) > 0:
+            out_mess = ('This file can be reformatted '
+                        'by following indentation.')
+            if json_sort:
+                out_mess = ('This file can be reformatted by sorting keys '
+                            'and following indentation.')
             yield Result(self,
-                         'This file can be reformatted by sorting keys and '
-                         'following indentation.',
+                         out_mess,
```
soumya-ranjan7: Include a comment describing the 'out mess' parameter. Just an one line description is enough. I would also suggest to change the name of the variable to something more meaningful (maybe err_message). Rest of the code looks fine.

naveentvelu: `format_message` Will do I guess?

naveentvelu:
@soumya-ranjan7 @li-boxuan Review this please

bears/js/JSONFormatBear.py
```diff
@@ -90,9 +90,15 @@ def run(self, filename, file,
         diff = Diff.from_string_arrays(file, corrected)
 
         if len(diff) > 0:
+            # Appropriate message is displayed to the user based on
```
soumya-ranjan7: Please change the comment describing the parameter to this format:
"""
:param format_message: (a brief description about the parameter)
"""

naveentvelu:
Sure, will add.

naveentvelu:
@li-boxuan I am facing a little issue with this. So need little clarification.
I added the test_cases and the messages are being checked correctly using `check_results`. But the `Diff` object is not matching, reason being the default value passed is `None`. In all other instances where the `check_results` is used, the desired expected value of `Diff` object is `None` and same was passes as expected value. 

But in this case we have to pass the expected `Diff` object. How can I do that?

abhishalya:
@naveentvelu `check_results` should work. Compare the strings from both of the `Result` objects yourself. (See the length difference, assert wherever required to find out the difference between the two). You  should be able to get your mistake, if not then post the entire pytest error on gitter.

naveentvelu:
@abhishalya 
The `check_results` will not work because, we have a `diff` object which is not `None`. So when you compare two of them their ids get compared rather than their attributes. 
If you see the above assertion error, 
It is because of `diffs`(one of the attributes of 'result' class) which belongs to `<class 'dict'>`. This has a key-value pair `filename : <class 'coalib.results.Diff.Diff'>`. 
In the current implementation of `check_results`, this key-value pair is directly being compared.This causes the assertion error.

naveentvelu:
> See the length difference, assert wherever required to find out the difference between the two.

This is not needed here because there will always be mismatch in the id's of class objects even if they have the same attribute.

naveentvelu:
@abhishalya @li-boxuan 
Finally one of the approaches wokr, I added a `def__repr__(self):` function in Diff class. And It solved the above assertion error.
Should I create a issue along with PR for modifying the Diff class? 
It's in coala/coala repo. https://github.com/coala/coala/blob/master/coalib/results/Diff.py

li-boxuan:
> Then I moved on to writing test using LocalBearTestHelper.
I was using check_results for the test cases.
But I got a assertion error like this ...

Post your code snippet, or just push with your unfinished code, so that people have a better understanding of your problem or find bugs with your code if there is any.

naveentvelu:
> > Then I moved on to writing test using LocalBearTestHelper.
> > I was using check_results for the test cases.
> > But I got a assertion error like this ...
> 
> Post your code snippet, or just push with your unfinished code, so that people have a better understanding of your problem or find bugs with your code if there is an

I got it right now. The mistake was on my side. I passed the wrong expected diff value. Sorry @li-boxuan @abhishalya

tests/js/JSONFormatBearTest.py
```diff
@@ -41,6 +42,12 @@
 # This will generate a line with 79 characters
 max_line_length_file2 = '{\n    "string": "' + 'a' * 63 + '"\n}'
 
+# This file is for checking the appropriate message displayed
+# to the user based on the json_sort value
+test_file5 = '{\n', '    "a":   5\n', '}\n'
+
+correct_test_file5 = '{\n', '    "a": 5\n', '}\n'
```
abhishalya: Some relevant names to the test files would be better.

bears/js/JSONFormatBear.py
naveentvelu: @li-boxuan Is the comment format wrong?

bears/js/JSONFormatBear.py
```diff
@@ -90,9 +90,15 @@ def run(self, filename, file,
...
+            # json_sort value
+            format_message = ('This file can be reformatted '
+                              'by following indentation.')
+            if json_sort:
+                format_message = ('This file can be reformatted by sorting '
+                                  'keys and following indentation.')
```
li-boxuan: What if `json_sort` is True, but keys are already sorted? The format message `This file can be reformatted by sorting keys ...` will be confusing.

The next problem is, how do we know if keys are already sorted? A naive approach would be to use `json.dumps` twice, change `sort_keys` value and keep other parameters unchanged, then compare the resulting diff. Any better idea?

bears/js/JSONFormatBear.py
naveentvelu: Yes, I thought so. Will remove it in next commit.

bears/js/JSONFormatBear.py
naveentvelu: Yes, that's a problem. Shouldn't we address that in another PR ? Because this PR does pass through the acceptance criteria mentioned in the issue https://github.com/coala/coala-bears/issues/2503.

li-boxuan: @Makman2 ping :)

jayvdb:
@gitmate-bot rebase

bears/js/JSONFormatBear.py
jayvdb: Those are *minimal* acceptance criteria.

If new additional problems are found in your code, they should be fixed so that your algorithm is stable code.

IMO it is ok to defer determine what type of problem occurred, e.g. by seeing if the keys are unsorted.

But the message needs to be accurate.  Now in light of that, the acceptance criteria are insufficient.  The indentation might be correct, and the keys all sorted, but the dump is different due to setting `escape_unicode`.

For the message to be accurate, it should indicate the values of the three parameters which is using, and be written so that the reader understands that the error could be caused by any of the three parameters being wrong.

naveentvelu: Yes, I agree. I will work towards a possible solution. Any input towards that would be really helpful.:)

tests/js/JSONFormatBearTest.py
naveentvelu: Will do in my next commit..

PR #2836, opened 2019-01-06
URLHeadBear.py: Change connection check URL

frextrite:
Changed URL that checks if client is connected to the internet or not
from Google's IP address 216.58.218.174
to DNS IP owned by Cloudfare 1.1.1.1

Fixes https://github.com/coala/coala-bears/issues/2833

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/URLHeadBear.py
+++ b/bears/general/URLHeadBear.py
@@ -69,8 +69,8 @@ class URLHeadBear(LocalBear):
     LICENSE = 'AGPL-3.0'
     CAN_DETECT = {'Documentation'}
 
-    # IP Address of www.google.com
-    check_connection_url = 'http://216.58.218.174'
+    # DNS IP by Cloudfare
+    check_connection_url = 'https://1.1.1.1/'
 
     @classmethod
     def check_prerequisites(cls):
--- a/tests/general/URLHeadBearTest.py
+++ b/tests/general/URLHeadBearTest.py
@@ -15,8 +15,9 @@ class URLHeadBearTestPrerequisites(unittest.TestCase):
     def test_check_prerequisites(self):
         with requests_mock.Mocker() as m:
             m.add_matcher(custom_matcher)
-            self.assertEqual(URLHeadBear.check_prerequisites(),
-                             'You are not connected to the internet.')
+            self.assertTrue(URLHeadBear.check_prerequisites())
+            self.assertNotEqual(URLHeadBear.check_prerequisites(),
+                                'You are not connected to the internet.')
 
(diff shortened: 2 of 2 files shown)


frextrite:
Adding to the previous comment, I backtracked to the point the original test case was replaced by a faulty one, I got the following results,

While sending a GET resuest using requests on some URL say,```http://random.url```, the requests module explicitly appends a trailing ```/```(forward-slash) if it is missing from the url, hence the final URL that actually gets called becomes ```http://random.url/```. Owing to this fact, in the ```custom_matcher``` defined in ```InvalidLinkBearTest.py``` at [L42](https://github.com/coala/coala-bears/blob/2888639aa2427ae77ab74486e8750770a1e219c1/tests/general/InvalidLinkBearTest.py#L42),```request.url == URLHeadBear.check_connection_url``` would always return ```False``` since the value of original ```check_connection_url``` was defined to be ```http://216.58.218.174```(Note: no forward-slash). Hence this condition would never be possible and the value of the ```status_code``` of the default connection url would always be ```None```, though this behaviour was **never intended**.

When the original test case for this was written, it was asserting True on the returned value of the output of ```URLHeadBear.check_prerequisites()```. It would always return the error string ```You are not connected to the internet.```, and when asserting this return, the test would pass, **incorrectly**. 

To overcome this, the test was modified in [this commit](https://github.com/coala/coala-bears/commit/9992f4b6ee49007c4c4de42e0313fb14011174e3#diff-6fe1c3cef020b4ffe41fa8bcce231e9b), to assert for the returned string instead of the asserting True. This too was **incorrect**, since in the custom matcher the ```check_connection_url``` URL was explicitly defined(atleast intended) to return a status code of ```200OK```, but this line was never hit as mentioned in the first paragraph.

Hence ```URLHeadBear.check_prerequisites()``` while testing would always return the error string ```You are not connected to the internet.``` and hence as mentioned in the above commit, instead of changing the mock properties to actually return ```200OK``` if the head request was to ```check_connection_url``` URL, the actual test case was modified to check against the error string which was **never intended**.

frextrite:
In my defense, I was sleepy when I made the commit

jayvdb:
@gitmate-bot rebase

jayvdb:
ack https://github.com/coala/coala-bears/pull/2836/commits/7c6ae737d1b19a28ecb2da70d797ca30f27cf8af

jayvdb:
@gitmate-bot ff

PR #2848, opened 2019-01-31
CI: Fix Pycodestyle warning W605

Naveenaidu:
W605 is a warning for invalid escape sequences.

In python 3.6, Unknown escapes consisting of `\` and an ASCII letter in regular expressions will
cause an warning.In the upcoming python versions this would be treated as an error.
Owing to it pycodestyle have added this check for future Python versions.

The alternative way to escape special characters now is to append a extra `\` infront of the
sequences. Escape the characters by double backticks.

Fixes https://github.com/coala/coala-bears/issues/2760

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/c_languages/codeclone_detection/ClangCountingConditions.py
+++ b/bears/c_languages/codeclone_detection/ClangCountingConditions.py
@@ -245,8 +245,8 @@ def in_sum(stack):  # pragma nt: no cover
 def in_product(stack):  # pragma nt: no cover
     """
     A counting condition returning true if the variable is used in a product
-    statement, i.e. within the operators \*, /, % and their associated compound
-    operators.
+    statement, i.e. within the operators \\*, /, % and their associated
+    compound operators.
     """
     return _stack_contains_operators(stack, ['*', '/', '%', '*=', '/=', '%='])
 
@@ -254,8 +254,8 @@ def in_product(stack):  # pragma nt: no cover
 def in_binary_operation(stack):  # pragma nt: no cover
     """
     A counting condition returning true if the variable is used in a binary
-    operation, i.e. within the operators \|, & and their associated compound
-    operators.
+    operation, i.e. within the operators \\|, & and their associated
+    compound operators.
     """
     return _stack_contains_operators(stack, ['&', '|', '&=', '|='])
 
--- a/bears/vcs/git/GitCommitBear.py
+++ b/bears/vcs/git/GitCommitBear.py
@@ -35,7 +35,7 @@ class GitCommitBear(_CommitBear):
         """
         This function creates a git command to fetch the
         unmerged parent commit shortlog from a commit generated
-        by GitHub in a refs/pull/(\d+)/merge git remote reference.
+        by GitHub in a refs/pull/(\\d+)/merge git remote reference.
         Visit https://github.com/travis-ci/travis-ci/issues/8400
         for more details.
 


Naveenaidu:
> commit message doesn't say why ignoring is appropriate solution

@jayvdb The reason I ignored the `W605`(Invalid escape sequence) was for a quick fix for travis CI. Which I realized was a wrong thing to do( Never silence an error to make your test pass :P)

I have been going through the past PR's from the pycodestyle repo and other repo's which were facing such issues. Looking at the history of out PR builds. Travis started to fail because of this error somewhere around December 2 -  2018. Before that W605 existed but never poised the current problem. I am yet to understand what caused it.

Reading the comment [here](https://github.com/PyCQA/pycodestyle/issues/766#issuecomment-451141810) and the [Issue](https://github.com/Python-Markdown/markdown/issues/611), the better solution to the current failure of Travis would be to either escape the sequences such as `'\d'` by adding an extra `\` or by `r'\d'`.

After trying out the above method on my fork, I realised that we have more than 4 files which needs to be fixed else the W605 warning will persist.

I would love to have your opinions on this.

frextrite:
The best option right now would be to wrap all the invalid escape sequences with `r" "`, in all the 4 files. 

Moreover, why did PycodestyleBear start emitting W605 all of a sudden? I couldn't find any erroneous commit that could've resulted in this behavior. Did something change upstream?

sladyn98:
According to me we could ignore the warning for now, if we run into regex expression issues later due to the wrapping of r'  ' we could fix it via the string literal append, as mentioned by @Naveenaidu.

jayvdb:
Why hasnt this been updated??  At the very least your commit message is useless, and it is referencing an issue which was closed as a duplicate.  Stop looking for other people to guide you to the best solution, especially not from maintainers.  Find the best possible solution yourself, or listen to the best advice from other community members.

Naveenaidu:
> Why hasnt this been updated?? At the very least your commit message is useless, and it is referencing an issue which was closed as a duplicate. Stop looking for other people to guide you to the best solution, especially not from maintainers. Find the best possible solution yourself, or listen to the best advice from other community members.

Thanks for telling me this :)

@sladyn98 IMHO Ignoring the warning didn't feel the best possible approach to the problem. I stand by my point that we should not ignore the errors only to make the build possible.

@frextrite  As discussed above, I first used a `r" "` to wrap all the invalid sequences in it. But then I faced two problems with the above solution
1. Sphinx documentation doesn't recognize `r" "`, and hence in the documentation, the sequences haven't been escaped rather they appeared as a string. eg:- (`r"/*"` appeared as `r"/*"`, whereas it should detect the expression as only `*` after the netlify build of documentation) (I would be raising an issue there to inform about it soon)
2. Pycodestyle still emitted the warning of `Invalid escape sequences`

That's the reason why i did not use `r` to wrap the invalid sequences.

An alternative solution to the above problem was to use an extra backslash `\` before the sequence. This solved the above two problems. Sphinx perfectly detected a `\\*` as `*` and Pycodestyle didn't complain too.

IMHO it is necessary to mention this somewhere in our documentation so that we don't encounter these errors in the future and cause build failures.

bears/c_languages/codeclone_detection/ClangCountingConditions.py
```diff
@@ -245,17 +245,17 @@ def in_sum(stack):  # pragma nt: no cover
 def in_product(stack):  # pragma nt: no cover
     """
     A counting condition returning true if the variable is used in a product
-    statement, i.e. within the operators \*, /, % and their associated compound
-    operators.
+    statement, i.e. within the operators \\*, /, % and their associated
+    compound operators.
```
Naveenaidu: This change has been made to keep the character count under 80

Naveenaidu: This change has been made to keep the character count under 80. The same goes for the below change.

sladyn98:
Yeah @Naveenaidu your right but  I suggested ignoring because you suggested that we would have to make changes in four files,else in my original comment on gitter I said that adding an extra '\' to avoid the escape sequences  error would be the best solution.Moreover @jayvdb mentioned in his orginal issue that sphinx would throw errors on ```r" " ``` wrapping ,my bad not to see that, anyways we would have to wait for the gometa linter issue to fixed for us to see the travis build and then proceed.

jayvdb:
ack https://github.com/coala/coala-bears/pull/2848/commits/25a2c43f7f0785518134dadf098495d52795c316

jayvdb:
@gitmate-bot ff

PR #2896, opened 2019-03-22
generate_bear_requirements.py: Add Perl Meta

Ishaan29:
This enhances generate_bear_requirements.py to support
perl bears and generate Makefile.PL using jinja2
template.
closes https://github.com/coala/coala-bears/issues/2483

### DIFF
--- a/.moban.dt/Makefile.PL.jj2
+++ b/.moban.dt/Makefile.PL.jj2
@@ -1,0 +1,6 @@
+use ExtUtils::MakeMaker;
+WriteMakefile(
+    NAME              => 'Coala::Bears',
+    VERSION           => '0.10',
+    PREREQ_PM         => {Perl::Critic => 1.126},
+);
--- a/bears/perl/PerlCriticBear.py
+++ b/bears/perl/PerlCriticBear.py
@@ -1,6 +1,10 @@
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.DistributionRequirement import (
     DistributionRequirement)
+from dependency_management.requirements.AnyOneOfRequirements import (
+    AnyOneOfRequirements)
+from dependency_managment.requirements.PerlRequirement import (
+    PerlRequirement)
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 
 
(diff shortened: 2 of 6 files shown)


KVGarg:
@Ishaan29 There should be a blank line between `Closes` And your commit description 😇

bears/perl/PerlCriticBear.py
```diff
@@ -1,6 +1,10 @@
 from coalib.bearlib.abstractions.Linter import linter
 from dependency_management.requirements.DistributionRequirement import (
     DistributionRequirement)
+from dependency_management.requirements.AnyOneOfRequirements import (
+    AnyOneOfRequirements)
+from dependency_managment.requirements.PerlRequirement import (
```
sangamcse: Is this dependency manager available in the [`coala package_manager`](https://gitlab.com/coala/package_manager/tree/master/dependency_management/requirements)?

Ishaan29: No it is not available, Thank you for pointing me in the right direction ... I will create `PerlRequirement` file.

sangamcse: BTW the issue is not about creating a dependency. This issue is to create just Perl meta.

Ishaan29: I don't think there is a need to add `PerlRequirement` as `CpanRequirement` is already present..

sangamcse: If `CpanRequirement` is the requirement here, then work with that. Adding `PerlRequirement` here without having it in package_manager will break it

jayvdb:
Continuation of https://github.com/coala/coala-bears/pull/2888

.moban.dt/Makefile.PL.jj2
```diff
@@ -0,0 +1,6 @@
+use ExtUtils::MakeMaker;
+WriteMakefile(
+    NAME              => 'Coala::Bears',
+    VERSION           => '0.10',
+    PREREQ_PM         => {Perl::Critic => 1.126},
```
jayvdb: this isnt a template if it has literal values in it.

bears/perl/PerlCriticBear.py
jayvdb: Will you be fixing this @Ishaan29 ?

PR #2912, opened 2019-06-04
bears/general: Add FileModeBear

### LINKED ISSUE #2370: Create FileModeBear
It's really easy to forget to set +x on scripts. How about a bear like this:

```ini
[all.mode]
bears = FileModeBear
require = rw

[all.shell.mode]
bears = FileModeBear
require = rwx
```

bkhanale:
Closes https://github.com/coala/coala-bears/issues/2370

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/bears/general/FileModeBear.py
+++ b/bears/general/FileModeBear.py
@@ -1,0 +1,48 @@
+import os
+import stat
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+
+
+class FileModeBear(LocalBear):
+    LANGUAGES = {'All'}
+    AUTHORS = {'The coala developers'}
+    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
+    LICENSE = 'AGPL-3.0'
+
+    def run(self,
+            filename,
+            file,
+            filemode: str,
+            ):
+        """
+        The bear will check if the file has required permissions provided by
+        the user.
+
+        :param filemode:
+            Filemode to check, e.g. `rw`, `rwx`, etc.
+        """
+        st = os.stat(filename)
+        permissions = {'r': stat.S_IRUSR,
+                       'w': stat.S_IWUSR,
+                       'x': stat.S_IXUSR,
+                       }
+
+        invalid_chars = [ch for ch in filemode if ch not in permissions]
+
+        if invalid_chars:
+            raise ValueError('Unable to recognize character `{}` in filemode '
+                             '`{}`.'.format(''.join(invalid_chars), filemode))
+
+        mode = st.st_mode
+        for char in filemode:
+            if not mode & permissions[char]:
+                message = ('The file permissions are not adequate. '
+                           'The permissions are set to {}'
+                           .format(stat.filemode(mode)))
+                return [Result.from_values(origin=self,
+                                           message=message,
+                                           severity=RESULT_SEVERITY.INFO,
+                                           file=filename)]
--- a/tests/general/FileModeBearTest.py
+++ b/tests/general/FileModeBearTest.py
@@ -1,0 +1,219 @@
+import os
+import platform
+import stat
+
+from queue import Queue
+
+from bears.general.FileModeBear import FileModeBear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.results.Result import RESULT_SEVERITY, Result
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+
+
+FILE_PATH = os.path.join(os.path.dirname(__file__),
+                         'filemode_test_files', 'test_file.txt')
+
+
+class FileModeBearTest(LocalBearTestHelper):
+
+    def setUp(self):
+        self.section = Section('')
+        self.uut = FileModeBear(self.section, Queue())
+
+    def test_r_to_r_permissions(self):
+        os.chmod(FILE_PATH, stat.S_IRUSR)
+        self.section.append(Setting('filemode', 'r'))
+        self.check_results(
+            self.uut,
+            [],
+            [],
+            filename=FILE_PATH,
+            )
+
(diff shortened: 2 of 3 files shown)


tests/general/FileModeBearTest.py
```diff
@@ -0,0 +1,212 @@
...
+        self.check_results(
+            self.uut,
+            [],
+            [],
+            filename=FILE_PATH,
+            )
+
+    def test_x_to_x_permissions(self):
+        os.chmod(FILE_PATH, stat.S_IXUSR)
+        if platform.system() != 'Windows':
```
abhishalya: Mention the reason wherever you have done this.

abhishalya (changes requested):
Overall looks good :+1:

tests/general/FileModeBearTest.py
```diff
@@ -0,0 +1,212 @@
...
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+
+
+def get_testfile_path(file):
+    return os.path.join(os.path.dirname(__file__),
+                        'filemode_test_files', file)
+
+
+FILE_PATH = get_testfile_path('test_file.txt')
```
abhishalya: You are using `get_testfile_path()` only once, so no need to keep the function.

abhishalya (changes requested):
Few improvements, other things look good.

bears/general/FileModeBear.py
```diff
@@ -0,0 +1,51 @@
...
+            if char not in permissions:
+                raise ValueError('Unable to recognize character `{}` in '
+                                 'filemode `{}`.'.format(char, filemode))
+
+        mode = st.st_mode
+        for char in filemode:
+            if not mode & permissions[char]:
+                message = ('The file permissions are not adequate. The '
+                           'permissions are set to {}'.format(stat.filemode(
+                                                              mode)))
```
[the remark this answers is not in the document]
bkhanale: Will improve this :+1:

jayvdb:
Current codecov failure at 99.91% can be ignored - https://travis-ci.org/coala/coala-bears/jobs/561920971 shows the new bear is getting 100% test coverage

tests/general/FileModeBearTest.py
jayvdb: No.  Get rid of it.  If this bear doesnt support Windows, that fact shouldnt be hidden inside the logic of the tests.

A `.coafile` is a cross-platform specification, and it must not cause failures on only one platform.

jayvdb: Perhaps create a separate test class for executable permissions, and the class should be skipped on Windows for tests which dont work on Windows, and potentially we need a second extra test class which does Windows specific testing to show that specifying executable bits in the .coafile results in a mild note in the coala log, but does not cause it to fail.

jayvdb (changes requested):

jayvdb:
It should be possible to get executable bit working correctly on Windows, and this PR cant close the issue without executable support, as executable support is the primary motivation for this bear, and the need to include Windows was a consistent message throughout the comments of the issue.

bkhanale:
@jayvdb The bear works fine for Windows (as it only reads the file permissions and doesn't change them), the problem was I couldn't test it properly on Windows since Python can only set read-only flags on it. One way to test this could be to rename the file to the extension `.exe` which automatically sets the executable flag.

tests/general/FileModeBearTest.py
```diff
@@ -0,0 +1,219 @@
...
+
+from bears.general.FileModeBear import FileModeBear
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+from coalib.results.Result import RESULT_SEVERITY, Result
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+
+
+FILE_PATH = os.path.join(os.path.dirname(__file__),
+                         'filemode_test_files', 'test_file.txt')
```
jayvdb: the issue asked for "...on scripts" .  Why are you using `.txt` ?

jayvdb:
@bkhanale ,

> ... the problem was I couldn't test it properly on Windows

If you can not test it, you have no way to determine that is works, and you should not attempt to argue that it does work.

I can see a mile away it does not achieve what the issue requested.

No you do not get to create a useless test by renaming files to .exe .  That isnt what the issue asked for.

And even ignoring the fact the bear doesnt work, the test method `test_x_to_x_permissions` on Windows does:
```python
os.chmod(FILE_PATH, stat.S_IXUSR)
os.chmod(FILE_PATH, stat.S_IRUSR)
```

That is not a test of the bear, it it.  That is wrong.  Fix it.

https://github.com/coala/coala-bears/pull/2912/files#r294035442 asked you two months ago to document why you were doing that.  You havent done that.  Why not?

Why are you using custom test logic instead of `verify_local_bear`?  Is there something about this bear which means `verify_local_bear` doesnt work for any of the tests?  If so, why not use `self.check_invalidity`?

tests/general/FileModeBearTest.py
bkhanale: I guess having no extension at all would've been better. This is because the bear is supposed to check for all file permissions as mentioned [here](https://github.com/coala/coala-bears/issues/2370#issuecomment-401603938) and is not just restricted scripts.

jayvdb: Start with the realistic scenario of scripts having `.sh` / `.bash` etc extensions and make sure you can get that working and tested problem, before trying extensionless voodoo.

abhishalya:
>And even ignoring the fact the bear doesnt work, the test method test_x_to_x_permissions on Windows does:

I guess that he did to roll back changes made to the file permissions (which I guess defaults to read permission).

>Why are you using custom test logic instead of verify_local_bear? Is there something about this bear which means verify_local_bear doesnt work for any of the tests? If so, why not use self.check_invalidity?

Yep, you can use `check_validity` wherever you're not checking for the message statements.

PR #2948, opened 2019-08-17
FilenameBear.py: Provide multiple patches

akshatkarani:
FilenameBear will be able to provide multiple patches whenever it cannot guess a suitable naming convention.
asciinema of [this](https://asciinema.org/a/M6q4Tz1UrDYNxekT2s5OcyUE8)


akshatkarani:
I think `codecov` needs to be ignored, see [this](https://github.com/coala/coala-bears/pull/2912#issuecomment-513734892)

PR #2968, opened 2020-01-18
Update mypy to >=0.590

daureg:
The version used right now is from April 2018 https://github.com/python/mypy/releases/tag/v0.590


utkarsh2102 (changes requested):
Thanks for this but I believe #2976 supersedes this and is a better fix, what do you think?

daureg:
> Thanks for this but I believe #2976 supersedes this and is a better fix, what do you think?

agreed

PR #3002, opened 2022-10-18
Upgrade infer

### LINKED ISSUE #2931: Upgrade infer
InferBear is using infer 0.7.0 , which is very old and requires a lot of hand holding to keep building.

It would be good if a newer version of infer was supported.

bailey-coding:
Closes #2931

### Checklist

      them.
      pass on every single commit!

### DIFF
--- a/.ci/deps.opam.sh
+++ b/.ci/deps.opam.sh
@@ -6,10 +6,10 @@ set -x
 .ci/deps.python27.sh
 
 # Infer commands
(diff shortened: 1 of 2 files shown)


bailey-coding:
I hoped this would trigger a CI run to see if it passes with this change, but I think more might be necessary.

Maybe we can simplify some of the dependency scripts for CI, since that's implied in #2931, and hopefully is true.

